PR #27, opened 2026-07-08
fix(mempool): gate activation on estimated tip distance

evan-forbes:
## Motivation

A Mainnet node syncing from scratch can start its mempool while it is far behind the network tip. The mempool then verifies current transactions against stale consensus rules and can ban honest peers.

`SyncStatus::is_close_to_tip()` measures sync throughput. A peer-starved syncer downloads zero blocks per round, so this signal reports that the node is close to the tip. Peer bans then reinforce the starvation. #473 also removed the transition that disabled an active mempool.

## Solution

- Require both near-tip sync throughput and an estimated tip distance of at most 100 blocks before activating the Mainnet mempool or crawler. Disable an active mempool only beyond `MAX_BLOCK_REORG_HEIGHT` (1000 blocks), so fluctuations between the thresholds do not toggle activation.
- Bypass the clock estimate on test networks, where mining can stop for long periods. Keep the sync-throughput gate. Preserve the debug activation override.
- Make the crawler wait for a tip update while the clock gate fails. It waits for sync status only when that status can open the gate, avoiding an immediately-ready loop on a stale tip.
- Sample the activation decision before consuming the tip action and reuse that decision during the state transition.
- Preserve the verification tip height in failed results. Score a peer only when that height still matches the current tip and the node passes the normal gate. Reject branch-ID and lock-time mismatches without penalizing the peer, even if the estimate reports a current tip.
- Broadcast the removed transaction IDs when distance-based disabling clears the mempool.

The timestamp estimate is a freshness heuristic, not a bound on actual height lag. Consensus-valid future timestamps can hide part of the lag. Exempting branch-ID mismatches fixes the reported upgrade-boundary ban without rejecting valid tip timestamps or treating the estimate as proof that the peer used the wrong rules. This PR does not reclassify every context-dependent verification error.

## Testing

Regression coverage exercises stale-tip crawler yielding and wakeup, verification started before catch-up, branch mismatches with a future-dated tip, sparse testnet activation and crawling, a sync-status change between activation sampling and state transition, and subscriber invalidation on disabling. Existing coverage checks threshold hysteresis and ensures invalid shielded proof sizes still score peers.

Validated with Rust 1.98.0:

- `cargo +1.98.0 test -p zakura --lib mempool --no-fail-fast`: 107 passed.
- `cargo +1.98.0 clippy -p zakura --lib --tests -- -D warnings`: passed.
- `cargo +1.98.0 fmt --all -- --check`: passed.
- `./scripts/changelog.py check`: passed.

## Changelog

`docs/changelog/unreleased/27.md`

- `check_synced_to_tip()` in `crates/zakura-rpc/src/methods/types/get_block_template.rs` combines sync throughput with a 100-block estimate and bypasses the estimate on test networks.
- #473 removed the disable transition restored here.

### Follow-up Work

- Review other context-dependent verification errors for peer-scoring exemptions.
- Add ban expiry or score decay to the address book.
- Stop sync rounds in which every fanout request failed from publishing a zero-length sample.
- Filter banned candidates before repeated address-book insertion attempts.


evan-forbes:
closing for now although I do still think this is a problem

ValarDragon (approved):
Nice job!

PR #36, opened 2026-07-08
chore(zakura): standardize operator-facing naming

p0mvn:
## Motivation

Complete the partial Zebra to Zakura rebrand for operator-facing surfaces so new installs, deploys, logs, config files, CI examples, and release artifacts consistently use Zakura names.

No linked issue was provided.

## Solution

- Rename runtime defaults from `zebrad.toml`, `zebrad.log`, and `ZEBRA_*` to `zakura.toml`, `zakura.log`, and `ZAKURA_*`, while keeping targeted `zebrad.toml` and `ZEBRA_*` compatibility fallbacks with deprecation warnings.
- Update deployer templates, systemd units, fleet configs, GitHub workflows, Docker examples, installers, zcashd-compat helpers, and release packaging to use `zakurad` / `zakura` paths by default.
- Keep compatibility aliases where needed: release archives still include `./bin/zebrad`, installer flags accept `--zebrad-path`, and the zcashd-compat deploy workflow refreshes a `/usr/local/bin/zebrad` symlink for existing host services.
- Update Valarbook copy and `CHANGELOG.md` for the operator-facing rename.

PR #38, opened 2026-07-08
fix(zcashd-compat): correct installer for the P2P sidecar and propagate --network

p0mvn:
## Motivation

`scripts/install-zcashd-compat.sh` is shipped to operators as a release asset and run via `curl … | bash`. After the switch from RPC-ingest to a P2P sidecar (#19), the installer's **printed start commands drifted** and still tell operators to launch zcashd with the obsolete RPC-ingest flags. This is a silent failure, verified against the pinned `v0.0.1-compat-alpha.3` zcashd: it has **no** `-zebra-compat*` flags, ignores the unknown args, and then DNS-seeds onto the public network **unpinned** — while the operator believes they are in compat mode. Nothing in CI executes, parses, or lints this script, so the drift shipped.

Rebased onto #36, so all changes use the new `ZAKURA_*` naming.

## Solution

Rework the printed commands to the P2P sidecar contract and fix the related first-run gaps:

- **zcashd P2P pinning** — emit `-connect=<zakura p2p>` + `-listen=0 -dnsseed=0 -listenonion=0 -discover=0`, matching the supervisor's own child args (`zcashd_compat/supervisor.rs`). Removes the obsolete `-zebra-compat` / `-zebra-compat-url` / `-zebra-compat-cookiefile`.
- **Propagate `--network`** — it was only used for datadir sizing, so `--network Testnet` printed a **Mainnet** command. Now emits `ZAKURA_NETWORK__NETWORK`, zcashd `-testnet` / `-regtest -regtestacceptunvalidatedpow`, and the network-correct P2P port. (`zebra-network`'s `Config::default()` hardcodes `[::]:8233` for every network, so testnet/regtest need an explicit `[::]:18233`.)
- **Bootstrap a minimal `zcash.conf`** — zcashd refuses to start without one (`"Before starting zcashd, you need to create a configuration file"`); the installer printed `-conf=<path>` but never created it.
- **chown Docker bind mounts for `docker-split-containers` too** — the mount-prep was guarded to `docker-supervised` only, so the split path's container (`10001:10001`) hit `EACCES`.
- **`ZAKURA_NETWORK__MAX_CONNECTIONS_PER_IP=8` on Docker** — behind `docker-proxy` all connections share one source IP and the default cap is 1, which can starve the sidecar's only slot (documented in `book/src/user/zcashd-compat.md`).
- **Export `ZAKURA_ZCASHD_COMPAT__COOKIE_DIR`** so the cookie lands where the printed command points.
- **sha256-pin the zakurad download** (`ZAKURA_ARCHIVE_SHA256`) — the zebrad archive was fetched with no integrity check; see follow-up on the artifact name.
- **`build-from-source`** — accept a `zcashd/` or `zcash/` source checkout (a plain `git clone` of `valargroup/zcashd` lands in `zcashd/`).
- **Cleanup** — drop the now-pointless `-p 8232` publish and the dead `/zakura-state` readonly mount; `local match` scoping fix.
- **New flags** — `--zakura-p2p-addr HOST:PORT`, `--compat-listen-addr ADDR` (with `--zebra-p2p-addr` accepted as an alias).

## Tests

Environment: 8 vCPU / 31 GiB, snapshots on a 893 GiB disk — zebra testnet state (v28) + zcashd testnet3 (h4129124), operator's own mainnet + testnet nodes running untouched.

- `bash -n scripts/install-zcashd-compat.sh` — passes.
- Dry-run matrix across all five modes × {Mainnet, Testnet} — verified the printed commands: correct `-connect` port per network, `-testnet` only on testnet, `ZAKURA_NETWORK__NETWORK`, no `-zebra-compat*` remaining, `MAX_CONNECTIONS_PER_IP` on Docker.
- Validation gates still fire: disk gate, unsupported mode/network, conf-is-a-directory, conf-is-a-dangling-symlink, `--download-binaries no` with absent binaries, `curl | bash` with no tty.
- **End-to-end (pre-#36 naming, since these prove the mechanism): `split-binary` reached the testnet tip** — Zakura `4153844` and zcashd `4153844` with **identical best-block hash**, and `getpeerinfo` on zcashd showed exactly **one** peer: `127.0.0.1:<zakura-p2p>` `subver=/Zakura:…/Zebra:…/`. zcashd synced the entire chain from the local Zakura node over P2P.
- **`build-from-source` mechanism confirmed** — zcashd peered (`peers=1`) and synced 5,000+ blocks from the local Zakura node before the *local* `roman/ironwood` zcashd build (`v6.20.0-115f1361c`) hit `InvalidChainFound` on a pre-ironwood block. That is a consensus mismatch of that particular local binary vs. the snapshot, not an installer issue; the compat-branch zcashd (`alpha.3`) reaches tip cleanly.

### Follow-up (not in this PR)

1. **`zakurad-*.tar.gz` release artifact does not exist yet** — #36 renamed the downloaded archive from `zebrad-*.tar.gz` (which 404s as `zakurad-*`). `split-binary` / `supervised` downloads will 404 until a release is cut with the new artifact name; `ZAKURA_ARCHIVE_SHA256` is left empty until then. This PR was validated with `--download-binaries no` + `--zakurad-path`.
2. **Supervised managed download is blocked upstream** — the released zebrad embeds a **stale** managed manifest pinning zcashd `v6.2.1-alpha.9` (the pre-P2P RPC-ingest build), whose handshake the current sidecar supervisor rejects. Source at HEAD embeds `alpha.3` correctly; a new zebrad release fixes it. Not installer-fixable.
3. **`ZAKURA_COMPAT_DOCKER_FALLBACK_IMAGE` (`:zcashd-compat-latest`) 404s** in the registry — the docker-supervised fallback can never succeed.
4. **No CI coverage for `scripts/`** — `lint.yml` is Rust-only and its `paths:` filter doesn't match `scripts/**`; adding `bash -n` + shellcheck there would catch this class of drift.

### AI Disclosure

- [x] AI tools were used: Claude Code — installer changes, rebase onto #36, and end-to-end testing against the local snapshots. Findings were verified by driving real Zakura+zcashd nodes to the testnet tip, not by inspection alone.

### DIFF
--- a/scripts/install-zcashd-compat.sh
+++ b/scripts/install-zcashd-compat.sh
@@ -14,6 +14,10 @@ UNITY_ROOT="$(cd "$REPO_ROOT/.." && pwd)"
 ZAKURA_RELEASE_TAG="v0.0.1-alpha.1"
 ZAKURA_ARCHIVE="zakurad-${ZAKURA_RELEASE_TAG}-linux-x86_64.tar.gz"
 ZAKURA_URL="https://github.com/zakura-core/zakura/releases/download/${ZAKURA_RELEASE_TAG}/${ZAKURA_ARCHIVE}"
+# sha256 of ZAKURA_ARCHIVE from the release's SHA256SUMS.txt. Pin it once the
+# zakurad-named release artifact is published: an unpinned download in a
+# `curl | bash` installer is a supply-chain hole. Empty = skip verification.
+ZAKURA_ARCHIVE_SHA256=""
 ZAKURA_MEMBER="./bin/zakurad"
 ZAKURA_DOCKER_IMAGE="valargroup/zakura:0.0.1-alpha.1@sha256:74f76366eed48bdfb15a3386d033a6e3e2d7481f40cb06c5c6ae3c5e9f77e4b5"
 ZAKURA_COMPAT_DOCKER_IMAGE="valargroup/zakura:zcashd-compat-0.0.1-alpha.1@sha256:f3f36dc215a15f3724690529244df8527cc0389ccf4bf9348206cb49388ac8c8"
@@ -36,9 +40,12 @@ ZCASHD_DATADIR="$ZCASHD_DEFAULT_DATADIR"
 INSTALL_DIR="${HOME}/.local/zcashd-compat"
 CACHE_DIR="${HOME}/.cache/zcashd-compat"
 COOKIE_DIR=""
+ZAKURA_P2P_ADDR=""
+COMPAT_LISTEN_ADDR="127.0.0.1:28232"
 ZCASHD_CONF=""
 ZAKURAD_PATH=""
 ZCASHD_PATH=""
+ZCASH_SRC_DIR=""
 ZCASHD_DOCKER_IMAGE=""
 DOWNLOAD_BINARIES=1
 DOWNLOAD_BINARIES_SET=0
@@ -142,6 +149,9 @@ Options:
   --install-dir DIR
   --cache-dir DIR
   --cookie-dir DIR
+  --zakura-p2p-addr HOST:PORT Zakura legacy P2P listener; zcashd is pinned to its port
+                             (default [::]:8233 mainnet, [::]:18233 testnet/regtest)
+  --compat-listen-addr ADDR  Zakura zcashd-compat RPC listener (default 127.0.0.1:28232)
   --zcash-conf FILE
   --zakurad-path PATH
   --zcashd-path PATH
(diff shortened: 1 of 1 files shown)


p0mvn:
## E2E update — new-naming validation against #36

Re-ran against the merged #36 (`ZAKURA_*`) using a clean `zakurad` built from #36 source (embedded manifest verified `v0.0.1-compat-alpha.3`), driving the installer's printed commands verbatim against the local testnet snapshots.

- **New `ZAKURA_*` naming is consumed correctly** — a #36 `zakurad` parsed every printed env var (config dump: `network=Testnet`, P2P `[::]:28250`, compat listener `127.0.0.1:38322`, `zcashd_compat.enabled`, `cookie_dir`).
- **split-binary (new naming): zcashd peers to Zakura and mirrors it block-for-block** — `getpeerinfo` shows exactly one peer `127.0.0.1:28250 subver=/Zakura:8.0.0/Zakura:0.0.1-alpha.2/`, height tracking the Zakura node's tip.
- **supervised now works with a correct binary** — the clean `zakurad` (alpha.3 manifest) **managed-downloaded `v0.0.1-compat-alpha.3`** (the P2P sidecar zcashd) and spawned it as `-testnet -connect=127.0.0.1:<zakura-p2p>` … `-listen=0 …`. This confirms follow-up #2: supervised is blocked only by *released* binaries embedding the stale `alpha.9` manifest — a fresh release fixes it, no installer change needed.
- Reaching the absolute network tip on these fresh nodes is gated on testnet **peer availability** (the throwaway nodes use default seeds and got 2 peers under contention, vs. 23 in an earlier run that reached exact tip `4153844` with matching best-block hash). Not an installer/compat concern.

Note the supervised/split-binary **download** paths still require `--download-binaries no` + `--zakurad-path` here because the `zakurad-*.tar.gz` release artifact 404s (follow-up #1).

PR #42, opened 2026-07-08
fix(zcashd-compat): fix docker-split zcashd arg formatting (spurious empty arg)

p0mvn:
## Motivation

The `docker-split-containers` printed command (added in #38) splices two zcashd flags onto one line, producing a spurious empty argument when an operator copy-pastes it. Found while building a #36 image locally to run docker-split end-to-end.

The zcashd container block used inline `$(while ... printf '  %s \\\n' ...)` command substitutions inside the heredoc. Command substitution strips the trailing newline, so the last emitted flag ends in ` \`; the heredoc's own `\`-continuation then joins it with the next literal line:

```
  -testnet \  -datadir=/home/zcashd/.zcash \
  ...
  -discover=0 \  -printtoconsole
```

A shell tokenizes `-testnet \  -datadir` as `-testnet`, a literal single-space argument (`\ ` = escaped space), then `-datadir`. zcashd then receives an empty parameter and errors. Verified argv before the fix:

```
9:[-testnet]  10:[ ]  11:[-datadir=...]   ...   17:[-discover=0]  18:[ ]  19:[-printtoconsole]
```

## Solution

Move the whole zcashd container flag block into a `print_docker_zcashd_flag_lines` helper (mirroring the working `print_zcashd_flag_lines` used by the binary modes), which emits one flag per line. No command-substitution/heredoc join can occur.

After the fix the pasted command tokenizes to exactly the intended argv, no empty tokens:

```
9:[-testnet] 10:[-datadir=...] 11:[-conf=...] 12:[-connect=127.0.0.1:18233] 13:[-listen=0] ... 17:[-printtoconsole]
```

## Tests

- `bash -n scripts/install-zcashd-compat.sh` — passes.
- Re-tokenized the printed `docker-split-containers` zcashd command through a stub `args()` function; confirmed no spurious `[ ]` tokens remain (Mainnet and Testnet).
- `docker-supervised` and the binary modes were already correct (they don't use the inline-loop pattern) — verified no other occurrences remain.

## Context

Follow-up to #38. Only `docker-split-containers` was affected.

### AI Disclosure

- [x] AI tools were used: Claude Code — found and fixed while setting up a local docker E2E of the zcashd-compat modes.

### DIFF
--- a/scripts/install-zcashd-compat.sh
+++ b/scripts/install-zcashd-compat.sh
@@ -1407,6 +1407,22 @@ print_zcashd_flag_lines() {
   printf '  -printtoconsole\n'
 }
 
+# zcashd container flag block (fixed in-container datadir/conf paths). Emits one
+# flag per line so no command-substitution/heredoc line-join can splice two
+# flags together with an escaped space.
+print_docker_zcashd_flag_lines() {
+  local arg
+  while IFS= read -r arg; do
+    [[ -n "$arg" ]] && printf '  %s \\\n' "$arg"
+  done < <(zcashd_network_args)
+  printf '  -datadir=/home/zcashd/.zcash \\\n'
+  printf '  -conf=/home/zcashd/.zcash/zcash.conf \\\n'
+  while IFS= read -r arg; do
+    [[ -n "$arg" ]] && printf '  %s \\\n' "$arg"
+  done < <(zcashd_p2p_pinning_args)
+  printf '  -printtoconsole\n'
+}
+
 print_split_binary_commands() {
   cat <<EOF
 $(style "$GREEN$BOLD" "Start Zakura in terminal 1:")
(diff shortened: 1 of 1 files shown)


p0mvn:
## Added a second docker-split fix (found via live E2E)

While running docker-split-containers end-to-end against a locally-built #36 image, a second gap surfaced and is now included in this branch:

**zcash.conf is never bootstrapped for docker-split.** zcashd refuses to start without a config file. `split-binary`/`build-from-source` bootstrap one via `ensure_zcashd_conf`, and `docker-supervised` is covered because the in-container Zakura node creates the supervised child's conf. But `docker-split-containers` runs a *standalone* zcashd container whose conf nothing creates — a fresh datadir (or a snapshot restore without a top-level `zcash.conf`) leaves the zcashd container unable to start. Fixed by extending `ensure_zcashd_conf` to `docker-split-containers`, called before `prepare_docker_mounts` so its recursive chown covers the new file.

### Live E2E status
Built a #36 zebra image locally (`docker build --target runtime`, BuildKit) and ran docker-split:
- The zebra container parsed all `ZAKURA_*` env vars correctly (config dump: `network: Testnet`, `listen_addr: [::]:18333`, `zcashd_compat: enabled: true`) — confirming the new naming works on a #36 image.
- The compat RPC published fine; the zcashd container started against the bootstrapped conf and is syncing from the zebra container over P2P.
(Test used non-default ports to avoid colliding with the operator's live nodes on 18233/28232.)

PR #54, opened 2026-07-09
fix(zcashd-compat): always gossip blocks to sidecar peers

p0mvn:
## Motivation

zcashd-compat runs zcashd as a pinned sidecar peer that depends on Zebra for block announcements. Zebra's normal committed-block gossip uses fractional peer sampling, so when Zebra has many ready peers the single zcashd sidecar can be skipped for multiple block inventory broadcasts. In practice that can leave zcashd believing it is at the tip while Zebra has advanced, causing multi-minute stalls until a later inventory announcement happens to reach it.

image

## Solution

- Add `zcashd_compat.block_gossip_peer_ips` for inbound sidecar peer IPs that must always receive block inventory gossip.
- Default the compat block gossip peer list to loopback addresses when zcashd-compat is enabled, while rejecting non-empty values when zcashd-compat is disabled.
- Route `AdvertiseBlock` broadcasts to the normal sampled peer set plus any configured inbound sidecar peers, without switching committed block gossip to advertise-to-all.
- Document the new config and add a changelog entry.

### Tests

- `cargo test -p zebra-network sidecar_peer_always_receives_block_gossip`
- `cargo test -p zebrad deserialize_block_gossip_peer_ips`
- `cargo test -p zebrad block_gossip_peer_ips_require_zcashd_compat`

Full `cargo fmt`, `cargo clippy`, and full `cargo test` were not run locally.

- Adds a regression test with 59 peers showing the configured sidecar receives block gossip while the rest of the network remains fractionally sampled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -422,6 +422,9 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   for the supervisor task before abandoning it. An interrupted shutdown was
   able to silently discard hours of ingested chainstate and force a long
   replay on the next start.
+- Always include configured `zcashd_compat.block_gossip_peer_ips` in block
+  inventory gossip, preventing pinned zcashd-compat peers from stalling when
+  normal fractional peer sampling misses them.
 - Make `zebra-rollback-state` rollback existing v5 databases without replaying
   note commitment trees from genesis for modern rollback targets whose removed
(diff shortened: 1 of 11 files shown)


p0mvn:
@cursor review

p0mvn:
@cursor review

evan-forbes (approved):

PR #56, opened 2026-07-09
fix(sync): mark mined block gossip after broadcast completes

p0mvn:
## Motivation

Closes [ZCA-759](https://linear.app/zcale/issue/ZCA-759/mined-block-advertisement-can-timeout-before-broadcast-completes).

When a miner submits a block via `submitblock`, the gossip task can take the mined-block fast path (`AdvertiseBlockToAll`) instead of the slower committed-tip path (`AdvertiseBlock`). Previously, the task marked the tip hash as already gossiped **immediately after spawning** the broadcast future—not after it completed.

That was unsafe because:

1. The peer set is wrapped in a small `tower::Buffer` (capacity 3) shared by sync, downloads, mempool gossip, and block gossip.
2. The broadcast call is wrapped in a 6s `TIPS_RESPONSE_TIMEOUT` and spawned without being awaited.
3. Under load—or when `broadcast_all` is waiting on unready peers—the spawned future can time out or be dropped before peers receive the block.

Once marked, the committed-tip fallback was suppressed: `ChainTipChange` treated the hash as handled and would not send `AdvertiseBlock` for that tip. The node could end up with **no reliable advertisement** of a locally mined block. This is especially painful in zcashd-compat setups where zcashd is a single pinned peer.

## Solution

Mark the tip hash as gossiped **only after** the mined-block broadcast future returns `Ok` within the timeout. The gossip loop itself stays non-blocking.

### Implementation details

**Mark-on-completion channel**

- Added an internal mpsc channel between the spawned broadcast task and the main gossip loop.
- On successful `AdvertiseBlockToAll`, the spawned task sends the block hash back; on timeout/error, it sends nothing.
- The main loop always applies `mark_last_change_hash` for received marks. A successful all-peers broadcast means peers were already told about that hash, so we do not drop the mark when another mined submission is still queued (or if tip has already moved).

**Non-blocking mark delivery**

- Marks are drained at the top of each loop iteration (`try_recv`).
- The mark receiver is also a third branch in the loop `select!`, so a completed broadcast can update dedup state promptly—even while the committed-tip path is sleeping through `PEER_GOSSIP_DELAY` (7s).

**Preserve dedup across spawned tip monitors**

- Each loop iteration clones `ChainTipChange` for the committed-tip async block.
- Plain `Clone` resets `last_change_hash` to `None` (intended for new monitors), which meant marks on the main monitor were invisible to spawned futures—dedup could not work reliably even before this fix.
- Added `ChainTipChange::clone_for_task()` to preserve `last_change_hash` across long-lived gossip iterations, and switched the gossip task to use it.

**Fallback behavior**

- If the mined broadcast times out (6s) before returning `Ok`, no mark is applied.
- The committed-tip path still runs after `PEER_GOSSIP_DELAY` and sends `AdvertiseBlock` as a partial fallback (~⅓ of ready peers).
- The 6s broadcast timeout finishing before the 7s gossip delay keeps dedup and fallback cleanly separated in the common case.

## Tests

Added integration tests in `zebrad/src/components/sync/tests/gossip.rs`:

- `mined_block_marks_tip_after_successful_broadcast` — after a successful `AdvertiseBlockToAll`, no duplicate `AdvertiseBlock` for the same hash after `PEER_GOSSIP_DELAY`.
- `mined_block_broadcast_timeout_uses_committed_tip_fallback` — if `AdvertiseBlockToAll` is held past `TIPS_RESPONSE_TIMEOUT`, the committed-tip path sends `AdvertiseBlock` as fallback.
- `mined_block_mark_survives_pending_submit_queue` — successful `AdvertiseBlockToAll` still suppresses committed-tip `AdvertiseBlock` when another mined notification is queued while the first broadcast completes.

### DIFF
--- a/zakura-state/src/service/chain_tip.rs
+++ b/zakura-state/src/service/chain_tip.rs
@@ -560,11 +560,6 @@ impl ChainTipChange {
         self.last_change_hash = Some(hash);
     }
 
-    /// Returns the most recent block hash provided by this instance.
-    pub fn last_change_hash(&self) -> Option<block::Hash> {
-        self.last_change_hash
-    }
-
     /// Clone this monitor for another async task in the same long-running service.
     ///
     /// Unlike [`Clone::clone`], this preserves [`Self::last_change_hash`], so callers can
--- a/zakura-state/src/service/chain_tip.rs
+++ b/zakura-state/src/service/chain_tip.rs
@@ -560,6 +560,23 @@ impl ChainTipChange {
         self.last_change_hash = Some(hash);
     }
 
+    /// Returns the most recent block hash provided by this instance.
+    pub fn last_change_hash(&self) -> Option<block::Hash> {
+        self.last_change_hash
+    }
+
+    /// Clone this monitor for another async task in the same long-running service.
+    ///
+    /// Unlike [`Clone::clone`], this preserves [`Self::last_change_hash`], so callers can
+    /// suppress duplicate gossip across multiple spawned futures.
+    pub fn clone_for_task(&self) -> Self {
+        Self {
+            latest_chain_tip: self.latest_chain_tip.clone(),
+            last_change_hash: self.last_change_hash,
+            network: self.network.clone(),
+        }
+    }
+
     /// Return an action based on `block` and the last change we returned.
     fn action(&self, block: ChainTipBlock) -> TipAction {
         // check for an edge case that's dealt with by other code
--- a/zakurad/src/components/sync/gossip.rs
+++ b/zakurad/src/components/sync/gossip.rs
@@ -10,7 +10,7 @@ use tokio::sync::{mpsc, watch};
 use tower::{timeout::Timeout, Service, ServiceExt};
 use tracing::Instrument;
 
-use zakura_chain::{block, chain_tip::ChainTip};
+use zakura_chain::block;
 use zakura_network as zn;
 use zakura_state::ChainTipChange;
 
@@ -40,20 +40,6 @@ pub enum BlockGossipError {
 
     #[error("permanent peer set failure")]
     PeerSetReadiness(zn::BoxError),
-}
-
-/// Mark the chain tip hash as gossiped after a successful mined block broadcast.
-///
-/// This suppresses the committed tip gossip path for the same hash, but only
-/// after the all-peers broadcast completes successfully.
-fn apply_mined_block_mark(
-    chain_state: &mut ChainTipChange,
-    mined_block_channel_empty: bool,
-    hash: block::Hash,
-) {
-    if mined_block_channel_empty && chain_state.latest_chain_tip().best_tip_hash() == Some(hash) {
-        chain_state.mark_last_change_hash(hash);
-    }
 }
 
 /// Run continuously, gossiping newly verified [`block::Hash`]es to peers.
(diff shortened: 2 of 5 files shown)


p0mvn:
@cursor review

zakurad/src/components/sync/gossip.rs
```diff
@@ -63,10 +85,23 @@ where
...
     loop {
+        while let Ok(hash) = mined_block_mark_receiver.try_recv() {
+            apply_mined_block_mark(
+                &mut chain_state,
+                mined_block_receiver
+                    .as_ref()
+                    .is_none_or(mpsc::Receiver::is_empty),
+                hash,
+            );
+        }
```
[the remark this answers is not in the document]
p0mvn: [2455fe9](https://github.com/zakura-core/zakura/pull/56/commits/2455fe97b58de182d3c5d2ad70e999aa2364a97e)

p0mvn:
@cursor review

evan-forbes (approved):

zakurad/src/components/sync/gossip.rs
```diff
@@ -21,6 +21,14 @@ use crate::{
 
 use BlockGossipError::*;
 
+/// How many completed mined block broadcasts can wait to mark the chain tip.
+/// In normal operations, we expect at most 1 pending mark.
+/// The main loop can be busy for several seconds in the committed-tip path.
+/// During that window, multiple mined-block broadcasts could finish and
+/// try to send marks. A capacity of 16 with 25-75s block times
+/// is chosen arbitrarily high to be safe.
+const MINED_BLOCK_MARK_CHANNEL_CAPACITY: usize = 16;
```
ValarDragon, line 30: Ah this is per peer
```suggestion
/// We create a bufferred channel per peer, to broadcast our knowledge of latest chain tip.
/// In normal operations, we expect at most 1 pending block hash to mark the chain tip.
/// This sets the capacity of that buffer.
/// The main loop can be busy for several seconds in the committed-tip path.
/// During that window, multiple mined-block broadcasts could finish and
/// try to send marks. A capacity of 16 with 25-75s block times
/// is chosen arbitrarily high to be safe. The buffered data is small, so we don't worry about this.
const MINED_BLOCK_MARK_CHANNEL_CAPACITY: usize = 16;
```

zakura-state/src/service/chain_tip.rs
```diff
@@ -560,6 +560,23 @@ impl ChainTipChange {
         self.last_change_hash = Some(hash);
     }
 
+    /// Returns the most recent block hash provided by this instance.
+    pub fn last_change_hash(&self) -> Option<block::Hash> {
+        self.last_change_hash
+    }
```
ValarDragon: Not sure I follow the point of this fn, its not used right?

ValarDragon: (fixed in my commit)

ValarDragon:
Lint is failing because Cargo is down, not due to this PR. (Its down for everyone)

ValarDragon (approved):
Nice job! This bug fix is a big deal

It'd explain a host of problems we've seen on testnet. (We've found bugs that would amplify an initial issue, but not explain the initial bug. This is a good explanation of the initial bug that would cause a rapidly advanced miner to stay ahead)

PR #64, opened 2026-07-10
fix(mempool): drain pending gossip txids from mempool on channel lag

p0mvn:
## Motivation

The mempool change broadcast channel is lossy. When its gossip receiver
lagged, unread `Added` events were discarded and their transaction IDs were
never advertised to peers. The loss was permanent because the channel payload
was the only source of IDs awaiting gossip.

Tracked as ZCA-760 and reported by OtterSec.

## Solution

- Keep an authoritative set of verified transaction IDs awaiting gossip in
  the mempool service. Add IDs before sending the channel notification.
- Treat `Added` events and `Lagged` errors as wakeups. The gossip task drains a
  bounded batch through
  `TakePendingGossipTransactionIds { limit: MAX_TX_INV_IN_SENT_MESSAGE }`.
- After a full batch, wait for `PEER_GOSSIP_DELAY` and drain again without
  requiring another wakeup. This preserves the existing outbound rate bound
  while guaranteeing progress through a backlog.
- Remove mined, invalidated, expired, and ZIP-401-evicted transactions,
  including eviction dependents, from the pending set before they can be
  advertised.
- Preserve the existing public `Storage::insert` signature; only a narrowly
  visible internal helper returns eviction IDs to the mempool service.
- Replace the pruned-block test's process-global tracing subscriber with
  test-local capture propagated into the buffered inbound service. This fixes
  the full-suite subscriber race exposed after rebasing.

The change adds lag and pending-drain metrics and raises the mempool change
channel capacity from 20 to 40 to absorb short bursts without making the
channel a correctness boundary.

## Testing

- `npx --yes markdownlint-cli changelog.d/64.md --config .trunk/configs/.markdownlint.yaml`
- `cargo clippy -p zakura -p zakura-node-services --all-targets -- -D warnings`
- `cargo test -p zakura --lib` — 309 passed
- `cargo test -p zakura-node-services` — 1 passed; doc tests passed

The new tests cover ordinary drains, lag recovery, bounded multi-cycle
recovery without another wakeup, empty drains, mined/invalidated pruning,
ZIP-401 eviction and dependent pruning, zero-limit and bounded requests, and
the disabled mempool response.

- [ZIP-401: Addressing Mempool Denial-of-Service](https://zips.z.cash/zip-0401)
- [Tokio broadcast lagging semantics](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html#lagging)

### Follow-up Work

The indexer subscriber still consumes full payloads from the lossy broadcast
channel. Giving it an independent recovery path remains out of scope.

### AI Disclosure

- [x] AI tools were used: Codex was used for the rebase, conflict resolution,
  implementation review, test-fix correction, and verification. The final
  changes were reviewed and tested as described above.


ValarDragon:
Taking over this PR

zakurad/src/components/mempool/storage.rs
```diff
@@ -427,7 +443,7 @@ impl Storage {
                 "returning cached error for transaction",
             );
 
-            return Err(error);
+            return (Err(error), HashSet::new());
```
ValarDragon, line 446: FYI for reader, HashSet::new() is allocation-free!

ValarDragon:
@v12 pls audit

ValarDragon:
@v12-auditor audit this

zakurad/src/components/mempool/gossip.rs
```diff
@@ -22,93 +28,114 @@ use crate::{
...
     // use the same timeout as tips requests,
     // so broadcasts don't delay the syncer too long
     let mut broadcast_network = Timeout::new(broadcast_network, TIPS_RESPONSE_TIMEOUT);
+    let mut drain_pending_without_wakeup = false;
 
     loop {
-        let mut combined_changes = 1;
+        // This count is only used in logs. It is zero for a pending-set drain
+        // that runs without consuming a channel notification.
+        let mut combined_changes = 0;
```
ValarDragon, line 66: IMO we delete / re-design this in a follow-up PR

ValarDragon (approved):
Spent awhile reviewing, plus sending many 5.6 runs at this

I think its good to go, the gossip loop likes very altered but is quite similar still

PR #74, opened 2026-07-10
feat(release): maintainer minisign signatures and CI provenance attestation

ebfull:
## Motivation

Zakura releases currently ship a `SHA256SUMS.txt`, but nothing signs it — a user has no way to check that the artifacts they downloaded are the ones I signed off on, and no way to check they came out of this repo's CI unmodified. I want a public, documented verification process for releases.

Separately, `SHA256SUMS.txt` currently lists `x86_64.sha256`, a build intermediate that is never uploaded as a release asset, so a plain `sha256sum -c` errors on a missing file.

## Solution

Two independent verification layers, plus the checksums fix:

- **Maintainer signature (endorsement).** After reviewing a release I sign its `SHA256SUMS.txt` with a minisign key held on a trusted machine, and upload the detached signature as a release asset. `scripts/sign-release.sh` (or `make sign-release TAG=v1.0.0`) drives this: it cross-checks every uploaded asset against the digests GitHub reports (refusing to sign on mismatch or unlisted assets), signs with a trusted comment naming the tag (so a signature can't be replayed on a different release), verifies the result against the published public key before uploading, then uploads `SHA256SUMS.txt.minisig`. The secret key never touches CI — a compromised runner can't forge an endorsement.
- **Build provenance (CI attestation).** The publish job now attests all release assets with `actions/attest-build-provenance`, so anyone can run `gh attestation verify <asset> --repo zakura-core/zakura` to confirm the asset was built by this repo's release workflow at the tagged commit.
- **`VERIFY.md`** documents the public key and the consumer-facing steps: `minisign -Vm SHA256SUMS.txt -P <key>`, then `sha256sum -c --ignore-missing SHA256SUMS.txt`. Intended to be linked from release notes and the website.
- **Checksums fix:** the `x86_64.sha256` intermediate (used to pin the installer) is deleted before `SHA256SUMS.txt` is generated, so the signed file only lists real assets.

Pre-releases may go unsigned; the absence of a `.minisig` marks a release as not yet endorsed. Docker images are not signed yet (follow-up).

### Tests

- Dry-ran the script's digest cross-check against the live `v1.0.0-rc0` release: all four assets match GitHub's reported digests, and the phantom `x86_64.sha256` entry is correctly flagged as a warning.
- Workflow YAML validated locally; the attestation step is additive and SHA-pinned (`v4.1.1`) like the other actions. Will confirm attestation and the corrected `SHA256SUMS.txt` on the next tag push.
- Planning to exercise the full sign-and-verify loop on `v1.0.0-rc0` as a live test.

### AI Disclosure

- [x] AI tools were used: Claude Code, for drafting the signing script, workflow changes, and verification docs.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -191,6 +191,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      id-token: write
+      attestations: write
     if: github.event_name == 'release' || startsWith(github.ref, 'refs/tags/v') || inputs.publish_assets_to_release
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
@@ -285,10 +287,22 @@ jobs:
           PY
           (
             cd ./publish/final
-            rm -f SHA256SUMS.txt
+            # x86_64.sha256 is an intermediate used to pin the installer above;
+            # it is not uploaded to the release, so keep it out of the signed
+            # checksums file.
+            rm -f SHA256SUMS.txt x86_64.sha256
             sha256sum ./* > ../SHA256SUMS.txt
             mv ../SHA256SUMS.txt SHA256SUMS.txt
           )
+
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@0f67c3f4856b2e3261c31976d6725780e5e4c373 #v4.1.1
+        with:
+          subject-path: |
+            ./publish/final/zakurad-*.tar.gz
+            ./publish/final/zakurad-manifest-*.json
+            ./publish/final/install-zakura.sh
+            ./publish/final/SHA256SUMS.txt
 
       - name: Ensure release exists
         env:
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 include make/zcashd-compat.mk
 include make/perf.mk
 include make/zakura-dev.mk
+include make/release.mk
 
 help:
 	@echo "Available targets:"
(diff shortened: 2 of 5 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -285,11 +287,23 @@ jobs:
...
-            rm -f SHA256SUMS.txt
+            # x86_64.sha256 is an intermediate used to pin the installer above;
+            # it is not uploaded to the release, so keep it out of the signed
+            # checksums file.
+            rm -f SHA256SUMS.txt x86_64.sha256
             sha256sum ./* > ../SHA256SUMS.txt
             mv ../SHA256SUMS.txt SHA256SUMS.txt
           )
 
+      - name: Attest build provenance
```
p0mvn, line 298: Non-blocking nit on republishing:

Stale signatures survive release repairs. 

If the release workflow is rerun:

- It replaces the binaries and SHA256SUMS.txt.
- It does not replace or remove SHA256SUMS.txt.minisig.
- The old signature remains attached but no longer validates the new checksums.

Users therefore see a signature asset and assume the release is endorsed, but verification fails. Removing the old signature immediately before replacing the assets makes the release correctly appear unsigned until the maintainer signs it again.

Consider adding this immediately before `gh release upload`, after building/attesting succeeds:
```bash
signature_asset_id="$(
  gh api "repos/${REPOSITORY}/releases/tags/${RELEASE_TAG}" \
    --jq '.assets[] | select(.name == "SHA256SUMS.txt.minisig") | .id'
)"
if [ -n "$signature_asset_id" ]; then
  gh api --method DELETE \
    "repos/${REPOSITORY}/releases/assets/${signature_asset_id}"
fi
```

p0mvn (approved):
LGTM. 

We can later add signature verification to the installer as well

PR #76, opened 2026-07-10
fix(zcashd-compat): keep sidecar connected and synced

p0mvn:
## Summary

This fixes two distinct production failure modes that caused the managed `zcashd` sidecar to stall behind Zakura.

### 1. Sidecar identity mismatch and block-notification starvation

Linux dual-stack listeners can report an IPv4 localhost connection as the IPv4-mapped IPv6 address `::ffff:127.0.0.1`, while `zcashd_compat.block_gossip_peer_ips` is normally configured as native IPv4 `127.0.0.1`. Direct equality treated these representations as different peers.

As a result, Zakura treated `zcashd` as an ordinary inbound peer. It was only selected by fractional random block-gossip sampling instead of receiving every committed block advertisement. Missed advertisements produced long no-progress windows followed by burst catch-up when a later advertisement finally reached the sidecar.

The fix canonicalizes only the operands used for sidecar identity checks. It deliberately leaves the accepted socket address unchanged, preserving existing peer keys, ban behavior, per-IP accounting, logs, and handshake metadata. Regression coverage verifies that a mapped localhost connection matches native IPv4 configuration and is always included in block gossip.

### 2. Sidecar admission starvation at the inbound limit

A restart exposed a separate admission bug. `zcashd` can spend several minutes loading its block index, while Zakura immediately accepts public peers. On the production node, all 100 inbound slots were occupied before `zcashd` completed startup. Every localhost connection was then reset before handshake, leaving `zcashd` with zero peers.

The listener now reserves one existing inbound slot when sidecar IPs are configured:

- ordinary peers can use only the public portion of the existing limit;
- one canonically matched sidecar can use the reserved slot;
- sidecar reconnects bypass the ordinary recent-IP throttle;
- bans are still checked first;
- the accepted address remains unchanged; and
- total inbound connections never exceed the configured limit.

The listener regression test fills public capacity, configures the sidecar with an alternate native/mapped representation, and verifies one public connection plus one sidecar connection are admitted while additional connections are rejected.

### Deployment cleanup

Remove the obsolete single-host `deploy-zcashd-compat.yml` workflow and its documentation references; the mainnet fleet is managed by `zakura-mainnet-deploy.yml` from the us-east0 deployer.

Drive-by: update the mainnet status dashboard to read `/root/logs/zakura-tracing.log` instead of the obsolete `/root/logs/zebrad.log` path.

## Production evidence

Before the fixes:

- `zcashd` block-delivery delays reached approximately 29 minutes;
- after restart, the listener had 100/100 inbound slots occupied;
- `zcashd` repeatedly received TCP resets and reported zero peer connections; and
- Zakura continued committing and advertising blocks while `zcashd` produced no `UpdateTip` events.

After deploying the corrected local build:

- all 100 inbound slots remained occupied;
- `zcashd` connected through its reserved slot;
- `getconnectioncount` returned exactly one pinned Zebra peer; and
- Zakura and `zcashd` both reached height 3,407,718 with drift 0.

## Test plan

- [x] `cargo test -p zebra-network listener_reserves_one_zcashd_compat_inbound_slot --locked`
- [x] `cargo test -p zebra-network sidecar_peer_always_receives_block_gossip --locked`
- [x] `cargo build --release --locked -p zebrad`
- [x] Deploy the locally built binary and restart `zebrad-compat.service`
- [x] Validate the sidecar stays connected with the inbound listener at capacity
- [x] Validate Zakura/zcashd height drift returns to zero
- [x] Check modified files for IDE diagnostics and run `git diff --check`


.github/workflows/deploy-zcashd-compat.yml
```diff
@@ -1,190 +0,0 @@
-name: Deploy zcashd-compat
```
p0mvn, line 1: Note to reviewer: removed as drive-by. This is obsolete

.github/workflows/zakura-mainnet-deploy.yml
```diff
@@ -170,7 +170,7 @@ jobs:
           # Read-only fields below: consumed by the status dashboard's SSH probe,
           # not written to any node (manage_config = false).
           rpc_listen_addr = "0.0.0.0:8232"
-          log_file        = "/root/logs/zebrad.log"
+          log_file        = "/root/logs/zakura-tracing.log"
```
p0mvn, line 173: Note to reviewer: drive-by

zebra-network/src/peer_set/initialize.rs
```diff
@@ -731,6 +733,18 @@ where
         config.peerset_inbound_connection_limit(),
         "Inbound Connections",
     );
+    let mut active_zcashd_compat_connections =
+        ActiveConnectionCounter::new_counter_with(1, "zcashd-compat Inbound Connection");
```
p0mvn, line 737: Note to reviewer (see full file diff)

Fix 2: reserve one p2p connection slot for zcashd compat. Due to zcashd taking over 7 minutes to start, all p2p spots were getting saturated by other peers. As a result, zcashd was not able to connect and got stalled.

scripts/install-zakura.sh
```diff
@@ -11,12 +11,12 @@ fi
 REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
 UNITY_ROOT="$(cd "$REPO_ROOT/.." && pwd)"
 
+ZAKURA_RELEASES_API="https://api.github.com/repos/zakura-core/zakura/releases?per_page=100"
```
p0mvn: Note to reviewer: the diff in this file avoids dependency on the release having been published.

So that the installer can continue to work against the latest pre-release, not published released.

p0mvn:
@cursor review

evan-forbes (approved):
lgtm makes a lot of sense to try and always keep it connected.

quickly queried ai for other evictions:

>   The one policy-driven eviction that could affect it is the FindBlocks/FindHeaders stall detector at zebra-network/src/
  peer_set/set.rs:785. Because the sidecar currently participates in ordinary P2C routing, Zebra could select it for those
  requests, classify it as stalled, and call self.remove(&addr) at line 798.

p0mvn:
> lgtm makes a lot of sense to try and always keep it connected.
> 
> quickly queried ai for other evictions:
> 
> > The one policy-driven eviction that could affect it is the FindBlocks/FindHeaders stall detector at zebra-network/src/
> > peer_set/set.rs:785. Because the sidecar currently participates in ordinary P2C routing, Zebra could select it for those
> > requests, classify it as stalled, and call self.remove(&addr) at line 798.

Addressed

PR #94, opened 2026-07-11
test(rpc): wait for getpeerinfo peer connection

p0mvn:
## Motivation

The `get_peer_info` acceptance test queried `getpeerinfo` as soon as the RPC port opened. Peer discovery is asynchronous, so slower or contended CI runners could return an empty peer list even though the node was still connecting normally.

## Solution

Poll `getpeerinfo` once per second until a peer is available, with a two-minute timeout and contextual timeout error. This matches the asynchronous peer startup behavior while retaining a bounded failure.

### Tests

- `cargo test -p zakura --test acceptance get_peer_info -- --exact --nocapture`
- IDE diagnostics for the changed file
- `git diff --check`

Full workspace clippy and tests were not run. This is a low-risk, isolated acceptance-test timing fix, so targeted verification was used under the repository's risk-based policy.

- Restores the upstream Zebra test's bounded polling behavior after the Zakura rename migration.

### Follow-up Work

None.

### AI Disclosure

- [x] AI tools were used: Cursor was used to diagnose the CI failure, restore the bounded polling behavior, and draft this PR description.

### PR Checklist

- [x] The solution is tested.
- [x] The documentation and changelogs are up to date (no user-visible change).


zakurad/tests/common/get_block_template_rpcs/get_peer_info.rs
```diff
@@ -38,16 +40,26 @@ pub(crate) async fn run() -> Result<()> {
 
     tracing::info!(?rpc_address, "zakurad opened its RPC port",);
 
-    // call `getpeerinfo` RPC method
-    let peer_info_result: Vec<PeerInfo> = RpcRequestClient::new(rpc_address)
-        .json_result_from_call("getpeerinfo", "[]")
-        .await
-        .map_err(|err| eyre!(err))?;
+    let peer_info = tokio::time::timeout(Duration::from_secs(2 * 60), async {
```
evan-forbes: if this is taking up to 2 minutes, even in CI, then that is a bug we should fix

evan-forbes (approved):

zakurad/tests/common/get_block_template_rpcs/get_peer_info.rs
evan-forbes: https://linear.app/zcale/issue/ZCA-773/reduce-2-minute-ci-wait-for-peer-discovery

ValarDragon: I'd almost rather keep this as 10s, and see if it still fails. If so, then investigate further.

I wonder if two minutes was just put as a default here

p0mvn: Made it not touch the network and be local

[7b59cb6](https://github.com/zakura-core/zakura/pull/94/commits/7b59cb68b9d690aa6731172c53e1f841703b6721)

PR #97, opened 2026-07-11
fix(state): avoid VCT successor body deadlock

evan-forbes:
## Motivation

Archive checkpoint sync can permanently stall at a hard-coded checkpoint boundary when the VCT finalized writer reaches height H before H+1 has crossed the next checkpoint verifier range.

The writer previously required the H+1 block body as the one-block VCT authentication witness. At a checkpoint boundary, H+1 cannot enter the finalized-body channel until the following 400-block range verifies, creating a circular wait that can saturate all 401 apply slots.

## Solution

- Represent a VCT successor witness using its header, explicit height, hash, and ZIP-244 auth-data root rather than requiring an entire block body.
- Prefer the buffered H+1 body when available, then fall back to the contextually validated Zakura header store and its stored auth-data root.
- Require the stored witness to be exactly H+1 and to link directly to H before using its header commitment.
- Route a genuinely missing witness through the existing VCT stall tracker, with one initial warning, trace-level polling, and error-level escalation after 30 seconds.
- Document the user-visible fix in the root and `zakura-state` changelogs.

### Tests

- `CXX=clang++ CXXFLAGS="-include cstdint" cargo check -p zakura-state --lib`
- `CXX=clang++ CXXFLAGS="-include cstdint" cargo clippy -p zakura-state --all-targets -- -D warnings`
- `CXX=clang++ CXXFLAGS="-include cstdint" cargo test -p zakura-state --lib` (342 passed, 2 ignored)
- Focused regression coverage confirms:
  - H roots are authenticated by a real H+1 header-only witness.
  - The stored H+1 witness is available while its body remains absent.
  - VCT look-ahead and tracked retry behavior remain correct.

The explicit C++ flags work around vendored RocksDB 8.10 headers failing to include `<cstdint>` under the host's GCC 15 / Clang 22 toolchains; no RocksDB source was changed.

This implements the VCT one-block authentication rule without coupling it to checkpoint body-range completion.

### AI Disclosure

- [x] AI tools were used: OpenAI Codex investigated the trace/code path and implemented the Rust changes, regression tests, changelog entries, validation, and PR description under human direction.

### PR Checklist

- [x] This change was discussed with the team before opening the draft PR.
- [x] The solution is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
+- Prevent verified-commitment-trees checkpoint sync from stalling at checkpoint
+  boundaries by authenticating a block's supplied roots with its already-validated
+  successor header, without waiting for the successor block body to finish checkpoint
+  verification.
 - Handle the `invalidateblock` and `reconsiderblock` state-control RPC edge
   cases (invalidating a non-finalized chain root, invalidating same-height
   sibling fork tips, and repeated reconsideration of the same block) without
--- a/zakura-state/CHANGELOG.md
+++ b/zakura-state/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- VCT checkpoint writes now use the validated header store for the H+1 root witness,
+  avoiding a circular wait for the next checkpoint range's block bodies.
 - `read::history_tree` now returns the finalized history tree only when the
   requested hash or height exactly matches the finalized tip.
 
--- a/zakura-state/src/service/finalized_state.rs
+++ b/zakura-state/src/service/finalized_state.rs
@@ -740,12 +740,11 @@ impl FinalizedState {
                         // write worker only buffers direct successors, so this should
                         // never fire.
                         let next_vct_block = next_vct_block.filter(|next_vct_block| {
-                            let links =
-                                next_vct_block.block.header.previous_block_hash == block_hash;
+                            let links = next_vct_block.header.previous_block_hash == block_hash;
                             if !links {
                                 tracing::warn!(
                                     ?height,
-                                    witness_parent = ?next_vct_block.block.header.previous_block_hash,
+                                    witness_parent = ?next_vct_block.header.previous_block_hash,
                                     expected_parent = ?block_hash,
                                     "VCT: ignoring a successor witness that does not link \
                                      to the block being committed"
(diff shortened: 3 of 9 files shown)


evan-forbes:
doing some ai audits otherwise lgtm

evan-forbes:
we probably don't need to be so defensive either 

> Route a genuinely missing witness through the existing VCT stall tracker, with one initial warning, trace-level polling, and error-level escalation after 30 seconds.

evan-forbes:
> we probably don't need to be so defensive either

one follow up change I'd be interested in after discussing with @p0mvn would be to simply always rely on having the H+1 header and removing a lot of stuff. Happy to do here in this PR or in a flup

I think that is likely the cleaner final design.

It would let us remove most of the VCT body-lookahead machinery:

  - lookahead: VecDeque<_>;
  - fill_successor;
  - non-linking body-discard routing;
  - body-derived witness construction;
  - retry-before-lookahead ordering;
  - two different H+1 witness sources.

 The writer would become approximately:

```rust
let witness = header_store.successor_witness(H, hash_H);

if witness.is_none() && H_requires_successor {
    tracked_retry(H);
    continue;
}

commit(H, witness);
```

ValarDragon (approved):
LGTM, this seems to solve a real stall risk to me

evan-forbes:
for reviewers, the main issue is outside of this code in 

> // ... we must 
 // commit blocks transactionally on a per-checkpoint basis. Otherwise, 
 // the checkpoint verifier's state could desync from the underlying 
 // state service
 
 ```
  For this trace:

  H                             = 2,176,698
  next checkpoint               = 2,177,098
  next range                    = H+1 ... H+400
  minimum simultaneous applies  = H + those 400 = 401

  Therefore 401 was not generous headroom. It was the exact theoretical minimum. Scheduling, duplicate/redelivery activity,
  or any apply occupying a slot at the wrong time could prevent the next range from reaching the state writer.

  More fundamentally, correctness had accidentally become dependent on downloader concurrency:

  state commit H
  requires apply capacity for the complete next checkpoint range

  That is the coupling the PR removes.
  
```

the claim is that when commit times are long, we're more likely to hit the downloader concurrency blocking the state, albiet still rare enough not to occur all the time.
 

https://github.com/zakura-core/zakura/blob/b2994c3c1f177b4941e4d98ef33124b09f328e75/zakura-consensus/src/checkpoint.rs#L1112-L1170

p0mvn (approved):

p0mvn:
> one follow up change I'd be interested in after discussing with @p0mvn would be to simply always rely on having the H+1 header and removing a lot of stuff. Happy to do here in this PR or in a flup

That sounds like the right direction. Had a similar change on a branch behind the "header sync roots validation" but did not get it in because we weren't seeing these vct stalls until recently.

Propose to generally go off of observability before the release (Monday) and refactor only what we evidently see as a problem to limit the risk of accidental breaks. After that, this sgtm.

PR #95, opened 2026-07-11
chore(installer): target v1.0.0-rc1

p0mvn:
## Motivation

Keep the standalone installer aligned with the published Zakura v1.0.0-rc1 release.

## Solution

- update the Zakura release and Docker image tags from rc0 to rc1
- pin the published SHA-256 checksum for the rc1 Linux x86_64 archive
- retain the existing zcashd v1.0.0-compat-rc2 release configuration

### Tests

- `bash -n scripts/install-zakura.sh`
- IDE diagnostics: no errors

This is a low-risk installer metadata update, so workspace-wide Rust tests and lints were not run.

### Specifications & References

- https://github.com/zakura-core/zakura/releases/tag/v1.0.0-rc1
- https://github.com/valargroup/zcashd/releases/tag/v1.0.0-compat-rc2

### Follow-up Work

None.

### AI Disclosure

- [x] AI tools were used: Cursor (GPT-5.6 Sol) updated the release metadata, verified published checksums, and drafted this PR.

### PR Checklist

- [x] The PR title follows [conventional commits](https://www.conventionalcommits.org/) format: `type(scope): description`
- [x] The PR follows the [contribution guidelines](https://github.com/ZcashFoundation/zebra/blob/main/CONTRIBUTING.md).
- [x] This change was discussed in an issue or with the team beforehand.
- [x] The solution is tested.
- [x] The documentation and changelogs are up to date.

PR #81, opened 2026-07-10
fix(release): report Zakura version and source revision

p0mvn:
## Motivation

The published daemon and container still reported Zebra's stale `5.0.0-rc.3` identity and did not consistently expose the source commit, making deployed revisions difficult to identify.

## Solution

- report `zakurad 1.0.0-rc0` with Git commit build metadata
- embed source revisions in release builds and add OCI version, revision, source, and title labels
- rename the Cargo packages to `zakura` and `zakura-watchdog`, including watchdog runtime and systemd identities
- update build commands, operator documentation, and the changelog

### Tests

- `cargo fmt --all -- --check`
- `cargo run --locked --package zakura --bin zakurad -- --version`
- `cargo test --locked -p zakura-watchdog` (18 passed)
- `cargo run --locked -p zakura-watchdog -- --version`
- `actionlint` on modified workflows
- `markdownlint` on modified Markdown files
- IDE diagnostics on modified source and workflow files

Full workspace Clippy and tests were not run because this is a low-risk metadata, naming, and CI wiring change with no consensus or validation-critical logic changes.

### Specifications & References

No linked issue; this change was requested directly by the maintainer.

### Follow-up Work

None.

### AI Disclosure

- [ ] No AI tools were used in this PR
- [x] AI tools were used: Cursor Agent implemented the version, provenance, package naming, workflow, documentation, and PR-description changes.

### PR Checklist

- [x] The PR title follows conventional commits format.
- [x] The solution is tested proportionally to its low-risk classification.
- [x] The documentation and changelog are up to date.

PR #98, opened 2026-07-11
ci(release): gate release builds on tag matching the zakura package version

ebfull:
## Motivation

v1.0.0-rc1 was tagged without bumping the `zakura` package version, so its binaries self-report `zakurad --version` = `1.0.0-rc0` and advertise `/Zakura:1.0.0-rc0/` — the same class of mismatch #81 fixed for rc0. Release binaries are built in Docker without `.git`, so the binary falls back to `CARGO_PKG_VERSION`, and nothing in the release pipeline checked that the tag matched it. This PR makes that mismatch impossible to ship again. (No pre-existing issue; this addresses the rc1 identity regression found during release prep.)

## Solution

- New `check-release-version` job in `release-binaries.yml`, gating all build jobs (and transitively every publish job): it fails fast if the tag (minus `v`) doesn't equal the `zakura` package version, so a mismatched tag builds and publishes nothing.
- The "Validate binary" step now also asserts the built binary's self-reported version (build metadata stripped) equals the tag — belt-and-suspenders against any stamping divergence.
- New reusable script `scripts/check-release-version.sh <tag>` used by CI and by the pre-tag runbook step.
- New `skip_version_check` workflow_dispatch input so asset-repair runs for releases tagged before this gate (rc0/rc1) aren't blocked. Tag pushes can't set inputs, so new tags are always checked.
- Release checklist: the "Update Zakura Version" section now has mandatory checkboxes to bump the `zakura` package and run the check before tagging (rc1 happened because the section explained release levels but never said to bump). Matching notes added at the tag-creation step and in `book/src/dev/release-process.md`. The hotfix checklist delegates to the same section, so it inherits the steps.

Deliberately out of scope: this gate checks only the node package.

### Tests

- `scripts/check-release-version.sh v1.0.0-rc0` passes against the current tree (package is `1.0.0-rc0`); `scripts/check-release-version.sh v1.0.0-rc1` fails with a clear remediation message — i.e. the script reproduces the exact rc1 mismatch.
- The binary-version parsing in "Validate binary" was exercised standalone with a simulated `zakurad 1.0.0-rc0+g05a8794c1234` output against both matching and mismatching tags.
- Workflow YAML validated with a YAML parser. The gate job follows the existing hygiene patterns (SHA-pinned actions, `persist-credentials: false`, tag passed via env, not template interpolation).
- Not tested: a live tag-push run of the workflow (requires pushing a `v*` tag, which publishes assets). The `skip_version_check` repair path can be verified with a `workflow_dispatch` run for `v1.0.0-rc1` once merged.

- rc1 mismatch documented in the release-prep survey (SUMMARY.md §6.2 / CRATE_SUMMARY.md §4, not committed)
- #81 (original version-identity fix for rc0), #95 (installer pinned to rc1)

### Follow-up Work

- Bump the `zakura` package before cutting the next tag — this PR gates, it doesn't bump.
- Version unification of the 11 fork-owned crates at the release version, plus the one-line `configured_user_agent` dedup (`zakura-network/src/peer/handshake.rs`).

### AI Disclosure

- [x] AI tools were used: Claude Code wrote the workflow gate, check script, and checklist/docs updates; reviewed by the author.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -99,6 +99,22 @@ Choose a release level for `zakurad`. Release levels are based on user-visible c
 - significant new features or behaviour changes; changes to RPCs, command-line, or configs; and deprecations or removals are `minor` releases
 - otherwise, it is a `patch` release
 
+**This step is mandatory for every release, including release candidates.**
+Release binaries are built without `.git`, so `zakurad --version` reports the
+`zakura` package version, not the tag — v1.0.0-rc1 was tagged without this bump
+and its binaries self-report `1.0.0-rc0`. The `release-binaries.yml` workflow
+refuses to build or publish assets for a tag that does not match the package
+version.
+
+- [ ] Bump the `zakura` package version to the release version:
+
+```sh
+cargo release version --verbose --execute --allow-branch '*' -p zakura patch # [ major | minor ]
+```
+
+- [ ] On the release commit, check that the version matches the tag you are
+      about to create: `./scripts/check-release-version.sh v<version>`
+
 ## Update Crate Versions and Crate Change Logs
 
 If you're publishing crates for the first time, [log in to crates.io](https://github.com/zakura-core/zakura/dev/crate-owners.html#logging-in-to-cratesio),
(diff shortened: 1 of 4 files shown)


ValarDragon:
@cursor review

evan-forbes (approved):

p0mvn (approved):

PR #99, opened 2026-07-11
fix(installer): repair default profile, low-spec flag, mode ambiguity, and arch handling

evan-forbes:
## Motivation

A user ran the documented one-liner on a fresh Ubuntu 24.04 box (`v1.0.0-rc1`) and reported six issues. This PR covers the installer-side ones. The headline problem is that **the first option in the interactive menu fails outright for any non-root user** — which is the default login on essentially every cloud AMI.

## Solution

**Critical — default profile hardcoded a path only root can create.** `ZAKURA_STANDALONE_STATE_DIR` was `/mnt/data/zakura-state`. `/mnt` is `root:root 755` on a stock image, so uid 1000 fails validation before downloading anything:

```
- zakura state directory path /mnt/data/zakura-state cannot be created:
  nearest existing ancestor /mnt is not writable by the current user
```

The zcashd-compat profile already had a capacity-aware directory search that falls back to `~/.cache/zakura`; the default profile simply never called it. It now does, sized with the standalone disk floor.

That search checked *capacity* but never *writability*, so it could still hand back a large root-owned volume. Fixing that was a prerequisite, not a bonus: without it, routing the default profile through the search would have reproduced the identical failure on any box with a big root-owned `/mnt/data`. Added a `path_is_creatable` predicate and filtered candidates and install roots through it.

**High — `--unsafe-low-specs` never reached the generated command.** It only downgraded the installer's *own* preflight errors to warnings. `zakurad` re-runs the same hardware checks at startup, so the command the installer printed still hard-exited on exactly the machines the flag was meant to rescue. It is now emitted into all five generated `zakurad start` commands.

**Medium — `--mode build-from-source` silently picked the wrong profile.** It is valid under both profiles, but `mode_implies_default_profile` claimed it, so a scripted run quietly built plain Zakura with no sidecar. It now implies neither: it prompts interactively, and requires an explicit `--install-profile` under `--non-interactive` (which `prompt_install_profile` already handled correctly).

**Minor — no host architecture check at all.** The report noted a bare Docker "no matching manifest" error on arm64. The native path is worse: it downloads a hardcoded `linux-x86_64` tarball that cannot exec. Both now fail early with an actionable message. `build-from-source` and the plain Zakura Docker image (which *does* publish an arm64 manifest) remain available on arm64.

**Minor — `zakurad --version` reported `1.0.0-rc0` against a `v1.0.0-rc1` tag.** Release binaries build inside Docker, where `.git` is unavailable, so vergen cannot derive the version and the binary falls back to `CARGO_PKG_VERSION`. Bumped `zakurad/Cargo.toml`, and added a guard to `release-binaries.yml` that fails the release if the tag and crate version disagree — the bump alone would just drift again next release.

**CI.** `install-zakura.sh` had no CI coverage whatsoever; its only self-check (`--self-test-disk-limits`) was never invoked. Added a `shellcheck` job covering `scripts/*.sh` that also runs that self-test.

### Tests

Exercised in fresh `ubuntu:24.04` containers, reproducing the reporter's environment:

- **As uid 1000** (`--install-profile default --mode native --network Testnet -y`): fails on `main` with the exact reported error; now resolves to `/home/ubuntu/.cache/zakura` and completes.
- **As root with a writable `/mnt`**: still selects `/mnt/data/zakura-state`, so the original intent (prefer a big volume) is preserved.
- `--unsafe-low-specs`: flag present in the generated command in all four compat modes; absent when not passed.
- `--mode build-from-source` with no profile: errors under `--non-interactive`; still works with an explicit profile.
- arm64 (simulated `uname -m`): blocked for native/compat-docker modes, allowed for `build-from-source` and default `docker`.
- `shellcheck` clean on all `scripts/*.sh` at default severity; `--self-test-disk-limits` passes.
- Version guard verified to pass on `v1.0.0-rc1` and fail on a mismatched tag.

### Follow-up Work

Real arm64 support for the zcashd-compat images is out of scope here — it is structural (the upstream zcashd archive is amd64-only, so `resolve-zcashd-compat-manifest.sh` hard-fails non-amd64). This PR only makes the failure legible.

### AI Disclosure

- [x] AI tools were used: Claude Code, for the implementation, container-based testing, and this description. Reviewed by me.

### PR Checklist

- [x] The PR title follows conventional commits format
- [x] The solution is tested
- [x] The documentation and changelogs are up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,14 +32,6 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   maturity, so a shipped checkpoint cannot be orphaned by a reorg Zakura would
   still follow. `MAX_BLOCK_REORG_HEIGHT` now lives in `zakura-chain` as the single
   source of truth. Backported from upstream Zebra PR #10719.
-- `install-zakura.sh` no longer fails for non-root users on the default install
-  profile. The standalone state directory defaulted to a hardcoded
-  `/mnt/data/zakura-state`, which cannot be created by anyone but root on a stock
-  cloud image, so the first interactive menu option failed before downloading
-  anything. The default profile now runs the same capacity-aware directory search
-  the zcashd-compat profile already used, falling back to `~/.cache/zakura`. That
-  search also filters on writability now, so it can no longer recommend a large
-  but root-owned volume that the installer then fails to create.
 - `install-zakura.sh --unsafe-low-specs` now appends the flag to every generated
   `zakurad start` command. Previously it only silenced the installer's own
   preflight warning, so the command it printed still hard-exited on the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,9 +40,6 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Changed
 
-- Added a `shellcheck` lint job covering `scripts/*.sh`, which also runs the
-  installer's `--self-test-disk-limits` self-check. `install-zakura.sh` previously
-  had no CI coverage at all.
 - Corrected Zakura's release identity and build provenance. `zakurad --version`
   now reports the `1.0.0-rc0` release with source-commit build metadata, and
   Docker images expose OCI version, revision, source, and title labels. The
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,18 +32,6 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   maturity, so a shipped checkpoint cannot be orphaned by a reorg Zakura would
   still follow. `MAX_BLOCK_REORG_HEIGHT` now lives in `zakura-chain` as the single
   source of truth. Backported from upstream Zebra PR #10719.
-- `install-zakura.sh --unsafe-low-specs` now appends the flag to every generated
-  `zakurad start` command. Previously it only silenced the installer's own
-  preflight warning, so the command it printed still hard-exited on the
-  under-specced machines that needed the override.
-- `install-zakura.sh --mode build-from-source` no longer silently resolves to the
-  default profile. The mode is valid under both profiles, so it now prompts for
-  the profile and requires an explicit `--install-profile` under
-  `--non-interactive`, instead of quietly building plain Zakura with no sidecar.
-- `install-zakura.sh` now fails with an actionable message on non-amd64 hosts
-  instead of downloading an x86_64 binary that cannot exec, or surfacing a bare
-  Docker "no matching manifest" error. `build-from-source` and the plain Zakura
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,26 @@ and this project adheres to [Semantic Versioning](https://semver.org).
...
   still follow. `MAX_BLOCK_REORG_HEIGHT` now lives in `zakura-chain` as the single
   source of truth. Backported from upstream Zebra PR #10719.
+- `install-zakura.sh` no longer fails for non-root users on the default install
+  profile. The standalone state directory defaulted to a hardcoded
+  `/mnt/data/zakura-state`, which cannot be created by anyone but root on a stock
+  cloud image, so the first interactive menu option failed before downloading
+  anything. The default profile now runs the same capacity-aware directory search
+  the zcashd-compat profile already used, falling back to `~/.cache/zakura`. That
+  search also filters on writability now, so it can no longer recommend a large
+  but root-owned volume that the installer then fails to create.
```
p0mvn:
```suggestion
```

CHANGELOG.md
```diff
@@ -32,6 +32,18 @@ and this project adheres to [Semantic Versioning](https://semver.org).
...
+  preflight warning, so the command it printed still hard-exited on the
+  under-specced machines that needed the override.
+- `install-zakura.sh --mode build-from-source` no longer silently resolves to the
+  default profile. The mode is valid under both profiles, so it now prompts for
+  the profile and requires an explicit `--install-profile` under
+  `--non-interactive`, instead of quietly building plain Zakura with no sidecar.
+- `install-zakura.sh` now fails with an actionable message on non-amd64 hosts
+  instead of downloading an x86_64 binary that cannot exec, or surfacing a bare
+  Docker "no matching manifest" error. `build-from-source` and the plain Zakura
+  Docker image (which publishes an arm64 manifest) remain available on arm64.
```
p0mvn:
```suggestion
```

CHANGELOG.md
```diff
@@ -40,6 +40,9 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Changed
 
+- Added a `shellcheck` lint job covering `scripts/*.sh`, which also runs the
+  installer's `--self-test-disk-limits` self-check. `install-zakura.sh` previously
+  had no CI coverage at all.
```
p0mvn:
```suggestion
```

p0mvn (approved):
While user facing, I do not think many users care about the installer changelog, and it would only be adding cogntiive overhead to the relevant changes.

So I removed them. If objections, propose to discuss and agree for the future

PR #100, opened 2026-07-11
fix(sync): keep the zcashd-compat sidecar syncing instead of stalling on a fresh install

evan-forbes:
## Motivation

On a fresh zcashd-compat install, the sidecar's sync stalls early and never recovers. This is the headline feature of compat mode, and it does not reliably work.

A user reported this and root-caused it to Zakura's inbound load-shedding silently dropping the sidecar's `getheaders`. **That diagnosis turns out to be wrong.** It is a real silent-failure path and this PR closes it, but it is not what stalls the sidecar. Reproducing the stall end to end and reading the wire showed something else:

- Every `getheaders` the sidecar sent **was answered** — the shed path never fired once.
- The final `headers` batch was *short* (fewer than `MAX_HEADERS_RESULTS`), so zcashd correctly concluded it had reached our tip and stopped asking. That is what a short batch means.
- From then on it waited for an `inv`. **None ever arrived.** It sat at 8,400 blocks while Zakura ran on past 440,000.

## Solution

**1. Gossip to the sidecar during initial block download (the actual fix).**

`gossip_best_tip_block_hashes` gates *every* announcement behind `SyncStatus::wait_until_close_to_tip()`. That is correct for ordinary peers: a peer far from the tip is expected to sync from a node that is near it, so announcing blocks we are still catching up on is noise. A sidecar cannot do that — it is pinned to us by `-connect` as its only peer, so an announcement is the only way it ever learns our tip moved. Gating it on being close to the tip means it hears nothing for the whole of our IBD.

Zakura now gossips on every tip change when zcashd-compat is enabled, and the peer set picks the audience: `select_block_broadcast_peers` returns **only** sidecar peers while we are far from the network tip, so gossip to ordinary peers is byte-for-byte unchanged.

This is what `zcashd_compat.block_gossip_peer_ips` has always documented — *"must always receive block inventory broadcasts"* — but the close-to-tip gate sat upstream of peer selection, so the sidecar was never even considered.

**2. Stop silently shedding the sidecar's inbound requests (defensive, the reported cause).**

All peers share one 200-slot `tower::Buffer` behind `load_shed`. On overload, `handle_inbound_overload` **silently ignores** the request ~95% of the time: no reply, no `notfound`, no reject, connection left open, and only a `pool.ignored.loadshed` counter. The `Overloaded` branch did not even log the request, while the timeout branch beside it logs at `INFO` *with* it — which is why this was invisible.

For an ordinary peer that is correct. A sidecar has no other peer to ask, and classic `getheaders` has neither a timeout nor a retry, so one ignored request black-holes it. Peers in the existing allowlist now get **backpressure instead of shedding**: the request is retried while the queue drains. If still overloaded after `MAX_COMPAT_INBOUND_RETRY_TIME` (20s), the connection is **closed with a `WARN`** rather than ignored — closing is recoverable (zcashd re-dials and restarts headers sync), ignoring is not.

Not a DoS vector: the allowlist is IP-based and defaults to loopback — a local zcashd this node supervises. Every other peer keeps the existing randomised drop/ignore shedding, which `connection_is_randomly_disconnected_on_overload` still covers.

The 20s budget deliberately sits *below* the sidecar's own 30s `HEADERS_RESPONSE_TIMEOUT` (valargroup/zcashd), so we always give up first and close, and no duplicate `getheaders` is ever in flight.

The IP predicate moves to `ConnectedAddr` so the inbound path and the block-gossip path share one definition of "who is the sidecar" instead of two.

### Tests

**End to end on live testnet**, both nodes local, zcashd pinned to zakurad exactly as `install-zakura.sh` prints it (`-connect=127.0.0.1:18233 -listen=0 -dnsseed=0`), i.e. zakurad on the real network and zcashd with zakurad as its only peer. This gives a clean A/B isolating the gossip fix:

| Run | Config | zcashd after 10 min |
|---|---|---|
| 1 | load-shed fix only | **stalled at 8,400** |
| 2 | + gossip fix | **111,469**, header chain at 349,040 vs Zakura's 349,600 |
| 3 | repeat (6 min) | **64,697**, tracking tip |

In runs 2 and 3 the sidecar's header chain tracks Zakura's tip in real time and pulls blocks in behind it — a healthy sidecar.

**Unit tests.** New: sidecar overload is retried and answered, not ignored; sidecar connection is closed (not ignored) once the retry budget is spent; the sidecar-IP predicate (including that an *outbound* peer at loopback must never inherit sidecar privileges). Both connection tests were mutation-checked — they fail when the sidecar flag is flipped off.

The peer-set gossip test was asserting the old behaviour while silently running in the "unknown tip distance" regime (`MockChainTip` only reports a distance when a tip height is also set). It is now split into the two regimes it should always have covered: sidecar-only before the tip, sidecar plus a random sample at the tip.

`cargo test -p zakura-network --lib`: 910 passed. `cargo test -p zakura --lib`: 303 passed. `fmt` and `clippy -D warnings` clean.

### Follow-up Work

The `zcashd` half — recovering from an unanswered `getheaders` instead of latching forever — is in valargroup/zcashd and is defence in depth against this class of bug from any cause.

### AI Disclosure

- [x] AI tools were used: Claude Code, for the investigation, implementation, end-to-end testing, and this description. Reviewed by me.

### PR Checklist

- [x] The PR title follows conventional commits format
- [x] The solution is tested
- [x] The documentation and changelogs are up to date

### DIFF
--- a/zakurad/src/components/sync/gossip.rs
+++ b/zakurad/src/components/sync/gossip.rs
@@ -52,12 +52,16 @@ pub async fn gossip_best_tip_block_hashes<ZN>(
     mut chain_state: ChainTipChange,
     broadcast_network: ZN,
     mut mined_block_receiver: Option<mpsc::Receiver<(block::Hash, block::Height)>>,
+    gossip_during_initial_sync: bool,
 ) -> Result<(), BlockGossipError>
 where
     ZN: Service<zn::Request, Response = zn::Response, Error = BoxError> + Send + Clone + 'static,
     ZN::Future: Send,
 {
-    info!("initializing block gossip task");
+    info!(
+        ?gossip_during_initial_sync,
+        "initializing block gossip task"
+    );
 
     // use the same timeout as tips requests,
     // so broadcasts don't delay the syncer too long
@@ -90,10 +94,23 @@ where
             // wait until we're close to the tip, because broadcasts are only useful for nodes near the tip
             // (if they're a long way from the tip, they use the syncer and block locators), unless a mined block
             // hash is received before `wait_until_close_to_tip()` is ready.
-            sync_status
-                .wait_until_close_to_tip()
-                .map_err(SyncStatus)
-                .await?;
+            //
+            // A zcashd-compat sidecar is the exception: it cannot "use the syncer and block
+            // locators", because it is pinned to this node as its only peer. Once it catches up
+            // to our in-progress tip it receives a short `headers` batch, correctly concludes it
+            // has reached our tip, stops sending `getheaders`, and then waits for an inventory
+            // announcement. Gating every announcement on being close to the tip means that
+            // announcement never arrives, and the sidecar sits idle for the whole of our initial
+            // block download. So when a sidecar is configured, gossip on every tip change.
+            //
+            // The peer set still restricts the audience: while we are far from the network tip,
+            // `select_block_broadcast_peers` sends only to the sidecar, never to ordinary peers.
+            if !gossip_during_initial_sync {
+                sync_status
+                    .wait_until_close_to_tip()
+                    .map_err(SyncStatus)
+                    .await?;
+            }
 
             // get the latest tip change when close to tip - it might be different to the change we awaited,
             // because the syncer might take a long time to reach the tip
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,24 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   maturity, so a shipped checkpoint cannot be orphaned by a reorg Zakura would
   still follow. `MAX_BLOCK_REORG_HEIGHT` now lives in `zakura-chain` as the single
   source of truth. Backported from upstream Zebra PR #10719.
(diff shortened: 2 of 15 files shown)


zakurad/src/components/sync/gossip.rs
```diff
@@ -90,10 +94,23 @@ where
...
-                .map_err(SyncStatus)
-                .await?;
+            //
+            // A zcashd-compat sidecar is the exception: it cannot "use the syncer and block
+            // locators", because it is pinned to this node as its only peer. Once it catches up
+            // to our in-progress tip it receives a short `headers` batch, correctly concludes it
+            // has reached our tip, stops sending `getheaders`, and then waits for an inventory
+            // announcement. Gating every announcement on being close to the tip means that
+            // announcement never arrives, and the sidecar sits idle for the whole of our initial
+            // block download. So when a sidecar is configured, gossip on every tip change.
```
p0mvn, line 104: We can ensure we gossip on every tip change to sidecar only. Let me look at the logs a bit first

p0mvn:
Going to inspect the logs first to confirm. Marking as draft for now

p0mvn:
Agreed to solve this on zcashd side only

PR #104, opened 2026-07-11
ci(release): create protected release tags via workflow

p0mvn:
## Motivation

The existing release workflow validates versions only after a tag has already been created. A mismatched release tag can therefore remain in the repository even when artifact publication is blocked. The source installer checksum also needs a safe, reviewable update after release artifacts are built.

## Solution

- Add a manually dispatched `Create release` workflow that runs only from `main` and verifies the requested tag against the `zakura` package version before creating anything.
- Use a narrowly scoped GitHub App installation token for tag and release creation, allowing a tag ruleset to block direct maintainer pushes.
- Make retries safe when a partially completed run already created the expected tag.
- After publishing release artifacts, update the source installer's archive checksum on an automation branch and open a follow-up PR instead of writing directly to `main`.
- Skip stale checksum updates when `main` no longer targets the release being built, and avoid duplicate follow-up PRs on workflow retries.
- Grant the release app only `Contents: Read and write` and `Pull requests: Read and write`, with installation access limited to `zakura-core/zakura` and webhooks disabled.
- Document GitHub App creation and installation, the protected `release` environment, creation-only app bypass, separate immutable-tag ruleset, and updated release process.

The GitHub App's tag push triggers `release-binaries.yml`. Its contents and pull-request permissions also let that workflow create the reviewed installer-checksum follow-up without bypassing branch protection on `main`.

### Follow-up Work

After merge, provision the organization-owned release GitHub App with the documented contents and pull-request permissions, configure the protected `release` environment variables/secrets, install the app only on `zakura-core/zakura`, and activate the two documented `v*` tag rulesets. Tag restrictions must not be enabled before the app is installed and configured.


docs/release-tag-protection.md
```diff
@@ -0,0 +1,57 @@
...
+- No webhook subscription
+
+Create a private key for the app. Configure a GitHub Actions environment named
+`release`, add the app's client ID as the environment variable
+`RELEASE_APP_CLIENT_ID`, and add its private key as the environment secret
+`RELEASE_APP_PRIVATE_KEY`. Configure required reviewers for the environment so
+that creating a release requires explicit approval.
+
+The app private key is a credential. Store its source copy in the team's secret
+manager and do not commit it or paste it into issues, pull requests, or logs.
```
p0mvn, line 44: Note to reviewers: org owners will need to create this

p0mvn:
Just a demo. If agreed on, will need to make a GitHub App. Then, we can merge this

ebfull:
Concept ACK.

.github/workflows/release-binaries.yml
```diff
@@ -417,6 +417,158 @@ jobs:
...
+          path: source-update
+
+      - name: Resolve release GitHub App user
+        id: app-user
+        env:
+          APP_SLUG: ${{ steps.app-token.outputs.app-slug }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: echo "id=$(gh api "/users/${APP_SLUG}[bot]" --jq .id)" >> "$GITHUB_OUTPUT"
+
+      - name: Update source installer checksum
```
p0mvn: Note to reviewer: this exists to avoid us having to manually update the main branch installer with the hash.

The hash is currently computed as part of the release process.

Ref: https://github.com/zakura-core/zakura/pull/106#discussion_r3565171799

docs/release-tag-protection.md
```diff
@@ -0,0 +1,68 @@
...
+
+## GitHub App
+
+In the `zakura-core` organization settings, open **Developer settings > GitHub
+Apps**, select **New GitHub App**, and create an organization-owned app with:
+
+- Name `zakura-release-bot`
+- Homepage URL set to the `zakura-core/zakura` repository
+- Repository permission `Contents: Read and write`
+- Repository permission `Pull requests: Read and write`
```
p0mvn, line 19: Note to reviewer: the PR permissions are for automated follow-up with a binary hash update

p0mvn:
@cursor review

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,108 @@
...
+
+          if gh release view "$RELEASE_TAG" --repo "$REPOSITORY" >/dev/null 2>&1; then
+            echo "GitHub release ${RELEASE_TAG} already exists; nothing to do."
+            exit 0
+          fi
+
+          release_flags=()
+          if [[ "$RELEASE_TAG" == *-* ]]; then
+            release_flags+=(--prerelease)
+          fi
```
p0mvn: Note to reviewer: only the options with a hyphen (e.g., `v1.0.2-rc2`) are marked as pre-release. The regular tags are made as proper releases.

.github/workflows/release-binaries.yml
```diff
@@ -241,6 +241,7 @@ jobs:
     name: Publish zakurad assets
     needs: [resolve-release, build-zakurad-assets]
     runs-on: ubuntu-latest
+    environment: release
```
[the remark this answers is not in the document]
p0mvn: [442db18](https://github.com/zakura-core/zakura/pull/104/commits/442db18b734787c6c9fbeb3f1ca337dd1f1d4816)

.github/workflows/release-binaries.yml
```diff
@@ -417,6 +418,158 @@ jobs:
...
+            gh pr list \
+              --repo "$REPOSITORY" \
+              --head "$branch" \
+              --state open \
+              --json number \
+              --jq '.[0].number // empty'
+          )"
+          if [ -n "$existing_pr" ]; then
+            echo "Installer checksum PR #${existing_pr} is already open."
+            exit 0
```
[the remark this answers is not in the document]
p0mvn: I think this is fine to be manual

p0mvn:
@cursor review

book/src/dev/release-process.md
```diff
@@ -130,4 +130,4 @@ To help ensure that you have sufficient time and a clear path to update, this is
...
 
-Every release tag must match the `zakura` package version: release binaries are built without `.git`, so `zakurad --version` reports the package version, not the tag. The `release-binaries.yml` workflow checks this and refuses to build or publish assets for a mismatched tag. Before pushing a tag, verify the release commit with `./scripts/check-release-version.sh v<version>`.
+Every release tag must match the `zakura` package version: release binaries are built without `.git`, so `zakurad --version` reports the package version, not the tag. Create tags using the protected `create-release.yml` workflow, which verifies the package version before creating the tag and GitHub release. The `release-binaries.yml` workflow independently checks the version again before building or publishing assets. See [Protected Release Tags](../../../docs/release-tag-protection.md) for configuration and operation details.
```
ebfull:
```suggestion
Every release tag must match the `zakura` package version: release binaries are built without `.git`, so `zakurad --version` reports the package version, not the tag. Create tags using the protected `create-release.yml` workflow, which verifies the package version before creating the tag and GitHub release. The `release-binaries.yml` workflow independently checks the version again before building or publishing assets. See `release-tag-protection.md` for configuration and operation details.
```

This isn't part of the book so the link is useless.

ebfull:
Oops, didn't mean to ask copilot to review. I'm still reviewing.

.github/workflows/release-binaries.yml
```diff
@@ -417,6 +417,180 @@ jobs:
...
+          gh pr create \
+            --repo "$REPOSITORY" \
+            --base main \
+            --head "$branch" \
+            --title "chore(release): update installer checksum for ${RELEASE_TAG}" \
+            --body "$(cat <<EOF
+          ## Motivation
+
+          Keep the installer served from \`main\` pinned to the archive published for ${RELEASE_TAG}.
+
```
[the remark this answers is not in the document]
ebfull: This is a worthless false positive.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,108 @@
...
+          ./scripts/check-release-version.sh "$RELEASE_TAG"
+
+          echo "release_tag=$RELEASE_TAG" >> "$GITHUB_OUTPUT"
+          echo "target_sha=$TARGET_SHA" >> "$GITHUB_OUTPUT"
+
+  create:
+    name: Create tag and GitHub release
+    runs-on: ubuntu-latest
+    needs: [validate]
+    environment: release
```
ebfull: This is the crux of the review: the tag is minted here *before* anything has been built, so a build or packaging failure afterward strands an immutable `v*` tag and the only recovery is burning the version number. The ordering can be inverted while keeping everything else in this PR:

1. `validate` — unchanged.
2. `build-assets` — call `release-binaries.yml` via `workflow_call` with a new `stage_release_assets` input and `checkout_ref: ${{ needs.validate.outputs.target_sha }}`. In stage mode, `publish-zakurad-assets` assembles and attests `./publish/final` exactly as it does today, but uploads it as a workflow **artifact** instead of touching any release. The Docker and zcashd-compat jobs already run in their no-push build mode on non-tag refs, so they join the build gate for free.
3. `publish` (this job) — download the staged artifact, re-verify it (`sha256sum --check --strict SHA256SUMS.txt`, the installer's embedded tag, expected asset names), attach everything to a **draft** release, then flip `draft=false`. Drafts have no git tag — the tag is created at publish time, after every asset is already attached. The `release` environment approval then lands *after* the whole build has succeeded, and the release is complete the instant its tag exists.

<details><summary>Core of the replacement publish step (idempotent; dry-run tested)</summary>

```bash
set -euo pipefail

# Refuse to touch a tag that points at a different commit; a tag
# already at the target means a previous attempt published it.
existing_sha="$(
  gh api "repos/${REPOSITORY}/git/ref/tags/${RELEASE_TAG}" \
    --jq '.object.sha' 2>/dev/null || true
)"
if [ -n "$existing_sha" ] && [ "$existing_sha" != "$TARGET_SHA" ]; then
  echo "Tag ${RELEASE_TAG} already points to ${existing_sha}, not ${TARGET_SHA}." >&2
  exit 1
fi

releases="$(gh api "repos/${REPOSITORY}/releases?per_page=100")"

published_count="$(
  echo "$releases" | jq --arg tag "$RELEASE_TAG" \
    '[.[] | select(.tag_name == $tag and .draft == false)] | length'
)"
if [ "$published_count" -gt 0 ]; then
  echo "Release ${RELEASE_TAG} is already published; nothing to do."
  exit 0
fi

# Reuse the notes from an existing draft for this tag (usually the
# release-drafter draft) before replacing it. Drafts have no git
# tag, so deleting one is always safe.
notes_file="$(mktemp)"
echo "$releases" | jq -r --arg tag "$RELEASE_TAG" \
  'first(.[] | select(.tag_name == $tag and .draft == true)).body // ""' \
  > "$notes_file"
if [ ! -s "$notes_file" ]; then
  echo "Release created by the protected Create release workflow." > "$notes_file"
fi

echo "$releases" | jq -r --arg tag "$RELEASE_TAG" \
  '.[] | select(.tag_name == $tag and .draft == true) | .id' \
| while read -r draft_id; do
    echo "Deleting stale draft release ${draft_id} for ${RELEASE_TAG}."
    gh api --method DELETE "repos/${REPOSITORY}/releases/${draft_id}"
  done

# Creating a draft does not create the git tag: the tag appears
# only when the draft is published below, after every asset is
# attached, so a failure here never leaves a tag pointing at an
# asset-less or partially built release.
draft_id="$(
  jq -n \
    --arg tag "$RELEASE_TAG" \
    --arg sha "$TARGET_SHA" \
    --rawfile body "$notes_file" \
    '{tag_name: $tag, target_commitish: $sha, name: $tag, draft: true, prerelease: true, body: $body}' \
  | gh api --method POST "repos/${REPOSITORY}/releases" --input - --jq '.id'
)"

# Upload SHA256SUMS.txt last, so its presence on the release marks
# a complete asset set (release-binaries.yml keys its rebuild-skip
# on it).
mapfile -t asset_files < <(
  find ./assets -maxdepth 1 -type f ! -name 'SHA256SUMS.txt' | sort
)
asset_files+=("./assets/SHA256SUMS.txt")
for asset in "${asset_files[@]}"; do
  name="$(basename "$asset")"
  echo "Uploading ${name}."
  curl --fail --silent --show-error \
    --request POST \
    --header "Authorization: Bearer ${GH_TOKEN}" \
    --header "Content-Type: application/octet-stream" \
    --data-binary "@${asset}" \
    --output /dev/null \
    "https://uploads.github.com/repos/${REPOSITORY}/releases/${draft_id}/assets?name=${name}"
done

# Publishing the draft is the step that creates the protected v*
# tag, through the release App's ruleset bypass. Releases are
# always published as pre-releases: promoting a final release
# (clearing the pre-release flag, setting it as latest) stays a
# manual checklist step after the release has been tested.
html_url="$(
  gh api --method PATCH "repos/${REPOSITORY}/releases/${draft_id}" \
    --field draft=false --jq '.html_url'
)"
echo "Published ${html_url}; tag ${RELEASE_TAG} now points to ${TARGET_SHA}."
```

</details>

Retry semantics stay at least as good as the current version: tag-points-elsewhere refuses, already-published no-ops, and a leftover draft from a failed attempt is deleted and recreated — always safe, because a draft carries no tag. Bonus: the release-drafter draft's notes get carried over instead of placeholder text.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,108 @@
...
+          if [[ "$RELEASE_TAG" == *-* ]]; then
+            release_flags+=(--prerelease)
+          fi
+
+          gh release create "$RELEASE_TAG" \
+            --repo "$REPOSITORY" \
+            --verify-tag \
+            --title "$RELEASE_TAG" \
+            --notes "Release created by the protected Create release workflow." \
+            "${release_flags[@]}"
```
ebfull: Standalone from the reordering: deciding release-vs-prerelease from the version string means the eventual `v1.0.0` goes live immediately as a full release (and becomes "latest") with placeholder notes and zero assets, hours before binaries exist — and the checklist's later "disable 'pre-release', then set as latest" step has nothing to operate on. Publishing everything as a pre-release keeps promotion a human decision after testing and signing (#73 already made promotion non-destructive):

```suggestion
          # Always publish as a pre-release, whatever the tag looks like:
          # promoting to a full release (and setting it as latest) stays a
          # manual step after the release has been tested and signed.
          gh release create "$RELEASE_TAG" \
            --repo "$REPOSITORY" \
            --verify-tag \
            --prerelease \
            --title "$RELEASE_TAG" \
            --notes "Release created by the protected Create release workflow."
```

.github/workflows/release-binaries.yml
```diff
@@ -417,6 +417,180 @@ jobs:
...
             --clobber
 
+  update-source-installer:
+    name: Update source installer checksum
+    needs: [resolve-release, publish-zakurad-assets]
+    runs-on: ubuntu-latest
+    environment: release-automation
+    permissions:
+      contents: read
+    if: startsWith(github.ref, 'refs/tags/v') || inputs.publish_assets_to_release
```
ebfull: Two things on this job:

1. As written it isn't wired into `release-binaries-success` (`needs` + `allowed-skips`), so the aggregate gate ignores it and a failed checksum-PR automation won't surface there. (The competing `cursor/release-environment-tag-publishing-fc19` branch got exactly this piece right — worth taking it and deleting that branch.)

2. Under the tag-last flow, the tag-push run should skip `build-zakurad-assets`/`publish-zakurad-assets` when the release already carries the complete asset set (probe for `SHA256SUMS.txt`, on push-triggered tag runs only, so `workflow_dispatch` repairs still rebuild — this also stops the rebuild from re-clobbering live assets). This job then becomes the main purpose of the tag-push run, so its condition has to tolerate the skipped dependency:

```yaml
    if: >-
      ${{ !cancelled()
          && needs.resolve-release.result == 'success'
          && (needs.publish-zakurad-assets.result == 'success'
              || needs.resolve-release.outputs.assets_published == 'true')
          && (startsWith(github.ref, 'refs/tags/v') || inputs.publish_assets_to_release) }}
```

Nice side effect: with assets attached before the tag exists, the `gh release download` at the top of this job can no longer race an in-flight asset upload.

docs/release-tag-protection.md
```diff
@@ -0,0 +1,75 @@
...
+
+1. Merge the release version bump into `main`.
+2. Open **Actions > Create release > Run workflow**.
+3. Select the `main` branch and enter the exact release tag.
+4. Approve the `release` environment deployment.
+5. Confirm that the workflow creates the tag and GitHub release.
+6. Confirm that `Release binaries` starts from the new tag.
+
+The workflow is safe to rerun after a partial failure when the tag already
+points to the same commit. It refuses to reuse a tag that points elsewhere.
```
ebfull: With the build-first ordering this runbook gains a step, and two guarantees become worth writing down:

```
4. Wait for the pipeline to build and verify the release assets and the
   no-push Docker builds. Nothing is tagged or published while it runs, so a
   failure here costs nothing — fix main and dispatch again.
5. Approve the release environment deployment. The workflow then publishes
   the pre-release — assets already attached — creating the protected tag as
   its final step.
6. Confirm that Release binaries starts from the new tag, skips the asset
   rebuild, publishes the Docker images, and opens the installer checksum PR.
```

The two guarantees: the tag points at the `main` commit that was validated and built even if `main` advances between dispatch and approval, and releases are always published as pre-releases — promotion stays a manual step after testing and signing.

book/src/dev/release-process.md
ebfull: The relative link escapes `book/src/`, so it will 404 in the rendered mdBook (it only works on GitHub); the release-checklist link earlier on this page uses a full GitHub URL for the same reason. (If we adopt the tag-last ordering, the "before creating the tag" wording here will want a matching update too.)

```suggestion
Every release tag must match the `zakura` package version: release binaries are built without `.git`, so `zakurad --version` reports the package version, not the tag. Create tags using the protected `create-release.yml` workflow, which verifies the package version before creating the tag and GitHub release. The `release-binaries.yml` workflow independently checks the version again before building or publishing assets. See [Protected Release Tags](https://github.com/zakura-core/zakura/blob/main/docs/release-tag-protection.md) for configuration and operation details.
```

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -181,20 +181,14 @@ The end of support height is calculated from the current blockchain height:
...
-- [ ] Mark the release as 'pre-release', until it has been built and tested
-- [ ] Publish the pre-release to GitHub using "Publish Release"
+- [ ] Run the [Create release workflow](https://github.com/zakura-core/zakura/actions/workflows/create-release.yml)
+      from `main`, entering the exact version tag, for example `v1.0.0-rc2`.
+      The workflow verifies that the tag matches the `zakura` package version,
+      then creates the protected tag and GitHub release.
+- [ ] Edit the new release and replace its placeholder description with the
+      final changelog you created, starting just _after_ the title
+      `## [Zakura ...` of the current version and ending just _before_ the title
+      of the previous release.
```
ebfull: Under the build-first flow this section splits naturally into: dispatch (build runs; nothing is tagged, so failures are free to fix and retry) → approve the `release` deployment (publishes the pre-release, minting the tag last) → review the notes. The description step also gets nicer: the publish step can carry the release-drafter draft's notes over instead of placeholder text, so this becomes review-and-touch-up rather than paste-from-scratch. Worth adding an explicit "merge the installer checksum PR the tag run opens" item to the Test section below, too.

ebfull (approved):
I'm uncomfortable with a process where a release gets tagged and then some kind of build step can screw up afterward. A `v*` tag is the one thing we can never take back: the rulesets in this PR (correctly!) make release tags immutable, and anyone who `git fetch`es before we notice keeps the bad ref locally forever — git doesn't overwrite tags on fetch. I'd rather we tag only after we know everything is okay: build and verify everything first, mint the tag as the last step, then I can check the release before I sign it, push my signature, and *at that point* flip it from pre-release to release — rather than deciding release-vs-prerelease up front from the shape of the version string.

I think this review steers us in that direction, but I'm not blocking the PR on it — the App/environment/ruleset design here is right and carries over unchanged, so this can also be improved later.

Two properties make the reordering cheap:

- Release binaries don't embed the tag (they're built without `.git`, and the version gate pins tag == package version), so building from the `main` commit *before* the tag exists produces identical artifacts.
- A **draft** release has no git tag — GitHub creates the tag at the moment the draft is published, which is exactly how our tags were minted before this PR (publishing the drafted release created the tag).

So the shape becomes: validate → build and stage everything from the validated SHA (the Docker jobs' existing no-push mode becomes part of the gate for free) → after the `release` environment approval, attach the staged assets to a draft and publish it, creating the tag last. The tag-push run of `release-binaries.yml` then skips the asset rebuild and only publishes Docker images and opens the installer-checksum PR. Details in the inline comments; I have a full working diff of this (actionlint-clean, scripts dry-run tested) that I'm happy to share.

**AI disclosure:** the analysis and code in this review were written by Claude (Claude Code); I reviewed them.

book/src/dev/release-process.md
ebfull: This is [duplicated](https://github.com/zakura-core/zakura/pull/104#discussion_r3567270997).

.github/workflows/create-release.yml
p0mvn: yeah, I was also thinking about this. Thus, noted it here https://github.com/zakura-core/zakura/pull/104#discussion_r3566860795

Adopting the recommendation

p0mvn: Addressed

.github/workflows/release-binaries.yml
p0mvn: Addressed

docs/release-tag-protection.md
p0mvn: Addressed

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
p0mvn: Addressed

p0mvn:
@cursor, please review the PR comprehensively as of the latest state Focus on prior comments left in the PR. Call out if any detail has gone unaddressed

.github/workflows/create-release.yml
p0mvn: Addressed

ebfull (approved):

PR #112, opened 2026-07-12
fix: clear remaining Zebra/ZF identity remnants

ebfull:
- `ZEBRA_COINBASE_MARKER`/`ZEBRA_COINBASE_SEPARATOR` → `ZAKURA_*`, plus the surrounding Zebra doc comments. Marker bytes unchanged.
- `zakura` package description no longer says "The Zcash Foundation's…" (rendered on crates.io and by `--help`).
- SECURITY.md routes reports to GitHub private vulnerability reporting on this repo instead of `security@zfnd.org`/ZF's PGP key; ZF's bilateral-agreement claims reworded as intent. No Zakura mailbox/key exists yet — swap in if we publish one.
- README: dead `zfnd-*` workflow badges removed; Docker Hub org corrected to `valargroup/zakura`.
- release-drafter drafts "Zakura" notes linking `zakura-core/zakura` instead of "Zebra"/`valargroup/zebra`.
- Book: title, nav, and user pages rebranded; Docker/mining pages now document the supported `ZAKURA_*` env prefix and the real image. Upstream RFCs and `ed25519-zebra` references deliberately untouched.
- Dual-stack nodes stop advertising `/Zakura:x/Zakura:x/`: the v2 token is only prepended when the user agent doesn't already carry it (unit test).
- CHANGELOG reduced to an initial-release note. Zakura has no prior release — the rc tags are pre-releases of 1.0.0 itself — so there is nothing to enumerate changes against; pre-fork history is credited to upstream's CHANGELOG.
- CODE_OF_CONDUCT.md removed. It described ZF's community venues and moderation process, which don't exist for this fork; we don't consider a code of conduct necessary at this stage. Nothing else in the repo referenced it.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, and `zakura-rpc`/`zakura-network` tests pass.

AI disclosure: implemented with Claude Code (entire branch).


README.md
```diff
@@ -24,7 +22,7 @@
 ## Getting Started
 
 You can run Zakura using our [Docker
-image](https://hub.docker.com/r/zakura-core/zakura/tags) or you can install it manually.
+image](https://hub.docker.com/r/valargroup/zakura/tags) or you can install it manually.
```
ebfull, line 25: Right? Can we move this to a different repo under `zakura-core` going forward? (**DOES NOT BLOCK 1.0.0**)

zakura-network/src/peer/handshake.rs
```diff
@@ -616,6 +616,13 @@ fn configured_user_agent(config: &Config, user_agent: String) -> String {
 
     if trimmed_user_agent.is_empty() {
         format!("/{zakura_token}/")
+    } else if trimmed_user_agent
+        .split('/')
+        .any(|token| token == zakura_token)
+    {
+        // The default user agent already carries the Zakura token, so
+        // prepending it again would advertise `/Zakura:x/Zakura:x/`.
+        format!("/{trimmed_user_agent}/")
```
ebfull, line 625: Please check that this is right, seems weird.

ebfull:
PR previously modified the code of conduct; rebased and cleaned up so that it just removed the code of conduct entirely. Ready for review.

README.md
p0mvn: https://linear.app/zcale/issue/ZCA-779/zakura-core-docker-hub

zakura-network/src/peer/handshake.rs
p0mvn: `zakurad` already supplies `/Zakura:1.0.0-rc2/`, while the v2-enabled handshake previously prepended its own Zakura token, producing `/Zakura:1.0.0-rc2/Zakura:1.0.0-rc2/`. The new branch detects that exact slash-delimited token and leaves the user agent unchanged.

This is correct for the current executable: `zakurad` and `zakura-network` both use `1.0.0-rc2`, and `zakurad` removes build metadata before constructing the token. It only affects the legacy `version` message for v2-enabled/dual-stack nodes; v2 capability negotiation still uses the `NODE_P2P_V2` service bit.

The weirdness is architectural: the networking layer is modifying a valid user agent already constructed by the application. This fixes the immediate duplication, but a cleaner design would give one layer sole ownership of composing the user agent. Exact matching could also duplicate tokens if the two crate versions ever diverge.

p0mvn: The bug was relevant before this PR. The PR merely discovered and fixed it.

The bug came from two independently developed changes:

The v2 stack was written when the application user agent was still `/Zebra:<version>/`. Prepending Zakura correctly produced: `/Zakura:<version>/Zebra:<version>/`

Later, the application user agent was renamed to `/Zakura:<version>/`. The unchanged v2 logic then produced: `/Zakura:<version>/Zakura:<version>/`/

No networking regression is expected.

Remote `/Zebra:<version>/` user agents are treated as opaque metadata. Zakura does not use them for compatibility or upgrade decisions.


Therefore, existing Zebra nodes continue to handshake normally and remain on the legacy transport.

Older Zakura nodes advertising Zebra but setting `NODE_P2P_V2` can still upgrade.

This patch only changes what our node advertises about itself.

p0mvn: TLDR; seems fine to me because this is just metadata

p0mvn (approved):

PR #113, opened 2026-07-12
ci: classify consensus-critical changes

ValarDragon:
## Motivation

Consensus-critical changes need stronger review guarantees than ordinary
changes, including cases where their impact is not obvious from the modified
path alone.

This change was requested and discussed directly with the project owner.

## Solution

- Add a trusted `pull_request_target` workflow that classifies consensus impact
  without checking out or executing pull request code.
- Immediately classify known consensus paths as critical.
- Classify other complete, bounded diffs through the OpenAI Responses API using
  `gpt-5.6-sol` with medium reasoning.
- Explicitly classify regtest, custom-testnet, network-parameter selection, and
  network fallback behavior as critical, even when an intentional change does
  not affect Mainnet.
- Report whether checkpoint-related changes are demonstrably isolated to
  checkpoint sync, may affect live semantic consensus, or lack enough context
  to establish that separation. Known consensus paths remain critical even
  when the model reports checkpoint isolation.
- Treat uncertain results, API failures, incomplete patches, and oversized
  diffs as consensus-critical.
- Apply mutually exclusive `consensus-critical` and
  `consensus-not-critical` labels.
- Create or update one sticky PR comment containing the enforced result,
  checkpoint scope, affected areas, and classification reasons.
- Require three eligible approvals in Mergify when `consensus-critical` is
  present.
- Publish a SHA-bound `Consensus Critical Classification` commit status and
  fail closed unless the classifier explicitly applies `consensus-not-critical`
  or the PR has three eligible approvals.

The workflow expects an Actions secret named `OPENAI_API_KEY`. The model can be
overridden with the `OPENAI_CONSENSUS_MODEL` repository variable.

### Tests

- `git diff --check`
- Parsed both changed YAML files with Ruby's YAML parser.
- Extracted and syntax-checked the embedded JavaScript with `node --check`.
- Replayed the deterministic floor against historical consensus changes:
  - Zakura #111 hit all four representative consensus entry points, including
    `zakura-network/src/zakura/header_sync/validation.rs`.
  - Zebra #8413, #8421, and #8474 each hit the floor after applying the current
    `zebra-*` to `zakura-*` crate rename.
- Resolved `actions/github-script` v8 to immutable commit
  `ed597411d8f924073f98dfc5c65a23a2325f34cd`.
- `actionlint` was not available locally.
- The live OpenAI call cannot be exercised until `OPENAI_API_KEY` is configured
  in GitHub Actions.

- OpenAI Responses API with strict JSON-schema output.
- Mergify conditional queue requirements using `#approved-reviews-by`.

### Follow-up Work

- Add `OPENAI_API_KEY` as an Actions secret before enabling the classifier as a
  required check.
- Validate the Mergify configuration in the Mergify configuration editor.
- After this workflow is merged and emits its first SHA-bound status, add
  `Consensus Critical Classification` to the active `main` ruleset's required
  status checks. Adding it before merge would deadlock this PR because
  `pull_request_target` uses the workflow from the base branch.

### AI Disclosure

- [x] AI tools were used: OpenAI Codex implemented the workflow, Mergify policy,
  validation, and PR description.

### PR Checklist

- [x] This change was discussed with the project owner beforehand.
- [x] The solution is tested as described above.
- [x] No changelog update is needed because this is repository CI policy rather
  than a user-visible node change.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -26,6 +26,11 @@ defaults:
       - label!=do-not-merge
       # has at least one approving reviewer
       - "#approved-reviews-by >= 1"
+      # Consensus-critical changes require three distinct reviewers with
+      # admin, write, or maintain permission.
+      - or:
+          - label = consensus-not-critical
+          - "#approved-reviews-by >= 3"
 
 
 # Allows to define the rules that reign over our merge queues
--- a/.github/workflows/consensus-critical.yml
+++ b/.github/workflows/consensus-critical.yml
@@ -1,0 +1,405 @@
+name: Consensus Critical Classification
+
+on:
+  pull_request_target:
+    branches: [main, "feat/**", "release/**"]
+    types:
+      - edited
+      - labeled
+      - opened
+      - ready_for_review
+      - reopened
+      - synchronize
+      - unlabeled
+
+# This workflow handles untrusted pull requests while holding an API secret.
+# Do not check out, import, or execute code from the pull request head.
+permissions:
+  contents: read
+  issues: write
+  pull-requests: read
+  statuses: write
+
+concurrency:
+  group: consensus-critical-${{ github.event.pull_request.number }}
+  cancel-in-progress: ${{ github.actor != 'github-actions[bot]' }}
+
+jobs:
+  classify:
+    name: classify consensus impact
+    if: >-
+      (github.event.action != 'labeled' && github.event.action != 'unlabeled') ||
+      github.actor != 'github-actions[bot]'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    env:
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+      OPENAI_MODEL: ${{ vars.OPENAI_CONSENSUS_MODEL || 'gpt-5.6-sol' }}
+    steps:
+      - name: Classify the pull request
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
+        with:
+          script: |
+            const criticalLabel = "consensus-critical";
+            const nonCriticalLabel = "consensus-not-critical";
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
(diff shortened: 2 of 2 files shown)


ebfull:
Review from Claude...

> **1. The Mergify gate fails open.** `label != consensus-critical` passes whenever no label is present: classifier not yet a required check, run failed before labeling (the `catch` labels via the same API that may have just failed), or label manually removed (no `labeled`/`unlabeled` triggers to re-apply it). Invert it:
> 
> ```yaml
> - or:
>     - label = consensus-not-critical
>     - "#approved-reviews-by >= 3"
> ```
> 
> and make `classify` a required status check — that also closes the stale-label window on `synchronize` (labels aren't SHA-bound; a required check is).
> 
> **2. The path floor misses known consensus surface.** `zakura-network/` isn't covered at all — `zakura-network/src/zakura/header_sync/validation.rs` was #111's third consensus entry point. The bug this workflow exists because of would be classified by the model alone. Add at least `^zakura-network/src/zakura/(header_sync|block_sync)/`. Also missing: `zakura-chain/src/` top-level consensus files (`ironwood.rs`, `amount.rs`, `value_balance*`, `primitives/` — outside both regexes) and workspace `Cargo.toml`/`Cargo.lock` (consensus dep graph, e.g. the `[patch.crates-io]` equihash pin). And `zakurad/src/components/consensus/` doesn't exist — dead entry; cross-check the list against the tree.
> 
> **3. `actions/github-script@v8` is tag-pinned.** The only floating `uses:` in the repo, inside a `pull_request_target` workflow holding `OPENAI_API_KEY` with `issues: write`. Pin the SHA.
> 
> Minor: add `edited` to the trigger types (title/body feed the model but don't retrigger). And since `pull_request_target` runs the base-branch workflow, this PR won't classify itself — do the historical-PR calibration from the description (replay #111 and the Regtest-Equihash PRs) before the Mergify rule enforces, and post the results here.
> 
> Verified, no action: Mergify's `approved-reviews-by` counts only admin/write/maintain reviewers, so the 3-approval comment is accurate.

PR #114, opened 2026-07-12
chore(network): trace block sync cleanup anomalies

p0mvn:
## Motivation

A mainnet node stalled immediately after checkpoint height 3,278,006 with the next body absent while roughly 99,000 later bodies remained buffered. Existing traces narrowed the failure to request cleanup and frontier-reset bookkeeping, but did not identify the exact terminal path that lost the floor-height work item.

## Solution

- classify frontier resets as growth, preserved stale updates, or destructive resets with the predicates used by the decision
- record detailed work-return outcomes for anomalous timeout, retry, reset, and peer-drop cleanup paths
- trace floor-watchdog cancellation outcomes and post-cleanup queue membership
- keep trace volume bounded by omitting normal successful work returns; the existing bounded JSONL channel still drops rows under writer pressure
- test reserved, held, already-pending, and missing work-return classifications

### Tests

- `cargo fmt --all -- --check` — passed on the final changes
- `cargo test -p zakura-network zakura::block_sync::tests --lib` — 179 passed before the final trace-volume filter
- `cargo clippy --workspace --all-targets -- -D warnings` — passed before the final trace-volume filter
- `cargo test --workspace` — started, then stopped before completion at operator request rather than waiting for long-running integration tests
- IDE diagnostics — clean

This is observability-only: it does not intentionally change protocol, scheduling, retry, timeout, or consensus decisions.

### Follow-up Work

Use these events during the next reproduced floor-gap stall to identify whether work was missing, held, already pending, or successfully returned at each cleanup boundary.

### AI Disclosure

- [x] AI tools were used: Cursor (GPT-5.6 Sol) investigated production traces, implemented the diagnostics, and updated tests.

### PR Checklist

- [x] The PR title follows conventional commits format: `type(scope): description`
- [x] The PR follows the contribution guidelines
- [x] This change was discussed with the team beforehand
- [x] The focused block-sync solution is tested
- [x] Documentation and changelogs are up to date; no user-visible changelog entry is required

### DIFF
--- a/zakura-network/src/zakura/block_sync/reactor.rs
+++ b/zakura-network/src/zakura/block_sync/reactor.rs
@@ -403,14 +403,35 @@ impl BlockSyncReactor {
             let released = self
                 .state
                 .work_queue
-                .release_reserved_and_return_items([claim.height]);
-            self.state.budget.release(released);
+                .release_reserved_and_return_items_detailed([claim.height]);
+            self.state.budget.release(released.released_bytes);
+            self.emit_trace(bs_trace::BLOCK_FLOOR_WATCHDOG_CANCELLED, |row| {
+                bs_insert_peer(row, bs_trace::PEER, &claim.peer);
+                bs_insert_height(row, bs_trace::HEIGHT, claim.height);
+                bs_insert_u64(row, bs_trace::ESTIMATED_BYTES, claim.meta.estimated_bytes);
+                bs_insert_u64(row, "released_bytes", released.released_bytes);
+                bs_insert_u64(row, "returned_count", released.returned_count);
+                bs_insert_u64(row, "already_pending_count", released.already_pending_count);
+                bs_insert_u64(row, "held_count", released.held_count);
+                bs_insert_u64(row, "released_count", released.released_count);
+                bs_insert_u64(row, "missing_count", released.missing_count);
+                bs_insert_u64(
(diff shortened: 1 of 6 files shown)


p0mvn:
@cursor review

zakura-network/src/zakura/block_sync/reactor.rs
```diff
@@ -403,14 +403,35 @@ impl BlockSyncReactor {
...
                 .work_queue
-                .release_reserved_and_return_items([claim.height]);
-            self.state.budget.release(released);
+                .release_reserved_and_return_items_detailed([claim.height]);
+            self.state.budget.release(released.released_bytes);
+            self.emit_trace(bs_trace::BLOCK_FLOOR_WATCHDOG_CANCELLED, |row| {
+                bs_insert_peer(row, bs_trace::PEER, &claim.peer);
+                bs_insert_height(row, bs_trace::HEIGHT, claim.height);
+                bs_insert_u64(row, bs_trace::ESTIMATED_BYTES, claim.meta.estimated_bytes);
+                bs_insert_u64(row, "released_bytes", released.released_bytes);
```
evan-forbes, line 412: Note to self: we should clean instances of this up via simply passing a structure with a macro etc

evan-forbes (approved):

PR #115, opened 2026-07-12
fix(sync): correlate header responses by request ID

p0mvn:
## Motivation

Header-sync v6 correlates `Headers` responses through independent FIFO state in the peer pipe and reactor. A delayed response can therefore be decoded against its original expectation but consume a newer reactor request, producing false link failures and triggering stale-anchor recovery.

This was observed on `temp-zakura-sync-test-1` and is tracked in [ZCA-777](https://linear.app/zcale/issue/ZCA-777/fixsync-correlate-v2-header-responses-with-their-requests). The related frontier-invariant follow-up is [ZCA-778](https://linear.app/zcale/issue/ZCA-778/fixsync-prevent-header-reanchors-below-the-verified-body-frontier).

### Backwards-compatible protocol negotiation

- Keep the header-sync v6 wire format and capability unchanged.
- Add header-sync v7 under a distinct capability while retaining stream kind 5.
- Allow one service to own multiple versions of a stream kind, selecting the highest mutually negotiated version and opening exactly one stream.
- Preserve the selected stream declaration through transport admission so header sync chooses the correct codec.
- Advertise both header-sync discovery service generations during migration.

Compatibility behavior:

- v6-only ↔ dual-stack peers select v6.
- dual-stack ↔ dual-stack peers select v7.
- a non-selected or unsupported stream version is rejected rather than silently downgrading an established connection.

### Exact v7 correlation

- Add non-zero, monotonically allocated `u64` request IDs to v7 `GetHeaders` and `Headers` messages.
- Parse response IDs before header allocation and correlate expectations by ID rather than FIFO order.
- Propagate IDs through reactor events, state-backed serving, and response emission so responses echo the initiating request.
- Retire completed/timed-out IDs into a bounded tombstone set: known late responses are dropped without scoring, while never-issued IDs fail closed. The unscored drop path relies on transport rate/byte caps and is now documented at the drop site.
- Track inbound serving requests by ID so reverse-order state completions and stale completions after reconnect cannot target another request.

### Safe v6 fallback

V6 cannot disambiguate a delayed response after timeout or cancellation. Ambiguous v6 header-sync sessions are therefore retired before newer work can reuse the stream. Acknowledged empty responses remain safe to retry without disconnecting because their original response has already been consumed.

Known limitation (documented in the changelog): only the connection initiator reopens a retired stream, so a v6-only peer that dialed this node loses header sync on that connection after an ambiguous timeout until it reconnects. Connections this node dialed, and all v7 sessions, are unaffected.

### Observability and fault-lane coverage

- Trace `header_get_headers_sent` rows now carry `session_id`, `stream_version`, and the v7 `request_id`.
- New `header-faults` e2e mode restarts a node with outstanding header-sync work and runs a trace-oracle check requiring v7 request IDs to be present, non-zero, and unique per stream session.
- Testkit additions: hostile peers can open, finish, reset, and reopen versioned ordered streams; test nodes can restrict advertised capabilities and override header-sync config and request timeout.

### Review follow-ups

- Ordered-stream reopen retries now back off exponentially (250 ms base, doubling to an 8 s cap, reset on a successful reopen) instead of retrying at a flat 250 ms forever.
- E2E CI wiring: each CI job routes to exactly one docker lane via `ZAKURA_E2E_MODE` (the fault-lane test and the dual-stack test self-skip when the mode selects the other), lanes are serialized with a process-wide lock for local unfiltered runs, and `header-faults` is added to the nightly mode matrix.
- The changed-files detector in `zakura-e2e.yml` reads the pulls-files API's `filename` field (the previous `.path` matched nothing).
- Un-ignored `native_stream5_v7_correlates_reversed_physical_responses`: the wire path was never broken — the test network had a lone checkpoint at height 4, and the scheduler defers (never truncates) finalized ranges that exceed a peer's advertised cap, so a cap-2 peer could never be scheduled. The victim now anchors at a height-2 checkpoint so a backward backfill bracket (1..=2) plus a forward range (3..=4) produce two concurrent in-cap v7 requests over real QUIC streams, answered in reverse order and correlated by request ID.

## Tests

Passed:

- `markdownlint CHANGELOG.md --config .trunk/configs/.markdownlint.yaml`
- `cargo clippy --workspace --all-targets -- -D warnings` (`-p zakura-network -p zakura` re-verified on the final branch)
- `trace_oracle.py --self-test`
- Focused regression coverage for:
  - delayed v6 responses after timeout
  - v6 timeout retiring and reopening the stream on the same connection
  - v7 request-ID codec round trips, missing/zero-ID rejection, and cross-version payload rejection
  - reverse-order v7 response correlation (reactor fixture and real QUIC streams)
  - retired and unknown v7 response IDs, tombstone bounds at the exact limit
  - request-ID reservation rollback on full/closed transport queues without ID reuse
  - permanent request-ID exhaustion failure
  - reverse-order serving completions echoing the correct IDs
  - multi-version registry selection
  - cluster-level v6 fallback, v7 preference, and unsupported-version rejection
  - reopen backoff growth and cap

Broad network run:

- `cargo test -p zakura-network --lib`: 947 passed; 3 unrelated macOS loopback-source tests failed because this host cannot bind their requested aliases (`AddrNotAvailable`). No header-sync, transport, codec, or correlation tests failed.

The `header-faults` docker e2e lane is wired into the nightly matrix; it can be run locally with `ZAKURA_E2E_MODE=header-faults cargo test -p zakura --test zakura_regtest_e2e -- --ignored`.

- [ZCA-777: correlate V2 header responses with requests](https://linear.app/zcale/issue/ZCA-777/fixsync-correlate-v2-header-responses-with-their-requests)
- [ZCA-778: prevent header reanchors below verified body frontier](https://linear.app/zcale/issue/ZCA-778/fixsync-prevent-header-reanchors-below-the-verified-body-frontier)

### Follow-up Work

- Enforce `verified_body <= best_header` centrally in the shared frontier exchange under ZCA-778.

### AI Disclosure

- [x] AI tools were used: Cursor assisted with incident analysis, implementation, tests, and PR description. Claude Code assisted with the review follow-ups (e2e CI wiring, reopen backoff, test root-cause analysis) and this description update. The changes were reviewed and verified locally as described above.

### PR Checklist

- [x] The change was discussed and tracked in ZCA-777/ZCA-778.
- [x] The solution has targeted and broad network test coverage.
- [x] The changelog is updated.


evan-forbes (approved):
approving to unblock and we can optionally apply some cleanups later

p0mvn:
Tested on mainnet and testnet clusters

.github/workflows/zakura-e2e.yml
```diff
@@ -69,9 +69,11 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            # The pulls/files API exposes changed paths as `filename` (unlike
+            # the compare API's `.files[].filename`, this array is top-level).
             changed=$(gh api --paginate \
               "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
-              --jq '.[].path')
```
p0mvn, line 74: Bug fix

PR #128, opened 2026-07-13
refactor(sync)!: remove header-sync v6

p0mvn:
## Motivation

Header-sync v6 correlates `Headers` responses by arrival order, so it cannot disambiguate a delayed response. That is exactly the bug #115 fixed by adding v7 request IDs — but #115 kept v6 for compatibility, which means keeping the ambiguity, plus everything built to work around it:

- a second capability bit and a second discovery service generation,
- a multi-version stream registry (one service owning two versions of a stream kind),
- a v6-only "retire the session and reopen the stream" fallback after an ambiguous timeout, because v6 has no safe way to fail,
- a documented known limitation: a v6-only peer that dialed this node loses header sync on that connection until it reconnects.

None of that is wanted. This removes v6 outright, so header sync speaks exactly one protocol.

## Solution

Header sync now negotiates **stream kind 5, version 7 only**, correlated by a non-zero `u64` request ID on `GetHeaders`/`Headers`. A peer that speaks only v6 no longer negotiates header sync at all; its connection, gossip, discovery, and block sync are unaffected.

**Protocol surface**
- `ZAKURA_HEADER_SYNC_STREAM_VERSION` bumped in place to `7` (the same hard-break pattern as the previous 5 → 6 bump in `c17be0099`); the `_V7` constant is gone.
- Sole capability bit is `1 << 4`, renamed to `ZAKURA_CAP_HEADER_SYNC`. Bit `1 << 1` is retired and commented as not-for-reuse.
- Sole discovery service id is `zakura.header_sync.v2`.
- Transport registry reverted to its pre-#115 single-version form.

**Request IDs are non-optional in the type system.** `ExpectedHeadersResponse.request_id` and the correlated events/actions carry a plain `HeaderSyncRequestId`, so a correlated frame cannot be constructed without one. This matters: the old no-arg `encode()`/`encode_frame()` defaulted to `request_id: None`, and once the version constant moved they would have returned `Err(MissingRequestId)` for **every** `GetHeaders` and `Headers` — compiling cleanly and failing only at runtime, with a live caller in `handler.rs`. Those helpers are deleted rather than repointed.

**Discovery compatibility (important).** `SUMMARY_TAG_HEADER_SYNC_V1` hard-binds summary tag `1` to whatever `ZakuraServiceId::header_sync()` returns. Simply re-pointing that at the v2 id would make every `Services` message from a pre-upgrade peer a `MismatchedSummaryService` wire violation **in both directions**, taking service discovery down network-wide rather than just header sync. Instead, tag 1 stays bound to a retained `SERVICE_ID_HEADER_SYNC_RETIRED`, and the live summary moves to a **new tag 4**, which peers that predate it ignore via the unknown-tag path.

**Ordered-stream reopen is retained.** Its trigger was never v6-specific: a header-sync stream can die while its connection is healthy (the responder parks it at capacity or with no demand — `reactor.rs` admission park and `handler.rs` no-demand park), and nothing else reopens it. The exit channel also prunes `opened_kinds`/`accepted_ordered_kinds`, without which a legitimate reopen trips `duplicate_stream` and kills the whole connection, taking block sync and gossip down with header sync.

### Behaviour changes worth calling out

- **Covered and timed-out ranges retire their request ID instead of tearing down the stream.** A late response is matched to the retired ID and dropped, so it cannot be mistaken for newer work. The peer stays connected and usable, and the v6 "known limitation" in the changelog disappears.
- **Unsolicited `Headers` are no longer a reactor-level misbehavior.** The peer-owned pipe fails closed on an ID the session never issued and protocol-rejects before the reactor sees anything. Coverage moved to the pipe unit test and the real-QUIC hostile test; the assertion was removed from the simulated reactor matrix rather than left asserting a dead path.
- **Peers may have several header requests in flight**, since requests are now unambiguous (`available_slots()` no longer pins v6 to one).

Net: **-967 lines** across 18 files.

### Tests

Passing:

- `cargo test -p zakura-network --lib` — **942 passed, 0 failed**, including the cluster e2e tests over real QUIC.

The e2e cluster tests caught a real defect during this work: the simulated transport was ferrying `GetHeaders`/`Headers` through the uncorrelated `WireMessage` path, so header sync silently stopped converging. It is now correlation-aware end to end, with an assertion in `inject` so the wrong path cannot be used by accident.

New/updated coverage:

- `retired_header_sync_summary_from_a_pre_upgrade_peer_stays_valid` — a pre-upgrade peer's `Services` message must not become a wire protocol violation.
- `local_correlation_map_drains_commands_and_matches_by_request_id` — replaces the v6 FIFO test; reservations are answered out of order and duplicates drop without scoring.
- `ordered_stream_reopen_*` — retained, retargeted to explain that reopen recovers from park/stream death, not from a v6 quirk.
- `unsupported_stream_version_is_reset_and_never_delivered` — the retired v6 is now just another unsupported version: reset, while the current version is delivered on the same connection.

Not run: the docker `header-faults` e2e lane (`ZAKURA_E2E_MODE=header-faults`) — docker is unavailable in my environment. The `session_id`/`request_id`/`stream_version` trace columns its oracle asserts on are unchanged, so it should still hold, but **it has not been executed and should be before merge.**

Pre-existing failures, not caused by this change: `zakura --test acceptance config_tests` and `pruned_storage_mode_prunes_during_regtest_sync`. Both fail/hang identically on unmodified `main` in a separate worktree; `pruned_storage` is a single-node regtest RPC test with no peers and therefore no header-sync surface at all.

- Builds directly on #115 (already squash-merged to `main` as `a28c8f67f`).
- ZCA-777 (correlate v2 header responses with their requests).

### Follow-up Work

- Run the `header-faults` docker lane in CI to confirm the trace oracle.
- Deleting v6 frees capability bit `1 << 1`; it is deliberately left unclaimed, since peers that have not upgraded still advertise it.

### AI Disclosure

- [x] AI tools were used: Claude Code performed the v6 removal, the test migration, and this description. Two defects in my own first pass were caught in review and fixed: the discovery summary-tag break described above, and an incorrect removal of the ordered-stream reopen machinery. The contributor has reviewed and verified the change as described.

### PR Checklist

- [x] The PR title follows conventional commits format (`!` marks the breaking wire change).
- [x] The solution is tested.


evan-forbes:
sg, yeah we definitely do not want to be advertising multiple headersync reactors as we're not actually running multiple under the hood, and managing the different behaviors across the same state would be quite messy even with two reactors. 

for all reactors, strictly run a single version. if we want to run two reactors, then we need to create at least one new quic stream, and encode logic such that we never run both for the same peer

p0mvn:
@cursor review

evan-forbes (approved):

PR #133, opened 2026-07-13
fix(state): stop serving fabricated commitment metadata

evan-forbes:
## Motivation

Pruned databases retain historical note-commitment trees after deleting raw block transaction data. For heights below the VCT serving-index upgrade marker, root serving derives note-tree roots from those retained trees but needs the block body for shielded transaction counts and the ZIP-244 auth-data root.

When that body was unavailable, both tree-derived producers substituted zero counts and an all-zero auth-data root. This made the response look complete and allowed Zakura header sync to send fabricated commitment metadata to a peer.

There is no linked issue yet; this draft captures the reproduced bug and proposed fix for maintainer review.

## Solution

- Stop tree-derived root production at the first missing block body instead of fabricating metadata.
- Apply the same defensive behavior to both implementations.
- Add a regression test modeling an upgraded pruned database with retained trees, unavailable bodies, and historical heights below the index marker.
- Update existing producer fixtures to store real block bodies and assert their real auxiliary metadata.

The existing header-sync coverage checks require one aligned root for every served header. A short root result is therefore converted immediately into an empty response; it does not wait for the inbound response timeout or serialize non-empty headers with an empty roots vector. Another fanout peer can answer the range.

### Tests

- `CXXFLAGS='-include cstdint' cargo test -p zakura-state --lib` — 346 passed, 2 ignored
- `CXXFLAGS='-include cstdint' cargo test -p zakura served_header_tree_aux_roots_require_complete_coverage`

The `CXXFLAGS` workaround supplies the standard `<cstdint>` declarations missing from the bundled RocksDB build in this environment.

- ZIP-221 history-tree leaf inputs
- ZIP-244 transaction auth-data commitment

### Follow-up Work

- Add root authentication against committed headers separately; this PR only ensures a node never invents missing auxiliary inputs.
- Consider advertising root coverage bounds in native discovery to avoid selecting peers for ranges they cannot serve.

### AI Disclosure

- [x] AI tools were used: OpenAI Codex was used to investigate the serving path, implement the fix and regression tests, run validation, and draft this PR description.

### PR Checklist

- [x] The solution is tested.


zakura-state/src/service/finalized_state/commitment_aux.rs
```diff
@@ -464,32 +464,29 @@ pub(crate) fn produce_block_roots(
...
+        let Some(block) = db.block(height.into()) else {
+            metrics::counter!("state.block_roots.missing_body").increment(1);
+            static MISSING_BODIES: AtomicU64 = AtomicU64::new(0);
+            let occurrences = MISSING_BODIES.fetch_add(1, Ordering::Relaxed) + 1;
+            if occurrences.is_power_of_two() {
+                tracing::error!(
+                    ?height,
+                    occurrences,
+                    "stopping tree-aux root serving because the finalized block body is unavailable"
+                );
```
ValarDragon, line 480: somewhat odd tracing design pattern, but is fine

ValarDragon (approved):
LGTM to me, as a straight forward fix.TL:DR pruned nodes don't store old tx counts by pool per block, or auth data roots, so they can't serve the aux-data.

Later on this will become a capability in header sync I imagine, just like supported block range in block sync?

p0mvn (approved):

evan-forbes:
fixing conflicts then will merge

PR #134, opened 2026-07-13
fix(network): contain legacy peer parser panics

ValarDragon:
## Motivation

A panic during synchronous legacy TCP message decoding can escape the peer
connection task instead of becoming a normal malformed-message failure. The
known V5 Orchard reserved-flag trigger was root-fixed in #121, but the same
availability risk can recur if a future deserializer, conversion, or message
formatter panics on peer-controlled bytes.

This change was requested during the legacy peer panic-containment audit. It is
defense-in-depth for P2P availability and does not alter consensus validation or
state mutation.

## Solution

- Catch unwinds around the complete synchronous legacy message-body decoding
  boundary, after the body is isolated and the codec state is reset.
- Include transaction deserialization and `Transaction -> UnminedTx`
  conversion, full-block deserialization and wrapping, and the other legacy
  body parsers in that boundary.
- Separately contain structured `p2pv2up` prelude decoding, which happens after
  the legacy codec returns its payload.
- Convert a caught panic into `SerializationError::Parse`. The existing
  `Connection::fail_with` path records the error, closes the affected peer's
  channels and transport, and removes only that peer.
- Record a bounded `peer.message.parse.panics` metric and an error log without
  assigning peer-misbehavior points.

The boundary deliberately stops before the connection's shared Tower service
and does not catch encoder, generic connection-task, generic handshake-task, or
heartbeat-task panics.

Heartbeat behavior remains:

- Normal heartbeat timeouts, channel errors, and bad/missing responses update
  the address book with `MetaAddr::new_errored` and disconnect that peer.
- Client-requested shutdown updates the address book with
  `MetaAddr::new_shutdown`.
- An actual heartbeat task panic continues to propagate as an internal
  invariant failure. Catching it only in `Client::poll_heartbeat` could leave
  stale `new_ping_sent` address-book state because that layer cannot perform
  the required asynchronous repair.

There are no public, `pub(crate)`, or production function-signature changes.

### Tests

Passed on commit `faa1db267`, rebased onto current `origin/main`:

- `cargo doc -p zakura-network --no-deps --locked`
- `cargo test -p zakura-network --lib --locked -- --skip listener_bans_zcashd_compat_peer_before_reserved_slot --skip listener_reserves_one_zcashd_compat_inbound_slot --skip listener_zcashd_compat_reconnect_bypasses_recent_ip_limit` — 952 passed, 0 failed, 4 ignored, 3 filtered

Focused regressions cover:

- body-dispatch panic to terminal parse error without unwinding the Tokio task;
- post-deserialization transaction conversion, including the location of the
  prior Orchard-triggered panic;
- full-block post-deserialization conversion;
- structured upgrade-prelude decoding;
- serialization error to the normal single-peer connection failure path; and
- preserved propagation of generic connection and heartbeat task panics.

The unfiltered library suite produced the same 952 passes plus three
environment-only failures because this macOS host cannot bind the synthetic
loopback source aliases used by those unchanged tests (`AddrNotAvailable`, OS
error 49).

- Requested during the legacy peer panic-containment audit.
- #121 root-fixes the known reserved V5 Orchard flag input.
- Zakura's development and release profiles require `panic = "unwind"`.

### Follow-up Work

If heartbeat panic containment is desired later, it should catch inside the
heartbeat async wrapper, retain an address-book updater, publish `new_errored`,
and then return a normal peer failure. That broader internal-task policy is
intentionally outside this parser-focused PR.

### AI Disclosure

- [x] AI tools were used: OpenAI Codex audited the panic and ownership
  boundaries, implemented the parser-only containment and regressions, reviewed
  heartbeat/address-book behavior, and drafted this PR description under human
  direction.

### PR Checklist

- [x] This change was discussed with the team during the legacy peer
  panic-containment audit.
- [x] The solution is tested.

---

> [!NOTE]
> **Medium Risk**
> Changes sit on the hot legacy P2P decode and handshake path, but scope is limited to panic containment with no consensus or API changes; mis-handling could affect peer disconnect behavior.
> 
> **Overview**
> **Defense-in-depth for P2P availability:** panics while parsing peer-controlled legacy TCP traffic no longer unwind the connection or handshake tasks. Only the offending peer is torn down via the existing serialization/parse failure path.
> 
> The legacy **`Codec`** now runs the full synchronous message-body decode (including block/tx deserialization and post-conversion wrapping) inside **`catch_unwind`** after the body is isolated and decode state is reset. A caught panic becomes **`SerializationError::Parse`**, increments **`peer.message.parse.panics`** (`legacy_body`), and logs before disconnect.
> 
> **Zakura `p2pv2up` prelude** decoding uses the same pattern via **`decode_upgrade_prelude` / `decode_upgrade_prelude_with`**, mapping panics to a terminal parse error instead of a handshake-task panic.
> 
> Tests cover injected body-dispatch and post-conversion panics, upgrade-prelude panic mapping, and connection run-loop behavior on inbound parser errors. The changelog documents the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1807135968532d519631b51a65374d87bf52d5cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
-### Fixed
-
-- Contained panics while decoding peer-controlled legacy messages to the
-  affected connection, which is disconnected through the normal peer failure
-  path.
-
 Initial release of Zakura.
 
 Zakura is a fork of the Zcash Foundation's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
+### Fixed
+
+- Contained panics while decoding peer-controlled legacy messages to the
+  affected connection, which is disconnected through the normal peer failure
+  path.
+
 Initial release of Zakura.
 
 Zakura is a fork of the Zcash Foundation's
--- a/zakura-network/CHANGELOG.md
+++ b/zakura-network/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Contained panics while decoding peer-controlled legacy messages to the
+  affected connection, which is disconnected through the normal peer failure
+  path.
+
 ### Added
 
 - Added `HeaderSyncEvent::VctRootRepairRequested` and
--- a/zakura-network/src/peer/connection/tests/vectors.rs
+++ b/zakura-network/src/peer/connection/tests/vectors.rs
@@ -347,6 +347,42 @@ async fn connection_run_loop_inbound_close() {
     assert_eq!(outbound_message, None);
 }
 
+/// Test that an inbound parser error takes the normal peer failure path.
+#[tokio::test]
+async fn connection_run_loop_inbound_parser_error() {
+    let _init_guard = zakura_test::init();
+
+    // The real stream and sink are from a split TCP connection,
+    // but that doesn't change how the state machine behaves.
+    let (mut peer_tx, peer_rx) = mpsc::channel(1);
+
+    let (connection, client_tx, mut inbound_service, mut peer_outbound_messages, error_slot) =
+        new_test_connection();
+
+    peer_tx
+        .try_send(Err(SerializationError::Parse("test peer parser panic")))
+        .expect("test peer channel should accept a parser error");
+
+    let connection = connection.run(peer_rx).shared();
+    let connection_guard = connection.clone();
+    assert_eq!(connection.now_or_never(), Some(()));
+
+    let error = error_slot
+        .try_get_error()
+        .expect("peer parser error should be stored");
+    assert_eq!(
+        error.inner_debug(),
+        "Serialization(Parse(\"test peer parser panic\"))"
+    );
+    assert!(client_tx.is_closed());
+    assert!(peer_tx.is_closed());
+
+    inbound_service.expect_no_requests().await;
+
+    std::mem::drop(connection_guard);
+    assert_eq!(peer_outbound_messages.next().await, None);
+}
+
 /// Test that the connection run loop fails correctly when the peer channel is dropped
 /// (We're not sure if tokio closes or drops the TcpStream when the TCP connection closes.)
 #[tokio::test]
--- a/zakura-network/src/peer/handshake.rs
+++ b/zakura-network/src/peer/handshake.rs
@@ -50,7 +50,8 @@ use crate::{
     types::MetaAddr,
     zakura::{
         P2pV2Upgrade, P2pV2UpgradeAccept, P2pV2UpgradeInit, P2pV2UpgradeReject,
-        ZakuraHandshakeConfig, ZakuraLegacyNonces, ZakuraPeerId, PRELUDE_MAGIC,
+        ZakuraHandshakeConfig, ZakuraLegacyNonces, ZakuraPeerId, ZakuraProtocolError,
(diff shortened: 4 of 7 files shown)


ValarDragon:
We are explicitly changing the semantics from: https://github.com/ZcashFoundation/zebra/issues/3199

This is intentional, I think the trade-offs they picked at the time don't match what I think is the correct decision to do right now.

The big thing to be wary of in review, which I've had both V12, GPT 5.6 Sol Ultra, and Fable help cross check, is that we can't have any partially applied block state in non-finalized consensus logic not get cleaned up / get panic-caught here. Thinking through more convincing tests to add in for this as well.

The peer state is as-is with the tests here pretty convincing imo as safe.

ValarDragon:
(AI comment)
A panic in a legacy peer connection task cannot expose a partially applied non-finalized block commit.

The production code path that would go through is:

  peer task
    → buffered inbound service
    → independent downloader/verifier
    → buffered state service
    → dedicated state-writer thread
    → detached candidate chain
    → complete published snapshot

  Why that specific path is safe:

  - The connection task never owns &mut NonFinalizedState; the mutable state is exclusively owned by a dedicated OS thread in /private/tmp/zakura-legacy-panic-fix/zakura-state/src/
    service/write.rs:199.

  - If the peer panics before dispatching work, nothing is committed. If work has already crossed the channel, the independent state writer may finish the complete commit.
  - commit_block first validates and modifies a detached clone/fork of the parent chain. It only replaces the live chain after success in /private/tmp/zakura-legacy-panic-fix/zakura-
    state/src/service/non_finalized_state.rs:346.

  - Chain::push performs incremental mutations, but consumes that detached candidate. An error or panic drops the candidate rather than exposing it in /private/tmp/zakura-legacy-panic-
    fix/zakura-state/src/service/non_finalized_state/chain.rs:360.

  - Readers receive a complete cloned snapshot through the watch channel; they do not share a mutably updated object.

  However, a panic inside the state-writer thread can occur after mutating its private state. In that case:

  - the writer terminates;
  - its unpublished private value is dropped;
  - readers retain the last complete snapshot;
  - subsequent writes fail with WriteTaskExited.

  That is fail-stop availability loss, not continued operation on a half-mutated chain.

  I also found two reasons we must not claim universal panic atomicity:

  1. Publishing the snapshot and updating the chain-tip channel are separate operations. A panic between them can leave a complete new snapshot with stale tip notification.
  2. There is a concrete existing edge case: invalidating the sole non-finalized chain can return success, after which update_latest_chain_channels() panics because best_chain() is
     empty. Readers retain the old snapshot and the writer dies. This is unrelated to peer parsing, but disproves the blanket claim.

  Finally, catching the entire connection-task panic is broader than catching parser panics. The connection task can poll an inbound read future, so this could turn a systemic database/
  read invariant panic into a peer disconnect. Such reads cannot partially mutate NonFinalizedState, but masking them may still be undesirable.

  So the peer PR’s state-corruption concern checks out specifically for block commits, but its rationale should say “independently owned state writer and staged publication,” not that
  everything executed by the connection task is necessarily peer-local.

---------------

Given this, I'm going to fix this two highlighted known gaps first

ValarDragon:
ok I think the path forward is to only panic catch on the areas we can prove are safe in the legacy stack, and separately try to fix more of the stack to be panic-safe.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
+### Fixed
+
+- Contained panics while decoding peer-controlled legacy messages to the
+  affected connection, which is disconnected through the normal peer failure
+  path.
+
```
p0mvn: I think Sean aimed for an ampty main CHANGELOG for the initial release

```suggestion
```

p0mvn (approved):

PR #138, opened 2026-07-13
fix(sync): retry missing header ranges indefinitely

evan-forbes:
## Motivation

Header sync can permanently strand a missing range after peer or local failures. The fixed-fanout scheduler recorded assignment history rather than explicit work ownership, so repeated timeouts, disconnects, partial external coverage, or local commit backpressure could leave useful peers idle while finalized progress stopped.

This draft preserves the requester implementation for continuous-sync testing and design review. Zakura does not currently use issues for this work.

## Solution

- Replace fixed-fanout bookkeeping with a bounded, single-owner header work queue whose explicit states are pending, wire-in-flight, buffered, and committing.
- Retry normal forward and backward work indefinitely, with short peer-local avoidance after timeout, empty/short/invalid responses, send failure, or disconnect.
- Fill bounded per-peer v7 request slots from shared work and retire request IDs before returning timed-out claims.
- Synchronously prepare the frame and response reservation, install a `Publishing` outstanding request, and assign its range before handing blocking transport publication to the generation-gated peer-local requester.
- Roll back the outstanding request, range assignment, and pipe reservation if requester enqueue fails.
- Reduce requester events to publication completion and requester stop notifications. Successful publication transitions `Publishing` to `AwaitingResponse`; an early response remains valid and cannot be recreated by a late completion.
- Preserve publication deadlines for requesters blocked on outbound capacity, while keeping global scheduling and retries reactor-owned.
- Buffer valid out-of-order responses while committing only from the durable contiguous frontier.
- Preserve local buffered/committing ownership across peer disconnects and correlate commit completion with the admitted session generation.
- Trim overlapping pending/in-flight/buffered work when `FullBlockCommitted`/`NewBlock` advances the durable frontier. Already validated buffered suffix vectors are preserved rather than re-downloaded; partially covered failed commits requeue their uncovered suffix.
- Clear a short response's inherited suffix anchor if ordered drain later rejects its parent prefix, including for backward work.
- Reserve action-channel capacity before marking work committing, so a full driver queue cannot orphan a commit.
- Use one-second empty-response pacing, prune expired retry avoidance before calculating maintenance, and replace the permanent 50 ms tick with deadline-driven wakeups.
- Add queue state/byte/stall observability and pin the driver's all-or-nothing commit-completion contract.

### Audit regressions

- Two resident batches plus partial external coverage recreate the interior suffix instead of wedging ordered drain.
- Partial coverage trims and commits an already buffered suffix without re-downloading it.
- A partially covered local commit failure requeues the uncovered suffix.
- A publishing request blocked on outbound capacity expires, cancels its session, and becomes available to another peer.
- A response received before publication completion is accepted, and the later completion does not reinstall it.
- Requester queue failure cancels the prepared pipe reservation and rolls back reactor ownership.
- Closed transport reports publication failure, retries the range, stops the requester, and disconnects the peer.
- Empty-response retry bookkeeping cannot be extended by a late publication-completion event and does not run at 20 Hz.
- Ordered-drain rejection clears a dependent suffix anchor; wrong-fork buffered work is rejected before commit.
- Expired retry avoidance cannot leave a maintenance deadline in the past.
- A saturated 128-entry action queue preserves the buffered range and emits exactly one commit after capacity returns.
- Truncated finalized responses explicitly request their suffix, and that suffix still enforces its checkpoint hash.
- `Height(u32::MAX)` one-header ranges, 4096-entry seen-hash eviction, and forward queue reset/index rebuilding have direct coverage.

### Tests

- `CXXFLAGS='-include cstdint' cargo test -p zakura-network header_sync::` — 158 passed
- `CXXFLAGS='-include cstdint' cargo test --workspace --exclude zakura` — passed, including 978 `zakura-network` library tests with 4 ignored; the node-launching `zakura` package was excluded
- Real-QUIC `header-faults` coverage for silence, delayed/out-of-order responses, empty/short responses, disconnect, and failover. The clean second-run trace corpus passed the direct oracle before this follow-up; the final post-follow-up rerun remains rollout evidence.

- Based on the v7-only protocol merged in #128: mandatory non-zero request IDs and explicit response correlation.
- Requester lifecycle refactor head: `b0e7f5b7a`.
- The ownership and retry invariants follow block sync, but the queue remains header-specific.
- Detailed audit handoff and regression evidence are maintained in the Valar artifact inbox for this work.

### Follow-up Work

- The serving-side bulk-read and concurrency changes are split into #139, stacked on this branch.
- Exercise this requester draft on the continuous-sync canary with preserved state and injected silent peers.

### AI Disclosure

- [x] AI tools were used: OpenAI Codex implemented the requester queue, sequencing, tests, observability, validation, audit fixes, lifecycle refactor, and draft PR description under developer direction.

### PR Checklist

- [x] This change was discussed with a core Zakura developer; no issue is used because issues are not the project's current workflow.
- [x] The solution is tested.
- [x] The documentation and changelog are up to date.

### DIFF
--- a/zakura-network/src/zakura/header_sync/requester.rs
+++ b/zakura-network/src/zakura/header_sync/requester.rs
@@ -28,8 +28,16 @@ pub(super) enum HeaderRequesterFailure {
     Encode,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub(super) enum HeaderRequesterEvent {
+    Register {
+        peer: ZakuraPeerId,
+        session_id: u64,
+        generation: u64,
+        command: HeaderRequesterCommand,
+        request_id: HeaderSyncRequestId,
+        registered: tokio::sync::oneshot::Sender<()>,
+    },
     Sent {
         peer: ZakuraPeerId,
         session_id: u64,
@@ -42,6 +50,7 @@ pub(super) enum HeaderRequesterEvent {
         session_id: u64,
         generation: u64,
         command: HeaderRequesterCommand,
+        request_id: Option<HeaderSyncRequestId>,
         reason: HeaderRequesterFailure,
     },
     Stopped {
@@ -94,15 +103,65 @@ async fn run_header_requester(
             }
         };
 
+        let prepared = match session.prepare_get_headers(
+            command.range.start_height,
+            command.expected_max_count,
+            command.range.want_tree_aux_roots,
+        ) {
+            Ok(prepared) => prepared,
+            Err(error) => {
+                let reason = if matches!(error, OrderedSendError::Closed) {
+                    HeaderRequesterFailure::Closed
+                } else {
+                    HeaderRequesterFailure::Encode
+                };
+                if events
+                    .send(HeaderRequesterEvent::Failed {
+                        peer: peer.clone(),
+                        session_id,
+                        generation,
+                        command,
+                        request_id: None,
+                        reason,
+                    })
+                    .is_err()
+                    || reason == HeaderRequesterFailure::Closed
+                {
+                    return;
+                }
+                continue;
+            }
+        };
+        let request_id = prepared.request_id();
+        let (registered, registration) = tokio::sync::oneshot::channel();
+        if events
+            .send(HeaderRequesterEvent::Register {
+                peer: peer.clone(),
+                session_id,
+                generation,
+                command,
+                request_id,
+                registered,
+            })
+            .is_err()
+        {
+            return;
+        }
+        let registered = tokio::select! {
+            biased;
+            _ = shutdown.cancelled() => return,
+            _ = cancel.cancelled() => return,
+            registered = registration => registered,
+        };
+        if registered.is_err() {
+            return;
+        }
+
         let result = tokio::select! {
             biased;
             _ = shutdown.cancelled() => return,
             _ = cancel.cancelled() => return,
-            result = session.send_get_headers(
-                command.range.start_height,
-                command.expected_max_count,
-                command.range.want_tree_aux_roots,
-            ) => result,
+            result = prepared.send() => result,
         };
         match result {
             Ok(request_id) => {
@@ -131,6 +190,7 @@ async fn run_header_requester(
                         session_id,
                         generation,
                         command,
+                        request_id: Some(request_id),
                         reason,
                     })
                     .is_err()
@@ -222,7 +282,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn successful_send_reports_correlated_request_metadata() {
+    async fn request_waits_for_reactor_registration_before_publication() {
         let peer = peer(1);
         let (session, mut frames, _cancel) = session(peer.clone(), 1);
         let (events_tx, mut events_rx) = mpsc::unbounded_channel();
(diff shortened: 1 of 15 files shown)


zakura-network/src/zakura/header_sync/work_queue.rs
```diff
@@ -0,0 +1,467 @@
+use super::{state::*, wire::*, *};
+
+#[derive(Clone, Debug, Default)]
+pub(super) struct HeaderHashDedup {
```
p0mvn, line 4: Would be good to add some unit tests for non-regression

p0mvn: Not sure if zakura-network/src/zakura/header_sync/tests.rs covers enough

evan-forbes: yeah agree 

should be good on https://github.com/zakura-core/zakura/pull/138/commits/5c6d96e0648f82f8c5cdb2d2d9a3952e2d5cf5da

zakura-network/src/zakura/header_sync/requester.rs
```diff
@@ -0,0 +1,378 @@
...
+                };
+                command
+            }
+        };
+
+        let result = tokio::select! {
+            biased;
+            _ = shutdown.cancelled() => return,
+            _ = cancel.cancelled() => return,
+            result = session.send_get_headers(
```
p0mvn: Could we make reactor-side request registration happen before this frame becomes visible to the peer, or otherwise serialize registration and responses through one ordered channel? `send_get_headers` publishes the frame before the requester emits `Sent`, while `HeaderSyncEvent`s and requester events use separate channels and the reactor selects wire events first. A fast response can therefore reach `handle_headers` before `OutstandingRange` is installed, be dropped as an unknown request ID, and leave the later-installed request to time out. A regression test that queues both events before polling the reactor should expose the ordering issue.

evan-forbes: nice https://github.com/zakura-core/zakura/pull/138/commits/f6667ff215c79c03ea0aee1d12e6dbb850388761

evan-forbes:
fixing ci and going through everything again now

p0mvn (commented):
## Review finding

**P1/P2: failed status publication can make the reactor spin continuously**

`next_maintenance_deadline()` returns `now` when a peer's status meter is already due (`reactor.rs:2126-2157`). However, `refresh_statuses()` only advances the unsolicited/keepalive meters after `try_send_status()` succeeds (`reactor.rs:2477-2533`, `2574-2621`).

If the peer's outbound channel is full or closed, the send fails and leaves `next_allowed` in the past. The next reactor iteration therefore computes another immediately-ready `sleep_until`, runs maintenance again, retries the same send, and repeats without yielding to a timer. A closed channel can sustain this indefinitely; a full channel spins until capacity returns. This can consume a core, inflate maintenance metrics/logging, and reduce reactor fairness. The previous fixed 50 ms tick implicitly bounded these retries.

Please add retry pacing or a capacity-triggered wakeup. If using backoff, keep failed-attempt timing separate from the successful-send meter so a failed publication is not recorded as delivered. A regression test should hold the outbound channel full/closed and verify that maintenance wakeups remain bounded while the status remains eligible for a later retry.

p0mvn (approved):
LGTM on design.

Makes perfect sense to create a more robust state machine that tracks header range ownership for reassigning and retrying the work.

evan-forbes:
the bug is fixed but there's still a more subtle issue around if the connection closes and we want to avoid looping 20 times a second until its properly closed by the peer lifecycle manager

we can just close immediately without waiting, which feels cleaner to me. However doing this consistently would fall out of scope for this changed, therefore opening an issue with as much context as we need to continue should we want to later

PR #139, opened 2026-07-13
fix(sync): serve header ranges without blocking sync

evan-forbes:
## Motivation

Pipelined v7 peers can issue several header-range requests concurrently, but the original serving path serialized state reads through the header-sync driver and used one-shot outbound enqueue. This makes serving concurrent and bounded end to end.

Stacked behind #138 so serving is reviewable apart from requester liveness. Zakura does not use issues for this work.

## Solution

The first two commits are the original author's work, rebased:

- **Serve header ranges without blocking sync** — one read-only `HeaderSyncRange` state request returning aligned headers, hashes, advisory body sizes, and optional commitment roots; one cloned non-finalized chain plus one RocksDB snapshot shared by bounded iterators; height-aligned stream-merge with consensus/full-block precedence and first-gap stop, replacing per-height point reads.
- **Bound header range serving end to end** — one absolute serving deadline carried through action enqueue, read admission, state work, preparation, and outbound enqueue; a global read bound over actual state-work and materialized-response lifetime; idempotent completion tokens and RAII guards; terminal-result observability.

The rest are review fixes. Each has a test verified to fail without it.

### Correctness

- **Stop fabricating roots on the header-range path.** When a block body was unavailable (pruned while trees were retained), the fallback served `(0, 0, 0)` transaction counts and an all-zero auth-data root. This is the #133 defect reappearing in a new parallel path. The roots are now left unset so the caller reports incomplete roots.
- **Report header-serve completion from an unpolled task.** The completion guard was built *inside* `tokio::spawn`, so a task the runtime dropped before its first poll — what happens to every pending task at shutdown — never built one. `HeaderRangeResponseToken` has no `Drop`, so the token was dropped unfinished and the driver's `finished()` wait parked forever holding a read permit.
- **Release the serving slot when admission never reports it.** Same shape, worse consequence: the slot is claimed before the spawn, and serving requires request ids to strictly increase, so a lost slot is never reclaimed for the life of the session. Enough of them fill the peer's cap, and the reactor then reports an **honest peer** for `GetHeadersSpam`.
- **Report serve completeness honestly.** `roots_complete` was read after the rejection path emptied the vectors, and empty roots trivially match empty headers — so every rejected response reported itself as a complete serve.

### Bounds and cost

- **Own the serve tasks.** They were detached, so the driver returned immediately on shutdown while state reads kept running against the database, and the endpoint reported the header-sync lifecycle stopped when it had not. They now live in a `JoinSet` that is reaped each iteration and drained on shutdown.
- **Make the read limit configurable.** It was a hardcoded 32, and pre-PR the driver awaited serially — a step change in wire-triggered blocking-pool and snapshot pressure with no operator control. New `network.zakura.header_sync.max_concurrent_range_reads` (default 32, floored at 1; a configured `0` would silently disable serving). Existing configs are unaffected.
- **Scope root derivation to the heights that need it.** The guards now run before any body read, so a range needing one root reads one block body instead of all of them.
- **Stop paying for columns the request drops.** `body_size` for non-finalized rows was a full re-serialization per block per request; the chain already records exactly that value at push time. `headers_by_height_range` was reimplemented on the roots-capable path and opened two extra CF iterators plus a `BlockInfo` decode per height only to discard them — a regression on a request type this refactor didn't need to touch. Requests now name the optional columns they read (`HeaderRangeColumns`).

## Testing

`cargo fmt --all -- --check`, `cargo clippy -p zakura-state -p zakura-network -p zakura --all-targets` clean; `zakura-state` 362, `zakura-network` 1004, `zakurad` 306 tests pass.

Every fix has a test verified to fail against the unfixed code by in-place revert, not by assertion alone:

- The unpolled-task tests drop the future without polling and assert the token is released; against the previous shape they fail with the exact production symptom — a `finished()` wait that never resolves.
- The concurrency test fails against the hardcoded cap with all 6 requests reading at once instead of the configured 2.
- The completeness test drives a driver-rejected `IncompleteRoots` through the reactor and asserts the emitted trace. The pure-function tests cannot catch that regression: moving the computation below the body-sizes clear leaves them green, and this is the only test in 1004 that turns red.
- The non-finalized shortcut asserts `chain_commitment_roots` directly, because the range fallback backfills any row the shortcut leaves unset — so a shortcut that silently stopped deriving would still return correct rows, at exactly the cost it exists to avoid.

## Notes for reviewers

Two things worth knowing, both corrections to claims made earlier in this branch's commit messages:

- The `service.rs` root fabrication that `fix(state): stop fabricating roots` removes was **dead code**, not a live bug. `Chain::block()` and `contains_block_height()` read the same map on the same snapshot, so the `.unwrap_or(zeros)` arm was unreachable. That commit's message overstates it; the change is behaviour-preserving robustness.
- An earlier version of the derivation-scoping comment justified the skip loop with a "prefix" argument that is **unsound** — the roots index can have holes below the upgrade marker. The real invariant is that rows ascend without duplicates and a drain consumes only entries at or below its height. Already corrected in the follow-up commit.

Deliberately **not** changed, and verified as intended rather than defective:

- Abandoned reads keep their semaphore permit past the deadline (`timed_out_header_reads_keep_the_real_concurrency_slot`) — the slot reflects real in-flight work.
- Serve admission waits on the actions channel rather than `try_send` (`full_action_queue_waits_without_manufacturing_an_empty_response`) — backpressure must not look like an unavailable range.

Known follow-ups, out of scope here:

- **Headers-only requests cannot be served.** `validate_tree_aux_roots_len` requires `roots.len() == headers.len()` strictly, on both encode and decode, so a `Headers` message with headers and no roots cannot be represented at all — the `has_roots` wire marker is dead for non-empty responses. This PR changes the failure from a silent clear to a counted encode error, but does not fix it. Latent either way: our requester always asks with roots.
- `commitment_aux.rs` range-iterator rewrite (belongs against main).
- `sync.header.serve.last_roots_complete` is a single global gauge written by every per-request completion across all peers; it belongs as a label on the `serve.finished` counter.
- The per-peer inflight cap is advertised on the wire, so re-tuning it against the global read limit needs its own compatibility review.


evan-forbes:
closing this for now as we'll likely refactor a lot of things in the future for headersync

that being said, many of the ideas here are not terrible

PR #142, opened 2026-07-13
fix(state): use retained hashes for block locators

evan-forbes:
## Motivation

Pruned checkpoint sync can permanently stall after a sync round restarts.

Checkpoint sync intentionally omits old transaction bodies while retaining finalized block height/hash indexes. Block locator construction was using the body-gated hash lookup, so a node whose locator window contained only body-pruned blocks produced an empty locator. Peers then responded from genesis, and the node discarded every returned hash as already finalized.

This behavior was introduced when `hash_by_height()` changed from the retained hash index to `db.body_hash()` in `64f405b4ee`.

No issue is linked yet; this draft captures the prototype while the team discussion is still in progress.

## Solution

Separate chain identity from block-body availability:

- construct block locators from the retained finalized hash index with `db.hash(height)`
- recognize finalized chain intersections through `db.height(hash).is_some()`
- keep the general `hash_by_height()` lookup body-gated for callers that advertise serveable block bodies
- add a regression test whose finalized tip/hash index exists while its transaction body is absent

This lets pruned nodes advertise their actual finalized tip after a round restart without claiming that pruned block bodies are serveable.

### Tests

- `CXXFLAGS='-include cstdint' cargo test -p zakura-state`
  - 352 passed
  - 3 ignored
  - 0 failed

The `CXXFLAGS` workaround force-includes the standard `<cstdint>` header required by bundled RocksDB under the local host compiler.

- Regression introduced in `64f405b4ee`
- Zcash `getblocks` / block locator chain-intersection behavior

### Follow-up Work

Before marking ready for review:

- link the relevant issue or team discussion
- decide whether this bug fix needs an `[Unreleased]` changelog entry
- run any additional full-workspace checks requested by maintainers

### AI Disclosure

- [x] AI tools were used: OpenAI Codex was used for artifact diagnosis, implementation, regression-test construction, local validation, and drafting this PR description.

### PR Checklist

- [x] The solution is tested.
- [ ] The documentation and changelogs are up to date.

### DIFF
--- a/zakura-state/src/service/read/find.rs
+++ b/zakura-state/src/service/read/find.rs
@@ -239,7 +239,7 @@ where
     chain
         .map(|chain| chain.as_ref().height_by_hash.contains_key(&hash))
         .unwrap_or(false)
-        || db.contains_hash(hash)
+        || db.height(hash).is_some()
 }
 
 /// Create a block locator from `chain` and `db`.
@@ -272,7 +272,11 @@ where
     let mut hashes = Vec::with_capacity(heights.len());
 
     for height in heights {
-        if let Some(hash) = hash_by_height(chain, db, height) {
+        let hash = chain
+            .and_then(|chain| chain.as_ref().hash_by_height(height))
+            .or_else(|| db.hash(height));
+
+        if let Some(hash) = hash {
             hashes.push(hash);
         }
     }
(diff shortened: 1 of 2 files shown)


zakura-state/src/service/read/find.rs
```diff
@@ -239,7 +239,7 @@ where
     chain
         .map(|chain| chain.as_ref().height_by_hash.contains_key(&hash))
         .unwrap_or(false)
-        || db.contains_hash(hash)
+        || db.height(hash).is_some()
```
ValarDragon, line 242: Somewhat odd these aren't equivalent

ValarDragon (approved):
LGTM, but makes me concerned about more `hash_by_height` breakage, seems like we should've redesigned that query to still work or rename & break each usage.

p0mvn (approved):

evan-forbes:
#142 is the urgent sync-stall fix, but it is necessary rather than sufficient for pruned mode. The same body-availability gate still affects shared chain-identity lookups: without the follow-up, pruned nodes can truncate peer getblocks responses and report incorrect depth/confirmations for canonical historical blocks. Phase 1 of that systemic cleanup is now in stacked draft #146; it also deletes the misleading composite gates so this class is harder to reintroduce.

p0mvn:
Ref on merge: https://zcashzcaloors.slack.com/archives/C0BG341Q9TQ/p1783982839582119

PR #144, opened 2026-07-13
fix(state): omit VCT roots when block bodies are missing

p0mvn:
## Summary
- stop tree-derived VCT root serving at the first unavailable block body
- avoid pairing valid commitment roots with fabricated zero transaction counts and auth-data roots
- add regression coverage for missing bodies and update tree-root fixtures to use real body-derived auxiliary data

## Test plan
- [x] `cargo test -p zakura-state commitment_aux::tests --lib`
- [x] verify no linter errors in the modified file

### DIFF
--- a/zakura-state/src/service/finalized_state/commitment_aux.rs
+++ b/zakura-state/src/service/finalized_state/commitment_aux.rs
@@ -464,32 +464,29 @@ pub(crate) fn produce_block_roots(
         // `ironwood_root_or_empty`'s doc comment).
         let ironwood_root = ironwood_root_or_empty(db.ironwood_tree_by_height(&height));
         // Below the upgrade height the serving index does not exist, so derive the
-        // auth-data root and the shielded tx-counts from the locally stored block (this
-        // archival node holds the body for these heights). Zero only if the body is somehow
-        // absent, in which case the recipient simply re-fetches from a node that has it.
-        let block = db.block(height.into());
-        let (sapling_tx, orchard_tx, ironwood_tx, auth_data_root) =
-            if let Some(block) = block.as_ref() {
-                (
-                    block.sapling_transactions_count(),
-                    block.orchard_transactions_count(),
-                    block.ironwood_transactions_count(),
-                    block.auth_data_root(),
-                )
-            } else {
-                metrics::counter!("state.block_roots.zero_aux_fallback").increment(1);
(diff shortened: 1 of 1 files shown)


zakura-state/src/service/finalized_state/commitment_aux.rs
```diff
@@ -464,32 +464,29 @@ pub(crate) fn produce_block_roots(
...
+            static MISSING_BODIES: AtomicU64 = AtomicU64::new(0);
+            let occurrences = MISSING_BODIES.fetch_add(1, Ordering::Relaxed) + 1;
+            if occurrences.is_power_of_two() {
+                tracing::error!(
+                    ?height,
+                    occurrences,
+                    "stopping tree-aux root serving because the finalized block body is unavailable"
+                );
+            }
+            break;
```
evan-forbes, line 482: this looks like the exact fix in #133 fwiw

zakura-state/src/service/finalized_state/commitment_aux.rs
```diff
@@ -661,6 +659,43 @@ mod tests {
...
+        for height in heights {
+            let test_body = test_block();
+            let transaction_hashes: Arc<[_]> =
+                test_body.transactions.iter().map(|tx| tx.hash()).collect();
+            let height_byte =
+                u8::try_from(height).expect("test heights fit in a byte for hash fixtures");
+            let finalized = FinalizedBlock::from_checkpoint_verified(
+                CheckpointVerifiedBlock(SemanticallyVerifiedBlock {
+                    block: test_body,
+                    hash: block::Hash([height_byte; 32]),
```
evan-forbes, line 681: very very minor thing and not blocking

while I get that the hash here isn't that important, we just need something there, its also very easy to use the real hash as well, which imo makes the testing code a smidge more general and realistic.

I could see how the mismatch could be confusing without reading and remembering this line, since both just look like hashes. one might assume we're somehow hitting some serialization issue etc when in reality were just setting an arbitrary hash

evan-forbes (commented):
could be missing something but this seems like a duplicate of #133 

if so, that's nbd except 133 is a bit more defensive via also fixing `zakura_db/block.rs::finalized_commitment_roots_by_height_range`, albiet that code isn't actually called atm afaiu

#133 also added an e2e test, although that might be overkill

will defer to you on which to merge

PR #146, opened 2026-07-13
fix(state): separate chain identity from body availability

evan-forbes:
## Motivation

PR #142 fixed the urgent empty-block-locator stall, but it routed around a broader state-layer semantic bug: identity-named helpers still treat a finalized block as absent when its transaction body has been pruned.

Pruned storage deliberately retains the finalized height/hash indexes and headers. Chain-identity queries must use those indexes; body availability must remain an explicit, separate question. Otherwise pruned nodes truncate `getblocks` responses, report canonical historical blocks with incorrect depth/confirmations, retry stale backup blocks, and fail to reject semantic commits at already-finalized heights through the intended duplicate path.

This PR builds on merged PR #142 by extending its pruned checkpoint fixture and routing block locator construction back through the corrected shared helper, replacing the temporary inline workaround.

## Solution

- make `read::find::hash_by_height` use the retained finalized hash index
- make `read::find::depth` use the retained hash-to-height index
- restore the semantic-commit duplicate guard to finalized-height membership
- ignore backup blocks whose hashes are already in the finalized index
- delete `ZakuraDb::body_hash` and `ZakuraDb::contains_hash`, whose identity-like names encoded body-availability gates
- route block locator construction through the now-correct shared `hash_by_height` helper
- preserve the separate semantics of provisional header-only rows
- extend the pruned checkpoint regression to cover hash lookup, depth, block locators, and peer `getblocks` hash collection while the body remains absent
- add a real-peer reverse-risk characterization proving that an advertised pruned block returns `Block(None)`, maps to missing inventory, and produces a wire `notfound`
- update the two yanked `spin` versions (`0.9.8` and `0.10.0`) to compatible patch releases (`0.9.9` and `0.10.1`), including their matching deny/vet policy entries; this unrelated `main`-baseline CI unblock is kept in a separate commit

The actual block-serving path is unchanged: it still checks body availability and returns `notfound` for a pruned body.

### Tests

- `cargo fmt --all -- --check` — passed after rebasing onto current `main`
- `CXXFLAGS='-include cstdint' cargo clippy --workspace --all-targets -- -D warnings` — passed after rebasing onto current `main`
- `CXXFLAGS='-include cstdint' cargo clippy -p zakura --all-targets -- -D warnings` — passed after adding the reverse-risk characterization
- `CXXFLAGS='-include cstdint' cargo test -p zakura-state` — passed after rebasing onto current `main` (352 passed, 3 ignored)
- `CXXFLAGS='-include cstdint' cargo test -p zakura components::inbound::tests::real_peer_set` — passed (8 passed), including the pruned-body wire `notfound` characterization
- `CXXFLAGS='-include cstdint' cargo check --workspace --locked` — passed with `spin 0.9.9` and `0.10.1`
- `cargo deny --workspace check bans --allow unmatched-skip-root` — passed with the updated `spin 0.10.1` duplicate-version exception
- `cargo vet check --locked` — passed after reviewing both `spin` patch deltas and advancing their existing `safe-to-deploy` exemptions
- `CXXFLAGS='-include cstdint' cargo test --workspace` — state, unit, pruned-regtest, and 48/49 default acceptance tests passed before the rebase; `activate_mempool_mainnet` timed out after four minutes because the live mainnet peer set did not advance past genesis. This network-dependent failure is unrelated to the state changes.
- `pruned_storage_mode_prunes_during_regtest_sync` exceeded 210 seconds locally before the timing run was stopped. It now has a dedicated post-merge CI profile and runs only on pushes to `main`, so it guards pruned sync end-to-end without blocking pull requests or the merge queue.

The `CXXFLAGS` workaround force-includes the standard `<cstdint>` header required by bundled RocksDB under the local host compiler.

- Builds on merged PR #142
- Root regression introduced in `64f405b4ee`
- Phase 1 of the pruned body-gate audit

### Follow-up Work

- keep the retained-transaction-hash RPC policy and VCT historical-tree error classification in separate follow-ups

### AI Disclosure

- [x] AI tools were used: OpenAI Codex was used for audit interpretation, implementation, regression-test updates, rebase/conflict resolution, local validation, and drafting this PR description.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -63,7 +63,7 @@ retries = 2
 # no-fail-fast without hiding fast unit failures.
 [profile.full-tests]
 failure-output = "immediate"
-default-filter = "not test(check_no_git_dependencies) and not test(=fully_synced_rpc_z_getsubtreesbyindex_snapshot_test) and not test(=lwd_rpc_test) and not test(=lwd_rpc_send_tx) and not test(=lwd_grpc_wallet) and not test(=lwd_integration) and not test(=lwd_sync_full) and not test(=lwd_sync_update) and not test(=lightwalletd_test_suite) and not test(=rpc_get_block_template) and not test(=rpc_submit_block) and not test(~generate_checkpoints_) and not test(=sync_one_checkpoint_mainnet) and not test(=sync_one_checkpoint_testnet) and not test(=sync_update_mainnet) and not test(=activate_mempool_mainnet) and not test(=restart_stop_at_height) and not test(=sync_large_checkpoints_mempool_mainnet) and not test(=config_tests) and not test(~zakura::testkit::cluster) and not test(~zakura::testkit::blocksync_fuzz)"
+default-filter = "not test(check_no_git_dependencies) and not test(=fully_synced_rpc_z_getsubtreesbyindex_snapshot_test) and not test(=lwd_rpc_test) and not test(=lwd_rpc_send_tx) and not test(=lwd_grpc_wallet) and not test(=lwd_integration) and not test(=lwd_sync_full) and not test(=lwd_sync_update) and not test(=lightwalletd_test_suite) and not test(=rpc_get_block_template) and not test(=rpc_submit_block) and not test(~generate_checkpoints_) and not test(=sync_one_checkpoint_mainnet) and not test(=sync_one_checkpoint_testnet) and not test(=sync_update_mainnet) and not test(=activate_mempool_mainnet) and not test(=restart_stop_at_height) and not test(=sync_large_checkpoints_mempool_mainnet) and not test(=pruned_storage_mode_prunes_during_regtest_sync) and not test(=config_tests) and not test(~zakura::testkit::cluster) and not test(~zakura::testkit::blocksync_fuzz)"
 
 # Retry the real-time Zakura block-sync scenario tests under load (see the note on the
 # `all-tests` override above).
@@ -93,6 +93,11 @@ retries = 2
 slow-timeout = { period = "5m", terminate-after = 1 }
 failure-output = "immediate"
 default-filter = 'package(zakura) and test(=restart_stop_at_height)'
+
+[profile.zakura-pruned-storage]
+slow-timeout = { period = "15m", terminate-after = 1 }
+failure-output = "immediate"
+default-filter = 'package(zakura) and test(=pruned_storage_mode_prunes_during_regtest_sync)'
 
 [profile.zakura-integration]
 fail-fast = false
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2093,7 +2093,7 @@ dependencies = [
  "futures-sink",
  "nanorand",
  "pin-project",
- "spin 0.9.8",
+ "spin 0.9.9",
 ]
 
 [[package]]
@@ -2218,7 +2218,7 @@ dependencies = [
  "diatomic-waker",
  "futures-core",
  "pin-project-lite",
- "spin 0.10.0",
+ "spin 0.10.1",
 ]
 
 [[package]]
@@ -2611,7 +2611,7 @@ dependencies = [
  "hash32",
  "rustc_version",
  "serde",
- "spin 0.9.8",
+ "spin 0.9.9",
  "stable_deref_trait",
 ]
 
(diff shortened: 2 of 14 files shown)


zakura-state/src/service/read/find.rs
```diff
@@ -272,11 +272,7 @@ where
     let mut hashes = Vec::with_capacity(heights.len());
 
     for height in heights {
-        let hash = chain
-            .and_then(|chain| chain.as_ref().hash_by_height(height))
-            .or_else(|| db.hash(height));
-
-        if let Some(hash) = hash {
+        if let Some(hash) = hash_by_height(chain, db, height) {
```
evan-forbes, line 275: then nice thing about this change is that we can go back to just calling hash_by_height, effectively reverting the change we just made

zakura-state/src/service/read/find.rs
```diff
@@ -151,7 +151,7 @@ where
     let tip = tip_height(chain, db)?;
     let height = chain
         .and_then(|chain| chain.as_ref().height_by_hash(hash))
-        .or_else(|| db.contains_hash(hash).then(|| db.height(hash)).flatten())?;
+        .or_else(|| db.height(hash))?;
```
evan-forbes, line 154: we can simply go straight to asking by height instead of checking if things exist

zakura-state/src/service/read/find.rs
```diff
@@ -219,7 +219,7 @@ where
 
     chain
         .and_then(|chain| chain.as_ref().hash_by_height(height))
-        .or_else(|| db.body_hash(height))
+        .or_else(|| db.hash(height))
```
evan-forbes, line 222: key fix

return the hash even if the block was pruned, as we're actually still keeping the hash:height map after pruning!!

.config/nextest.toml
```diff
@@ -63,7 +63,7 @@ retries = 2
...
+default-filter = "not test(check_no_git_dependencies) and not test(=fully_synced_rpc_z_getsubtreesbyindex_snapshot_test) and not test(=lwd_rpc_test) and not test(=lwd_rpc_send_tx) and not test(=lwd_grpc_wallet) and not test(=lwd_integration) and not test(=lwd_sync_full) and not test(=lwd_sync_update) and not test(=lightwalletd_test_suite) and not test(=rpc_get_block_template) and not test(=rpc_submit_block) and not test(~generate_checkpoints_) and not test(=sync_one_checkpoint_mainnet) and not test(=sync_one_checkpoint_testnet) and not test(=sync_update_mainnet) and not test(=activate_mempool_mainnet) and not test(=restart_stop_at_height) and not test(=sync_large_checkpoints_mempool_mainnet) and not test(=pruned_storage_mode_prunes_during_regtest_sync) and not test(=config_tests) and not test(~zakura::testkit::cluster) and not test(~zakura::testkit::blocksync_fuzz)"
```
evan-forbes, line 66: added the pruned_storage_mode_prunes_during_regtest_sync to run after we merge to main / merge queue

evan-forbes (commented):
added two more tests

.config/nextest.toml
ValarDragon: delegating review of this CI change to Roman

zakura-state/src/service/read/find.rs
ValarDragon: Much simpler, nice

.config/nextest.toml
p0mvn: lgtm

p0mvn (approved):

zakura-state/src/service.rs
```diff
@@ -824,7 +824,7 @@ impl StateService {
         if self
             .read_service
             .db
-            .contains_body_at_height(semantically_verified.height)
+            .contains_height(semantically_verified.height)
```
ValarDragon, line 827: theres no risk of writing here with just a header?

Cargo.lock
```diff
@@ -6854,18 +6854,18 @@ dependencies = [
 
 [[package]]
 name = "spin"
-version = "0.9.8"
+version = "0.9.9"
```
ValarDragon, line 6857: why did the lock file update?

zakura-state/src/service.rs
ValarDragon: I guess it makes sense that a pruned node's duplicate check doesn't depend on block data. But are these DB entries written atomically for non-pruned nodes?

ValarDragon: Ah yes, this is only written when finalized.

PR #147, opened 2026-07-13
fix(network): ignore inbound source ports for address book identity

p0mvn:
## Summary
- Stop treating inbound direct TCP source endpoints as durable address-book identities.
- Keep transient connection identity behavior intact and add regression coverage for every `ConnectedAddr` variant.

## Test plan
- `cargo test -p zakura-network connected_addr_`
- `cargo test -p zakura-network peer::handshake`
- `cargo test -p zakura-network`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2093,7 +2093,7 @@ dependencies = [
  "futures-sink",
  "nanorand",
  "pin-project",
- "spin 0.9.8",
+ "spin 0.9.9",
 ]
 
 [[package]]
@@ -2218,7 +2218,7 @@ dependencies = [
  "diatomic-waker",
  "futures-core",
  "pin-project-lite",
- "spin 0.10.0",
+ "spin 0.10.1",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


p0mvn:
Measurement update: findings are inconclusive so far.

We tried passive peer-cache monitoring on the fixed legacy node and an old-commit legacy control node, including a run after emptying `mainnet.peers` on both. The nodes are running and the peer caches refill, but the passive cache shape is too noisy: both nodes accumulate non-default-port entries, likely via gossip/canonical-address paths as well as any inbound behavior. The direct metric (`active_inbound_newly_cached_since_baseline`) has remained `0` in the samples so far, so passive observation has not caught a currently active inbound source endpoint being cached on either node.

To prove the bug, we should stop relying on passive cache shape and do the controlled test: make known inbound legacy connections to both nodes, record the exact source `ip:port`, and check whether those exact entries appear in `mainnet.peers`.

ValarDragon:
Closing for now, we'll re-encounter / imagine more carefully when doing our refined pass on handshake audit. If it comes up again in stability we will fix then

PR #148, opened 2026-07-13
fix(memory): limit glibc malloc arenas

p0mvn:
## Motivation

A mainnet sync run on `temp-zakura-sync-test-7` reached approximately 32 GiB RSS with sustained memory and I/O pressure. Health probes repeatedly exceeded their 8-second timeout even though the process remained active.

The investigation separated live block memory from allocator retention:

- An eBPF allocation profile identified block-body deserialization—not headers—as the allocation workload, dominated by transaction byte vectors, Orchard actions, raw block buffers, and transaction collections.
- Calling `malloc_trim(0)` while the node remained active reduced RSS from 12.1 GiB to 3.0 GiB. The block-sync wire budget changed by only about 100 MB during that interval, showing that most reclaimed memory had already been freed by the application but retained in glibc arenas.
- The process had hundreds of Tokio worker threads, allowing glibc to spread short-lived, variable-sized allocations across many arenas and retain fragmented pages.

Header-sync concurrency exposed the latent allocator behavior by increasing sustained body throughput, but reducing header concurrency did not address the cause.

## Solution

- On Linux/glibc only, call `mallopt(M_ARENA_MAX, 4)` before worker threads start.
- Respect an operator-provided `MALLOC_ARENA_MAX` environment variable instead of overriding it.
- Keep the header-sync concurrency default at 10.
- Record the allocator parameter and operational evidence in `CHANGELOG_PARAMS.md`.

macOS and musl builds are unchanged.

### Operational validation

The affected node was restarted with `MALLOC_ARENA_MAX=4` and header concurrency 10. Over 30 minutes it:

- synced 1,057,436 blocks (~578 blocks/s, ~2.08M/hour),
- held process RSS near 3.7 GiB,
- accumulated a 61,108-block applying backlog without allocator amplification,
- reported zero sustained memory or I/O PSI,
- remained healthy throughout.

This compares with 12.1 GiB RSS after roughly 21 minutes under the earlier allocator default, and a previous peak near 32 GiB.

### Tests

Passed:

- `cargo check -p zakura --bin zakurad --locked`
- `markdownlint CHANGELOG_PARAMS.md --config .trunk/configs/.markdownlint.yaml`
- `git diff --check`
- IDE diagnostics for changed files
- 30-minute mainnet operational validation described above

Linux CI provides direct compilation coverage for the glibc-gated path; the local build host was macOS.

- Operational node: `temp-zakura-sync-test-7` (`dual`, mainnet)
- Header request IDs and concurrency: #115
- Header v7-only transition: #128

### Follow-up Work

- Continue long-duration RSS observation on Linux/glibc.
- Consider instrumented, threshold-triggered `malloc_trim` only if arena limiting does not remain sufficient over longer runs.
- Treat bounded/lazy block decoding as a separate cross-platform optimization rather than part of this minimal allocator fix.

### AI Disclosure

- [x] AI tools were used: Cursor assisted with incident diagnosis, live profiling, operational validation, implementation, verification, and PR description.

### PR Checklist

- [x] This change was discussed during operational incident response.
- [x] The focused build and formatting checks pass.
- [x] The parameter changelog is up to date.


zakurad/src/commands/start/zakura/block_sync_driver.rs
```diff
@@ -1192,7 +1192,10 @@ where
 
     BlockApplyCompletion {
         class,
-        checkpoint_refresh_floor: (class == BlockApplyClass::Checkpoint
+        // Probe applies never reach state, so a delayed refresh only adds state reads and makes
+        // probe tests race the refresh timer.
+        checkpoint_refresh_floor: (throughput_probe.is_none()
```
p0mvn, line 1197: This suppresses the delayed checkpoint-frontier refresh for synthetic probe applies. The refresh timer could fire while the test was flushing traces, causing an unexpected `FinalizedTip` read and making the test depend on CI timing.

evan-forbes (approved):

p0mvn:
To recap:
- MALLOC_ARENA_MAX=4 was observed to keep RSS limited while improving throughput
- Tried malloc_trim(0) -> Reclaimed: 9.1 GiB (75%). This confirms that the problem was fragmented and unreleased to OS memory, rather than a leak 

The initial AI recommendation set it to 4. This parameter is probabilistic. We cannot say 4 arenas translate to max X Gb RSS and Y reduction in throughput. It is selected by tuning. The concern was that by dropping it this low, we were improving the performance around light blocks 2M+ while making it worse for the early heights.

This ended up being true. Compared `MALLOC_ARENA_MAX=8,16,64` over the first 100K heights. 16K is the best. 8 was 15% less throughput.

I also ruled out header sync update as the cause of memory problems by running an A/B over 100K heights.

On throughput, a lower parameter can lead to a major improvement when memory is fragmented (2M+ heights), but having it too low also introduces thread contention fighting for space at larger allocations (early heights).

I’ve run into OOM a couple of more times on the default config, especially during sandblast.

Based on all of the above, the parameter of MALLOC_ARENA_MAX=16 has shown the greatest balance, so I have merged it.

PR #155, opened 2026-07-14
fix(consensus): classify wrapped commit errors as duplicate requests

ebfull:
## Motivation

Duplicate-block commit errors from the state were losing their classification on the semantic verification path. The state returns `CommitSemanticallyVerifiedBlock` failures boxed as `CommitSemanticallyVerifiedError` — a newtype around `CommitBlockError` — but the semantic block verifier only tried a direct `downcast_ref::<CommitBlockError>()`, which can never match the wrapper. Every commit error, including `CommitBlockError::Duplicate`, fell through to the generic `VerifyBlockError::StateService` variant, where `is_duplicate_request()` returns `false` and `duplicate_location()` returns `None`.

Downstream, this makes benign duplicate commits (e.g. an inbound-gossip/syncer race on the same block near tip) look like real verification failures: `should_restart_sync` restarts the syncer instead of continuing, and the Zakura block-apply path reports `Rejected` instead of `Duplicate`.

The checkpoint verifier had the same gap in `is_duplicate_request()`: its `duplicate_location()` already unwrapped `CommitCheckpointVerifiedError`, but `is_duplicate_request()` did not handle the `CommitCheckpointVerified` variant at all.

This is an availability fix only — no block-acceptance rule changes.

## Solution

- Add an `inner()` accessor to the state's `CommitSemanticallyVerifiedError` and `CommitCheckpointVerifiedError` newtypes.
- Route semantic commit failures through a new `map_commit_error` helper that unwraps `CommitSemanticallyVerifiedError` (with a fallback to a bare `CommitBlockError`) before classifying the error as `VerifyBlockError::StateService`.
- Teach `VerifyCheckpointError::is_duplicate_request()` to unwrap `CommitCheckpointVerifiedError`, matching its existing `duplicate_location()`.

### Tests

- New regression test on the semantic path (`block::tests::state_commit_duplicate_errors_are_duplicate_requests`): boxes a duplicate commit error exactly the way the state's request handler does and asserts it maps to `VerifyBlockError::Commit` with `is_duplicate_request() == true`, the correct `duplicate_location()`, and a zero misbehavior score.
- New regression test on the checkpoint path (`checkpoint::tests::state_commit_duplicate_errors_are_duplicate_requests`); verified it fails without the `is_duplicate_request()` fix and passes with it.
- `cargo test --release -p zakura-consensus --lib` (140 passed), `cargo test --release -p zakura-state --lib` (350 passed, 3 ignored), `cargo fmt --all -- --check`, and `cargo clippy --workspace --all-targets -- -D warnings` all pass locally.

### Follow-up Work

None.

### AI Disclosure

- [x] AI tools were used: Claude Code — root-cause investigation, the fix, regression tests, and this PR description (maintainer-directed and reviewed).

### PR Checklist

- [x] The solution is tested.
- [x] The documentation and changelogs are up to date (changelog intentionally untouched pre-1.0.0).

### DIFF
--- a/zakura-consensus/src/block.rs
+++ b/zakura-consensus/src/block.rs
@@ -127,6 +127,25 @@ impl VerifyBlockError {
     }
 }
 
+/// Converts an error from a `CommitSemanticallyVerifiedBlock` state request
+/// into a [`VerifyBlockError`].
+///
+/// The state boxes commit errors as [`zs::CommitSemanticallyVerifiedError`], a
+/// newtype around [`zs::CommitBlockError`], so the wrapper must be unwrapped
+/// here for `is_duplicate_request()` and `misbehavior_score()` to classify
+/// duplicate blocks as benign.
+fn map_commit_error(source: BoxError, hash: block::Hash) -> VerifyBlockError {
+    if let Some(commit_err) = source
+        .downcast_ref::<zs::CommitSemanticallyVerifiedError>()
+        .map(zs::CommitSemanticallyVerifiedError::inner)
+        .or_else(|| source.downcast_ref::<zs::CommitBlockError>())
+    {
+        return VerifyBlockError::Commit(commit_err.clone());
+    }
+
+    VerifyBlockError::StateService { source, hash }
+}
+
 /// The maximum number of transparent signature operations allowed in a block.
 ///
 /// # Consensus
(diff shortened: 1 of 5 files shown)


ValarDragon (approved):

ValarDragon:
Good fix! We should probably revisit this error semantics in the future, this is a sign the error handling and operating on error types is over complicated right now

PR #157, opened 2026-07-14
ci(release): adopt crates.io Trusted Publishing in create-release

ebfull:
## Motivation

The release tag and the crates.io artifacts must come from one commit. The rc3 publish demonstrated the failure mode this design removes: the crates were published by hand from a different commit than the one eventually tagged — and after the rc sweep, those registry versions now correspond to no tag at all. Manual publishing also depends on a long-lived `cargo login` token sitting on a maintainer machine.

This PR makes the `Create release` workflow publish the crate graph itself, from the exact SHA it tags, authenticated via [crates.io Trusted Publishing](https://crates.io/docs/trusted-publishing) (short-lived OIDC-minted tokens, no stored registry credential).

**Draft on purpose:** the mechanics work (see Tests), but I want maintainer eyes on the design decisions below before this becomes the release path.

## Solution

New `publish-crates` job in `create-release.yml`:

- Runs only for **stable (non-hyphenated) tags** — pre-releases never reach crates.io, and hyphenated tags skip the job entirely (no approval prompt is even requested for them).
- Runs **after the GitHub release publishes**, and checks out `target_sha` — the same commit the tag points to.
- Gated by a new **`crates-io` environment** with required reviewers. The environment holds no secrets; it exists so the crates.io Trusted Publishing configuration can be restricted to it, and so a human decides *when* the registry publish happens (intended: after the release is tested, signed, and promoted — registry publishes can only be yanked, never rolled back).

The work itself is in `scripts/publish-crates.sh`, with three modes so each step is separately runnable and locally testable:

1. **`plan`** — derives the publish set from `cargo metadata` (every workspace crate without `publish = false`, so there is no hardcoded crate list to drift), then checks each crate against the sparse index. Versions already on the index are skipped, which makes re-running the workflow resume a partially-published graph. Hard errors: a publishable crate that doesn't exist on crates.io (must be reserved + configured first), and a selected crate whose version differs from the release version.
2. **`verify`** — packages the selected crates with their verify builds. This runs *before* any registry token exists, because the builds take longer than the 30-minute Trusted Publishing token lifetime.
3. **`publish`** — mints the token and publishes everything in one dependency-ordered multi-package `cargo publish --no-verify` (cargo ≥ 1.90; CI is on `stable`). The script refuses pre-release versions no matter how it is invoked, as a second layer under the job-level tag condition.

Also updated:

- `docs/release-tag-protection.md` — new "crates.io Trusted Publishing" section covering the environment, the per-crate crates.io configuration, and the operational notes.
- `.github/PULL_REQUEST_TEMPLATE/release-checklist.md` — the manual `cargo login` + `cargo release publish` loop is replaced by "approve the `crates-io` deployment after promoting", plus the existing `cargo install` verification.

### Design decisions I'd like feedback on

1. **Publish timing.** The job waits on the `crates-io` approval, intended to be granted after test/sign/promote. Consequence: a run left waiting blocks the `create-release` concurrency group until approved or rejected, and unapproved deployments expire after 30 days. The alternative — publishing immediately after the tag — removes the human gate on an irreversible action.
2. **`--no-verify` at publish time.** Verification happens in the `verify` stage minutes earlier on the same checkout; re-verifying inside `cargo publish` would blow the 30-minute token TTL. Is that separation acceptable?
3. **Metadata-derived crate set** vs. an explicit list (as in `check-crate-packaging.sh`). Metadata can't drift, and the plan-time "crate not on crates.io" hard error catches new crates before they half-publish; but it means adding a publishable crate implicitly adds it to the release publish.
4. **Version-mix hard error**: every crate selected for publish must carry the release version; unchanged crates simply stay at their older published version. This matches how partial releases should look on the registry, but it's stricter than the old checklist flow.

### Tests

- `shellcheck` (the CI lint gate for `scripts/*.sh`) passes; the workflow YAML parses; `markdownlint` passes on the changed docs.
- `./scripts/publish-crates.sh plan` run live against crates.io from the current tree: correctly skips the 12 crates already published at `1.0.0-rc3` and selects only `zakura` (currently `1.0.0-rc4`), exercising both the index-skip path and the selection path.
- `cargo package --locked --no-verify -p zakura` packages cleanly; full 13-crate workspace packaging was already validated by the #122 preflight.
- **Not tested end-to-end:** the job can only run for real on a stable tag with the console configuration in place (see Follow-up Work). That, plus the design questions above, is why this is a draft.

- crates.io Trusted Publishing: <https://crates.io/docs/trusted-publishing> (token minted via `rust-lang/crates-io-auth-action`, SHA-pinned)
- Multi-package publish (dependency-ordered, waits for index availability): stabilized in cargo 1.90

### Follow-up Work

Owner-only console setup, required before the job can ever succeed:

- Create the `crates-io` GitHub environment: required reviewer(s), deployments from `main` only, no variables or secrets.
- Add the Trusted Publishing configuration on each of the 13 publishable crates on crates.io: repository `zakura-core/zakura`, workflow `create-release.yml`, environment `crates-io`.

Note: this PR and #156 both append a section to `docs/release-tag-protection.md`; whichever lands second takes a trivial append conflict on rebase.

### AI Disclosure

- [x] AI tools were used: Claude Code (Claude Fable 5) wrote the workflow job, publish script, documentation, and this PR description; design reviewed by the author, who is seeking further review via this draft.

### PR Checklist

- [ ] This change was discussed in an issue or with the team beforehand — this draft PR is that discussion.
- [x] The solution is tested (see Tests; not yet exercised end-to-end on a real stable release).

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -207,17 +207,17 @@ The end of support height is calculated from the current blockchain height:
 
 ## Publish Crates
 
-- [ ] [Run `cargo login`](https://github.com/zakura-core/zakura/dev/crate-owners.html#logging-in-to-cratesio)
-- [ ] It is recommended that the following step be run from a fresh checkout of
-      the repo, to avoid accidentally publishing files like e.g. logs that might
-      be lingering around
-- [ ] Publish the crates to crates.io; edit the list to only include the crates that
-      have been changed, but keep their overall order:
+Stable releases publish the crate graph automatically: the `Publish crates to
+crates.io` job in the same Create release run uses crates.io Trusted
+Publishing (no `cargo login`, no long-lived token) and publishes from the
+exact tagged commit. It skips crates whose version is already on crates.io,
+so re-running the workflow resumes a partial publish. See
+[`docs/release-tag-protection.md`](https://github.com/zakura-core/zakura/blob/main/docs/release-tag-protection.md#cratesio-trusted-publishing).
 
-```
-for c in zakura-test zakura-tower-fallback zakura-jsonl-trace zakura-chain zakura-tower-batch-control zakura-node-services zakura-script zakura-state zakura-consensus zakura-network zakura-rpc zakura-utils zakura; do cargo release publish --verbose --execute -p $c; done
-```
-
+- [ ] After promoting the release, approve the `crates-io` environment
+      deployment on the Create release run.
+- [ ] Check the job log: unchanged crates are skipped as already published,
+      and every selected crate is at the release version.
 - [ ] Check that Zakura can be installed from `crates.io`:
       `cargo install --locked --force --version <version> zakura && ~/.cargo/bin/zakurad`
       and put the output in a comment on the PR.
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,6 +1,8 @@
 # This workflow is the only supported path for creating v* release tags.
 # It builds and verifies every release asset before a protected environment
 # allows the release GitHub App to publish a draft and create the immutable tag.
+# For stable (non-hyphenated) tags it then publishes the crate graph to
+# crates.io from that same commit, behind a second environment approval.
 name: Create release
 
 on:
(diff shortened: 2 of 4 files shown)


ValarDragon:
> Publish timing. The job waits on the crates-io approval, intended to be granted after test/sign/promote

This makes sense, also because we have different signing flows on both.

> --no-verify at publish time. Verification happens in the verify stage minutes earlier on the same checkout; re-verifying inside cargo publish would blow the 30-minute token TTL. Is that separation acceptable?

Yeah, the way this should look is we've verified on our machines already. We should get the verify faster ideally, but not a this week problem.

> (3), (4)

These both sound fine to me, I think its lower cognitive burden for people to get all of these shared in tags. (4) may be worth re-considering when we have much greater API stability. Right now many API's are inherited from Zebra and need more substantial API safety thinking.

ValarDragon:
Superceded by #654

PR #165, opened 2026-07-14
fix: consensus, state-integrity, and sync-liveness fixes with expanded test coverage

ebfull:
## Motivation

This branch collects a batch of correctness fixes found while writing
adversarial test coverage for the newer Zakura-specific paths (V6/Ironwood,
NU6.1–NU6.3, verified-commitment-trees fast sync, and the offline
prune/rollback tooling). The history is organized so the two kinds of change
are easy to review separately: a small number of **test-only** commits that
add coverage without touching logic, and **one commit per behavioral change**,
each self-contained.

Nothing here is believed to be remotely exploitable on Mainnet or Testnet as
shipped — see the classification on each fix below for why. The changes are a
mix of genuine bugs on configured/local networks, consensus-code hardening
that is currently fail-closed elsewhere, and robustness/liveness fixes.

## Behavioral changes and bugfixes

Each fix is one commit. Tagged **🔒 consensus/security-relevant** when the code
it touches is consensus- or security-critical (with a note on why it is not
currently exploitable), or **robustness/liveness** otherwise.

### Consensus- and security-relevant

**🔒 `fix(consensus): validate zero-subsidy lockbox outputs`** — `subsidy_is_valid`
returned early with a zero deferred-pool change whenever the expected block
subsidy was zero, which skips the founders-reward, funding-stream, and NU6.1
one-time ZIP-271 lockbox-disbursement checks entirely. The lockbox disbursement
is a fixed amount that does not scale with the subsidy, so a network activating
NU6.1 at a zero-subsidy height would accept an activation block missing its
mandatory disbursement outputs, and even a compliant block had its deferred-pool
debit reported as zero, corrupting value-pool accounting. The early return is
removed so every subsidy-output rule runs regardless of the amount. *Not
exploitable on Mainnet/Testnet:* both activate NU6.1 at funded (non-zero-subsidy)
heights, so only custom-configured networks are affected — but it is a real
consensus-rule skip, hence the tag.

✅ **🔒 `fix(state): bind vct prevalidation to auth roots`** — the verified-commitment-trees
fast-commit look-ahead cached its prevalidated successor as `(height, hash)`, but
at NU5 and later the block hash does not commit to authorizing data — the header
commitment does, via the auth-data root. A header-only successor witness carries
that root separately from the later body, so a body with the same header hash but
different authorizing data could reuse the witness's prevalidation and skip the
one commitment check that binds its authorizing data to the validated header
chain. The cache is now keyed on `(height, hash, auth_data_root)` and the
committed block's own root must match before the check is skipped (below NU5 the
root is not a commitment input and stays `None`). *Not independently exploitable:*
the fast-sync path is checkpoint-gated, so an attacker cannot substitute an
arbitrary body — but this closes the binding gap directly.

✅ **🔒 `fix: lock checkpoint auth roots to wrapped blocks`** — `CheckpointVerifiedBlock`
exposed `DerefMut` alongside its cached `auth_data_root`, so a holder could swap
the block or the cached root after construction and pair a stale auth-data root
with a different block; a header matching the stale root could then finalize a
block without proving the header binds that block's actual authorizing data.
`DerefMut` is removed and the only legitimate use now goes through
`with_precomputed_auth_data_root()`, which computes the root from the wrapped
block itself. *Not a live bug:* no current caller desynced the pair, so this is
compile-time enforcement of a consensus invariant. (Also drops a since-obsolete
TODO about mempool spent-output ordering, which can no longer occur.)

✅ **🔒 `fix(script): enforce p2sh sigop input alignment`** — `p2sh_sigop_count` pairs
inputs with spent outputs via `zip()`, which silently truncates on a length
mismatch and undercounts P2SH sigops; the block sigop limit is a consensus rule,
so a silent undercount could admit an over-limit block. The alignment invariant
was guarded only by a `debug_assert`, which release builds skip — it is now a hard
`assert` with a documented panic contract. *Not a live bug:* all current callers
build `spent_outputs` positionally per input, so this is release-build hardening
of a consensus check.

✅ **🔒 `fix(chain): reserve V6 Orchard cross-address flag`** — the V6 transaction codec
applied the Ironwood-permissive flag grammar to both shielded slots, so a V6
Orchard bundle carrying the cross-address bit (bit 2) round-tripped through the
wire layer even though only the Ironwood slot defines that bit. The codec now
keeps bit 2 reserved in the V6 Orchard slot on both serialization and
deserialization. *Not an acceptance bypass:* the semantic verifier and the
librustzcash conversion already reject such bundles, and V6/NU6.3 is a future
upgrade — this is wire-format spec conformance, fail-closed downstream.

✅ **🔒 `fix(chain): reject common equihash proofs on regtest`** — `Solution::check()`
accepted a 1344-byte `Common` solution on Regtest and verified it under the
`(200, 9)` parameters, even though Regtest's active Equihash parameters are
`(48, 5)`; a Mainnet-shaped proof must not become valid merely because the caller
supplies a Regtest network. The checker now uses an exact network/variant mapping
so any mismatch is rejected before Equihash verification. *Not a live bypass:*
canonical Regtest disables PoW, so this is conformance with zcashd and safety for
direct callers or a future PoW-enabled Regtest.

### Robustness and liveness

**`fix(state): preserve chunked header root coverage`** — the native header-sync
driver issued a single `BlockRoots` read spanning the whole gap between the
verified-block tip and the best header tip, but the state caps each `BlockRoots`
response at `MAX_HEADER_SYNC_HEIGHT_RANGE`. Once the tip lag exceeded the cap, the
coverage check could never observe a full response and the driver kept restarting
from the verified-block tip. The driver now walks the span in cap-sized chunks,
and `BlocksByHeightRange` is bounded by the same constant at the state boundary so
one request cannot ask the database for an unbounded range.

✅ **`fix(state): fill finalized root ranges across unchanged trees`** —
`block_roots_by_height_range` iterated the per-height Sapling tree column family,
which is sparse (a row exists only where that pool's tree changed). Heights whose
blocks added no Sapling commitments were silently omitted from `BlockRoots`
responses, so the header-sync coverage check saw gaps and could stall. It now
resolves every requested finalized height through the backwards tree lookup, and
resolves Ironwood the same way instead of defaulting an absent tree to the empty
root.

✅ **`fix(network): honor disable_pow in native header sync`** — native header sync's
wire-level PoW gating keyed on Regtest parameters rather than the effective PoW
mode, so a configured Testnet with `disable_pow = true` had its headers rejected
and could never sync, even though semantic and checkpoint verification
intentionally skip Equihash and difficulty checks in that mode. Both helpers now
return early when `disable_pow()` is set; canonical Regtest (which always sets it)
keeps its prior behavior, and PoW-enabled networks are unaffected.

✅ **`fix(chain): satisfy the lockbox rule on local-genesis networks activating NU6.1 or later`**
— `NetworkUpgrade::activation_height()` falls through to the next configured
upgrade, so skipping NU6.1 on generated local testnets did not actually skip the
one-time lockbox-disbursement rule: subsidy validation still fired at the shared
activation height, found no configured disbursements, and rejected every possible
activation block. The generator now activates NU6.1 alongside the other upgrades
with a zero-value P2SH lockbox marker output (satisfying the structural rule
without creating funds), and rejects a requested upgrade that has no compiled
consensus branch ID instead of building a network that panics downstream. Affects
the local-testnet generation tool only.

**`fix(state): harden offline prune and rollback integrity`** — three integrity
fixes in the operator-run offline prune/rollback commands: (1) moving the pruning
marker backwards now requires each affected height to fully reconstruct through
the marker-aware block reader, so a partially missing body cannot be reclassified
as retained history; (2) rollback's deferred-pool recomputation no longer skips
heights inside the slow-start interval, where the fixed NU6.1 lockbox disbursement
still debits the deferred pool (skipping it corrupted value-pool accounting on
networks activating NU6.1 during slow start); (3) tree-index pruning deletes
subtrees by exact key instead of an `index..index + 1` range, which wrapped in
release builds for the supported `u16::MAX` subtree index and left that subtree in
place. Also replaces a block-size `as u32` cast with a checked conversion.

**`fix(state): harden finalized disk format checks`** — truncated finalized-state
rows could be misread as complete rows of an older format: a `BlockInfo` row
between the NU6.1 and Ironwood widths was parsed as legacy (reinterpreting the
first Ironwood amount bytes as the block size), and a partial
`CommitmentRootsByHeight` row was parsed as a pre-Ironwood layout with defaulted
roots and counts. Both readers now accept only the exact legacy widths (plus
forward-compatible extensions of the current width) and panic with the offending
length otherwise, surfacing local database corruption instead of silently
mangling value pools or serving zeroed commitment roots. The genesis-roots
upgrade check also now distinguishes the benign VCT fast-sync race from a
genuinely missing genesis tree.

✅ **`fix(state): remove all forks when invalidating shared root`** — invalidating a
block at a chain's non-finalized root removed only the chain whose tip matched, so
sibling forks that also contained the invalidated block survived and the node
could keep extending a chain built on it. Invalidation now retains only chains
that do not contain the invalidated hash, removing every fork rooted in that block
together. Reachable via the operator-facing invalidate request, not remotely.

✅ **`fix(state): reject unknown archive storage-mode config keys`** —
`[storage_mode.archive]` deserialized its table value as `IgnoredAny`, so a
pruned-only or misspelled setting placed under the archive table was silently
ignored rather than failing config parsing. It now deserializes as an empty
`deny_unknown_fields` struct, so such configs are rejected at startup.

**`fix(chain): keep sapling commitment constructors strict`** — Sapling
`ValueCommitment` had two constructors that accepted points its own deserializer
rejects: `FromHex` accepted any decodable byte string, and
`From<jubjub::ExtendedPoint>` encoded whatever point it was handed. Both now
enforce the canonical, not-small-order requirement (the hex path returns an error;
the `From` conversion documents and asserts its contract), so internal callers
cannot build commitments that wire deserialization would reject. Neither path
parses untrusted input.

✅ **`fix(chain): reject invalid Sapling transmission key bytes`** —
`TransmissionKey::try_from` is a fallible parse but called
`AffinePoint::from_bytes(bytes).unwrap()`, panicking on an off-curve `pk_d`
encoding instead of returning `Err`, contradicting the `TryFrom` contract. It now
returns an error. No production path reaches this conversion with untrusted bytes,
so this is API hardening rather than a crash fix.

✅**`fix(state): share vct safety state across clones`** — `VctCommitState`'s
prevalidation cache and frozen-frontier flag were plain fields, so each
`FinalizedState` clone received an independent snapshot: a clone could miss that
another froze the frontier and fall back to legacy recomputation, or miss a
prevalidation update. The production node has a single finalized writer, so this
was a latent hazard rather than a live bug; both fields are now shared across
clones so the clone contract holds regardless of caller topology.

✅ **`fix(state): avoid overflow when a subtree range starts at an excluded u16::MAX`**
— the shared subtree-range read helper computed an excluded start bound as
`start_index.0 + 1`, which panics in debug builds (and wraps in release) when the
excluded bound is `u16::MAX`. It now uses checked arithmetic and returns an empty
subtree map for such a range.

## Tests

<details>
<summary>The remaining commits are test-only and can be skipped by most reviewers — expand for the coverage summary.</summary>

These commits add coverage without touching runtime logic. Where a test lives in
a non-test file, the only non-test changes are doc-comment corrections (noted
below); everything else is inside `#[cfg(test)]` / test modules or test-vector
data.

- ✅ **`test(chain): strengthen history and note-commitment tree coverage`** — history
  tree version selection across upgrade epochs, the reset at NU6.3 activation, V3
  extension/pruning/cloning, and parts-API parity with genesis and header
  commitments; block-level Orchard/Ironwood accessor pool assignment; parallel
  per-pool tree update independence and ordering; and batched Orchard appends at
  exact tree capacity.

- ✅ **`test: pin network parameters and cover consensus admission paths`** — pins
  NU6.3 activation parameters and its consensus branch ID against librustzcash and
  BIP-70 network names per network kind; covers cross-pool nullifier collisions in
  spend-conflict checks, the public NU6.3 branch-ID admission boundary, checkpoint
  verification of generated local seed chains, and the quoted NU6.3
  activation-height config key in `zakurad`.

- ✅ **`test(chain): cover V5/V6 wire format and ZIP-244 digest paths`** — V6
  Orchard-style flag rules on the wire and the finalized V6 version group ID; V6
  Ironwood note-recovery routing; and V5 NU6.3 native ZIP-244 txid/auth digests
  pinned against the librustzcash oracle, plus the native digest wrappers and V6
  bundle slots. *(Non-test change: corrects stale doc comments on the
  txid/auth-digest helpers.)*

- ✅ **`test: cover lazy Sapling commitment paths and batch capacity edges`** — lazy V4
  Sapling output and spend commitment verification paths in the transaction
  verifier, the fallible Sapling binding-key helper, and Sapling incremental-tree
  batch capacity plus the Ironwood flag edge case.

- ✅ **`test(chain): cover Ironwood pool isolation and value balance edges`** — Ironwood
  value balances stay isolated from the Orchard pool, including add/subtract edge
  cases, plus Regtest Equihash proof-binding coverage.

- ✅ **`test(state): cover pruned-chain reads across the read service`** — best-chain
  depth, hash, and header reads (and the read-service request vectors) keep serving
  finalized blocks whose bodies were pruned, since pruning removes body rows but
  keeps the verified chain-identity indexes.

</details>

## Test evidence

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
and `cargo test --workspace` all pass locally (release profile).

## AI disclosure

Used Claude Code to rebase and reorganize the branch history, to review the
changes, and to draft this PR description. All code and tests are authored by the
contributor, who is the sole responsible author.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Initial release of Zakura.
 
 ### Fixed
 
+- Native Zakura header sync now honors `disable_pow = true` on configured
+  Testnets, matching semantic and checkpoint block verification.
 - Pruned finalized blocks remain visible to chain-identity queries, including peer
   block-hash responses and RPC confirmation lookups, after their bodies are removed.
 
--- a/docs/design/verified-commitment-trees.md
+++ b/docs/design/verified-commitment-trees.md
@@ -412,10 +412,9 @@ construction_ — but the public API previously let it be desynced after constru
 (`pub auth_data_root`, `DerefMut`, both re-exported). A holder could swap the block while
 keeping a stale root, and a header matching the stale root would finalize a block without
 proving the header binds the block's actual authorizing data. The (block, auth-data-root) pair
-is now locked together: `auth_data_root` is `pub(crate)`, `CheckpointVerifiedBlock` drops
-`DerefMut`, the one legitimately-post-set field goes through
-`set_deferred_pool_balance_change`, and the semantic verifier builds blocks through
-`from_semantic_data` (auth-data root left unset). Compile-time enforced (fix in commit #192).
+is now locked together: `CheckpointVerifiedBlock` drops `DerefMut`, and the checkpoint
+verifier can only fill the optional cache through `with_precomputed_auth_data_root`, which
+computes the value from that same wrapped block rather than accepting arbitrary bytes.
 
 ## 7. The fast commit path and checkpoint last checkpoint height
 
--- a/docs/design/verified-commitment-trees.md
+++ b/docs/design/verified-commitment-trees.md
@@ -393,9 +393,12 @@ The ZIP-221 MMR does not authenticate everything, so three gaps are closed by di
 
 A block's own commitment check `C(X, T_{X-1})` is the _identical_ computation the previous
 fast block already ran as its look-ahead one commit earlier. The committer caches the
-look-ahead result as `(next_height, next_hash)` and skips a block's own check when the prior
-look-ahead validated exactly it. The guard is hash identity and heights are monotonic, so a
-stale or cloned cache entry can never cause a false skip. Steady state drops from two
+look-ahead result as `(next_height, next_hash, next_auth_data_root)` and skips a block's own
+check when the prior look-ahead validated exactly it. Below NU5 the auth-data-root component is
+unused because it is not an input to the header commitment; at NU5 and later it binds a
+header-only successor witness to the later body, so a same-header body with different
+authorizing data cannot reuse the earlier prevalidation. Heights are monotonic, so a stale or
+cloned cache entry can never cause a false skip. Steady state drops from two
 commitment checks per block to one (legacy parity) while still attesting every root before it
 is persisted. A non-last checkpoint height fast block with no buffered successor is deferred by the write
 worker until the successor arrives; the checkpoint last checkpoint height is the only no-successor fast commit
@@ -429,7 +432,8 @@ logic. For a checkpoint-verified block at `height`:
    - apply the direct below-Heartwood/below-NU5/below-Nu6_3 checks (§6.1);
    - build a candidate history tree with the roots folded in (`HistoryTree::push`);
    - **verify-before-commit:** either check the buffered successor's commitment against the
-     candidate (the one-block-lag confirmation) and cache `(height+1, next_hash)` as
+     candidate (the one-block-lag confirmation) and cache
+     `(height+1, next_hash, next_auth_data_root)` as
      pre-validated, or, at the checkpoint last checkpoint height only, verify the embedded final
      frontiers — including Ironwood — against this height's roots; a failure means _this_
      height's root is bad → reject and evict (§8);
--- a/zakura-chain/CHANGELOG.md
+++ b/zakura-chain/CHANGELOG.md
@@ -34,6 +34,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   serialize and deserialize successfully, while unsupported later placeholders
   are rejected.
 
+### Fixed
+
+- Local-genesis networks that activate NU6.1 or later now configure the
+  required zero-value lockbox marker output, so their first semantic
+  activation block can pass subsidy validation. Requests for an upgrade with
+  no compiled consensus branch ID now fail during generation.
+
 ## [9.0.0] - 2026-06-02
 
 ### Added
--- a/zakura-chain/src/block/tests/vectors.rs
+++ b/zakura-chain/src/block/tests/vectors.rs
@@ -5,18 +5,20 @@ use std::{
 };
 
 use chrono::{DateTime, Duration, LocalResult, TimeZone, Utc};
+use halo2::pasta::pallas;
 
 use crate::{
     amount::{Amount, NonNegative, MAX_MONEY},
     block::{
         serialize::MAX_BLOCK_BYTES, Block, BlockTimeError, Commitment::*, Hash, Header, Height,
     },
+    ironwood, orchard,
     parameters::{Network, NetworkUpgrade::*},
     sapling,
     serialization::{
         sha256d, SerializationError, ZcashDeserialize, ZcashDeserializeInto, ZcashSerialize,
     },
-    transaction::{LockTime, Transaction},
+    transaction::{arbitrary::v5_transactions, LockTime, Transaction},
     transparent,
 };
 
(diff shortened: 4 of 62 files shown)


ValarDragon:
Closing as most are through, and remainder can be re-triaged as we keep going through reports!

PR #166, opened 2026-07-14
fix(network): ability to re-connect reactor kicked peers

evan-forbes:
## Motivation

Ordered service demand was only consulted during initial connection setup. If block sync,
header sync, or discovery declined or ended its stream while sibling services kept the QUIC
connection alive, the stream could remain absent for the connection's lifetime.

The first re-admission implementation exposed additional lifecycle gaps: retries could create
unbounded detached tasks, header-sync advisory backoff could wait forever on a stale snapshot,
stale stream exits could clear newer sessions, and deferred streams could exceed the negotiated
aggregate inbound queue limit.

## Solution

- Add an ordered-session demand contract supporting immediate open, deadline retry,
  event-driven retry, and retirement without changing connection ownership.
- Keep at most one connection-owned wake per ordered stream kind. Reactor demand replaces
  transport backoff, adoption/retirement cancels pending wakes, and generation checks discard
  stale queued wakes.
- Track active ordered streams by `kind -> session_id`, so an old worker exit cannot erase a
  newer local or peer-opened generation.
- Publish header-sync candidate state at advisory expiry and include advisory deadlines in
  reactor maintenance scheduling, so time alone can turn `RetryAt` into `OpenNow`.
- Validate aggregate inbound queue depth against every negotiated ordered stream before
  registration, including streams whose initial demand is deferred.
- Store one block-sync `PeerPark { conn_id, until }` record in the shared peer registry. The
  owning routine writes it synchronously at the no-progress decision site, and its generation
  check prevents a superseded routine from parking a replacement session.
- Make block-sync demand a pure read of the registry park and the existing reactor watches.
  Same-connection re-admission consumes the park once; connection cleanup only disassociates
  its connection while preserving the peer-level cooldown. Teardown never writes park state.
- Keep only active block-sync sessions in the service map and enforce direction caps by counting
  and installing them under the same lock. The reactor watch is only used to wake capacity waiters.
- Permit one bounded re-admission after a no-progress park; a repeated no-progress stall on the
  same connection disconnects instead of cycling indefinitely.
- Keep completed discovery exchanges retired for their connection, with teardown-safe state
  transitions.

## Testing

Focused regressions cover:

- one pending wake per stream kind, reactor-over-transport precedence, stale wake rejection,
  adoption/retirement cancellation, and connection cancellation;
- time-only header advisory expiry changing demand from `RetryAt` to `OpenNow`;
- stale ordered-stream exits arriving after a newer generation is active;
- registry-owned park consumption, connection cleanup, and stale-routine generation rejection;
- same-connection cooldown/body-work gating and fresh-connection admission;
- one bounded no-progress re-admission followed by disconnect;
- discovery teardown preceding retirement;
- concurrent block-sync admissions respecting the per-direction cap;
- complete negotiated ordered-stream queue-depth enforcement; and
- cooldown/tip-gated block-sync reopening on a still-live connection.

Local validation:

- `CXXFLAGS='-include cstdint' cargo test -p zakura-network` — 1,038 passed, 4 ignored

- [V12 audit findings](https://github.com/zakura-core/zakura/pull/166#issuecomment-4997307892)

### AI Disclosure

AI tools were used for investigation, implementation, review, and test execution.


evan-forbes:
@v12-auditor

evan-forbes:
@v12-auditor

evan-forbes:
@v12-auditor

evan-forbes:
fixing conflicts now

marking R4R as we are now shifting to getting v2 p2p stable again

ValarDragon:
My understanding was this PR's changes was split into 3 smaller PR's that have all been merged in!

evan-forbes:
Ahh this was actually a separate change, there was a different process that was split up for the 3 peer discovery issues

evan-forbes:
all v12 and local audits have been addressed, however I'll do hopefully one last ai-audit @v12-auditor as this code touches every reactor

zakura-network/src/zakura/handler.rs
```diff
@@ -1469,34 +1468,223 @@ struct AdmittedOrderedStream {
...
+}
+
+#[derive(Debug)]
+struct PendingOrderedSessionWake {
+    generation: u64,
+    source: OrderedSessionWakeSource,
+    abort: Option<AbortHandle>,
+}
+
+struct OrderedSessionWakeScheduler {
```
p0mvn: This is a general discussion comment and not specific to or blocking this PR. 

We keep abstracting/introducing new system components when I am not sure the abstractions are correct.

Each time we add a new component, the system becomes more difficult to reason about.

It also seems like the same problem as with the header sync reactor where we have one file with conflicting responsibilities.

p0mvn (approved):
The problem and the solution to it make sense to me. Approving, pending CI being green.

I did not run the branch, focusing on the conceptual understanding of the original bug and how we address it.

Left a non-blocking discussion comment.

Following up with a non-blocking AI review comment as well.

zakura-network/src/zakura/header_sync/service.rs
```diff
@@ -456,27 +471,106 @@ impl Service for HeaderSyncService {
...
+        let Some(node_id) = header_peer_node_id(peer) else {
+            return OrderedSessionDemand::Retire;
+        };
+        if self
+            .header_sync
+            .candidate_state()
+            .backed_off_node_ids
+            .contains(&node_id)
+        {
+            return OrderedSessionDemand::RetryAt(Instant::now() + HEADER_SYNC_ADVISORY_BACKOFF);
```
p0mvn: ## Non-blocking: Header-sync advisory reopen can overshoot the real backoff deadline

### From the top

Header sync sometimes marks a peer as temporarily not useful (**advisory backoff**). The reactor stores a real expiry time `backoff_until`. When the header-sync stream is absent, reopen demand should sleep until that time, then open again.

Instead, demand only sees “this peer is backed off” (a boolean list) and schedules **a fresh 60s from now**. Asking late in the window delays reopen past the real expiry — up to almost another full backoff.

### 1. Reactor owns the real deadline

Backoff is armed with an absolute time in `header_sync/state.rs` (`record_unconfirmed` → `backoff_until = Some(now + HEADER_SYNC_ADVISORY_BACKOFF)`).

Maintenance wakes on that deadline (`reactor.rs` includes `advisory.backoff_until` in the maintenance deadline). At expiry it clears backoff and republishes candidates so the peer leaves `backed_off_node_ids`.

Published candidate state only exposes the set of IDs — **not** `backoff_until`:

```rust
pub struct ZakuraHeaderSyncCandidateState {
    pub target_height: block::Height,
    pub admitted_node_ids: Vec<NodeId>,
    pub backed_off_node_ids: Vec<NodeId>,
}
```

### 2. Demand invents a new timer (this line)

```rust
return OrderedSessionDemand::RetryAt(Instant::now() + HEADER_SYNC_ADVISORY_BACKOFF);
```

The handler then `sleep_until` that instant (`handler.rs` `schedule_demand` → `RetryAt`).

### 3. Why it sticks

If a Demand wait is already pending, a newer (possibly shorter) one is ignored (`install_demand_wait` returns early when a Demand-sourced wait is already pending). So a mid-window `RetryAt(now+60s)` is not shortened when the reactor clears the peer at the real `backoff_until`. Demand also does not `WaitForChange` on candidate updates, so expiry publication does not cancel the sleep.

### Timeline

```text
T0      backoff_until = T0+60s; peer in backed_off_node_ids
T0+50s  stream exit / demand sample
        → RetryAt(T0+110s)   // wrong; should be T0+60s
T0+60s  reactor clears backed_off_node_ids
        scheduler still sleeping
T0+110s wake → OpenNow → reopen ~50s late
```

### Why tests didn’t catch it

`advisory_backoff_expiry_reopens_demand_without_an_unrelated_event` asserts `RetryAt(_)`, advances a **full** `HEADER_SYNC_ADVISORY_BACKOFF`, then polls demand → `OpenNow`. That proves “eventually OpenNow after a full backoff,” not “scheduled wake == real `backoff_until`.”

### Recommendation

**Preferred:** expose the real deadline and use it.

1. Extend candidate state, e.g. `backed_off: Vec<(NodeId, Instant)>` or a map `node_id → backoff_until` (keep `backed_off_node_ids` if needed for callers).
2. In `ordered_session_demand`:
   - if backed off: `RetryAt(backoff_until)` (not `now + HEADER_SYNC_ADVISORY_BACKOFF`)
   - optionally also `WaitForChange` on `subscribe_candidate_state()` so clearance can wake early if the sleep was wrong/stale
3. Test: enter backoff, sample demand at mid-window (e.g. advance 50s), assert `RetryAt` is within a small epsilon of the original `backoff_until`, not `now + 60s`. Optionally drive the handler scheduler and assert the wake fires at expiry.

**Acceptable alternative:** return `WaitForChange` on candidate state while the id is in `backed_off_node_ids` (reactor already publishes at expiry). That avoids inventing a timer entirely; pair with a unit test that advances only to real expiry and expects `OpenNow` without another full 60s.

This contradicts the PR’s claim that time alone turns `RetryAt` into `OpenNow` **at advisory expiry** for the scheduled wake path.

evan-forbes:
@v12-auditor

PR #174, opened 2026-07-14
fix(config): hide experimental sync tuning from generated configs

p0mvn:
## Motivation

The native Zakura header-sync and block-sync settings are experimental and likely to change. Emitting every tuning value from `zakurad generate` pins users to a generated snapshot instead of allowing new code defaults to take effect.

## Solution

- Remove `network.zakura.header_sync` and `network.zakura.block_sync` from generated configuration skeletons.
- Keep both sections deserializable so advanced users can still override them manually.
- Verify omitted sections resolve to the defaults in code and update the current generated-config fixture.

## Testing

- `cargo test -p zakura generated_config_omits_experimental_sync_config_and_uses_defaults` — verifies both sections are omitted, other Zakura settings remain, and parsing restores code defaults.
- Generated `zakura.toml` with the rebuilt binary and confirmed neither section was present.
- IDE diagnostics and `git diff --check` passed.

Full workspace tests were skipped because this is a low-risk generated-output change with focused coverage.

## AI Disclosure

Used Cursor to implement the generated-config filtering, add the focused test, and prepare this PR description.

### DIFF
--- a/zakurad/src/commands/generate.rs
+++ b/zakurad/src/commands/generate.rs
@@ -72,7 +72,8 @@ impl Runnable for GenerateCmd {
 
         // this avoids a ValueAfterTable error
         // https://github.com/alexcrichton/toml-rs/issues/145
-        let conf = toml::Value::try_from(default_config).unwrap();
+        let mut conf = toml::Value::try_from(default_config).unwrap();
+        remove_experimental_sync_config(&mut conf);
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):
I'm fine with this approach for now 

however, if we don't want users to have the footgun of changing something in a config then it seems like the best solution is to actually remove that option via only using constants

PR #190, opened 2026-07-15
fix(network): bound block-sync memory with a serialized apply backlog and wire-honest admission accounting

p0mvn:
## Motivation

Mainnet sync nodes approached OOM (12–32 GiB RSS on a 32 GiB host). #148 root-caused the allocator side and shipped the minimal Linux/glibc mitigation (`M_ARENA_MAX=4`), and its parameter changelog names the durable fix: "account decoded applying memory directly, or keep excess bodies serialized until a bounded applying slot is available." This PR is that fix.

The application-side defect: the block-sync pipeline decodes every body into `Arc<Block>` as soon as it becomes contiguous (`Sequencer::drain_ready_into_applying`) and holds the entire apply backlog decoded — 61k blocks in the incident, while the verifier only ever works on one submission window (401 blocks) at a time. Admission cannot see this: it estimates resident use as `wire_bytes × 4` uniformly (the `DESERIALIZED_MEM_FACTOR` TODO), which under-counts transaction-dense bodies exactly when blocks are heavy.

Rerunning the incident configuration (no arena cap) on a controlled single-feeder harness reproduces it: 6.17 GiB RSS by height ~194k in 40 minutes with a 34k-block decoded backlog. With the arena cap alone, a long run still climbs monotonically with block era — 3.3 GiB @100k up to 6.36 GiB at height 825k and still rising, with a 205k-block decoded backlog. The cap bounds allocator retention, not live decoded memory.

## Solution

Three commits, intended to be reviewed in order:

1. **Bound decoded bodies to the submission window.** Applying-backlog bodies stay in their serialized wire form (`BufferedBlockBody::RawFramePayload` — the raw payload is already carried end-to-end and already used this way for the reorder backlog). Decode happens at `prepare_submit` and the decoded copy is cached on the entry until apply-finished/unsubmit. Decoded residency is now structurally ≤ `submitted_apply_limit` (401) plus the bounded sequencer input channel, regardless of backlog depth. The steady-state path (body arrives next-in-sequence and submits immediately) keeps the peer routine's decoded copy, so the common case never decodes twice. `previous_block_hash` is captured at receipt so reset-conflict checks never decode a backlog body.

2. **Honest resident accounting.** With decoded residency bounded, admission charges serialized pools (reorder, applying backlog past the decode window, reservations) at wire size, and only the two structurally bounded decoded pools (submission window, sequencer input channel) at `× DESERIALIZED_MEM_FACTOR`. `submitted_applying_bytes` is threaded into `AdmissionSnapshot` from the sequencer view.

3. **Size the default budget as a memory target.** `DEFAULT_BS_MAX_REORDER_LOOKAHEAD_BYTES` becomes 1.5 GiB: under wire-honest accounting this is (to close approximation) the byte ceiling of the speculative backlog itself, so total RSS tracks `budget + decode window + process overhead`, independent of block era. The old checkpoint-range clamp (range × 4) is relaxed to the range's wire bytes — the commit-window admission exemption already guarantees range assembly, so the decoded-multiple floor was obsolete.

The #148 arena cap remains as the complementary layer: it addresses allocator retention of the (now much smaller) churn, which app-level accounting cannot; this PR addresses live decoded memory, which the arena cap demonstrably cannot.

### Tests

- `cargo test -p zakura-network zakura::block_sync` (225 tests, including a new invariant test `sequencer_bounds_decoded_bodies_to_submission_window` and updated admission tests encoding the wire-honest formula)
- `cargo fmt --all -- --check`, `cargo clippy -p zakura-network --all-targets -- -D warnings`
- **Operational validation** on an 8-core/31 GiB host, fresh mainnet sync against a local archive feeder (~825k blocks), identical configs across arms:
  - **This PR + `MALLOC_ARENA_MAX=4`:** genesis → height 701,649 in ~2h50m (93 blk/s second half), RSS **flat at 3.3–3.5 GiB** across the Overwinter/Sapling/Blossom transitions, absorbing a 116,192-block backlog with no RSS movement.
  - **main + `MALLOC_ARENA_MAX=4` (pure hotfix):** same harness, RSS climbs monotonically with era: 3.32G @100k, 4.98G @200k, 5.50G @400k, 6.19G @600k, 6.30G @700–800k — ~2× this PR at every height above 200k, still rising at corpus end.
  - **main without the hotfix (incident config):** 6.17 GiB by height ~194k in 40 minutes.

| height | 100k | 200k | 300k | 400k | 500k | 600k | 700k |
| --- | --- | --- | --- | --- | --- | --- | --- |
| this PR + arena cap | 2.08G | 3.01G | 3.12G | 3.33G | 3.33G | 3.49G | 3.34G |
| main + arena cap | 3.32G | 4.98G | 5.23G | 5.50G | 5.75G | 6.19G | 6.30G |


evan-forbes (approved):
nice

curious if we could ever make the decoded block not ~4x bigger lol

will defer to you on the v12. an agent could replicate it but still wasn't sure if this is handled by the other reorg stuff

p0mvn:
v12 has been addressed, though it seems to have been present before the PR

PR #183, opened 2026-07-14
chore(release): prepare v1.0.0-rc6

ebfull:
## Motivation

Stage a v1.0.0-rc6 release candidate. Fourteen PRs have merged since the rc5 cut — including consensus-relevant hardening (#168, #171), the release-metadata finalization (#169, #173, #179), and the release-tooling changes (#175, #177) — so the next cut rehearses the full 13-crate graph exactly as the 1.0.0 fork-point publish will ship it.

## Solution

- Bump all 13 publishable crates to `1.0.0-rc6` — the full-graph rehearsal of the 1.0.0 fork-point exception — including every internal dependency requirement, and sync `Cargo.lock`.
- Point the installer at the `v1.0.0-rc6` tag (checksum stays the placeholder that release CI replaces in the published asset copy) and move the Docker image pins to `1.0.0-rc6`.
- Update the README's `cargo install --tag` line, which `cargo release version` does not rewrite (see the note in `zakurad/Cargo.toml`).

Merging this PR does **not** cut the release: the `create-release` dispatch and the `release` environment approval remain the controls, and no crates.io publish happens as part of this change.

## Testing

- `make pre-release RELEASE_TAG=v1.0.0-rc6 BASE_TAG=v1.0.0-rc5` (the #177 preflight, first release exercise):
  - release tag matches the `zakura` package version
  - the changed-crate advisory enumerated all 13 publishable crates with the expected mixed baselines (`1.0.0-rc5 →` for zakura/network/state/consensus, `1.0.0-rc3 →` for the other nine) and found no missing bumps
  - all 13 crates package cleanly
- `cargo metadata --locked` (resolving) passes, so the lockfile matches the new graph
- `bash -n scripts/install-zakura.sh`
- Workspace tests/clippy not run locally for a metadata-only version bump; CI provides those checks.

### Specifications & References

- `.github/PULL_REQUEST_TEMPLATE/release-checklist.md`
- `docs/release-tag-protection.md`

### Follow-up Work

- Dispatch `create-release` for `v1.0.0-rc6` after merge (separate decision).
- When cut, rc6's artifacts join the pre-1.0.0 sweep; any crates.io publish of rc6 is a separate, explicit decision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #177, opened 2026-07-14
chore(release): add unified pre-release checks

p0mvn:
## Motivation

Release preparation currently requires separate version, crate-change, and packaging checks, which makes it easier to miss a step or overlook advisory output.

## Solution

- add a `make pre-release` entry point for version and packaging checks
- warn locally when changed publishable crates retain their previous versions
- use the unified preflight in the release workflow while keeping the crate-version advisory local-only
- update the release checklist and improve output grouping and notice visibility

## Testing

- `make pre-release RELEASE_TAG=v1.0.0 BASE_TAG=v1.0.0-rc5`
  - release tag matched the package version
  - all 13 changed publishable crates had version bumps
  - all 13 crates packaged cleanly
- `bash -n scripts/check-crate-version-bumps.sh scripts/check-crate-packaging.sh`
- `make -n pre-release RELEASE_TAG=v1.0.0 BASE_TAG=v1.0.0-rc5`
- `markdownlint .github/PULL_REQUEST_TEMPLATE/release-checklist.md --config .trunk/configs/.markdownlint.yaml`
- IDE diagnostics: no errors in changed files

Risk classification: low. This changes release tooling, CI orchestration, and documentation without modifying validator runtime behavior, so workspace Rust tests and clippy were not run.

### Success

```bash
roman@Romans-MacBook-Pro-2 zakura % make pre-release RELEASE_TAG=v1.0.0 BASE_TAG=v1.0.0-rc5

==> [1/2] Checking release version

/Applications/Xcode.app/Contents/Developer/usr/bin/make pre-release-version RELEASE_TAG="v1.0.0" BASE_TAG="v1.0.0-rc5" PRE_RELEASE_WARN_CRATE_VERSION_BUMPS="1"
./scripts/check-release-version.sh "v1.0.0"
Release tag v1.0.0 matches the 'zakura' package version (1.0.0).
Crate version bump advisory found no unchanged versions across 13 changed published crate(s).

NOTICE: The working tree has local changes.
        This advisory compared those changes against v1.0.0-rc5.

==> [2/2] Checking crate packaging

/Applications/Xcode.app/Contents/Developer/usr/bin/make pre-release-packaging CRATE_PACKAGING_VERIFY="0"
Packaging 13 crates...
   Packaging zakura-test v1.0.0 (/Users/roman/projects/zakura/zakura-test)
...

Package archive sizes:
  zakura-test-1.0.0.crate: 1718077 bytes
...

All 13 crates package cleanly.
```

### Warn

[...]

PR #169, opened 2026-07-14
chore(release): prepare v1.0.0

p0mvn:
## Motivation

Prepare the first stable Zakura release after the v1.0.0 release-candidate series. This release was discussed with the team beforehand.

## Solution

- promote the workspace crates and internal dependency constraints to `1.0.0`
- finalize the v1.0.0 changelog and installation references
- prepare stable installer and Docker image metadata
- add the generated v1.0.0 configuration snapshot

### Tests

- `./scripts/check-release-version.sh v1.0.0`
- `./scripts/check-crate-packaging.sh`
- `markdownlint CHANGELOG.md README.md --config .trunk/configs/.markdownlint.yaml`
- `bash -n scripts/install-zakura.sh`
- `git diff --check main...HEAD`
- confirmed the v1.0.0 config snapshot matches v1.0.0-rc5
- IDE diagnostics: no errors

Risk classification: low risk because this is a release metadata and generated-config update with no runtime logic changes. Full workspace formatting, Clippy, and tests were not run locally; CI will provide those checks.

### Specifications & References

- `.github/PULL_REQUEST_TEMPLATE/release-checklist.md`

### Follow-up Work

- Complete the remaining release checklist and publish v1.0.0 after this PR merges.

### AI Disclosure

- [x] AI tools were used: Cursor reviewed the release metadata and prepared the commit and PR description.

### PR Checklist

- [x] The PR title follows conventional commits format.
- [x] The PR follows the contribution guidelines.
- [x] This change was discussed with the team beforehand.
- [x] Focused release checks pass; full workspace checks are delegated to CI.
- [x] The documentation and changelog are up to date.

PR #194, opened 2026-07-15
chore(release): prepare v1.0.0

ebfull:
## Motivation

Re-stage the v1.0.0 release. `main` has sat at `1.0.0-rc6` since #183, and the rc6 cut completed the full dress rehearsal: the tag gate passed its fifth exercise, #177's pre-tag packaging ran in CI for the first time, and all 13 crates published to crates.io from the tagged SHA. This PR moves the package graph back to the fork-point `1.0.0` so the real release can be dispatched.

## Solution

- Bump all 13 publishable crates `1.0.0-rc6` → `1.0.0` via `cargo release version` (all internal dependency requirements included) and sync `Cargo.lock`.
- Point the installer at the `v1.0.0` tag with the placeholder-zero checksum (release CI substitutes the real value into the published asset copy; the zeroed pin fails cleanly — the #169 design). Docker pins move to `zakuracore/zakura:1.0.0` / `zcashd-compat-1.0.0` on the new org (#189/#191).
- Update the README `cargo install --tag` line via `cargo release replace` (see the note in `zakurad/Cargo.toml`).
- Regenerate the stored config snapshot `zakurad/tests/common/configs/v1.0.0.toml` from this branch (the `last_config_is_stored` acceptance test derives the filename from the package version). The regenerated file is byte-identical to the rc6 snapshot that passed CI at the rc6 cut; versus the #169-era copy the only change is trailing-newline normalization.
- Fix the release checklist's snapshot-generation command: its per-field `sed`s miss `cookie_dir`, which also embeds the default cache path — the test does global path-string replacement, so the documented command produced a snapshot the test rejects. The command now mirrors the test's semantics (found live while preparing this PR).
- Refresh `ESTIMATED_RELEASE_HEIGHT` 3,412,800 → **3,414,300**: the mainnet tip passed the old estimate this morning (3,413,133 at 11:58 UTC, per Blockchair). The new value assumes a cut within ~1 day; with `EOS_PANIC_AFTER = 7`, end-of-support lands at ≈3,422,364 — still ~5,800 blocks (~5 days) before Ironwood activation at 3,428,143, preserving the follow-up-release forcing function. **Re-run this estimate if dispatch slips past ~July 17.**
- Refresh the frozen "Initial release" changelog entries' date (root + 12 crate changelogs) to the expected tag date, 2026-07-15.

Merging this PR does **not** cut the release: the `create-release` dispatch and the `release`-environment approval remain the controls, and no crates.io publish happens as part of this change.

## Testing

- `make pre-release RELEASE_TAG=v1.0.0 BASE_TAG=v1.0.0-rc6`:
  - release tag matches the `zakura` package version
  - the changed-crate advisory enumerated all 13 publishable crates against the rc6 baseline with no missing bumps
  - all 13 crates package cleanly
- `cargo metadata --locked` resolves, so the lockfile matches the new graph
- freshly generated config matches the committed `v1.0.0.toml` byte-for-byte (the `last_config_is_stored` contract)
- `bash -n scripts/install-zakura.sh`
- Workspace tests/clippy not run locally for a metadata-only version bump; CI provides those checks. `cargo semver-checks` / `cargo public-api diff` skipped: 1.0.0 is the initial release baseline (changelogs frozen per `CHANGELOG_GUIDELINES.md`).

- `.github/PULL_REQUEST_TEMPLATE/release-checklist.md`
- `docs/release-tag-protection.md`
- #183 (the rc6 full-graph rehearsal), #169 (the original 1.0.0 prep)

### Follow-up Work

- Dispatch `create-release` for `v1.0.0` after merge (separate decision, per the release checklist).
- Checkpoints: `zakura-chain/src/parameters/checkpoint/main-checkpoints.txt` tops out at height 3,358,006 (last substantive update: upstream v4.5.0, 2026-05-28) — ~55k blocks behind tip. No release in this repo has bumped them; deferring keeps this PR a pure re-bump, but it should be an explicit decision (or a separate PR) before or shortly after 1.0.0.
- The pre-1.0.0 rc sweep, promotion to "Latest", signing, and the manual crates.io publish follow the release checklist after the cut.
- At promotion time, `zakuracore/zakura:zcashd-compat-latest` (the installer's compat fallback image) does not exist yet on the new org — push or re-point it.

## AI Disclosure

Prepared with Claude Code (version bumps via `cargo release`, metadata edits, and preflight runs); human review required before merge and dispatch.

### DIFF
--- a/zakurad/src/components/sync/end_of_support.rs
+++ b/zakurad/src/components/sync/end_of_support.rs
@@ -13,7 +13,7 @@ use zakura_chain::{
 use crate::application::release_version;
 
 /// The estimated height that this release will be published.
-pub const ESTIMATED_RELEASE_HEIGHT: u32 = 3_412_800;
+pub const ESTIMATED_RELEASE_HEIGHT: u32 = 3_414_300;
 
 /// The maximum number of days after `ESTIMATED_RELEASE_HEIGHT` where a Zebra server will run
 /// without halting.
--- a/zakurad/src/components/sync/end_of_support.rs
+++ b/zakurad/src/components/sync/end_of_support.rs
@@ -13,7 +13,7 @@ use zakura_chain::{
 use crate::application::release_version;
 
 /// The estimated height that this release will be published.
-pub const ESTIMATED_RELEASE_HEIGHT: u32 = 3_414_300;
+pub const ESTIMATED_RELEASE_HEIGHT: u32 = 3_413_500;
 
 /// The maximum number of days after `ESTIMATED_RELEASE_HEIGHT` where a Zebra server will run
 /// without halting.
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -124,10 +124,15 @@ cargo release version --verbose --execute --allow-branch '*' -p zakura patch # [
 ```sh
 cargo build --bin zakurad &&
 ./target/debug/zakurad generate |
-sed 's/cache_dir = ".*"/cache_dir = "cache_dir"/' |
-sed 's/identity_dir = ".*"/identity_dir = "identity_dir"/' \
+sed "s#${XDG_CACHE_HOME:-$HOME/.cache}/zakura#cache_dir#g" |
+sed "s#$HOME/.zakura#identity_dir#g" \
   > zakurad/tests/common/configs/v<version>.toml
 ```
+
+      The replacements are global path-string substitutions, mirroring
+      `last_config_is_stored` — the default cache path also appears in
+      fields other than `cache_dir` (for example `cookie_dir`), so
+      per-field rewrites produce a snapshot the test rejects.
 
 - [ ] On the release commit, run the pre-release checks for the tag you are
       about to create, using the previous release tag as the base:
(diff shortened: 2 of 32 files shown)


ebfull:
CI note: `spell-check` fails on a pre-existing typo that today's codespell dictionary update started flagging (main's lint has been red since 10:01 UTC) — fix in #195; re-run Docs Check here after it merges. The early `* success` aggregate failures are from runs the label-add re-trigger cancelled; the replacement runs are in flight.

zakurad/src/components/sync/end_of_support.rs
```diff
@@ -13,7 +13,7 @@ use zakura_chain::{
 use crate::application::release_version;
 
 /// The estimated height that this release will be published.
-pub const ESTIMATED_RELEASE_HEIGHT: u32 = 3_412_800;
+pub const ESTIMATED_RELEASE_HEIGHT: u32 = 3_414_300;
```
ValarDragon: Im checking this now

ValarDragon: Block at time of comment is: https://cipherscan.app/block/3413172

So this seems roughly right up to 24 hours. More accurate may just be 3_413_500, but I don't think we really care, because this first release is a very short EOS intentionally

ValarDragon (approved):

zakurad/src/components/sync/end_of_support.rs
ValarDragon:
```suggestion
pub const ESTIMATED_RELEASE_HEIGHT: u32 = 3_413_500;
```

PR #208, opened 2026-07-15
fix(state): harden VCT checkpoint validation

evan-forbes:
## Motivation

VCT fast sync used a weaker identity for prevalidation reuse than for commitment validation. That could reuse a successful check for a checkpoint body whose authorizing-data commitment had not been validated. An initial correction also exposed a liveness failure by classifying a permanently invalid body as a retryable missing-root condition.

This draft isolates packets 14 and 22 from #165 and fixes both the validation invariant and the retry classification.

## Solution

- Bind VCT prevalidation reuse to the height, canonical block hash, and normalized NU5+ authorizing-data root.
- Keep the cached checkpoint root tied to the wrapped block by computing it through a consuming API and removing mutable dereference access.
- Reject a cached same-block/different-root body with a dedicated non-retryable error, without invalidating authenticated VCT roots.
- Add end-to-end regressions showing the invalid body is not parked and a valid replacement for the same block identity can subsequently commit.
- Update the VCT design documentation with the cache and error-handling invariants.

### Review tradeoffs and risk

- Removing `DerefMut` from `CheckpointVerifiedBlock` is source-breaking for public `zakura-state` consumers. It makes the cached-root invariant structural and preserves parallel root computation. Alternatives are to retain mutable access while clearing a private trusted-root field, or to recompute the body root during serialized commit.
- Residual risk is low to medium because this remains a consensus-sensitive checkpoint path. The regression covers both safety and recovery/liveness behavior.
- Security disclosure and release timing must be coordinated under `SECURITY.md` before merge or public release. This PR intentionally avoids operational reproduction details.

## Testing

- `CXXFLAGS='-include cstdint' cargo test --workspace --locked`
- `CXXFLAGS='-include cstdint' cargo clippy --workspace --all-targets --locked -- -D warnings`

The workspace run included 356 passing `zakura-state` tests, the hostile-body write-loop regression, and the compile-fail test for mutable checkpoint access.

- Related review and source decomposition: #165 (packets 14 and 22)
- `SECURITY.md`
- `docs/design/verified-commitment-trees.md`

### Follow-up Work

- Confirm coordinated disclosure/advisory handling before merge.
- Confirm the public API compatibility choice for `DerefMut`.
- Changelog text is omitted while the repository's v1.0 changelog freeze applies; coordinate any security release-note treatment with maintainers.

## AI disclosure

OpenAI Codex was used to analyze #165, draft the regression tests, implement the fixes, run verification, and prepare this PR description. The contributor reviewed the changes and test results.

### DIFF
--- a/zakura-consensus/src/checkpoint.rs
+++ b/zakura-consensus/src/checkpoint.rs
@@ -1152,11 +1152,12 @@ where
             if req_block.block.auth_data_root.is_none()
                 && NetworkUpgrade::current(&network, req_block.block.height) >= NetworkUpgrade::Nu5
             {
-                let block = req_block.block.block.clone();
-                if let Ok(auth_data_root) =
-                    tokio::task::spawn_blocking(move || block.auth_data_root()).await
+                let block = req_block.block.clone();
+                if let Ok(block) =
+                    tokio::task::spawn_blocking(move || block.with_precomputed_auth_data_root())
+                        .await
                 {
-                    req_block.block.auth_data_root = Some(auth_data_root);
+                    req_block.block = block;
                 }
             }
 
--- a/.lychee.toml
+++ b/.lychee.toml
@@ -27,6 +27,8 @@ exclude = [
     "^https://eprint\\.iacr\\.org/",
     # Tor's GitLab returns 403 to automated link checkers
     "^https://gitlab\\.torproject\\.org/",
+    # The ZIP site returns 521 to GitHub-hosted automated link checkers
+    "^https://zips\\.z\\.cash/",
 
     # These routes will become available when the Zakura website launches.
     "^https://zakura\\.com/",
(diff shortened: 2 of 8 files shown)


zakura-consensus/src/checkpoint.rs
```diff
@@ -1152,11 +1152,12 @@ where
...
-                let block = req_block.block.block.clone();
-                if let Ok(auth_data_root) =
-                    tokio::task::spawn_blocking(move || block.auth_data_root()).await
+                let block = req_block.block.clone();
+                if let Ok(block) =
+                    tokio::task::spawn_blocking(move || block.with_precomputed_auth_data_root())
+                        .await
                 {
-                    req_block.block.auth_data_root = Some(auth_data_root);
+                    req_block.block = block;
```
evan-forbes, line 1160: here's the core fix!

DerefMut + pub auth_data_root let any holder desync the pair. A header matching the stale root could then finalize a block whose actual authorizing data differs, without ever proving the header binds that data. Therefore, making the only way to populate the root a method that derives it from the wrapped block fixes the issue

evan-forbes:
the lint checker is just due to a temporary dead link

```markdown
### Errors in book/src/dev/zakura-dependencies-for-audit.md

* [504] <https://hackmd.io/@qedit/zcash-nu5-audit> | Rejected status code: 504 Gateway Timeout (configurable with "accept" option)

### Errors in README.md

* [521] <https://tachyon.z.cash/> | Rejected status code: 521 Unknown status code (configurable with "accept" option)
```

p0mvn (approved):

PR #209, opened 2026-07-15
fix(network): preserve block-sync peer accountability

p0mvn:
## Motivation — the problem

Block sync conflated two things that are not the same: **global scheduling ownership** of a height (who is responsible for fetching it) and a **peer-local request that has already been sent on the wire**. Once a `GetBlocks` is on the wire it cannot be cancelled — the peer will answer it (or not) regardless of what the scheduler decides next.

The old design *removed* a peer's request whenever the height stopped being its responsibility — another peer delivered it first, the floor advanced past it, or it timed out. Removing an in-flight request caused a family of failures:

- **False exile of healthy peers.** A peer's liveness deadline is armed when it is issued a request. If the request is removed out from under it, the deadline is left armed for work the peer no longer owns, and because its last delivery predates its latest (now-removed) request, the old "clear on progress" condition never fires. The peer is then disconnected for "no accepted block progress" despite having nothing left to deliver. In live testing this drained peer count from five to zero, one healthy peer at a time.
- **Zombie / capped peers.** The counterpart failure: a peer left capped by `requests_without_block_progress` with no path to recover — a connection that stays up but can never be scheduled again.
- **Lost late-response correlation.** A body or terminator that arrives *after* its request was removed has nothing to match against, risking duplicate submission to the verifier or mis-accounted byte budget.
- **Silent peers hiding behind local stalls.** A global apply-backlog exemption let a local commit stall excuse *every* peer from liveness — so a genuinely dead peer could not be distinguished from a slow-committing pipeline.
- **Reset / reorg hazards.** A body claimed before a destructive chain reset could be admitted afterward (a stale-fork block handed to the verifier); the shared byte budget could leak or double-release across resets, timeouts, and watchdog cancellations.

## Previous attempts

- **#192 (superseded).** Fixed the false-exile symptom by updating the liveness deadline when the apply queue was non-empty, obsolete requests were removed, and the peer had no outstanding work, plus a global apply-backlog exemption for local commit stalls. Review (Evan) rejected it: it did not reset `requests_without_block_progress`, so it reintroduced the zombie-peer bug for a peer at its cap; and the apply-backlog exemption let silent peers dodge liveness entirely. The recommended direction was structural — **do not remove an in-flight request; it cannot be cancelled** — and to cleanly separate active peer state from global shared state.

This PR implements that direction.

## Architecture introduced

**Core rule: a request that has been sent is never removed — it is *retired*.** A retired request keeps a bounded correlation record (a "tombstone") while losing all active scheduling ownership.

- **Active vs Retired state** (`OutstandingRequests` in `outstanding.rs`). Each request is `Active` or `Retired { reason, retired_at, correlation_deadline }`.
  - *Active* owns: a peer request slot, BBR in-flight accounting, published registry claims, and byte reservations.
  - *Retired* owns none of those, but remains for: inbound body/terminator matching, response-timeout cleanup, peer-liveness accountability, and diagnostics.
  - Cached `active_count` / `retired_count` / `active_reserved_bytes` counters are updated on every mutation and checked with debug asserts plus a randomized invariant test.
- **Retirement reasons:** `Covered` (floor passed the whole range), `RequestTimeout`, `FloorWatchdog` (the reactor cancelled the claim), and `ViewReset` (a destructive chain reset).
- **Correlation deadline** = `min(liveness_deadline, now + request_timeout)`. It bounds how long a tombstone lives, so a late terminator/body can close its window before the same height is safely re-issued to that peer.
- **Generation-scoped request identity.** Every request gets a monotonic `BlockRequestToken`; paired with the routine's `generation` it forms a `ReservationOwner`. A stale watchdog/timeout/drop carrying an old `(generation, token)` cannot retire newer work, resurrect a retired claim, or release a replacement request's reservation.
- **Reservation ownership in the WorkQueue.** Each reserved height is stamped with its `ReservationOwner` at take time; cleanup releases only owner-matching reservations (mismatches are counted and skipped). This makes byte-budget release exactly-once even under ABA (a height returned to `pending` and re-taken by another peer).
- **Atomic late-body claim + reset epoch.** `claim_late_body` classifies pending/reserved/held and performs the state transition under a single queue lock, so two peers racing the same height cannot both win and forward. It captures the WorkQueue's authoritative `reset_epoch` atomically at claim time and returns it with the body.
- **Single authoritative reset epoch.** `WorkQueueInner.reset_epoch` is the source of truth, bumped only by `destructive_reset_above`; the sequencer *adopts* it rather than keeping its own, so the two cannot diverge. Forward (growth) advances use `advance_floor` and do not bump the epoch, so valid in-flight successors survive. The sequencer rejects any queued body whose epoch predates a destructive reset (stale-fork rejection) and releases its held bytes exactly once.
- **Reset classification.** A chain-tip reset is *growth* (advance the floor, keep successors) when it advances onto our own chain with a linking successor, and *destructive* (drop successors, bump the epoch) on a genuine reorg or rollback. Held and still-in-channel successors are recognized so a forward advance is not misclassified as a reorg.
- **Liveness accountability.** Only *accepted* block progress resets `requests_without_block_progress` and re-arms the deadline. A useful late body is credited; a discarded body for an already-committed height is a correlated response, not progress. The global apply-backlog exemption is removed; liveness is peer-specific and bounded.
- **Byte/height lifecycle.** The WorkQueue owns each height once, through `Pending → Reserved(estimate) → Held(actual) → Released`. Each transition yields an exact budget delta; held bytes travel with the forwarded body and release once on commit.

The module `README.md` documents the request lifecycle, retirement reasons, the active-vs-retired visibility table, and the core invariants.

## Risk areas / edge cases closed

Each of the following was found and fixed during review and is covered by a regression test:

- **Deadline-arm busy-spin.** An expired retired tombstone (and an expired retry-avoid entry) fed the wake deadline but were pruned only on the fill path, which is skipped while the outbound queue is full — a 100% CPU spin that also hammered the registry lock. Now pruned on the deadline arm.
- **Liveness-credit farming.** A peer that only delivers already-committed (discarded) bodies no longer resets its no-progress accounting, so sybil peers cannot farm connection slots indefinitely.
- **Hash-bound settlement.** A matched body settles ownership through the hash-checking claim path, so a stale-fork body cannot settle a different fork's reservation for the same height.
- **Budget double-release.** The `try_fill` mismatch unwind derives its release from the ledger (Held-aware) instead of a fixed amount, preventing an under-count → over-admission.
- **Stale reservation release (ABA).** Reservation-owner tagging prevents a stale timeout/drop/watchdog from releasing a replacement peer's re-taken reservation.
- **Forward-reset dropping valid successors.** A forward advance with a downloaded successor is classified as growth, so a valid in-flight/held successor body is preserved instead of dropped and re-downloaded; a genuine conflicting reorg still drops it.
- **Stale-fork body admission across a reset.** The reset epoch is captured atomically at claim time (not via a racy later read of the view), so a body claimed pre-reset is rejected post-reset.
- **Reset correlation loss.** A destructive reset retires active requests as `ViewReset` tombstones (instead of draining them), so a late terminator closes its correlation window before the same height can be reissued.
- **Dual-epoch divergence.** The sequencer adopts the WorkQueue's epoch, so the body-guard epoch and the reset epoch cannot drift apart.
- **First-completion-wins.** When one peer times out and another delivers, a late duplicate from the first peer is submitted exactly once, charges the budget once, and is never scored as misbehavior.

## Testing

- `cargo test -p zakura-network` — full crate green (1008 passed, 0 failed, 4 ignored), including the loopback listener tests that previously required specific machine networking.
- New integration coverage: discarded-body-does-not-rescue-a-silent-peer (farming exile), cross-peer first-completion single-submission, conflicting-reorg-drops-held-successor, plus forward-advance-preserves-successor and the reset-epoch guard.

## Specifications & references

Supersedes #192. Folds in #211 (bind block-sync cleanup to the reservation owner).

## Follow-up work

Add block-sync wire request IDs in a future stream-version upgrade to remove the remaining bounded same-start correlation ambiguity (today two requests with the same start height are disambiguated only by the retirement/tombstone window).


p0mvn:
Was worried about perf regressions/resource contention. Profiles do not support the concern

p0mvn:
Evidence from profiling the current PR release binary during active mainnet checkpoint sync:

- Instruments Time Profiler captured 180.302s at 1ms sampling resolution: 136.171 CPU-seconds (0.755 average cores).
- `PeerRoutine::publish_outstanding` accounted for 178ms inclusive (0.131% of sampled CPU).
- Building/scanning the outstanding `BTreeMap` accounted for 140ms (0.103%); replacement/drop accounted for 30ms (0.022%).
- Only 2ms of sampled mutex waiting was attributable to the registry calls under `publish_outstanding` (0.0015%). Time Profiler samples running threads, so this is supporting evidence rather than a complete blocked-time measurement.
- Across 180 one-second process samples, CPU averaged 73.4% of one core (median 44.4%, p95 338.6% during parallel verification bursts). RSS averaged 905MiB and peaked at 1,014MiB.
- The independent 60-minute PR-node run also completed successfully: 3,825 blocks, 970.6MiB peak RSS, 0 restarts, and no logged errors or panics.

The profile is dominated by expected Equihash verification (19.5%), RocksDB reads (15.9%), and state writes (10.1%). The new outstanding-publication/retirement path is measurable but not a material CPU or contention source in these runs.

p0mvn:
@v12-auditor review

PR #212, opened 2026-07-16
fix(state): harden VCT authorizing-data validation

p0mvn:
## Motivation

VCT fast commits had two related failure-classification problems around NU5+ authorizing data:

- After a queue reset, a malformed current block body could be misclassified as a retryable supplied-root failure and parked indefinitely.
- A buffered successor body with the canonical header/hash but altered authorizing data could provide a different `auth_data_root`. Its failed look-ahead check was then attributed to the current height, evicting a valid supplied root and stalling the frozen-frontier retry path.

The second case is possible because ZIP-244 transaction IDs, and therefore the block hash, do not bind signatures, proofs, or transparent unlocking scripts.

## Solution

- Preserve whether VCT verification failed on the current block body or on supplied roots, keeping current-body failures non-retryable.
- Source production successor witnesses exclusively from the contextually validated header store and its persisted precomputed `auth_data_root`; buffered bodies remain queued only for their own commits.
- Remove body-derived `NextVctBlock` construction while preserving supplied-root fast commits and successful-look-ahead deduplication.
- Reject obsolete short commitment-root rows so a missing persisted auth-data root cannot silently decode as zero.
- Extend regression coverage for cache-empty malformed bodies, forged wrappers, malformed same-hash successors, header-only prevalidation, and buffered-body behavior.

A malformed successor body can no longer evict the current block's root. It is rejected as a hard auth-data mismatch when it becomes the current block, while an honest same-hash replacement can still commit.

## Testing

- `CXXFLAGS='-include cstdint' cargo test -p zakura-state --lib`
  - 358 passed; 3 ignored.
  - Exercises malformed current and successor bodies, verifies valid supplied roots remain usable, and covers strict commitment-root row decoding.
- VCT-focused state tests
  - 41 passed; 1 ignored.
- IDE diagnostics for all modified files

- Follow-up to #208
- ZIP-244 authorizing-data commitments

### Follow-up Work

None.

## AI disclosure

Cursor GPT-5.6 Sol was used to investigate the VCT failure paths, implement the fixes and regression coverage, run verification, and draft this PR description. The contributor reviewed the changes and test results.

### DIFF
--- a/zakura-state/src/service/finalized_state/disk_format/shielded.rs
+++ b/zakura-state/src/service/finalized_state/disk_format/shielded.rs
@@ -90,7 +90,7 @@ impl FromDisk for orchard::tree::Root {
 /// The per-height Sapling and Orchard note-commitment roots, as stored in the
 /// `commitment_roots_by_height` index (keyed by [`Height`]).
 ///
-/// Every node persists this 64-byte value for each committed block — including a
+/// Every node persists this 152-byte value for each committed block — including a
 /// verified-commitment-trees fast-synced node, which folds these roots in but writes no
 /// per-height note-commitment trees. It lets such a node still serve the `tree_aux`
 /// `BlockRoots` read from a compact index rather than from the (absent) trees.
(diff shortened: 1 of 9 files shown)


p0mvn:
@v12-auditor review

zakura-state/src/service/finalized_state/disk_format/shielded.rs
```diff
@@ -136,41 +136,21 @@ impl IntoDisk for CommitmentRootsByHeight {
...
-            )
-        } else {
-            (
-                ironwood::tree::NoteCommitmentTree::default().root(),
-                0,
-                0,
-                0,
-            )
-        };
+        let bytes: [u8; 152] = bytes
```
p0mvn, line 139: Note to reviewers on the diff below:

Since I have resynced the snapshots, we now have a consistent format, so we can drop the historical migration pre-launch workarounds

p0mvn:
V12 has been addressed

evan-forbes (approved):

PR #218, opened 2026-07-16
fix(network): correct header-sync maintenance timing

p0mvn:
## Motivation

Header sync could repeatedly wake on a stale VCT repair retry timestamp while the repair request was still in flight. Separately, `HeaderSyncStartup::new` ignored the configured status refresh interval, requiring production wiring to correct it after construction.

## Solution

- Exclude `next_attempt_at` from the VCT repair maintenance deadline while an attempt is in flight, while retaining request and overall repair deadlines.
- Initialize the startup status refresh interval from `ZakuraHeaderSyncConfig` and remove the production-only override.
- Add focused regression coverage for both timing behaviors.

## Testing

- `cargo fmt --all -- --check` — passed.
- `cargo clippy -p zakura-network --all-targets -- -D warnings` — passed.
- Both new focused regression tests passed. They verify that active repairs ignore stale retry deadlines and constructors preserve custom refresh intervals.
- `cargo test -p zakura-network` — 992 passed; 3 unrelated listener tests failed because this macOS host could not bind their requested loopback source addresses (`AddrNotAvailable`).

### DIFF
--- a/zakura-network/src/zakura/handler.rs
+++ b/zakura-network/src/zakura/handler.rs
@@ -2998,7 +2998,6 @@ pub async fn spawn_zakura_endpoint_with_header_sync_driver(
         config.zakura.header_sync.clone(),
         limits.max_frame_bytes,
     );
-    startup.status_refresh_interval = config.zakura.header_sync.status_refresh_interval;
     startup.trace = trace.clone();
     startup.frontier_updates = sync_frontier
         .as_ref()
--- a/zakura-network/src/zakura/header_sync/state.rs
+++ b/zakura-network/src/zakura/header_sync/state.rs
@@ -307,6 +307,17 @@ impl VctRootRepair {
             && now >= self.next_attempt_at
     }
 
+    pub(super) fn next_maintenance_deadline(&self) -> Instant {
+        let repair_deadline = self.started_at + VCT_ROOT_REPAIR_MAX_WALL_TIME;
+        // An in-flight attempt has its own request timeout, so ignore its stale
+        // retry timestamp while retaining the overall repair deadline.
+        if self.in_flight.is_none() {
+            repair_deadline.min(self.next_attempt_at)
+        } else {
+            repair_deadline
+        }
+    }
+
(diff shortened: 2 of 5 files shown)


zakura-network/src/zakura/handler.rs
```diff
@@ -2998,7 +2998,6 @@ pub async fn spawn_zakura_endpoint_with_header_sync_driver(
         config.zakura.header_sync.clone(),
         limits.max_frame_bytes,
     );
-    startup.status_refresh_interval = config.zakura.header_sync.status_refresh_interval;
```
ValarDragon, line 3001: so were just removing a degree of freedom here?

zakura-network/src/zakura/header_sync/state.rs
```diff
@@ -307,6 +307,17 @@ impl VctRootRepair {
...
 
+    pub(super) fn next_maintenance_deadline(&self) -> Instant {
+        let repair_deadline = self.started_at + VCT_ROOT_REPAIR_MAX_WALL_TIME;
+        // An in-flight attempt has its own request timeout, so ignore its stale
+        // retry timestamp while retaining the overall repair deadline.
+        if self.in_flight.is_none() {
+            repair_deadline.min(self.next_attempt_at)
+        } else {
+            repair_deadline
+        }
```
ValarDragon, line 318: > Exclude next_attempt_at from the VCT repair maintenance deadline while an attempt is in flight, while retaining request and overall repair deadlines.

Agreed this solves this goal.

ValarDragon (approved):

zakura-network/src/zakura/handler.rs
p0mvn: Yes, it removes an accidental inconsistent default, not the ability to override it.

Previously:

`HeaderSyncStartup::new(config)` ignored the configured refresh interval and used the 30-second default.

Production immediately corrected it with a separate assignment.

The PR moves that correction into the constructor.

Production behavior is unchanged. The redundant assignment is removed. 

Other callers now cannot accidentally ignore their configured interval.

PR #220, opened 2026-07-16
fix(consensus): use PoWLimit at averaging-window height

ValarDragon:
## Motivation

`ThresholdBits(height)` must equal `PoWLimit` for every height at or below
`PoWAveragingWindow` (17).

The early-chain return also required the difficulty context to contain fewer
than 17 entries. A candidate at height 17 has exactly 17 contiguous ancestors,
so that guard was skipped and the normal difficulty-adjustment formula ran. On
promptly mined early chains, this produced a harder threshold and rejected a
consensus-valid height-17 header.

## Solution

Make the early-chain return depend only on the candidate height, matching the
consensus rule independently of context length.

Add a regression test with exactly 17 contiguous, target-spaced ancestors. This
context previously selected the full-window adjustment branch; it now returns
`PoWLimit`. Update the related comments and changelog entry.

## Testing

- `cargo test -p zakura-state pow_limit_threshold`
- `cargo clippy -p zakura-state --tests -- -D warnings`
- `git diff --check`

The regression exercises the off-by-one boundary directly: candidate height 17
with a full 17-entry context.

- [Zcash protocol specification: difficulty adjustment](https://zips.z.cash/protocol/protocol.pdf#diffadjustment)
- [zcashd `GetNextWorkRequired`](https://github.com/zcash/zcash/blob/master/src/pow.cpp)

---

> [!NOTE]
> **Medium Risk**
> Touches consensus proof-of-work threshold logic; the change is a narrow spec-alignment fix for early heights only, with limited exposure on mainnet after checkpoints.
> 
> **Overview**
> **Early-chain difficulty** now returns the network **PoW limit** whenever the candidate height is at or below `PoWAveragingWindow` (17), without also requiring fewer than 17 context entries. That removes an off-by-one case at height 17 with a full ancestor window, where the adjustment formula could run and reject consensus-valid headers during header sync or on non-checkpointed test networks.
> 
> A regression test covers candidate height 17 with exactly 17 contiguous, target-spaced ancestors; related comments and the changelog are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d641788010edf0dedd3a1c856307e903289e3a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


evan-forbes (approved):
makes sense

p0mvn (approved):
Cross-checked against zcashd
https://github.com/zcash/zcash/blob/95bc5e5c368cf9a4c46fd2f1da51cdf1c2333e19/src/pow.cpp#L46-L71

ValarDragon:
@codex fix change log

PR #229, opened 2026-07-17
fix(sync): reconcile promoted header forks

evan-forbes:
## Motivation

Recent v2 P2P traces showed a terminal header-sync stall after a non-finalized fork reset. Four peers repeatedly returned the requested successor header, but every state commit failed with `UnknownAnchor` for the locally selected tip.

The reset promoted a branch whose intermediate full-block headers had never been seeded. State attempted to seed only the newly promoted tip, silently skipped the write when its parent was missing, and then published that missing tip. The header reactor also treated forward resets as growth and only replaced its in-memory hash when height increased. Finally, the driver classified the resulting local `UnknownAnchor` as peer misbehavior.

## Solution

- Detect when a successful non-finalized commit changes the best branch rather than directly extending it.
- Atomically find the durable divergence, remove the stale provisional suffix and reverse indexes, and write every promoted full-block header through the new tip before publishing the chain update.
- Reject unlinked one-header seeds instead of reporting a successful no-op.
- Preserve `VerifiedReset` for same-height, backward, and forward resets.
- Reanchor the header reactor on every verified reset, clearing forward scheduled, buffered, committing, and outstanding work before requesting from the new durable tip.
- Treat `UnknownAnchor` as a local header-frontier failure so honest peers are not scored.
- Document the operator-visible stall fix in the changelog.

This restores the invariant that any header tip published as a sync anchor already exists as a contiguous durable header chain.

## Testing

- `cargo test -p zakura-state side_chain`
  - Reproduces a work-1 side branch overtaking a work-10 branch.
  - Verifies every promoted header, the new forward/reverse indexes, removal of the stale branch index, and the durable best header tip.
- `cargo test -p zakura-state header_store_coherence`
  - 37 scripted, audit, restart, and randomized coherence tests.
- `cargo test -p zakura-network verified_reset`
  - Verifies same-height hash replacement and forward-reset work cancellation/restart.
- `cargo test -p zakura zakura_header_sync_driver_tests`
  - 48 driver tests, including reset identity and local `UnknownAnchor` classification.
- `cargo clippy -p zakura-state -p zakura-network -p zakura --tests -- -D warnings`

Local state builds used the repository-documented `CXXFLAGS="-include cstdint"` workaround for the bundled RocksDB headers.

### Follow-up Work

The independent block-sync token-bucket behavior that cancels a shared connection after message throttling remains out of scope for this PR and should be fixed separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 - Prevent initial sync from stalling at checkpoint boundaries by refilling the
   verifier submission window after stale apply completions.
+- Prevent header sync from stalling after a verified-chain fork by atomically
+  persisting the promoted header branch before reanchoring network requests.
 
 ### Changed
 
--- a/zakura-network/src/zakura/header_sync/reactor.rs
+++ b/zakura-network/src/zakura/header_sync/reactor.rs
@@ -537,9 +537,7 @@ impl HeaderSyncReactor {
 
     async fn handle_frontier_update(&mut self, update: FrontierUpdate) {
         match update.change {
-            FrontierChange::Snapshot
-            | FrontierChange::VerifiedGrow
-            | FrontierChange::VerifiedReset => {
+            FrontierChange::Snapshot | FrontierChange::VerifiedGrow => {
                 let frontier = update.frontier;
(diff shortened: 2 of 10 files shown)


evan-forbes:
closing as this will be supersceded by header-chain

PR #231, opened 2026-07-17
fix(network): deliver committed-tip block gossip to unready zcashd-compat sidecar peers

evan-forbes:
## Root cause

A zcashd-compat sidecar peer connects to exactly one Zakura node and, in this
zcashd lineage, has **no `sendheaders`/BIP130** — it learns of a new tip *only*
from block `inv` gossip (`inv` → `getheaders` → `headers` → fetch). It also has no
at-tip timeout that recovers a missed `inv`.

Committed-tip block gossip (a node *following* the network) is sent as
`Request::AdvertiseBlock`, which the peer set routes via
`select_block_broadcast_peers` — and that function's "always include configured
sidecar peers" carve-out **iterates `ready_services` only**. When the sidecar's
outbound peer service is momentarily *unready* (any in-flight/hung request to it,
up to `REQUEST_TIMEOUT` = 20s), it is silently excluded from the broadcast. Unlike
`AdvertiseBlockToAll` (used only for locally-mined blocks), `AdvertiseBlock` has no
unready-queue fallback. Because the sidecar is inv-only and never retries, it then
stalls until it becomes ready *and* a later gossip happens to coincide — then it
burst-catches-up.

This was confirmed on a live testnet node (read-only): during stalls the sidecar
received **zero** block invs while every `getheaders` it sent was answered in <1ms
(13/13), `blocks == headers` stayed frozen with `inflight == []` (ruling out a
block-download timeout), the TCP connection never dropped, and gossip continued
reaching *other* Zakura peers. Recovery was a single inv → one `getheaders` → a
batch of headers → a burst of block requests.

## Fix

Give the sidecar carve-out the same unready-capable delivery that
`AdvertiseBlockToAll` already has, scoped to configured sidecar IPs:

- `route_block_broadcast` now also queues the latest `AdvertiseBlock` for any
  configured sidecar peer that is currently unready (`queue_block_gossip_for_unready_sidecars`).
- `poll_ready` delivers the queued gossip to those peers the moment they are ready
  again (`deliver_queued_sidecar_block_gossip`), mirroring `broadcast_all_queued`.
- Only the newest hash is retained — a sidecar that missed earlier hashes learns
  them from its next `getheaders` locator (exactly the observed burst recovery).
- Sidecar identification for unready peers canonicalizes the key IP, matching
  `LoadTrackedClient::is_inbound_direct_from_ip` so mapped/native loopback compare
  equal.

This bounds the sidecar's stall to one readiness cycle instead of leaving it
unbounded. Ordinary peers are unchanged (still sampled ready-only).

## Tests

`zakura-network/src/peer_set/set/tests/vectors.rs`:
- `unready_sidecar_block_gossip_is_queued_not_dropped` — an unready sidecar's
  gossip is queued (not dropped) while a ready peer still receives it immediately.
- `queued_sidecar_block_gossip_delivered_once_ready` — driving the peer set
  re-readies the sidecar and delivers the queued gossip through the `poll_ready`
  wiring.

The existing `prop::sidecar_peer_always_receives_block_gossip` (ready-sidecar path)
still passes. `cargo test`/`clippy`/`fmt` clean on `zakura-network`; `zakura` builds.

## Notes

- Distinct from the stale `fix/zcashd-compat-sidecar-stalls` branch (mapped-address
  identity canonicalization + reserved inbound slots for reconnect); `main` already
  canonicalizes, and the deployed stall is not a reconnect (constant TCP port).
- The deployed zcashd having no at-tip `getheaders` timeout is a separate,
  lower-priority hardening (would bound damage from any *single* missed inv);
  fixing the gossip is the real fix.

Full investigation + evidence: `art/debug/zcashd-recovers/FINDINGS.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   tip. Startup rejects configured anchors above that base, and no longer
   backfills headers below a checkpoint anchor.
 
+### Fixed
+
+- Deliver committed-tip block gossip to configured zcashd-compat sidecar peers
+  even when they are momentarily unready. The "always include sidecars" carve-out
+  in block broadcasts only covered ready peers, so a sidecar that was unready when
+  a block was gossiped was skipped; because it follows a single upstream and
+  learns the tip only from block `inv`s, it then stalled until a later gossip
+  coincided with a ready service. The latest hash is now queued for an unready
+  sidecar and delivered once it is ready again, bounding the stall to one
+  readiness cycle.
+
 ## [1.0.1] - 2026-07-17
 
 ### Added
--- a/zakura-network/src/peer_set/set.rs
+++ b/zakura-network/src/peer_set/set.rs
@@ -98,7 +98,7 @@ use std::{
     convert,
     fmt::Debug,
     marker::PhantomData,
-    net::IpAddr,
+    net::{IpAddr, SocketAddr},
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
@@ -137,7 +137,7 @@ use crate::{
         InventoryChange, InventoryRegistry,
     },
     protocol::{
-        external::InventoryHash,
+        external::{canonical_socket_addr, InventoryHash},
         internal::{Request, Response},
     },
     BoxError, Config, PeerError, PeerSocketAddr, SharedPeerError,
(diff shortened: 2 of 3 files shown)


evan-forbes:
the v12 thing is unrelated but valid so handling in a different PR

ValarDragon (approved):
Good fix this makes sense to me

p0mvn (approved):

PR #236, opened 2026-07-17
fix(network): deliver mined-block gossip to peers unready at broadcast time

evan-forbes:
## Motivation

`PeerSet::broadcast_all` — the `AdvertiseBlockToAll` (mined/submitted-block) gossip path — queues a re-send for peers that were **unready** when a block was advertised, then drains the queued `send_multiple` futures in a tail loop:

```rust
while queued_broadcast_fut_receiver.recv().await.is_some() {}   // drops each future unpolled
```

That loop received each queued `send_multiple` future and **dropped it unpolled**. `Client::call` enqueues the peer request into the connection task **synchronously at call time** but keeps the response receiver *inside* the returned future. Dropping the future therefore drops the receiver, and when the connection task later dequeues the request it sees `tx.is_canceled()` and skips writing the block `inv` (`Connection::handle_client_request`). So the re-send to an originally-unready peer is racily lost.

For an ordinary peer this is benign (it re-syncs via headers). For a **zcashd-compat sidecar** — single upstream, learns the tip only from block `inv`s — a dropped tip `inv` stalls it until the next block. This is the **mined-block twin** of the committed-tip stall fixed in #231, by a different mechanism on a different code path; #231 does not touch this path.

### Reachability

- **Standard following nodes: unaffected.** Network blocks are advertised via `AdvertiseBlock` (committed tip), not `AdvertiseBlockToAll`.
- **Local mining: affected.** `AdvertiseBlockToAll` fires only for locally submitted/mined blocks, so regtest, docker e2e (`generate`), and any local-mining deployment running a sidecar exercise this path.

## Solution

Stop dropping the queued futures — **spawn** each one so it runs to completion and keeps its response receivers alive, mirroring `deliver_queued_sidecar_block_gossip` from #231:

```rust
while let Some(send_fut) = queued_broadcast_fut_receiver.recv().await {
    tokio::spawn(send_fut);
}
```

Spawning (rather than awaiting serially) keeps each send independent of a slow/dead peer and leaves the existing channel/backpressure/ban-scrub structure of `broadcast_all_queued` untouched. `ResponseFuture` is `Send + 'static`, so `tokio::spawn` compiles as-is. One-line behavioral change.

## Testing

`cargo test -p zakura-network peer_set` (81 tests) and `cargo clippy -p zakura-network --all-targets -- -D warnings` pass; `cargo fmt --all -- --check` clean.

Two new tests in `set/tests/vectors.rs`:

- **`mined_block_gossip_to_unready_peer_is_delivered_not_canceled`** — the regression test. Two peers, both forced unready (so the initial ready-peer send is empty and the drain loop runs immediately), `broadcast_all(AdvertiseBlockToAll(hash))` spawned and driven through `poll_ready` cycles until the broadcast future completes, then asserts each once-unready peer received `AdvertiseBlockToAll(hash)` **and** that its response channel is **not** canceled.
  - ⚠️ A plain "peer received the inv" assertion does **not** catch this bug: the mock enqueues the `ClientRequest` synchronously, so it arrives even on the buggy code. The distinguishing signal is `!client_request.tx.is_canceled()`. Verified this test **fails on the unfixed drain loop** (at the `is_canceled` assertion) and **passes with the fix**.
- **`mined_block_gossip_reaches_ready_peers`** — guards the common path: a block advertised while peers are ready reaches all of them synchronously.

Twin of #231 (committed-tip sidecar gossip). Same `C-bug` block-gossip-liveness class.

### Follow-up Work

None. Independent of #231 (touches only `broadcast_all`'s drain loop; no shared lines / no merge conflict).

### DIFF
--- a/zakura-network/src/peer_set/set.rs
+++ b/zakura-network/src/peer_set/set.rs
@@ -1199,7 +1199,18 @@ where
 
         async move {
             let _ = send_multiple_fut.await?;
-            while queued_broadcast_fut_receiver.recv().await.is_some() {}
+            // Each queued item is a `send_multiple` future whose response receivers
+            // keep the enqueued peer requests from being treated as canceled (see the
+            // `tx.is_canceled()` skip in `Connection::handle_client_request`).
+            // `send_multiple` calls `Client::call`, which enqueues each request to the
+            // connection task synchronously but holds the response receiver inside the
+            // returned future. Dropping that future unpolled — as `.is_some() {}` did —
+            // drops the receiver before the connection task drains the request, so it
+            // sees a canceled request and racily skips the block `inv`. Spawn each
+            // future so it is polled to completion and keeps its receivers alive.
+            while let Some(send_fut) = queued_broadcast_fut_receiver.recv().await {
+                tokio::spawn(send_fut);
+            }
             Ok(Response::Nil)
         }
         .boxed()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,19 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   tip. Startup rejects configured anchors above that base, and no longer
   backfills headers below a checkpoint anchor.
 
+### Fixed
+
+- Deliver mined/submitted block gossip to peers that were momentarily unready
+  when the block was advertised. A block broadcast via `AdvertiseBlockToAll`
+  queued a re-send for unready peers, but the queued send future was dropped
+  before the connection wrote the `inv`, so the connection treated the request
+  as canceled and silently skipped it. Because a zcashd-compat sidecar follows a
+  single upstream and learns the tip only from block `inv`s, it could then stall.
+  The queued send now runs to completion. Only local mining paths (regtest, e2e,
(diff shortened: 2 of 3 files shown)


evan-forbes:
the v12 is true but resolved in 6s with the current timeout so nbd imo

zakura-network/src/peer_set/set.rs
```diff
@@ -1220,7 +1220,18 @@ where
...
         async move {
             let _ = send_multiple_fut.await?;
-            while queued_broadcast_fut_receiver.recv().await.is_some() {}
+            // Each queued item is a `send_multiple` future whose response receivers
+            // keep the enqueued peer requests from being treated as canceled (see the
+            // `tx.is_canceled()` skip in `Connection::handle_client_request`).
+            // `send_multiple` calls `Client::call`, which enqueues each request to the
+            // connection task synchronously but holds the response receiver inside the
+            // returned future. Dropping that future unpolled — as `.is_some() {}` did —
+            // drops the receiver before the connection task drains the request, so it
```
p0mvn, line 1229: When looking at this comment in the future, without git digg, I'd be confused why `is_some` is relevant

p0mvn (approved):

PR #240, opened 2026-07-17
fix(state): complete database upgrades before startup

p0mvn:
## Motivation

`ZakuraDb::new` previously started the initial database format operation on a background thread and returned immediately. The state service could therefore begin reading the database or committing blocks while an upgrade was still changing its contents.

A database upgrade establishes the on-disk invariants expected by the running code. Exposing the database before it finishes creates a race between normal state access and the migration. This is particularly unsafe for migrations such as the planned Sprout-history repair, which must be complete before any caller observes the repaired state.

## Solution

Startup now has two distinct phases:

1. Run the one-time format operation synchronously. This applies any required upgrade, handles a newly created or downgraded database, and validates the resulting format. `ZakuraDb::new` does not return until this phase finishes.
2. Start the existing background worker only for optional periodic validity checks.

A periodic check is not an upgrade and does not search for a newer database version. When `debug_validity_check_interval` is configured, it periodically audits blocks written by the current process against the current database-format invariants. These checks remain asynchronous because they can repeat for the lifetime of the node. They are disabled by default.

The regression test blocks a generic startup upgrade and proves that `ZakuraDb::new` remains blocked until the upgrade is released and the new on-disk version is visible.

This PR is independently mergeable and has no dependency on the VCT Sprout persistence or repair changes.

## Testing

### Follow-up Work

The VCT Sprout-history repair migration will rely on this startup guarantee in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
+### Fixed
+
+- Database format upgrades now finish before startup exposes the finalized
+  state database; only configured periodic format checks continue in the
+  background.
+
 ### Changed
 
 - Header sync now schedules only forward ranges from the durable verified block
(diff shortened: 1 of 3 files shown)


p0mvn:
@cursor review

evan-forbes (approved):
straightfoward dope

thanks for splittin these up!!

PR #241, opened 2026-07-17
feat(state): add authenticated VCT Sprout repair artifacts

p0mvn:
## Motivation

Repairing historical Sprout anchors requires a reviewable, deterministic source of canonical Mainnet history. Peer-delivered VCT roots are not sufficient because they do not contain Sprout frontiers, and repair data must not be accepted from an unauthenticated runtime source.

## Solution

- Define a bounded binary artifact containing only canonical blocks that change Sprout.
- Authenticate its payload with SHA-256 and bind every record to canonical block indexes.
- Bind the terminal Sprout root and handoff identity to the separately embedded VCT frontier.
- Add an offline generator that opens a complete Mainnet archive read-only, generates the artifact, and validates its own output.
- Harden finalized-frontier generation so all four trees come from one stable finalized tip identity.
- Document the artifact format, trust boundary, validation, and independent reproduction procedure.

`MAINNET_ARTIFACT` remains `None`: this PR does not enable repair or change node startup behavior.

## Testing

- Focused artifact codec, canonical-prefix, handoff, generator, and finalized-tip race tests.

The tests reject malformed or oversized artifacts, digest and replay-root mismatches, canonical index mismatches, incorrect handoffs, historical frontier requests, and finalized-tip changes during generation.

### Follow-up Work

- Add the startup repair migration using the merged Sprout persistence and synchronous startup guarantees.
- Independently reproduce and review the canonical Mainnet artifact.
- Embed the reviewed bytes in a final activation PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9175,6 +9175,7 @@ dependencies = [
  "serde",
  "serde-big-array",
  "serde_json",
+ "sha2 0.10.9",
  "spandoc",
  "tempfile",
  "thiserror 2.0.18",
--- a/docs/design/verified-commitment-trees.md
+++ b/docs/design/verified-commitment-trees.md
@@ -822,3 +822,66 @@ cargo test -p zakura-state final_frontier
 cargo test -p zakura-utils --features zakura-checkpoints
 cargo test -p zakura --features zakura-checkpoints checkpoints
 ```
+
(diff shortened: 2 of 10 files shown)


evan-forbes (approved):

ValarDragon (approved):

p0mvn:
The v12 do not seem very useful

PR #244, opened 2026-07-17
fix(state): repair legacy VCT Sprout history

p0mvn:
## Motivation

The original VCT fast-sync path did not persist historical Sprout anchors. Databases produced by that path can therefore reject later JoinSplit spends whose anchors should be available.

The prevention fix protects new syncs, but existing affected Mainnet databases need a deterministic repair before the state service can safely use them.

## Solution

- Add database format upgrade `28.0.1` for affected pre-repair Mainnet VCT databases.
- Validate the reviewed artifact against its handoff frontier and the database canonical indexes before writing.
- Replay the empty anchor and every retained historical Sprout frontier in one atomic RocksDB batch.
- Restore the current Sprout tip only for databases whose finalized tip is below the VCT marker; preserve truthful post-marker tip state otherwise.
- Complete repair synchronously before exposing the database.
- Fail closed for missing artifacts, invalid inputs, read-only opens, or explicitly disabled upgrades.
- Keep repair cancellable, crash-safe, repeatable, and Mainnet-only.

`MAINNET_ARTIFACT` remains `None`. Until separately reviewed bytes are embedded, affected databases receive an actionable refusal and must use a snapshot or genesis resync. This PR does not activate automatic production repair.

> [!NOTE]
> This is stacked on #241. After #241 merges, this branch will be rebased onto `main` and the PR base changed to `main`.

## Testing

The migration tests cover eligibility, artifact absence, read-only and disabled-upgrade refusal, canonical mismatches, pruned block bodies, prefix-tip restoration, post-marker tip preservation, atomic replay, cancellation, interrupted reopen, repeatability, and startup visibility.

### Follow-up Work

- Independently reproduce and review the canonical Mainnet artifact.
- Embed the reviewed bytes in a separate activation PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   state database; only configured periodic format checks continue in the
   background.
 - Preserve Sprout note-commitment history during fresh verified-commitment-tree
-  fast sync, so later JoinSplit spends can use historical anchors.
+  fast sync, so later JoinSplit spends can use historical anchors. Affected
+  Mainnet databases that previously ran v2 p2p + fast mode require repair at
+  startup from a reviewed trusted artifact, snapshot redownload, or genesis
+  resync.
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


evan-forbes (approved):
highly doubt we ever need to think about downgrades, but would the worst case just be that the migration could be performed twice?

ValarDragon (approved):

p0mvn:
For this specific migration, yes: a downgrade could make the disk version repair-eligible again, causing the repair to run twice.

PR #246, opened 2026-07-18
refactor(network): correlate header operations by exact identity

p0mvn:
## Motivation

Header-sync state completions were correlated using peer/session/range geometry. That relies on implicit serialization and cannot safely distinguish retries from the separate header-commit and root-authentication operations that can originate in a single wire response.

This sets up work for further integration of the vct auth roots: https://github.com/zakura-core/zakura/pull/230

### 1. Split HeaderRangeCommitted

Split the old `HeaderRangeCommitted` path because it represented two semantically different events:

- `BestHeaderTipLoaded`: state reports its durable tip during startup/refresh. No reactor-owned operation completed.
- `HeaderRangeOperationCompleted`: a specific submitted commit completed, identified by peer, session, request ID, and operation kind.

### 2. Add `operation` field to `CommitHeaderRange`

In a future PR, we will utilize the same path for root auth and committing.

### 3. Introduced exact end-to-end identities for wire requests and state operation

This will allow detailed peer attribution with peers in the future.

Makes commit success/failure affect only the exact matching pending operation, preventing stale or overlapping events from corrupting scheduling.

https://github.com/zakura-core/zakura/pull/230

### DIFF
--- a/zakura-network/src/zakura/header_sync/events.rs
+++ b/zakura-network/src/zakura/header_sync/events.rs
@@ -286,25 +286,24 @@ pub enum HeaderSyncEvent {
         /// State repair generation that was resolved.
         generation: u64,
     },
-    /// State successfully committed a header range.
-    HeaderRangeCommitted {
-        /// First committed height.
-        start_height: block::Height,
-        /// New best header tip height.
+    /// State returned the durable best header tip during startup or refresh.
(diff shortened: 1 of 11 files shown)


p0mvn:
v12 is pre-existing

evan-forbes (approved):

PR #252, opened 2026-07-18
chore(release): prepare v1.0.2

p0mvn:
## Motivation

Publish the VCT Sprout-history persistence, repair, embedded artifact, and validation work as Zakura v1.0.2 with a consistent crates.io package graph.

## Solution

- bump Zakura and affected workspace crates to their SemVer-compatible release versions
- update dependency requirements, the lockfile, release metadata, and the v1.0.2 test configuration
- finalize user-facing and crate changelogs, including automatic writable-startup repair behavior and safe failure conditions
- keep the reviewed repair artifact embedded in official builds while omitting it from crates.io packages, implemented in stacked PR #253
- update the release support height and installation example

## Testing

- `cargo test -p zakura-state embedded_mainnet_artifact_matches_handoff --locked`
- `./scripts/check-crate-packaging.sh --verify` — all 13 packaged archives rebuilt successfully; `zakura-state` is 569,365 bytes
- `make pre-release RELEASE_TAG=v1.0.2 BASE_TAG=v1.0.1`
- verified the packaged archive omits `mainnet-sprout-history.bin` and the official-build assertion rejects a missing artifact
- Markdown lint on all changed Markdown files
- `git diff --check origin/feat/validate-pre-repair-vct-sprout-history...HEAD`
- IDE diagnostics for the edited release files

Risk classification: high because the release changes distribution of database-repair data. Official builds require the reviewed artifact; crates.io builds omit it and fail closed only for affected unrepaired databases.

### Specifications & References

Builds on #239, #240, #241, #244, #247, #250, and #253.

### AI disclosure

Prepared with AI assistance.

PR #253, opened 2026-07-18
fix(release): finalize v1.0.2 package graph

p0mvn:
## Summary

- exclude the large Mainnet Sprout-history artifact from the `zakura-state` crates.io package
- embed and require the reviewed artifact in official Docker/release builds
- make artifactless crates.io builds fail closed on databases that require repair
- bump `zakura-network` to 3.0.0 for its breaking public error-enum change and update dependent requirements
- document the channel-specific behavior, breaking API, and recovery options

## Motivation

The reviewed repair artifact is required by affected legacy VCT databases, but including it makes `zakura-state` exceed crates.io package constraints. Official builds must guarantee that the artifact is present, while crates.io builds must never download or substitute unreviewed bytes.

The release also adds a variant to the exhaustive public `HeaderSyncStartError` enum, which requires a major `zakura-network` version and a coherent dependent package graph.

## Tests

- `make pre-release RELEASE_TAG=v1.0.2 BASE_TAG=v1.0.1 CRATE_PACKAGING_VERIFY=1`
- `cargo semver-checks -p zakura-network --default-features`
- `ZAKURA_REQUIRE_VCT_SPROUT_HISTORY=1 cargo test -p zakura-state`
- Markdown lint on all changed Markdown files
- IDE diagnostics on changed source files

## Risk

High risk: this changes release packaging and the availability of a database-repair artifact. Source and official Docker builds require the artifact; crates.io builds intentionally omit it and fail closed when repair is needed.

## AI disclosure

Prepared with AI assistance.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -129,10 +129,10 @@ sed "s#$HOME/.zakura#identity_dir#g" \
   > zakurad/tests/common/configs/v<version>.toml
 ```
 
-The replacements are global path-string substitutions, mirroring
-`last_config_is_stored` — the default cache path also appears in fields other
-than `cache_dir` (for example `cookie_dir`), so per-field rewrites produce a
-snapshot the test rejects.
+      The replacements are global path-string substitutions, mirroring
+      `last_config_is_stored` — the default cache path also appears in
+      fields other than `cache_dir` (for example `cookie_dir`), so
+      per-field rewrites produce a snapshot the test rejects.
 
 - [ ] On the release commit, run the pre-release checks for the tag you are
       about to create, using the previous release tag as the base:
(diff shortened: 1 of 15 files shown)


p0mvn:
Addressed v12 to fail closed during cargo install if no artifact is present.

p0mvn:
Ah, I made it fail for everyone in [7658022](https://github.com/zakura-core/zakura/pull/253/commits/765802260638258be7e3d0be3ca87e599259df99) when addressing v12. The original was the way I intended

ebfull:
Auto-closed by the same history rewrite — see the note on #252. Two things for the redesign:

- The artifact is gone from git history, so the embed/exclude split here no longer has a source file to embed.
- As written, `build.rs` panics whenever the artifact file is missing — which is every crates.io build, for every user, not just affected databases. Artifact absence must be a runtime fail-closed condition (`MAINNET_ARTIFACT = None`), never a build failure.

PR #263, opened 2026-07-19
refactor: move mempool check

p0mvn:
## Motivation

ZIP-317 mempool fee policy was evaluated only when constructing `VerifiedUnminedTx`, after script and proof checks had completed. This made the policy ordering later than other mempool admission checks.

## Solution

- Calculate and validate mempool transaction fees after spent UTXOs are loaded and before script and proof checks.
- Reuse the calculated fee when constructing the verified transaction while retaining the constructor check as an invariant.
- Keep block transaction fee calculation and block validation behavior unchanged.
- Add coverage for early ZIP-317 rejection, constructor validation, and sufficiently funded mempool script rejection.

## Testing

- `cargo clippy -p zakura-chain -p zakura-consensus --all-targets -- -D warnings`
- `cargo test -p zakura-chain -p zakura-consensus`
- `markdownlint CHANGELOG.md --config .trunk/configs/.markdownlint.yaml`
- IDE diagnostics for changed files

### DIFF
--- a/zakura-consensus/src/transaction.rs
+++ b/zakura-consensus/src/transaction.rs
@@ -521,13 +521,14 @@ where
                 check::mempool_standard_input_scripts(tx.as_ref(), &spent_outputs)?;
             }
 
+            let mempool_transaction = req.mempool_transaction();
             let mut miner_fee = None;
-            if let Some(unmined_tx) = req.mempool_transaction() {
+            if let Some(unmined_tx) = mempool_transaction.as_ref() {
                 // Apply ZIP-317 policy before expensive cryptographic verification.
                 // VerifiedUnminedTx::new() repeats this check to preserve its constructor
                 // invariant.
                 let fee = Self::miner_fee(tx.as_ref(), &spent_utxos)?;
-                let unpaid_actions = transaction::zip317::unpaid_actions(&unmined_tx, fee);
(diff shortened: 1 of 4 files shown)


p0mvn:
V12 report is valid but was not triggered by this PR, addressed here: https://github.com/zakura-core/zakura/pull/265

This PR has run on a real node per the above comment.

zakura-consensus/src/transaction.rs
ValarDragon, line 514: Wow its crazy that these mempool checks are littered across like this. We should plan for a refactor to clean this.

zakura-consensus/src/transaction.rs
```diff
@@ -521,6 +521,19 @@ where
                 check::mempool_standard_input_scripts(tx.as_ref(), &spent_outputs)?;
             }
 
+            let mempool_transaction = req.mempool_transaction();
+            let mut miner_fee = None;
+            if let Some(unmined_tx) = mempool_transaction.as_ref() {
```
ValarDragon, line 526: This seems like it should be req.is_mempool as well?

ValarDragon: Nvm, The existing anchor/nullifier query already uses: if let Some(unmined_tx) = req.mempool_transaction() { ... }

ValarDragon (approved):
Great work!

PR #266, opened 2026-07-19
test(sync): make checkpoint coverage network-independent

p0mvn:
## Motivation

The ignored large-checkpoint acceptance test downloaded blocks from public Mainnet peers. On CI runners, sparse or unreliable peer connectivity could leave the node at genesis until the four-minute command timeout, failing otherwise healthy post-merge unit-test runs.

## Solution

- replace the public-network test with a local Regtest producer that mines 800 blocks without seed peers or a peer cache
- derive checkpoints at heights 0, 400, and 800, then sync a persistent consumer exclusively from the loopback producer
- preserve the restart-at-lower-height assertion after the consumer checkpoint-verifies both maximum-size ranges
- allow the sync harness to accept custom configurations and use stable network metadata in log assertions
- rename and tighten the dedicated nextest profile, removing the obsolete `TEST_LARGE_CHECKPOINTS` CI switch

## Testing

- `cargo test -p zakura --test acceptance sync_large_checkpoints_local --release --features default-release-binaries -- --ignored --nocapture`
  - verifies local P2P catch-up, checkpoint ranges 0–400 and 400–800, persistence, and restart behavior
- `cargo clippy -p zakura --test acceptance --features default-release-binaries -- -D warnings`
- `actionlint .github/workflows/tests-unit.yml .github/workflows/coverage.yml`

The generated checkpoints use `MAX_CHECKPOINT_HEIGHT_GAP`, preserving the original two-range checkpoint coverage without depending on external peers.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -114,9 +114,11 @@ failure-output = "immediate"
 default-filter = 'package(zakura-network) and test(~zakura::testkit::blocksync_fuzz)'
 
 
-[profile.sync-large-checkpoints-empty]
-slow-timeout = { period = "60m", terminate-after = 2 }
-default-filter = 'package(zakura) and test(=sync_large_checkpoints_empty)'
+[profile.sync-large-checkpoints-local]
+test-threads = 1
+slow-timeout = { period = "15m", terminate-after = 1 }
+failure-output = "immediate"
+default-filter = 'package(zakura) and test(=sync_large_checkpoints_local)'
 
 [profile.sync-full-mainnet]
 slow-timeout = { period = "30000m", terminate-after = 1 }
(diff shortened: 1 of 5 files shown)


p0mvn:
```
 Nextest run ID 34d45f6f-97a5-4202-8711-ae439ccffb50 with nextest profile: sync-large-checkpoints-local
    Starting 1 test across 8 binaries (430 tests and 25 binaries skipped, including 42 tests and 25 binaries via profile.sync-large-checkpoints-local.default-filter)
        PASS [ 566.801s] zakura::acceptance sync_large_checkpoints_local
────────────
     Summary [ 566.805s] 1 test run: 1 passed, 430 skipped
```

ValarDragon (approved):

PR #272, opened 2026-07-19
fix(zcashd-compat): clean up sidecar on shutdown

ValarDragon:
## Motivation

SIGINT and SIGTERM were handled by the Tokio runtime, which dropped `StartCmd::start` before its managed zcashd-compat cleanup ran. The zcashd child therefore kept its data-directory lock after Zakura exited. Startup also did not observe graceful shutdown until it reached the steady-state task loop, and an ordinary post-sidecar startup error or panic could bypass the bounded supervisor wait.

## Solution

Pass a cancellation token to the start command on a shutdown signal, then await orderly cleanup. Startup awaits now stop promptly when the token is cancelled.

After spawning a managed zcashd supervisor, a private lifecycle owner retains its task, shutdown sender, and bounded wait. Post-sidecar startup only shuts the owner down on cancellation, an ordinary error, or a catchable panic; on successful startup, it transfers that owner to the steady-state task loop. That loop continues monitoring the same supervisor and performs the normal bounded shutdown when Zakura exits.

## Testing

- `cargo test -p zakura commands::start::tests::startup_shutdown_interrupts_pending_initialization -- --exact`
- `cargo test -p zakura commands::start::tests::post_startup_completion_keeps_zcashd_supervisor_running -- --exact`
- `cargo test -p zakura commands::start::tests::post_startup_error_requests_zcashd_supervisor_shutdown -- --exact`
- `cargo test -p zakura commands::start::tests::post_startup_panic_requests_zcashd_supervisor_shutdown -- --exact`
- `cargo test -p zakura components::tokio::tests::graceful_shutdown_waits_for_root_future_cleanup -- --exact`
- `cargo test -p zakura zcashd_compat::supervisor::tests --lib`

The post-startup tests prove that successful startup keeps supervision active, while an error or a catchable panic notifies and awaits the supervisor before control continues.

### Follow-up Work

An end-to-end process test using a stub zcashd remains valuable, but is outside this focused lifecycle fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 - Reject transactions that do not meet ZIP-317 mempool fee policy before
   running script and proof checks. Block validation is unchanged.
+- Stop a managed zcashd-compat sidecar cleanly after Zakura receives SIGINT or
+  SIGTERM, instead of leaving it running and holding its data-directory lock
+  ([#272](https://github.com/zakura-core/zakura/pull/272)).
 
 ## [1.0.2-rc0] - 2026-07-19
 
--- a/zakurad/src/commands/start.rs
+++ b/zakurad/src/commands/start.rs
@@ -76,6 +76,7 @@
 pub(crate) mod zakura;
 
 use std::{
+    future::Future,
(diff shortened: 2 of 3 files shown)


czarcas7ic:
Closing in favor of separate PR

PR #274, opened 2026-07-19
fix(zakurad): reap managed zcashd on shutdown

czarcas7ic:
## Motivation

On SIGINT or SIGTERM, Zakura could drop `StartCmd::start` before its cleanup block ran, leaving the managed zcashd alive and holding its data directory lock.

## Solution

Route shutdown through `StartCmd` and await the existing supervisor cleanup once zcashd can have been spawned, while preserving prompt exit during earlier startup. The Linux acceptance test now exercises the real child process and restarts its exact command against the same data directory.

## Testing

On unpatched `f4d51970f`, the test stopped zcashd, terminated Zakura, and reproduced `Cannot obtain a lock ... Zcash is probably already running` on direct restart. With this change, Zakura terminated and reaped the stopped child, and the same restart became authenticated-RPC ready. Clippy, shutdown unit tests, formatting, and diff checks also passed.

AI disclosure: OpenAI Codex was used to investigate the shutdown path, implement the fix, and write the regression test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   running script and proof checks. Block validation is unchanged.
 - Streamline mempool script error handling so invalid scripts are reported as
   script verification errors.
+- Shut down a managed zcashd-compat process before Zakura exits on SIGINT or
+  SIGTERM.
 
 ## [1.0.2-rc0] - 2026-07-19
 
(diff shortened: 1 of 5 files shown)


evan-forbes (approved):
nice!

if we so desire in the future, we could add a timeout to the root future and not await indefinitely, along with the ability to press ctrl + c again to force cancel everything

definitely no need to overcomplicate things atm

PR #228, opened 2026-07-17
refactor(tracing): add lazy typed trace events

evan-forbes:
## Motivation

Zakura's production JSONL trace events were assembled through mutable raw JSON maps at call sites. This scattered schema ownership across reactors and drivers, made required fields implicit, and performed some peer hashing and diagnostic projection work before knowing whether the trace channel could accept an event.

## Solution

- add typed JSONL tables, events, and a reserve-first `JsonlEventEmitter` while preserving the public raw-map compatibility API
- move block reactor, peer-routine, sequencer, header/block driver, chain-tip mirror, handler, discovery, exchange, and legacy trace construction into owner-local trace modules
- reduce production control flow to semantic `trace_*` calls while leaving metrics and ordinary `tracing` logs beside the state changes they observe
- replace the zakurad commit-state property bag and key-dispatch helpers with typed owner projections
- absorb newly added exact header-operation identities and legacy peer-attribution traces into typed, reserve-first owner projections
- construct peer labels, display hashes, error strings, elapsed durations, collection counts, and event projections only after reserving channel capacity
- preserve existing table names, file names, event names, field representations, ordering, and null-versus-omitted behavior
- add literal JSON-order tests, parsed-shape tests, laziness tests, and inventory tests that reject serializer construction in core owner files

The block-sync sparse row projection remains internal to block-sync trace modules; it is no longer imported or constructed by reactor, peer-routine, or sequencer core logic.

## Size

The initial PR state was GitHub `+3,798/-2,309` (net `+1,489`) and `+754` production Rust LoC by the counter below. The current PR is GitHub `+6,344/-3,960` (net `+2,384`) and `+1,289` production Rust LoC.

Production LoC is defined as non-blank, non-comment physical Rust lines excluding `tests`, `testkit`, `benches`, `bench.rs`, `tests.rs`, and brace-delimited items immediately preceded by `#[cfg(test)]`; strings and comments are stripped before brace counting. The semantic owner split was kept despite the increase because it removes serialization policy and mutable field assembly from production control flow.

## Compatibility results

- exact key-order and parsed-shape tests pass for the typed owner projections and optional branches
- fixed handler null fields, sparse omissions, legacy-request null peers, numeric widths, hashed peers, display hashes, and error representations remain covered
- disabled, full, and closed emitters do not invoke event builders
- the smoke trace oracle passed after a 203-block from-scratch kind-6 catch-up and reorg recovery
[...]

PR #275, opened 2026-07-19
fix(network): do not advertise pruned block bodies

evan-forbes:
## Motivation

Pruned nodes retain canonical hash/height indexes after deleting raw block transactions. `FindBlockHashes`, which backs legacy `getblocks`, read those retained indexes without checking body availability. As a result, a pruned node could announce a block hash in an `inv` and then return `notfound` when the peer requested that block body.

The existing `inbound_pruned_advertised_block_notfound` integration test explicitly asserted this contradictory behavior. In production it can create large `notfound` retry storms during legacy sync.

## Solution

Make `FindBlockHashes` return only the contiguous prefix of hashes whose full block bodies are currently serveable. The response stops at the first pruned or header-only height; it never skips a gap because `getblocks` represents a contiguous extension from the requester's locator.

This change is limited to peer-facing `FindBlockHashes` behavior. Internal chain identity remains based on the retained indexes for `KnownBlock`, locators, depth, and hash/height lookups, preserving the state/body separation introduced in #146.

The PR also adds `legacy_peer_request.jsonl` when `[network.zakura] trace_dir` is configured. It attributes `FindBlocks` hash lists and block download outcomes to privacy-preserving process-local peer IDs, records exact-inventory versus speculative routing, and captures handshake/local heights and typed `notfound` outcomes.

## Testing

- `cargo test -p zakura-network legacy_peer_trace --lib`
- `cargo test -p zakura-network peer_set::set::tests --lib`
- `CXXFLAGS='-include cstdint' cargo test -p zakura-state chain_identity_uses_retained_hashes_when_checkpoint_bodies_are_skipped --lib`
- `CXXFLAGS='-include cstdint' cargo test -p zakura inbound_pruned_block_is_not_advertised --lib`
- `CXXFLAGS='-include cstdint' cargo test -p zakura-state service::tests --lib`
- `CXXFLAGS='-include cstdint' cargo clippy -p zakura-state --lib -- -D warnings`

The scoped `CXXFLAGS` workaround is required on this host because the bundled RocksDB 8.10 C++ headers do not include `<cstdint>` before using `uint64_t`; it does not modify the repository.

The renamed inbound integration test now verifies all three boundaries:

- state returns no hashes beyond a locator when the next body was pruned;
- legacy inbound maps that empty result to `Nil`, so no `inv` is sent; and
- an explicit request for the unavailable body still correctly returns `notfound`.

The pruning state test verifies that retained chain identity still works while peer-facing `FindBlockHashes` remains empty for unavailable bodies. Archive/full-body state request tests continue to return their normal hash sequences.


zakura-state/src/service/read/find.rs
```diff
@@ -463,10 +463,30 @@ where
...
+        if !body_available {
+            tracing::debug!(
+                ?height,
+                ?intersection,
+                ?stop,
+                ?max_len,
+                "stopping peer FindBlocks response before unavailable block body",
+            );
+            return None;
+        }
```
evan-forbes, line 486: main fix

evan-forbes (commented):
I can separate out the tracing here if that makes it easier, as most LoC are that

also fwiw, tracing improvement is in the pipeline after get all the important stuff finished! https://github.com/zakura-core/zakura/pull/228

p0mvn (approved):

p0mvn:
Non-blocking comment below

p0mvn:
I found one concrete resource-bounding issue in the tracing portion of this PR.

`LegacyPeerTrace::find_blocks_finish` serializes the entire peer-controlled `Response::BlockHashes` vector into one JSONL event. A peer can return up to `MAX_INV_IN_RECEIVED_MESSAGE` (50,000) block hashes, producing a roughly 3.3 MB row. The JSONL queue is bounded by event count (16,384), not total bytes, so a slow writer can retain many of these large rows and cause substantial memory and disk amplification. This only affects nodes that opt into tracing; the default disabled configuration is unaffected.

Recommended fix: keep `hash_count`, but bound the serialized hash list to a small fixed sample (or record only first/last hashes) and add a `hashes_truncated` field. Please also add a test using a 50,000-hash response that asserts the resulting event size/list length remains bounded. Alternatively, splitting the tracing feature from this PR would let the pruning fix ship independently; the pruning fix itself looks sound.

evan-forbes:
switched to just using a height range btw and no hashes

PR #276, opened 2026-07-19
chore: use Zakura snapshot endpoints

czarcas7ic:
Point installer and book links at the canonical snapshots page, and move benchmark and node-bake defaults to Zakura-named objects on the new machine origin. Remove the HTTP/1.1 workaround now that range handling lives at the edge, and make the checkpoint cache-version check accept the active database major instead of the retired fixed path.


ValarDragon (approved):

czarcas7ic:
DNM until snapshots have migrated

PR #277, opened 2026-07-19
perf(rpc): cache Sapling proving keys to disk on first use

ValarDragon:
## Motivation

`getblocktemplate`'s coinbase construction (`TransactionTemplate::new_coinbase` in `zakura-rpc/src/methods/types/transaction.rs`) called `LocalTxProver::bundled()` **on every template refresh**, re-parsing the ~50 MiB of Sapling proving parameters embedded in the binary each time. While internal mining is active this happens roughly every two seconds, so the node repeatedly paid the full parse cost just to build a coinbase transaction.

## Solution

Introduce a process-wide Sapling prover, used for **proof construction**, that caches its proving parameters to disk the first time they are needed and then reuses a single prover for the lifetime of the process:

- `zakura_consensus::sapling_prover()` returns a `&'static LocalTxProver`, lazily initialized on first use (e.g. the first coinbase build in `getblocktemplate`).
- On first init, the bundled Sapling spend/output parameters are written to the default Zcash parameters folder (`~/.zcash-params`) if a valid cache file isn't already present, then loaded from disk. Writes are **atomic** (temp file + rename) and validated by **exact file size**, so a truncated or wrong-sized cache is rewritten rather than loaded (avoiding a panic in `LocalTxProver::new`). If the folder can't be determined or written, it falls back to the bundled parameters in memory so proof construction still works.
- `new_coinbase` now uses this shared prover instead of building a fresh `LocalTxProver::bundled()` per call.

The consensus-critical **verification** path is deliberately left unchanged: the Sapling verifier only reads the prover's *verifying keys*, so it keeps using the bundled parameters in memory and never depends on any on-disk parameter files. This keeps block validation independent of disk state — only the proving path (mining) touches the disk cache.

The `sapling-spend.params` / `sapling-output.params` bytes are sourced from the already-bundled `wagyu-zcash-parameters` crate (added as a direct dependency; it was already present transitively via `zcash_proofs`'s `bundled-prover` feature), so no runtime network access is introduced.

## Testing

- `cargo build -p zakura-consensus -p zakura-rpc` — clean.
- `cargo clippy -p zakura-consensus -p zakura-rpc --all-targets` — no warnings.
- `cargo test -p zakura-consensus --lib primitives::sapling::tests` — added a unit test (`write_cache_file_is_idempotent_and_size_checked`) that exercises the cache logic: a missing file is written atomically, an already-cached file of the correct size is reused (not overwritten), and a file whose size doesn't match the expected length is treated as stale.

- `LocalTxProver` / bundled parameters: [`zcash_proofs`](https://docs.rs/zcash_proofs).
- The disk cache uses the same default parameters folder and file names (`sapling-spend.params`, `sapling-output.params`) as `zcashd`'s `fetch-params`, so pre-existing valid parameter files are reused.

https://claude.ai/code/session_01TRwWn98njQ6nrdyBg9Cscn

---
_Generated by [Claude Code](https://claude.ai/code/session_01TRwWn98njQ6nrdyBg9Cscn)_

---

> [!NOTE]
> **Low Risk**
> Changes affect mining/`getblocktemplate` proof construction only; consensus Sapling verification still uses bundled verifying keys and never reads the disk cache, with bundled fallback if caching fails.
> 
> **Overview**
> **Stops re-parsing ~50 MiB of bundled Sapling proving parameters on every `getblocktemplate` refresh** by introducing a lazy, process-wide `zakura_consensus::sapling_prover()` used when building coinbase transactions.
> 
> On first use, spend/output parameter files are written to the default Zcash params directory (reusing valid existing files), loaded via `LocalTxProver::new`, with **atomic temp-file writes**, **exact byte-size checks**, and **in-memory bundled fallback** if the cache cannot be created. **Sapling batch verification is unchanged** and still uses only the in-binary bundled prover so validation does not depend on disk.
> 
> Adds `wagyu-zcash-parameters` for cache population, exports `sapling_prover` from `zakura-consensus`, and includes a unit test for idempotent/size-gated cache writes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66fd2436491c840c911be5b063baf7bf36a0026b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


zakura-consensus/src/primitives/sapling.rs
```diff
@@ -27,8 +31,130 @@ use zcash_protocol::value::ZatBalance;
...
+///    construction still works.
+fn init_coinbase_sapling_prover() -> LocalTxProver {
+    match cache_bundled_sapling_params() {
+        Ok((spend_path, output_path)) => {
+            info!(
+                ?spend_path,
+                ?output_path,
+                "loading Sapling proving parameters from the on-disk cache"
+            );
+            LocalTxProver::new(&spend_path, &output_path)
```
p0mvn, line 91: `is_cached` only validates the file length, but a correctly sized file can still contain corrupted bytes. In that case this call reaches `LocalTxProver::new`, which validates/deserializes the contents and panics on an invalid hash or encoding. That panic bypasses the `Err` fallback below and poisons `COINBASE_SAPLING_PROVER`, so subsequent proving attempts will also panic until the cache is repaired manually. Could we validate the hash/content before treating the file as cached, or otherwise ensure an invalid cache safely falls back to `LocalTxProver::bundled()`?

p0mvn: > an invalid cache safely falls back to LocalTxProver::bundled()

Perhaps, this is best

zakura-consensus/src/primitives/sapling.rs
```diff
@@ -27,8 +31,130 @@ use zcash_protocol::value::ZatBalance;
...
+/// The exact size in bytes of a valid `sapling-output.params` file. See [`SAPLING_SPEND_BYTES`].
+const SAPLING_OUTPUT_BYTES: u64 = 3_592_860;
+
+/// The Sapling prover used to *construct* Sapling proofs, for example the shielded outputs of a
+/// coinbase transaction built by `getblocktemplate`.
+///
+/// Unlike [`SAPLING`], this prover is only initialized when the node actually needs to create
+/// Sapling proofs, and its proving parameters are cached to disk on first use so they don't need to
+/// be re-parsed from the bundled copy on every template refresh.
+static COINBASE_SAPLING_PROVER: Lazy<LocalTxProver> = Lazy::new(init_coinbase_sapling_prover);
```
p0mvn, line 61: The process-wide `Lazy` is the valuable optimization here, but I don't think persisting the bundled bytes to disk provides an additional parsing benefit. In `zcash_proofs`, `LocalTxProver::bundled()` loads the embedded bytes and calls `parse_parameters()`, while `LocalTxProver::new()` opens the files and then calls the same `parse_parameters()` path (including deserialization and full hash validation). Thus after every restart the disk-backed path still performs the expensive ~50 MiB parse; the speedup within a process comes entirely from retaining one `LocalTxProver` in a `Lazy`.

Could this be simplified by reusing the existing bundled `SAPLING` singleton for proving as well as verification, for example by having `sapling_prover()` return `Lazy::force(&SAPLING)`? That would still reduce parsing from once per template refresh to once per process, while removing the cache writes/`sync_all`, filesystem and corruption failure modes, the second process-wide prover (and potential second parse), the duplicated size constants, and the direct `wagyu-zcash-parameters` dependency. Verification would remain independent of disk because the shared prover would still be initialized exclusively from the bundled parameters.

ValarDragon:
Agreed, w/ this PR I was testing claude code from phone. 

This went in a significantly different direction than I wanted. If we wrote to disk it only made sense to do an uncompressed version. But tbh just an in-memory uncompressed form suffices here

PR #280, opened 2026-07-19
fix(state): guard pruning for zcashd compatibility

evan-forbes:
> **Status:** This is an experimental implementation kept for future reference. We are not pursuing this approach now, and this PR will be closed after creation.

## Motivation

A pruned Zakura node can delete raw block transaction data before its configured zcashd compatibility sidecar has downloaded it. Once deleted, Zakura can no longer satisfy the sidecar's block requests, so zcashd can remain permanently behind.

This PR documents one small implementation approach in executable form.

## Solution

- Create an in-memory height watermark when zcashd compatibility mode is enabled.
- Seed it from the protected sidecar's `version.start_height`.
- Advance it after Zakura successfully serves blocks requested through the existing P2P connection.
- Reset it conservatively when the sidecar disconnects.
- Cap checkpoint and online raw-transaction pruning at the exclusive `watermark + 1` boundary.
- Preserve existing pruning behavior when compatibility mode is disabled.

The implementation assumes one configured sidecar. It adds no new P2P messages or active probing.

## Testing

- `CXXFLAGS='-include cstdint' cargo check -p zakura-state --all-targets`
- `CXXFLAGS='-include cstdint' cargo check -p zakura --all-targets`
- `cargo check -p zakura-network --all-targets`
- `CXXFLAGS='-include cstdint' cargo clippy -p zakura-network -p zakura-state -p zakura --all-targets -- -D warnings`
- Focused network test verifies the served-block watermark only advances.
- Focused state tests verify the exclusive boundary and pruning-range cap.

The `CXXFLAGS` setting works around missing `<cstdint>` includes in the locally bundled RocksDB build and is not part of this change.

### Follow-up Work

If this design is revisited, decide whether a served-block watermark is sufficiently strong evidence of zcashd's validated active-chain height, or whether an explicit active P2P probe is required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
+- Prevent pruned zcashd-compat nodes from deleting block data ahead of the
+  configured sidecar's observed block-serving watermark.
 - Reject transactions that do not meet ZIP-317 mempool fee policy before
   running script and proof checks. Block validation is unchanged.
 - Streamline mempool script error handling so invalid scripts are reported as
--- a/zakura-network/CHANGELOG.md
+++ b/zakura-network/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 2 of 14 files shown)


evan-forbes:
Closing this experimental draft because we are not pursuing the served-block watermark approach at this time. Keeping the branch and PR as an implementation reference if we revisit zcashd-compatible pruning in the future.

PR #281, opened 2026-07-19
Remove redundant sighashes

czarcas7ic:
## Motivation

Zakura reuses one `SigHasher` per transaction, but signature checks still rebuilt transaction-wide ZIP 243 components for each V4 signature and ZIP 244 amounts and scriptPubKeys digests for each non-`ANYONECANPAY` V5/V6 signature. With a signature on each transparent input, that repeated work grows quadratically.

## Solution

Cache the shared signature-hash components inside `zakura-chain`. V5/V6 extend Zakura's native ZIP 244 digest tree with amounts, scriptPubKeys, per-input, and `SIGHASH_SINGLE` output digests. V4 uses a private ZIP 243 cache for prevouts, sequences, outputs, JoinSplits, and Sapling components. V3 and earlier transactions continue to use the existing librustzcash fallback.

Input index, runtime hash type, `SIGHASH_SINGLE` selection, and legacy `scriptCode` remain finalized for each requested digest. The interpreter callback and `is_valid` API stay unchanged, and Zakura continues to use released librustzcash crates.

The published `HashType` bitflags and `sighash_v4_raw` return type remain unchanged. Cached paths convert to a private canonical type, while V5/V6 script validation continues to accept only the six defined hash-type bytes.

- [ZIP 243](https://zips.z.cash/zip-0243)
- [ZIP 244](https://zips.z.cash/zip-0244)

## Testing

- `cargo test -p zakura-chain --locked`
- `cargo test -p zakura-script --locked`
- `cargo test -p zakura-consensus --locked script::tests`
- `cargo clippy -p zakura-chain -p zakura-script --all-targets --locked -- -D warnings`
- `markdownlint zakura-chain/CHANGELOG.md --config .trunk/configs/.markdownlint.yaml`


czarcas7ic:
@codex review

czarcas7ic:
@codex review

zakura-chain/src/transaction/zip244.rs
```diff
@@ -362,15 +420,12 @@ fn hash_transparent_txid(
     inputs: &[transparent::Input],
     outputs: &[transparent::Output],
 ) -> Blake2bHash {
-    let mut h = hasher(ZCASH_TRANSPARENT_HASH_PERSONALIZATION);
-    // The transparent bundle is absent (and the digest is the bare
-    // personalization hash) only when there are no inputs and no outputs.
```
ValarDragon, line 367: cref https://github.com/valargroup/zebra/pull/385

Would like to bring that over sooner than later

zakura-chain/src/transaction/sighash.rs
```diff
@@ -117,6 +126,29 @@ impl SigHasher {
...
+                .signature_hash(
+                    canonical_hash_type.encode(),
+                    input_index_script_code
+                        .as_ref()
+                        .map(|(index, script_code)| (*index, script_code.as_slice())),
+                )
+                .expect(
+                    "sighash precondition violated: callers must pass an in-bounds input_index",
+                );
+        }
```
czarcas7ic: These handle the v3 through v6 cases. v1 and v2 would fall through to the original sighash call but currently gets picked up by sighash_v4_raw before it reaches here, so the fall through here shouldn't get hit. That being said, AI recommends to keep the fall through here, which I am.

ValarDragon: v3 can't be hit on mainnet, we can just pass that off to LRZ - it'd only affect sync and hasn't been a problem

zakura-chain/src/transaction/zip244.rs
czarcas7ic: https://github.com/zakura-core/zakura/pull/288

ValarDragon:
Can we do a new typed enum for:   zcash_transparent::sighash::SighashType

Concern is another caller could bypass and put an incorrect data type.

zakura-chain/src/transaction/sighash/legacy.rs
```diff
@@ -0,0 +1,402 @@
+//! Native ZIP-143 and ZIP-243 signature hashing for V3 and V4 transactions.
```
ValarDragon: Why is this called legacy.rs?

Also do we save anything with removing v3 signing here?

Shouldn't this just be named v4.rs?

ValarDragon:
@v12 please review

evan-forbes:
@v12-auditor

ValarDragon:
committed one type change to use LRZ constants instead of re-defining, and then an extra test

zakura-chain/src/transaction/sighash/legacy.rs
czarcas7ic: Was called legacy due to it not just working for v4. I am changing to just work for v4 and then will change to v4.rs

zakura-chain/src/transaction/sighash.rs
```diff
@@ -95,8 +100,12 @@ impl SigHasher {
         nu: NetworkUpgrade,
         all_previous_outputs: Arc<Vec<transparent::Output>>,
     ) -> Result<Self, Error> {
+        let precomputed_tx_data = PrecomputedTxData::new(trans, nu, all_previous_outputs.clone())?;
+
         Ok(SigHasher {
-            precomputed_tx_data: PrecomputedTxData::new(trans, nu, all_previous_outputs)?,
+            precomputed_tx_data,
+            v4: V4Sighash::new(trans, nu, &all_previous_outputs),
```
ValarDragon, line 182: Why do we do this if its not a v4 tx type tho?

zakura-chain/src/transaction/sighash/legacy.rs
czarcas7ic: https://github.com/zakura-core/zakura/pull/281/commits/555ffde3c0a73eea76b3db498704e1ab8e061624

zakura-chain/src/transaction/sighash.rs
czarcas7ic: We dont, it returns None for non V4 transactions

evan-forbes (approved):
ran a benchmark locally and saw the expected speed improvements

I, fable, and sol couldn't find any consensus breaking changes, but we're running a fuzzer with librustzcash to make sure 

very nice

ValarDragon (approved):
Lgtm after multiple review passes

I still would like to update to not pay allocation or wasted hashing overhead when it’s not needed though

PR #286, opened 2026-07-19
fix(network): mutate peer ban list in place

ValarDragon:
## Motivation

The ban list was shared through immutable `Arc<IndexMap<...>>` snapshots in a
watch channel. Each new ban therefore cloned the complete retained list while
the peer-set and inbound listener retained their snapshots. This made filling
the ban list increasingly expensive.

## Solution

Share one `RwLock`-protected bounded FIFO ban list and mutate it in place.
The list uses `HashSet<IpAddr>` for membership checks and `VecDeque<IpAddr>`
for O(1) FIFO eviction. Peer-set and inbound-listener checks use a read-only
`BannedIps` handle. This removes snapshot cloning and linear eviction from the
ban path while preserving the existing 20,000-IP bound.
The ban-list write lock is held only for insertion and eviction, so peer checks
remain unblocked during address-book cleanup.

## API changes

`AddressBook::bans` and `AddressBookUpdater::spawn` now return `BannedIps`
rather than a watch receiver/snapshot. The handle exposes membership checks
only; it does not expose mutation of the shared ban map.

## Testing

- `npx --yes markdownlint-cli CHANGELOG.md zakura-network/CHANGELOG.md --config .trunk/configs/.markdownlint.yaml`
- `cargo test -p zakura-network ban_list_evicts_the_oldest_ip_at_capacity --lib`
- `cargo test -p zakura-network misbehavior_ban_does_not_panic_with_max_connections_per_ip_above_one --lib`
- `cargo test -p zakura-network broadcast_all_queued_removes_banned_peers --lib`

## AI disclosure

Used Codex for the implementation, test updates, and PR description.

---

> [!NOTE]
> **Medium Risk**
> Breaking `zakura-network` API and a shared `RwLock` on the ban-check path in peer networking; behavior (FIFO bound, IP-level bans) is intended to stay the same.
> 
> **Overview**
> Fixes ban-list cost as the 20,000-IP cap fills by **mutating one shared list in place** instead of cloning full `Arc<IndexMap<…>>` snapshots on every ban and pushing updates through a watch channel.
> 
> **`BannedIps`** (re-exported) wraps `Arc<RwLock<BanList>>` with **`contains` only** for peer-set and inbound checks. **`BanList`** uses a **`HashSet`** for membership and a **`VecDeque`** for FIFO eviction at **`MAX_BANNED_IPS`**, replacing **`shift_remove_index(0)`** on an ordered map.
> 
> **Breaking:** `AddressBook::bans()` and **`AddressBookUpdater::spawn`** now return **`BannedIps`** instead of `watch::Receiver<Arc<IndexMap<IpAddr, Instant>>>`. Ban timestamps are no longer exposed on the public API. The address-book updater no longer republishes ban snapshots after misbehavior updates—callers share the same handle from spawn.
> 
> **`PeerSet`** and **`accept_inbound_connections`** take **`BannedIps`** and call **`contains`** on the live list rather than borrowing cloned maps from a watch receiver.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 468e6ebea0439a5cd06221ec6ae75dfc33734dfc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/zakura-network/src/address_book_updater.rs
+++ b/zakura-network/src/address_book_updater.rs
@@ -1,13 +1,7 @@
 //! The timestamp collector collects liveness information from peers.
 
-use std::{
-    cmp::max,
-    net::{IpAddr, SocketAddr},
-    sync::Arc,
-    time::Instant,
-};
+use std::{cmp::max, net::SocketAddr, sync::Arc};
 
-use indexmap::IndexMap;
 use thiserror::Error;
 use tokio::{
     sync::{mpsc, watch},
@@ -17,7 +11,7 @@ use tracing::{Level, Span};
 
 use crate::{
     address_book::AddressMetrics, meta_addr::MetaAddrChange, types::PeerServices, AddressBook,
-    BoxError, Config,
+    BannedIps, BoxError, Config,
 };
 
 /// The minimum size of the address book updater channel.
@@ -40,6 +34,7 @@ impl AddressBookUpdater {
     ///
     /// Returns handles for:
     /// - the address book,
+    /// - the shared list of banned peer IP addresses,
     /// - the transmission channel for address book update events,
     /// - a watch channel for address book metrics, and
     /// - the address book updater task join handle.
@@ -51,7 +46,7 @@ impl AddressBookUpdater {
         advertised_services: PeerServices,
     ) -> (
         Arc<std::sync::Mutex<AddressBook>>,
-        watch::Receiver<Arc<IndexMap<IpAddr, Instant>>>,
+        BannedIps,
         mpsc::Sender<MetaAddrChange>,
         watch::Receiver<AddressMetrics>,
         JoinHandle<Result<(), BoxError>>,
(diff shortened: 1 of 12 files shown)


zakura-network/src/address_book_updater.rs
```diff
@@ -73,6 +68,10 @@ impl AddressBookUpdater {
         .with_expose_peer_addresses(config.expose_peer_addresses);
         let address_metrics = address_book.address_metrics_watcher();
         let address_book = Arc::new(std::sync::Mutex::new(address_book));
+        let bans = address_book
+            .lock()
+            .expect("mutex should be unpoisoned")
+            .bans();
```
ValarDragon: Something seems off here

ValarDragon:
Lets leave this out of next tag, I want to swap the data-structure before we do this

evan-forbes (approved):

PR #291, opened 2026-07-19
perf(rpc): reuse process-wide Sapling prover

p0mvn:
## Summary
- parse bundled Sapling proving parameters once per process and share the prover between verification and coinbase construction
- remove per-template prover construction without adding an on-disk parameter cache or new direct dependency

Supersedes #277 with the simplified design from review. Blocking-pool isolation is separately scoped in #292.

## Test plan
- [x] `cargo test -p zakura-consensus --lib primitives::sapling::tests::sapling_prover_is_reused`
- [x] `cargo test -p zakura-rpc --lib methods::types::get_block_template`
- [x] `cargo clippy -p zakura-consensus -p zakura-rpc --all-targets -- -D warnings`
- [x] `cargo build -p zakura-consensus -p zakura-rpc`

### DIFF
--- a/zakura-rpc/src/methods.rs
+++ b/zakura-rpc/src/methods.rs
@@ -2562,16 +2562,23 @@ where
                     // Respond instantly with an empty block upon a chain tip change so that
                     // the miner doesn't waste their effort trying to extend a shorter
                     // chain.
-                    return Ok(BlockTemplateResponse::new_internal(
-                        &self.network,
-                        precomputed_coinbase,
-                        miner_params,
-                        &chain_info,
-                        server_long_poll_id,
-                        vec![],
-                        submit_old,
-                    )
-                    .into())
+                    let network = self.network.clone();
+                    let miner_params = miner_params.clone();
+                    let response = tokio::task::spawn_blocking(move || {
+                        BlockTemplateResponse::new_internal(
+                            &network,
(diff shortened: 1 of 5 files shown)


zakura-rpc/src/methods.rs
```diff
@@ -2562,16 +2562,23 @@ where
...
-                        precomputed_coinbase,
-                        miner_params,
-                        &chain_info,
-                        server_long_poll_id,
-                        vec![],
-                        submit_old,
-                    )
-                    .into())
+                    let network = self.network.clone();
+                    let miner_params = miner_params.clone();
```
ValarDragon: Why the new clones?

zakura-rpc/src/methods.rs
```diff
@@ -2562,16 +2562,23 @@ where
...
+                    let response = tokio::task::spawn_blocking(move || {
```
ValarDragon: Why the spawn blocking change?

ValarDragon:
Much better approach! Tho I don't follow the spawn blocking update

zakura-rpc/src/methods.rs
p0mvn: AI auto-addressed v12 from #277 , moving it out

ValarDragon (approved):
nice job

PR #292, opened 2026-07-19
perf(rpc): isolate block template construction

p0mvn:
## Problem

`getblocktemplate` is an async RPC, but parts of its work ran synchronously on a Tokio async worker:

- **Final response construction.** `BlockTemplateResponse::new_internal` was called directly in both final response paths. `new_internal` constructs a coinbase transaction whenever it is not given a usable precomputed coinbase. For shielded miner addresses that work includes Sapling proof construction and can take seconds, and with #291 the first construction in a process can also initialize the shared prover by parsing the bundled Sapling parameters. While this synchronous work runs, the Tokio worker cannot poll unrelated RPC futures. This affected:
  - the normal template response, which always passes `None` and builds its coinbase inline;
  - the tip-change response, where the provisional coinbase is discarded if the predicted height no longer matches, causing `new_internal` to build a replacement inline.
- **Redundant provisional coinbase precomputation.** The long-poll loop precomputed a provisional coinbase via `spawn_blocking` inside the `wait_for_new_tip` future, racing it in the loop's `select!`. Whenever another branch won (the mempool poll timer fires every few seconds during a long poll), that future was dropped — but `spawn_blocking` tasks are not cancellable, so every loop iteration leaked a fresh proving task whose result was discarded. With a shielded miner address, that is seconds of wasted proving CPU per poll tick, repeatedly queued onto Tokio's blocking pool for the duration of the long poll.

## Solution

- Run `BlockTemplateResponse::new_internal` (together with mempool transaction selection in the normal path) through `tokio::task::spawn_blocking` in both final response paths. This moves coinbase proof generation and first-use prover initialization onto Tokio's blocking pool while the RPC future yields its async worker.
- Cache the provisional coinbase precompute across long-poll loop iterations as a height-keyed `Shared` future held outside the loop. A new proving task is spawned only when the predicted height changes, so mempool poll ticks reuse the in-flight (or completed) precompute instead of leaking a new proving task each iteration. The existing height-mismatch check still discards the provisional coinbase when the `+2` prediction turns out wrong.

`Network` and `MinerParams` are cloned into owned values because a blocking task must own `Send + 'static` inputs; the remaining response inputs are already owned and are moved into the task. Caching by height is sound because the provisional coinbase depends only on the network, the predicted height, and the miner parameters, all of which are constant within one RPC call. The precomputed-coinbase fast path in `new_internal` and all response contents remain unchanged.

This is the separately scoped follow-up to the blocking-RPC concern raised on #277 and is independent of the shared-prover optimization in #291.

## Test plan

- [x] `cargo test -p zakura-rpc --lib methods::types::get_block_template`
- [x] `cargo clippy -p zakura-rpc --all-targets -- -D warnings`


p0mvn:
Blocked on next tag

evan-forbes (approved):
depending on if the rpc is exposed for non-miners, then we might want to go ahead and fix the exhaustion issue brought up by v12

will defer to author on when, if, and where we do that. could also be a flup

p0mvn:
@v12-auditor review

p0mvn:
@v12 audit

p0mvn:
@v12this audit

crates/zakura-rpc/src/methods/types/get_block_template.rs
```diff
@@ -601,6 +617,23 @@ where
...
+            .map_err(|_| {
+                ErrorObject::borrowed(
+                    ErrorCode::InternalError.code(),
+                    "timed out waiting for block template construction capacity",
+                    None,
+                )
+            })?;
+
+        Ok(permit.expect("template construction semaphore is never closed"))
+    }
```
ValarDragon: Why did this get added in? The change seemed more right without this

p0mvn: I was worried about resource exhaustion with the async work and was trying things to limit it.

That has led to more complexity than I desired, with only a marginal benefit, so I removed it.

p0mvn:
@v12this audit

p0mvn:
@v12this audit

p0mvn:
@v12this audit

PR #297, opened 2026-07-20
docs(spec): define fork-aware header chain engine

evan-forbes:
## Motivation

PR #229 exposed that the current single-chain header overlay cannot safely reconcile promoted forks across header sync, full-state resets, finalization, restart, and asynchronous completion boundaries.

This draft makes the replacement contract reviewable before implementation. It also captures the consensus boundary of a headers-only client: observable header validity is not proof of transaction or state validity.

## Solution

Add a normative, fork-aware header-chain engine specification that defines:

- a durable header DAG with independent `finalized`, `header_best`, and `verified_best` frontiers;
- production consensus checks, settled network-upgrade pins, contextual validation, and deterministic greatest-work fork choice;
- mode-specific finality and bounded deterministic retention;
- generation-aware state transitions, body-failure evidence, VCT repair, and crash recovery;
- a versioned native v8 discovery protocol with exact auxiliary-record encoding; and
- a conformance, differential, property, crash, protocol, resource, audit-replay, and fuzzing matrix.

This document is the redesign oracle for PR #229; it intentionally does not preserve the current patch structure.

## Testing

Documentation-only change.

- Confirmed the repository copy is byte-for-byte identical to the audited source.
- `git diff --check`
- Mechanical audit: 111 unique normative rule IDs, balanced code fences, and no `TBD`, `TODO`, or `FIXME` markers.

- Follow-up to #229.
- [Rendered implementation and testing plan](https://gist.github.com/evan-forbes/7e052145bd8e4bc48b2335911e8a6b40).

### Follow-up Work

Implement the engine in reviewable phases after the normative contract is accepted.

### DIFF
--- a/docs/specs/fork-aware-header-chain-engine.md
+++ b/docs/specs/fork-aware-header-chain-engine.md
@@ -1,8 +1,8 @@
 # Fork-aware headers-only chain engine specification
 
-Status: normative design oracle for the replacement of PR #229  
-Version: 1.2  
-Date: 2026-07-20  
+Status: normative design oracle for the replacement of PR #229<br>
+Version: 1.3<br>
+Date: 2026-07-21<br>
 Scope: Zakura native header sync and its integration with Zakura full state
 
 ## Document overview
@@ -85,7 +85,7 @@ Lists the Zakura code and official Zcash specifications that define the required
 ## 1. Scope and authority
 
 - **ZC — Zcash consensus:** mandatory production Mainnet/Testnet behavior required to accept the same observable header rules as Zcash full consensus. Exceptions for custom networks or disabled proof of work are never included in a ZC rule.
-- **ZP — Zcash deployment security:** production-network trust pins or operational requirements imposed by the Zcash specification on full validators, but not derivable from header validity alone.
+- **ZP — Zcash deployment security:** trust pins or operational requirements with the exact scope imposed by the Zcash specification on full validators, but not derivable from header validity alone. Stronger Zakura deployment policy is classified as LS.
 - **ZF — Zakura fork choice:** deterministic local policy used by Zakura when consensus does not prescribe a deterministic choice.
 - **ZW — Zakura wire/integration:** native protocol or full-node integration behavior.
 - **LS — local safety/liveness:** crash consistency, resource, scheduling, and operator policy.
@@ -209,7 +209,9 @@ Validation proceeds in the following order so cheap bounds precede CPU work and 
 
 **LC-COMMIT-02 [ZW] — Custom-schedule commitment parsing.** Regtest and configured custom networks MUST run the same commitment parser using their configured activation schedule, including overlapping activation-height behavior. A custom activation schedule MUST NOT bypass or guess the commitment variant.
 
-**LC-VAL-05 [ZC] — Target and hash validation.** On production Mainnet/Testnet, compact `nBits` MUST decode to a positive, non-overflowing target no easier than the network PoW limit, and the little-endian integer value of the header hash MUST be less than or equal to that target. LC-VAL-05 runs before LC-VAL-04 so the free target checks reject a candidate before any Equihash CPU cost, matching the check order in `zakura-consensus::block::check`.
+**LC-VAL-05 [ZC] — Target and hash validation.** On production Mainnet/Testnet, compact `nBits` MUST decode to a positive, non-overflowing target no easier than the network PoW limit, and the little-endian integer value of the header hash MUST be less than or equal to that target.
+
+Non-normative implementation advice: run LC-VAL-05 before LC-VAL-04 so the inexpensive target checks reject a candidate before any Equihash CPU cost, matching the current check order in `zakura-consensus::block::check`. Validation order is not a Zcash consensus rule.
 
 **LC-VAL-04 [ZC] — Equihash verification.** On production Mainnet/Testnet, the Equihash solution size and parameters MUST match the network and height, and the Equihash proof MUST verify. Production headers MUST NOT use the short Regtest proof shape.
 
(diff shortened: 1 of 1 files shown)


docs/specs/fork-aware-header-chain-engine.md
```diff
@@ -0,0 +1,756 @@
...
+
+**LC-GEN-04 [LS] — Stale-result rejection.** Before any effect, completion handling MUST compare generation, branch identity, request identity, and pending owner. A stale or unowned result MUST have no frontier, coverage, retry, repair, scheduling, publication, body-task, or peer-score effect.
+
+**LC-GEN-05 [ZW] — Single frontier publisher.** The header engine MUST be the sole publisher of accepted header-frontier transitions. A driver or state-response task MUST NOT independently publish the raw result of a range commit.
+
+### 2.4 Transactions and restart
+
+**LC-TXN-01 [LS] — Atomic frontier mutation.** A frontier-affecting mutation MUST atomically update DAG rows, reverse indexes, height/child indexes, selected projections, eligibility, auxiliary provenance, generations, and frontier metadata before any observer is notified.
```
p0mvn, line 184: In the prior implementations, I found that our state update logic is every scatterred.

Given how important atomicity is here, we should have a clean, unified store API

docs/specs/fork-aware-header-chain-engine.md
```diff
@@ -0,0 +1,756 @@
...
+
+### 1.1 Definitions
+
+A **headers-only client** downloads and validates the observable block-header chain. It does not validate transaction bodies or the state transition they express.
+
+An **integrated engine** is that same reusable header-chain engine embedded in a full node and augmented by authenticated body-validation feedback. It is not a second overlay with different fork choice.
+
+An **anchor** is a trusted `(height, hash)`: configured genesis, a mandatory settled-upgrade activation pin, a configured local sync checkpoint, or the finalized frontier. The **work anchor** is the current finalized frontier from which retained suffix work is summed. Settled-upgrade pins and local sync checkpoints are separate configuration namespaces even when they contain the same tuple.
+
+**Integrated mode** embeds the engine in a full validator and advances finality only from fully verified state. **Headers-only mode** has no body-validity oracle and advances a disclosed local finality pin from sufficiently deep `header_best` history under LC-FINAL-03.
```
p0mvn, line 104: Do we need to implement header-only mode from the start?

Could we just design an API such that it can be plugged in later?

Mostly thinking about the pre-maturely implementing something we don't need and adding code complexity that is difficult to navigate. An example of that is backwards sync.

It sounds like the header-only mode has more specific finalization requirements and, likely, other edge cases.

Starting smaller and building up incrementally might help.

docs/specs/fork-aware-header-chain-engine.md
```diff
@@ -0,0 +1,756 @@
...
+
+### 5.3 Server snapshot contract
+
+**LC-V8-05 [ZW] — Snapshot-bound path serving.** On accepting a request, the server MUST acquire a retained-path snapshot for the exact target hash. It MUST select the first locator entry, in requester order, that lies on that target’s ancestor path, and serve only the contiguous path after that ancestor toward that target.
+
+**LC-V8-06 [ZW] — No target substitution.** The accepted target MUST remain snapshot-bound even if the server’s selected tip changes concurrently. The server MUST either complete that exact path through one or more responses/continuations or return one explicit `HeadersOutcomeV8`; it MUST NOT silently substitute its new selected path.
```
p0mvn, line 482: How long are the snapshots retained, and what is their eviction policy?

docs/specs/fork-aware-header-chain-engine.md
```diff
@@ -0,0 +1,756 @@
...
+
+**LC-RETAIN-02 [LS] — Deterministic eviction order.** On pressure, the engine MUST remove permanently ineligible subtrees first. It MUST then evict unprotected candidate tips in ascending order of cumulative work, breaking equal-work eviction ties by the smallest raw tip hash. Shared ancestors MUST be removed only when no retained path or validation-context window references them.
```
p0mvn, line 285: What correctness guarantee survives fork eviction?

A currently weak fork may be evicted and later become strongest. If no peer re-advertises or retains its old ancestry, can the engine still claim the greatest-work “visible” chain?

Also, how do we define "pressure"?

p0mvn (approved):

docs/specs/fork-aware-header-chain-engine.md
evan-forbes: this is a fair concern

> Do we need to implement header-only mode from the start?

inherently not header only already exists already no? this is just block propagation or are you referring to something else?

> It sounds like the header-only mode has more specific finalization requirements and, likely, other edge cases.

more specific in that they are different yes, but most of it is identical and simply pending some verification ie downloading the block data. meaning we need a DAG of blocks somewhere in any case. eventually when we write our own block prop, we won't need the old logic any more. it just appends onto the header DAG!

the reasoning for using headers before blocks as we're getting credible commitments to data before we download, which is required for high throughput block prop / blocksync. With out that, we must rely on rate limits and other (I would argue equally complex but more delicate) mechanisms to actually be safe.

imo we're actually net saving a lot of complexity in the long run by implementing headersync now correctly. we're effectively completely removing a huge swath of attacks by requiring data we download to be verifiable against a credible commitment

docs/specs/fork-aware-header-chain-engine.md
evan-forbes: agree, the implementation plan specifies this, we should go through that to ensure we're happy there before starting code generation

docs/specs/fork-aware-header-chain-engine.md
evan-forbes: this is an implementation detail left to the implementation plan atm

currently in that plan we're reusing the exact same constants used for block prop

docs/specs/fork-aware-header-chain-engine.md
evan-forbes: > A currently weak fork may be evicted and later become strongest. If no peer re-advertises or retains its old ancestry, can the engine still claim the greatest-work “visible” chain?

this assumes that all nodes in the network prune the weak fork afaiu. however inherently if there is a weak fork that eventaully overtakes the longer forks that fork has been maintained and can be re-downloaded. hidden data is handled in the normal way as well

pressure just means when we're hitting our configured limits on DAG storage

PR #300, opened 2026-07-20
fix(state): prevent same-hash rejected blocks from stalling sync

evan-forbes:
## Motivation

A rejected NU5+ block body can share its header hash with a later canonical block because the header hash does not commit to authorizing data. If contextual verification rejects the poisoned body first, the state write task records its hash in `parent_error_map`.

Even after the canonical same-hash block commits successfully, the stale map entry remains and causes the canonical block's children to inherit the earlier error. A remote peer that wins the propagation race can therefore stall chain synchronization.

This ports the GHSA-specific fix from ZcashFoundation/zebra#10995.

## Solution

Remove a block hash from `parent_error_map` immediately after a block with that hash commits successfully. Contextual verification remains unchanged: the poisoned body is still rejected, while the successful canonical commit supersedes the stale error associated with its hash.

Add an end-to-end Regtest regression test that:

1. Builds a canonical four-block NU5 chain.
2. Mutates the height-3 coinbase scriptSig to create a contextually invalid body with the canonical header hash.
3. Submits the poisoned body and verifies it is rejected.
4. Resubmits the canonical block and its child.
5. Verifies both are accepted and the chain reaches height 4.

## Testing

- `CXX=clang++ CXXFLAGS='-include cstdint' cargo test -p zakura --test acceptance rejected_block_does_not_reject_same_hash_block_children`
- `CXX=clang++ CXXFLAGS='-include cstdint' cargo clippy -p zakura-state --lib -- -D warnings`
- `CXX=clang++ CXXFLAGS='-include cstdint' cargo clippy -p zakura --test acceptance -- -D warnings`
- `git diff --check`

The explicit C++ flags work around vendored RocksDB 8.10 missing direct `<cstdint>` includes with this host's current C++ toolchain.

- [GHSA-8gxx-hc65-vv82](https://github.com/ZcashFoundation/zebra/security/advisories/GHSA-8gxx-hc65-vv82)
- [Upstream combined security fix PR](https://github.com/ZcashFoundation/zebra/pull/10995)
- [ZIP-244](https://zips.z.cash/zip-0244)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   checks instead of hashing them again for every signature
   ([#281](https://github.com/zakura-core/zakura/pull/281)).
 
+### Security
+
+- Prevent a peer from stalling chain synchronization by delivering a rejected
+  block body that shares its header hash with a later valid block
+  ([GHSA-8gxx-hc65-vv82](https://github.com/ZcashFoundation/zebra/security/advisories/GHSA-8gxx-hc65-vv82)).
+
 ## [1.0.2-rc1] - 2026-07-19
 
 ### Added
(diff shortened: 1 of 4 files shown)


evan-forbes:
rebasing in a second to keep original author's commit

otherwise r4r

ValarDragon (approved):

PR #306, opened 2026-07-20
ci: add golden zcashd compatibility PR nodes

czarcas7ic:
## Motivation

Managed zcashd compatibility tests still depended on manually downloading and unpacking roughly 270 GiB of state. Besides being slow, that made it easy to pair zcashd with Zakura state outside the pruned node's 10,000-block retention window.

## Solution

Extend the existing PR-node system with a `zcashd-compat` profile that combines the Golden base image with timestamp-matched Zakura and zcashd volume snapshots. The twice-weekly bake clones and advances the prior zcashd fixture, verifies height, hash, database format, retention, and clean shutdown, writes provenance manifests, and keeps the newest two complete snapshot families. The guarded first-seed path copies only stopped chain state; its current fixture remains `candidate` because the source archive checksum was not completed.

Compatibility runs validate the paired manifests, perform a cold managed-wrapper install, require both nodes to advance with exactly one supervised zcashd child and connection, then prove clean shutdown and warm-cache reuse. Ordinary Zakura tests retain their existing image and state path; the large zcashd state stays on a separate 350 GiB volume. Repository guidance, operator documentation, and an agent skill route future real-node tests through these Golden assets instead of ad hoc downloads.

The fixture intentionally excludes `peers.dat` and `banlist.dat`. zcashd recreates them on first start, so those two exact self-healing startup messages are reported separately without hiding unexpected errors.

## Testing

- `actionlint` on the PR-node workflows
- `bash -n` and `shellcheck` on the PR-node shell scripts
- Python compilation and `python3 -m unittest deploy/deployer/test_deploy.py` (9 tests)
- Monitor verdict fixtures, live-log classification, repository skill validation, and Markdown lint
- [Golden bake run 29782655645](https://github.com/zakura-core/zakura/actions/runs/29782655645) passed: seeded the candidate fixture at height 3,417,532, refreshed Zakura and zcashd to height 3,419,404 with zero lag, created the paired state snapshots and base image, and completed resource cleanup
- [Compatibility smoke run 29787849898](https://github.com/zakura-core/zakura/actions/runs/29787849898) passed against exact target SHA `361aff729ca1fa05615d814f2fc5a042108c9503`: both nodes advanced 3,419,446 → 3,419,453, ended at zero drift with one child and one connection, reused the unchanged managed binary after all four clean-shutdown markers, reported zero unexpected errors and zero pruned-sidecar diagnostics, and completed immediate teardown

## Follow up

Replace the candidate zcashd fixture with a checksum-verified seed; scheduled refreshes intentionally preserve the original verification status.

## AI disclosure

Codex was used to implement and review the workflows, deployment harness, tests, documentation, and agent guidance.

### DIFF
--- a/.agents/skills/test-zakura-pr-node/SKILL.md
+++ b/.agents/skills/test-zakura-pr-node/SKILL.md
@@ -1,0 +1,81 @@
+---
+name: test-zakura-pr-node
+description: Dispatch and validate Zakura real node tests with the repository's Golden DigitalOcean image and state fixtures. Use for fresh server or Droplet tests, real node tests of a Zakura PR or ref, managed zcashd wrapper or zcashd-compat tests, reproducing node or sync issues, and requests to use a Golden image or Golden assets instead of downloading chain state.
+---
+
+# Test a Zakura PR Node
+
+Use `.github/workflows/zakura-pr-node.yml` for DigitalOcean tests. Let the
+workflow create the Droplet and clone the maintained Golden assets. Never
+provision an ad hoc test Droplet or manually download or unpack chain state.
+Route provisioning and asset refreshes through `gh workflow run`; do not call
+`doctl` or the DigitalOcean API directly.
+
+Read `docs/pr-node-do-setup.md` before changing workflow inputs or refreshing
+assets.
+
+## Choose a profile
+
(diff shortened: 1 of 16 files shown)


czarcas7ic:
DNM, this conflicts with me moving over everything out of DO storage

PR #307, opened 2026-07-20
refactor(state): centralize commitment root index access

p0mvn:
## Motivation

Peer-supplied and body-derived commitment roots share one database index, but its reads, writes, and deletion policies were distributed across header commits, body commits, VCT repair, rollback, and startup repair. That made the later verified-only persistence boundary vulnerable to accidental bypasses.

## Solution

- add a state-owned commitment-root database façade for disk-row conversion, point and contiguous reads, and policy-specific mutations
- route body-derived replacement, legacy header insertion, reorg cleanup, rollback truncation, and startup repair through the façade
- preserve body-derived precedence and existing serving behavior while making contiguous reads stop at the first gap
- add regression coverage for gap handling, reorg and rollback suffix deletion, and direct column-family access

## Testing

- IDE diagnostics report no errors in the changed state files

- Phase 0.8 of the proposed header-sync VCT root-authentication design in #230

### Follow-up Work

- make history-tree snapshot decoding fallible
- introduce the sealed verified-root type together with the database migration and durable authentication frontier

### DIFF
--- a/zakura-state/src/service/finalized_state/zakura_db/rollback.rs
+++ b/zakura-state/src/service/finalized_state/zakura_db/rollback.rs
@@ -1100,11 +1100,11 @@ struct RetainedShieldedRoots {
 fn retained_shielded_roots(db: &ZakuraDb, target_height: Height) -> RetainedShieldedRoots {
     let mut retained = RetainedShieldedRoots::default();
 
-    for (_height, roots) in db.commitment_roots_for_migration(..=target_height) {
+    db.visit_commitment_roots_for_migration(..=target_height, |_height, roots| {
         retained.sapling.insert(roots.sapling_root);
         retained.orchard.insert(roots.orchard_root);
         retained.ironwood.insert(roots.ironwood_root);
-    }
+    });
 
     for (_height, tree) in db.sapling_tree_by_height_range(..=target_height) {
         retained.sapling.insert(tree.root());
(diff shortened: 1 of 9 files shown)


zakura-state/src/service/finalized_state/zakura_db/rollback.rs
```diff
@@ -1100,17 +1100,10 @@ struct RetainedShieldedRoots {
...
-        .db
-        .zs_forward_range_iter::<_, Height, CommitmentRootsByHeight, _>(
-            &commitment_roots_by_height,
-            ..=target_height,
-        )
-    {
-        retained.sapling.insert(roots.sapling);
-        retained.orchard.insert(roots.orchard);
-        retained.ironwood.insert(roots.ironwood);
+    for (_height, roots) in db.commitment_roots_for_migration(..=target_height) {
```
evan-forbes: [ai assisted]

does this mean it would return the entire set of roots from genesis until target height?

if so, is that needed for the rollback? if so, do we need to stream this?

evan-forbes (approved):
no blocking comments and appreciate the direction

likely worth waiting for v12 before merging

did notice that maybe we could be requesting quite a lot of data when we might not need to, but will defer to author on how big of an issue this is irl.

zakura-state/src/service/finalized_state/zakura_db/rollback.rs
p0mvn: Addressed by streaming

PR #308, opened 2026-07-20
ci(sync): attribute halted audits to the failed run

Zk-nd3r:
## Motivation

Scheduled audits replay durable halts without identifying the commit or run that failed. That can misattribute an old failure to a newer release.

## Solution

Include the failure-bound SHA, run ID, and failure time. Use the legacy running pair only when both failed-run fields are absent. Partial failed state never mixes with current state. State without provenance keeps the old line.

## Testing

- Four focused provenance tests pass.
- Both changed files parse with ast.parse.
- git diff --check passes.
- Full file result: 23 tests pass. Two unchanged symlink tests are blocked by native Windows WinError 1314.

### DIFF
--- a/deploy/continuous-sync/deploy.py
+++ b/deploy/continuous-sync/deploy.py
@@ -401,7 +401,20 @@ def audit_problem(data: dict[str, Any], max_completion_age: int) -> str | None:
     state = data.get("controller_state") or {}
     sample = data.get("sample") or {}
     if state.get("failed"):
-        return f"controller halted: {state.get('failure')}"
+        failed_sha = state.get("last_failed_sha")
+        failed_run = state.get("last_failed_run")
+        if failed_sha in (None, "") and failed_run in (None, ""):
+            failed_sha = state.get("running_sha")
+            failed_run = state.get("current_run")
+        provenance = (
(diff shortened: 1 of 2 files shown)


ValarDragon:
Hey, were you trying to setup monitoring for your own fleet with the repo?

A bit confused on what problem is trying to be solved here 😅

Zk-nd3r:
No. This is the repo's existing continuous-sync audit path, not fleet-specific monitoring. A halt persists until explicit resume, so later scheduled audits replay the old failure. `audit_problem()` currently drops `last_failed_sha`, `last_failed_run`, and `failed_at` even though controller state already stores them. This patch includes that stored provenance and preserves the old line for legacy state.

ValarDragon:
Really appreciate you making a PR for this, and excited about you contributing!

For this one, I'd like to close since I don't quite grok how it helps with a flow we are running, or you are right now. Seems like a devops concern, but the general take is its much easier to just understand and fix those when they come up as a problem.

Maybe a weird take, but right now the list of things to merge in the project is huge 😅, so want to keep the PR list trimmed to stuff that will converge quick

Thank you, and please do contribute again!!

PR #309, opened 2026-07-20
refactor(network): align header range records internally

p0mvn:
## Motivation

Header-sync responses still cross internal async boundaries as separate header, body-size, and root vectors after v7 decoding. Keeping those vectors independent makes trimming and buffering vulnerable to per-height misalignment even though the wire payload has already been validated.

## Solution

- introduce `HeaderRangeEntry` to bind one header to its body-size hint and optional roots
- convert v7 parallel vectors into entries immediately after decoding
- store and trim one entry vector in `HeaderRangePayload`
- carry the existing `HeaderSyncWireRequestIdentity` through `WireHeaders`
- keep `HeaderSyncMessage`, the v7 byte layout, and stream version unchanged

## Testing

- `cargo fmt --all`
- `cargo test -p zakura-network --lib zakura::header_sync` (165 passed)
- `cargo clippy -p zakura-network --lib --tests -- -D warnings`
- IDE diagnostics: clean

- Follow-up to #298

### Follow-up Work

A stacked v8 PR will make correlated wire messages self-contained and serialize aligned entries atomically.

### DIFF
--- a/zakura-network/src/zakura/header_sync/range.rs
+++ b/zakura-network/src/zakura/header_sync/range.rs
@@ -6,6 +6,58 @@ use super::{
     *,
 };
 
+/// One header and all data associated with its height.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct HeaderRangeEntry {
+    /// Header at this entry's implicit range height.
+    pub header: Arc<block::Header>,
+    /// Advisory serialized body-size hint.
+    pub body_size: u32,
+    /// Commitment roots when the response includes roots.
+    pub tree_aux_root: Option<BlockCommitmentRoots>,
+}
+
+impl HeaderRangeEntry {
+    /// Convert parallel vectors into aligned records at a system boundary.
+    pub fn from_parallel(
+        headers: Vec<Arc<block::Header>>,
+        body_sizes: Vec<u32>,
+        tree_aux_roots: Vec<BlockCommitmentRoots>,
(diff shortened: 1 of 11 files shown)


zakura-network/src/zakura/header_sync/range.rs
```diff
@@ -6,6 +6,58 @@ use super::{
...
+/// One header and all data associated with its height.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct HeaderRangeEntry {
+    /// Header at this entry's implicit range height.
+    pub header: Arc<block::Header>,
+    /// Advisory serialized body-size hint.
+    pub body_size: u32,
+    /// Commitment roots when the response includes roots.
+    pub tree_aux_root: Option<BlockCommitmentRoots>,
+}
```
evan-forbes, line 18: do we also want the height here?

evan-forbes (approved):
lfg

zakura-network/src/zakura/header_sync/range.rs
p0mvn: Added

p0mvn: Added in 2da1a8895. Each entry now carries an authoritative height; `HeaderRangePayload` derives its range from validated contiguous entries, and roots must match their entry height. The stacked v8 PR also encodes the height in each atomic response record.

p0mvn:
@cursor review

PR #313, opened 2026-07-20
refactor(network)!: encode aligned header entries in v8

p0mvn:
## Motivation

The v7 codec correlates requests through an API side channel and serializes headers/body sizes separately from commitment roots. Although the internal layer now binds those values into entries, the message model and byte layout can still represent them independently.

## Solution

- bump the header-sync peer protocol from v7 to v8
- embed request IDs directly in correlated `GetHeaders` and `Headers` messages
- encode each response entry atomically as header, body size, and optional roots
- remove redundant request-ID side channels from codec and test actions
- reject v7 header-sync negotiation while leaving other stream versions unchanged
- update the Unreleased changelog

This PR is intentionally stacked on #309.

## Testing

- `cargo fmt --all`
- `cargo test -p zakura-network --lib zakura::header_sync` (166 passed)
- `cargo test -p zakura-network --lib header_sync_stream_version_contract_is_exact`
- relevant transport/testkit v8 compatibility tests
- `cargo clippy -p zakura-network --lib --tests -- -D warnings`
- `git diff --check`
- IDE diagnostics: clean

- Base PR: #309

### Follow-up Work

Retarget this PR to `main` after #309 merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   preventing ban-list maintenance from slowing as the 20,000-IP bound fills
   ([#286](https://github.com/zakura-core/zakura/pull/286)).
 
+### Changed
+
+- Require header-sync peer protocol v8, with self-contained correlated messages and
+  atomically interleaved header response records. Header-sync v7 peers are incompatible.
+
 ## [1.0.2] - 2026-07-20
 
 ### Added
--- a/zakura-network/src/zakura/discovery/service.rs
+++ b/zakura-network/src/zakura/discovery/service.rs
@@ -1066,13 +1066,11 @@ mod tests {
             loop {
                 if let Some(HeaderSyncAction::SendMessage {
                     peer,
-                    request_id,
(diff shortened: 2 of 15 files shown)


evan-forbes (commented):
you'd likely have found it later with an ai audit, but fwiw

```
v8 byte-budget clamping omits the new 4-byte height field.

zakura-network/src/zakura/header_sync/config.rs:154 budgets header + body_size + roots, while zakura-network/src/zakura/header_sync/wire.rs:152 now also serializes a height per entry.

     At the 2 MiB payload cap on Mainnet:
      - Rooted count returned by the clamp: 1,273
      - Actual v8 maximum: 1,270
      - Resulting payload exceeds the cap by 4,585 bytes

The scheduler can therefore request an allegedly valid range whose response fails encoding, causing repeated timeout/retry behavior. Add HEADER_SYNC_HEIGHT_BYTES to the denominator and an exact-boundary regression test.
```

p0mvn:
I think we can come back to this as part of / post-DAG. Mainly did this to split up the network-breaking component into its own PR, separately from the internals

PR #314, opened 2026-07-20
fix(network): canonicalize banned peer IPs

ValarDragon:
## Motivation

Peers using an IPv4-mapped IPv6 address can bypass an existing IP ban. The
misbehavior path records the canonical IPv4 address, while ban checks can use
the original mapped IPv6 representation, so the equivalent addresses do not
match in the ban set.

## Solution

Canonicalize IP addresses at the `BannedIps` boundary for both insertion and
lookup. Reuse #238's `canonical_ip` implementation from the shared address
canonicalization module, so the ban list, peer set, and handshake paths apply
the same normalization rule.

Add a regression test covering both directions: inserting IPv4 and looking up
mapped IPv6, and inserting mapped IPv6 and looking up IPv4.

## Testing

- `npx --yes markdownlint-cli CHANGELOG.md --config .trunk/configs/.markdownlint.yaml`
- `cargo test -p zakura-network --lib banned_ips_match_ipv4_and_ipv4_mapped_ipv6`
- `cargo test -p zakura-network --lib address_book::tests`
- `cargo test -p zakura-network --lib mapped_and_canonical`
- `cargo test -p zakura-network --lib broadcast_all_queued_bans_mapped_ipv6_against_canonical_ban`
- `cargo test -p zakura-network --lib`: 1,017 passed, 3 failed, and 4 ignored.
  The three failures bind a test client to `127.0.0.2`, which is unavailable on
  this macOS host. The same
  `listener_bans_zcashd_compat_peer_before_reserved_slot` failure reproduces on
  unchanged `main` before reaching ban logic.

---

> [!NOTE]
> **Medium Risk**
> Touches peer misbehavior enforcement (security-sensitive); change is narrow and aligns ban storage with existing canonicalization elsewhere in the network stack.
> 
> **Overview**
> **Fixes a ban bypass** where peers could reconnect using IPv4-mapped IPv6 (`::ffff:x.x.x.x`) while the ban list held the plain IPv4 form (or the reverse), so `contains` missed the match.
> 
> **`BannedIps` now canonicalizes on both insert and lookup** via shared `canonical_ip` from the address canonicalization module. Duplicate local helpers in handshake and peer set are removed in favor of that single export.
> 
> A regression test covers ban/lookup symmetry for IPv4 vs mapped IPv6. CHANGELOG documents the fix ([#314]).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45c4543e4e352e98621379d65b974f6d97db8a34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/zakura-network/src/address_book/tests.rs
+++ b/zakura-network/src/address_book/tests.rs
@@ -6,7 +6,7 @@ use std::net::{IpAddr, Ipv4Addr};
 
 use crate::constants::MAX_BANNED_IPS;
 
-use super::BanList;
+use super::{BanList, BannedIps};
 
 mod prop;
 mod vectors;
@@ -27,4 +27,13 @@ fn ban_list_evicts_the_oldest_ip_at_capacity() {
     assert!(bans.ips.contains(&newest));
     assert_eq!(bans.ips.len(), MAX_BANNED_IPS);
     assert_eq!(bans.insertion_order.len(), MAX_BANNED_IPS);
+}
+
+#[test]
+fn banned_ips_match_ipv4_and_ipv4_mapped_ipv6() {
+    let ipv4 = IpAddr::V4(Ipv4Addr::LOCALHOST);
+    let ipv4_mapped = IpAddr::V6(Ipv4Addr::LOCALHOST.to_ipv6_mapped());
+
+    assert!(BannedIps::with_banned_ip(ipv4).contains(ipv4_mapped));
+    assert!(BannedIps::with_banned_ip(ipv4_mapped).contains(ipv4));
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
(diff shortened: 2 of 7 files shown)


evan-forbes (approved):
not actually redundant to #238, we can want to merge merge both, just need to update names after we merge one

zakura-network/src/address_book/tests.rs
```diff
@@ -28,3 +28,12 @@ fn ban_list_evicts_the_oldest_ip_at_capacity() {
...
 }
+
+#[test]
+fn banned_ips_match_ipv4_and_ipv4_mapped_ipv6() {
+    let ipv4 = IpAddr::V4(Ipv4Addr::LOCALHOST);
+    let ipv4_mapped = IpAddr::V6(Ipv4Addr::LOCALHOST.to_ipv6_mapped());
+
+    assert!(BannedIps::with_banned_ip(ipv4).contains(ipv4_mapped));
+    assert!(BannedIps::with_banned_ip(ipv4_mapped).contains(ipv4));
+}
```
evan-forbes, line 39: optional but we could add things closer to integration tests as well

PR #316, opened 2026-07-21
fix(state): make history tree snapshot decoding fallible

p0mvn:
## Motivation

History-tree snapshots were decoded through an infallible database conversion. Malformed current or legacy bytes panicked, snapshots written for another network hit an assertion, and empty or pre-Heartwood caches could reach assertions in tree reconstruction. The later durable root-authentication frontier needs startup to reject incoherent snapshots as typed local-state errors.

## Solution

- add a typed snapshot error for malformed current/legacy bytes, network mismatch, and `HistoryTreeError`
- preserve current-first decoding with legacy-width fallback
- add raw-byte-backed `try_history_tree()` and fallible full-tip reads
- return `HistoryTreeError` for empty and pre-Heartwood non-empty caches
- keep the existing current and legacy serialized formats unchanged

No database format version or migration is included.

## Testing

- `cargo test -p zakura-state history_tree --lib`
- `cargo clippy -p zakura-chain -p zakura-state --all-targets -- -D warnings`

The tests exercise current and legacy snapshots, legacy fallback after a non-EOF current decoder error, malformed bytes, wrong-network snapshots, empty and pre-Heartwood caches, and raw database reads for both valid formats and corrupt data.

- [Header-sync VCT root authentication design](https://github.com/zakura-core/zakura/blob/6a0e701064b226724e9ca67e2e5f61bc080dc22d/docs/design/header-sync-vct-root-authentication.md)

### Follow-up Work

Introduce the sealed verified-root type together with the atomic database cleanup migration and durable authentication-frontier initialization.

### DIFF
--- a/zakura-chain/src/history_tree.rs
+++ b/zakura-chain/src/history_tree.rs
@@ -36,6 +36,9 @@ pub enum HistoryTreeError {
 
     #[error("I/O error: {0}")]
     IOError(#[from] io::Error),
+
+    #[error("invalid cached history tree: {reason}")]
+    InvalidCachedTree { reason: &'static str },
 }
 
 impl PartialEq for HistoryTreeError {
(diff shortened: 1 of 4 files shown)


evan-forbes (approved):
utACK, but would wait for v12 cause why not

PR #322, opened 2026-07-21
fix(sync): recover near-tip legacy sync stall with a FindHeaders fallback

p0mvn:
## Motivation

An operator migrating a Zebra 6.1.0 archive node to Zakura 1.0.2 reported that legacy sync made zero progress from a valid restored state: 29 consecutive sync rounds ended in `exhausted prospective tip set` while the committed tip stayed at 3,419,261 and `/ready` degraded to 503, even though the canonical successor block existed before the syncer started.

**Root cause** — a structural dead zone in `obtain_tips`, inherited verbatim from upstream Zebra: every Mainnet `FindBlocks` response is stripped of its last hash (zcashd append-quirk guard), and a `CheckedTip` requires two remaining unknown hashes (`rchunks_exact(2)`). A peer N blocks ahead answers with N hashes, so N=1 strips to nothing and N=2 strips to one hash that cannot mint a tip — and the tip-creation `continue` also skips queueing the hash for download. Block gossip cannot fill the gap either: the missing blocks were announced while the node was still cloning/upgrading its database, and gossiped children cannot commit without their parent. The node is stuck until the network gets 3+ blocks ahead. Upstream documents the exact failure in its regtest exception comment but only fixed it for regtest; the Zebra "control run" in the report succeeded purely because it was restarted after the network was 3 blocks ahead (`extending tips ... in_flight=2`).

The Zakura migration workflow (stop Zebra → clone 275 GB → format upgrade → fresh peer cache) makes entering this dead zone near-deterministic, which is why it bites Zakura operators but is latent in Zebra.

## Solution

`obtain_tips` now counts responses that offered new blocks but were too short to survive the strip and the two-hash tip requirement. When the fanout ends with an empty download set and at least one such response, it falls back to a `FindHeaders` fanout with the same block locator and queues only the run of offered headers that **provably extends our chain** — the first unknown header's parent must be a block in our state, and every later header must link to its predecessor (`anchored_header_hashes`, same anchoring rule as the watchdog's `legacy_peers_blocks_ahead`). Anchored hashes enter the download set without minting a prospective tip and flow through the existing state re-check, download, and full verification pipeline.

Because headers name their parents, the zcashd trailing-hash quirk cannot poison this path, so even a single-block extension is safe — closing both the 1-behind and 2-behind cases while leaving the ≥3 bulk path byte-identical. Foreign-fork, gapped, out-of-order, and oversized header responses contribute nothing.

**Security posture** (audited against a fully adversarial peer model): the fallback only selects hashes to download; all blocks still pass full semantic + contextual verification, and no `CheckedTip` is ever minted from headers. Forged header chains (hash linkage is not PoW-checked) can queue junk hashes, but this is strictly *more* filtered than the existing `FindBlocks` path, which already accepts up to ~499 unanchored hashes from any peer; consequences (bounded notfound retries, round restart) are identical in kind. The trigger cannot fire while any peer returns a usable ≥3-hash response. Work per response is bounded (500-header cap, sequential state lookups, 6 s per-request timeout inside the 45 s round budget).

The trigger logs at `info` with `too_short_responses` and anchored counts (the reporter's scenario becomes diagnosable from default logs) and emits `sync.obtain.header_fallback.count` / `.hash.count` metrics.

## Testing

- `sync_near_tip_one_block_uses_header_fallback` — full mock-service reproduction of the report: state at block 1, peers answer `FindBlocks` with a single hash (discarded today, stalling forever); asserts the syncer issues `FindHeaders`, anchors the offered header at the tip via its parent link, downloads and commits the block. Fails on `main` by construction.
- `malformed_find_headers_responses_are_rejected` — negative cases for `anchored_header_hashes`: foreign-fork parent, unknown-then-known ordering, gapped (non-linking successor) run, and oversized response each contribute zero hashes.
- `cargo test -p zakura --lib components::sync` — all 60 tests pass; `cargo fmt` and `cargo clippy -p zakura --all-targets` clean; `markdownlint-cli2` clean on `CHANGELOG.md`.
- Field validation still pending: stop a tip-synced node, wait 1–2 Mainnet blocks (~75–150 s), restart — it should log the fallback line and commit within one sync round instead of exhausting for minutes.

- External bug report: "legacy sync repeatedly exhausts prospective tips without advancing from a valid Zebra state" (Zakura 1.0.2, Zebra 6.1.0 format-28 state).
- Upstream context: ZcashFoundation/zebra `zebrad/src/components/sync.rs` (identical strip + `rchunks_exact(2)` logic; regtest-only exception), upstream issue #3375 (slow syncing near tip).

### Follow-up Work

- Optional hardening: verify header PoW (difficulty threshold + Equihash, as native header sync does) on fallback headers before queueing, raising forgery cost from zero.
- `/ready` lag currently comes from the 75 s time extrapolator; the fallback's anchored-run count is network truth and could feed readiness instead.
- Offer the fallback upstream to Zebra: any zebrad restart spanning 1–2 block arrivals hits the same window, just more rarely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
+- Advance legacy sync from a state restarted 1-2 blocks behind the network
+  tip: when `FindBlocks` responses are too short to survive the zcashd
+  trailing-hash guard, the syncer now falls back to `FindHeaders` and
+  downloads the blocks whose header parent links anchor in its own chain,
+  instead of repeatedly exhausting the prospective tip set without progress.
 - Keep valid internal-miner work running across mempool-only block template
   updates ([#226](https://github.com/zakura-core/zakura/pull/226)).
 - Honor `disable_pow = true` during native header sync on configured Testnets,
--- a/zakurad/src/components/sync.rs
+++ b/zakurad/src/components/sync.rs
@@ -63,6 +63,49 @@ pub use status::SyncStatus;
 /// Controls the number of peers used for each ObtainTips and ExtendTips request.
 const FANOUT: usize = 3;
 
+/// Returns the hashes from a bounded header response that continuously extends
+/// a block already present in state.
+fn anchored_header_hashes(
+    headers: Vec<block::CountedHeader>,
+    known_hashes: &HashSet<block::Hash>,
+) -> Vec<block::Hash> {
+    if headers.len() > MAX_TIPS_RESPONSE_HASH_COUNT {
+        debug!(
+            headers.len = headers.len(),
+            max_hashes = MAX_TIPS_RESPONSE_HASH_COUNT,
+            "discarding oversized FindHeaders response"
+        );
+        return Vec::new();
+    }
+
+    let mut run = Vec::new();
+    let mut run_parent = None;
+    for counted in headers {
+        let hash = counted.header.hash();
+        if known_hashes.contains(&hash) {
+            if !run.is_empty() {
+                return Vec::new();
+            }
+            run_parent = Some(hash);
+            continue;
+        }
+
+        let parent = counted.header.previous_block_hash;
+        let anchored = match run_parent {
+            Some(run_parent) => parent == run_parent,
+            None => known_hashes.contains(&parent),
+        };
+        if !anchored {
+            return Vec::new();
+        }
+
+        run.push(hash);
+        run_parent = Some(hash);
+    }
+
+    run
+}
+
 /// Controls how many times we will retry each block download.
 ///
 /// Failing block downloads is important because it defends against peers who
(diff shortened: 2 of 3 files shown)


p0mvn:
Post-draft verification (local, CI-equivalent slice):

- `cargo test -p zakura --lib`: 303/304 passed. The single failure, `inbound::tests::real_peer_set::inbound_pruned_block_is_not_advertised_and_getdata_logs_error`, is a pre-existing shared-process flake ("global default trace dispatcher has already been set") unrelated to this diff — it passes in isolation (0.18s) and its file is untouched here; CI's nextest process-per-test isolation avoids it.
- `cargo clippy -p zakura --all-targets`: clean.
- `cargo fmt --check`: clean.
- `markdownlint-cli2` with `.trunk/configs/.markdownlint.yaml` on `CHANGELOG.md`: clean.

czarcas7ic:
Closing in favor of https://github.com/zakura-core/zakura/pull/452

PR #325, opened 2026-07-21
fix(zakurad): write generated configs atomically

ValarDragon:
## Motivation

The `generate` subcommand opened its output with `File::create` and wrote the config directly. A failed or interrupted write could therefore truncate an existing operator-specified config and leave partial contents behind.

## Solution

Write generated config files with `zakura_chain::common::atomic_write`, which stages the contents in the destination directory and atomically replaces the destination only after the write and sync succeed.

The existing `generate_args` acceptance test now pre-populates the destination and verifies that the generated config replaces those prior contents.

## Testing

None.

### Follow-up Work

None.

---

> [!NOTE]
> **Low Risk**
> Limited to config file generation CLI behavior; reuses an existing atomic-write helper already used elsewhere in the codebase.
> 
> **Overview**
> The `generate` subcommand no longer overwrites an existing output path with `File::create` and a direct write. It now persists through **`zakura_chain::common::atomic_write`**, staging the full config in the destination directory and swapping it in only after the write succeeds, so a failed or interrupted run is less likely to leave a truncated operator config.
> 
> The **`generate_args`** acceptance test seeds the target `zakura.toml` with placeholder text and checks that a successful run replaces it with a fresh generated skeleton (header present, old contents gone).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3b3212e7979b0b9cc7b7afbc198f7c1f5c0410c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/zakurad/src/commands/generate.rs
+++ b/zakurad/src/commands/generate.rs
@@ -3,6 +3,7 @@
 use crate::config::ZakuradConfig;
 use abscissa_core::{Command, Runnable};
 use clap::Parser;
+use zakura_chain::common::atomic_write;
 
 /// Generate a default `zakura.toml` configuration
 #[derive(Command, Debug, Default, Parser)]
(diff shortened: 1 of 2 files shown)


ValarDragon:
Actually I don't want to this, just in case there is any FS permissioning issue

PR #118, opened 2026-07-13
fix(rpc): prevent read-secondary synchronization races

ValarDragon:
## Motivation

Zakura is missing the RPC read-secondary synchronization fixes from Zebra
commits `676a4fddf` and `0c3f1eb85`. Under primary-node sync or a restarted
gRPC stream, the secondary can race finalized-state catch-up against
non-finalized commits, repeatedly receive blocks it already has, or fail to
bridge a finalized-state gap. These failures interrupt RPC and indexer
availability.

This path maintains the standalone RPC read secondary. It does not change
validator consensus checks, block acceptance, or fork choice.

## Solution

- Resume non-finalized subscriptions from the secondary's known chain tips and
  apply bounded backpressure instead of dropping streamed blocks.
- Add the indexer `GetBlock` RPC so a lagging secondary can fetch finalized
  blocks needed to bridge the gap to the streamed non-finalized chain.
- Make finalized-tip updater takeover an acknowledged barrier before the stream
  syncer performs catch-up and commits.
- Re-check streamed block heights after catch-up and gap bridging, safely prune
  and publish newly finalized state, and consume already-finalized messages
  without entering a resubscribe loop.
- Bound subscriptions and stream reads with timeouts and HTTP/2 keepalive.
- Drop side chains that become empty when their shared root is finalized.
- Compare generated protobuf artifacts as bytes and refresh the descriptor with
  source information.

The port includes Zebra follow-up `6493e7205`, which prevents already-finalized
stream messages from causing a retry loop. The handoff and post-catch-up checks
are strengthened beyond the upstream patches to close their remaining
in-flight catch-up windows.

### API changes

- `ReadRequest::NonFinalizedBlocksListener` is now a struct variant with
  `known_chain_tips`.
- `NonFinalizedBlocksListener::spawn` now takes a state receiver and known chain
  tips, without a `Network` argument.
- The generated indexer API adds `BlockRequest`,
  `NonFinalizedStateChangeRequest`, and `GetBlock`.
- The generated `NonFinalizedStateChange` client/server request changes from
  `Empty` to `NonFinalizedStateChangeRequest`. Empty protobuf requests remain
  wire-compatible.
- The generated `Indexer` server trait gains the required `get_block` method.

There are no new or changed `pub(crate)` or feature-gated APIs.

### Tests

- `cargo fmt --all -- --check`
- `cargo clippy -p zakura-rpc -p zakura-state --all-targets --locked -- -D warnings`
[...]

PR #326, opened 2026-07-21
fix(rpc): validate read-secondary synchronization

ValarDragon:
## Motivation

PR #118 added resumable RPC read-secondary synchronization, but several
validation and publication invariants remained:

- Concurrent finalized-tip publishers could publish a newer snapshot followed
  by a stale lower snapshot, briefly regressing the reported tip.
- Finalized-gap filling trusted each `GetBlock` response and assumed every
  successful commit advanced the local best tip, so an invalid or incompatible
  trusted-primary response could cause repeated retries without progress.
- Both the streamed and finalized-gap paths constructed a
  `SemanticallyVerifiedBlock` directly from cleartext gRPC input, then called
  low-level non-finalized commit methods. This bypassed semantic verification
  and the normal write path's initial contextual checks. An invalid compact
  difficulty could therefore reach the chain-work `expect` and panic the sync
  task, while a representable header with insufficient PoW could be served from
  the read-secondary state.

The primary is configured as trusted and normally runs on loopback, so this
requires a compromised or misconfigured primary, or an on-path actor when the
gRPC endpoint is exposed across an untrusted network. Invalid input must still
fail closed rather than crash or enter served state.

## Solution

- Serialize finalized-tip publication and track the highest published finalized
  height under the same lock, rejecting stale lower snapshots without changing
  non-finalized reorg behavior.
- Require every finalized-gap block to match the requested height, extend the
  expected parent, and match its advertised hash.
- Require streamed blocks to match their advertised hash as well.
- Run both streamed and finalized-gap blocks through the existing consensus
  semantic block and transaction verifier before preparing them for commit.
- Route every live non-finalized commit through a shared state method that runs
  the normal initial and chain-specific contextual checks before mutation.
- After each bridge commit, require the local tip to equal the requested block
  before continuing the loop.
- Add focused regression tests for stale finalized tips, malformed gap
  responses, no-progress commits, invalid compact difficulty, and insufficient
  PoW. The invalid compact test also calls the checked contextual commit path
  directly to prove it returns an error before cumulative work is updated.

## API Surface

- Re-export `SemanticBlockVerifier` from `zakura-consensus` so the RPC syncer
  can reuse the consensus implementation.
- Add `NonFinalizedState::validate_and_commit_block` as the checked commit entry
  point shared by the normal state writer and RPC syncer.

## Testing

Passed locally:

- `git diff --check`
- `cargo test -p zakura-rpc --locked
  trusted_sync_rejects_invalid_difficulty_and_pow_without_panicking`
- `cargo test -p zakura-rpc --locked`
- `cargo test -p zakura-state --locked --lib`
- `cargo clippy -p zakura-rpc -p zakura-state -p zakura-consensus
  --all-targets --locked -- -D warnings`
- `cargo test -p zakura --test acceptance
  --locked trusted_chain_sync_handles_forks_correctly -- --exact --nocapture`

- Follow-up to #118.

### Follow-up Work

None currently identified.

### DIFF
--- a/crates/zakura-consensus/src/lib.rs
+++ b/crates/zakura-consensus/src/lib.rs
@@ -52,7 +52,10 @@ pub use block::check::difficulty_is_valid;
 #[cfg(any(test, feature = "proptest-impl"))]
 pub use checkpoint::CheckpointVerifier;
 
-pub use block::{subsidy::funding_stream_address, Request, VerifyBlockError, MAX_BLOCK_SIGOPS};
+pub use block::{
+    subsidy::funding_stream_address, Request, SemanticBlockVerifier, VerifyBlockError,
+    MAX_BLOCK_SIGOPS,
+};
 pub use checkpoint::{VerifyCheckpointError, MAX_CHECKPOINT_BYTE_COUNT, MAX_CHECKPOINT_HEIGHT_GAP};
 pub use config::Config;
 pub use error::BlockError;
--- a/crates/zakura-rpc/src/sync.rs
+++ b/crates/zakura-rpc/src/sync.rs
@@ -2,22 +2,29 @@
 //! [`ChainTipSender`] via RPCs.
 
 use std::{
+    future,
     net::SocketAddr,
     sync::{Arc, Mutex},
+    task::{Context, Poll},
     time::Duration,
 };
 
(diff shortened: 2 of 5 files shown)


ValarDragon:
Closing this to save attention for now

PR #329, opened 2026-07-21
fix!(rpc): use consistent best-chain snapshots

ValarDragon:
## Motivation

Verbose `getblockheader` and `gettxout` previously combined several sequential
read-state responses. Each request sampled the latest best chain independently,
so a tip advance or reorganization could produce one RPC response containing
data from different chain views.

This PR is stacked on #328, which provides the narrow resolved-hash fix.

## Solution

- Add a composite block-header read that returns the header, resolved identity,
  next hash, Sapling tree, and depth using one captured best-chain handle.
- Add a composite unspent-output read that returns the creating transaction and
  the tip hash used for its confirmations after checking the output against the
  same best-chain handle.
- Migrate verbose `getblockheader`, its `getblock` callers, and `gettxout` to
  those composite reads.
- Keep the existing state requests intact for downstream compatibility.

### API surface

This adds the public `ReadRequest::BlockHeaderData` and
`ReadRequest::BestChainUnspentOutput` variants, the corresponding
`ReadResponse` variants, and the public `BestChainUnspentOutput` response type.
No existing public item is removed or has its signature changed. However,
adding variants to these exhaustive public enums is source-breaking for
downstream exhaustive matches, so the PR title uses the conventional-commit
breaking-change marker.

## Testing

- `git diff --check`
- `cargo test -p zakura-rpc rpc_gettxout` (2 passed)
- `cargo test -p zakura-rpc rpc_getblock` (9 passed)
- `cargo test -p zakura-state service::read::tests::vectors` (9 passed)
- `cargo clippy -p zakura-state -p zakura-rpc --all-targets -- -D warnings`
- Mock RPC coverage verifies verbose block-header context comes from one state
  response and `gettxout` receives its tip and transaction context together.

## Changelog

- Added `changelog.d/329.md` with root and `zakura-state` entries.

- https://github.com/ZcashFoundation/zebra/issues/10550
- Stacked on #328

### Follow-up Work

The legacy granular state request variants remain available. Removing any that
become unused should be considered separately because it would be a breaking
public API change.

---

> [!NOTE]
> **Medium Risk**
> Touches RPC correctness on the best chain during reorgs and adds public enum variants (source-breaking for exhaustive matches); scope is limited to read paths and well-covered by tests.
> 
> **Overview**
> Verbose **`getblockheader`** and **`gettxout`** no longer stitch together several independent read-state calls. Each RPC now uses one composite request so header, Sapling tree, depth, transaction, and tip hash all come from the **same** best-chain snapshot during tip moves or reorgs.
> 
> **`zakura-state`** adds **`ReadRequest::BlockHeaderData`** and **`ReadRequest::BestChainUnspentOutput`**, with matching **`ReadResponse`** variants and the public **`BestChainUnspentOutput`** type. Handlers capture **`latest_best_chain()`** once and run the related lookups against that handle. Legacy granular requests stay for compatibility.
> 
> RPC tests now expect a single state round-trip for verbose block-header context, and a new test asserts **`gettxout`** gets tip and transaction together via **`BestChainUnspentOutput`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f9479cbb2dcfb1ffab6fa02be8ee8c4e59e2aee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/changelog.d/329.md
+++ b/changelog.d/329.md
@@ -1,0 +1,15 @@
+## root
+
+### Fixed
+
+- Keep verbose block-header and `gettxout` responses on one best-chain view
+  during tip changes and reorganizations
+  ([#329](https://github.com/zakura-core/zakura/pull/329)).
+
+## zakura-state
+
+### Added
+
+- Add atomic best-chain read requests for verbose block-header data and unspent
+  transaction output context
+  ([#329](https://github.com/zakura-core/zakura/pull/329)).
--- a/zakura-rpc/src/methods.rs
+++ b/zakura-rpc/src/methods.rs
@@ -1503,15 +1503,15 @@ where
                 // Reference for the legacy error code:
                 // <https://github.com/zcash/zcash/blob/99ad6fdc3a549ab510422820eea5e5ce9f60a5fd/src/rpc/blockchain.cpp#L629>
                 .map_error(server::error::LegacyCode::InvalidParameter)?;
-        let zakura_state::ReadResponse::BlockHeader {
-            header,
-            hash,
-            height,
-            next_block_hash,
-        } = self
+        let request = if verbose {
+            zakura_state::ReadRequest::BlockHeaderData(hash_or_height)
+        } else {
+            zakura_state::ReadRequest::BlockHeader(hash_or_height)
+        };
+        let state_response = self
             .read_state
             .clone()
-            .oneshot(zakura_state::ReadRequest::BlockHeader(hash_or_height))
+            .oneshot(request)
             .await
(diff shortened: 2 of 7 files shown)


ValarDragon:
Closing for now in favor of #360. The composite-read prototype is useful context, but it does not yet provide an immutable finalized-database snapshot, and the current mock tests only prove one service request rather than atomicity. We will revisit this with deterministic state-layer interleaving tests and a deliberate decision about the public state API.

PR #333, opened 2026-07-21
fix(dummy CPU miner): invalidate stale templates on generation errors

ValarDragon:
## Motivation

`generate_block_templates` retained its last successful block template when a
later `get_block_template` call failed. Because the watch channel remained at
`Some(Block)`, `run_mining_solver` continued mining and submitting obsolete
work while the node was unsynced or unable to construct valid templates.

An isolated or eclipsed miner could therefore spend sustained Equihash work on
a stale template after producing one valid template.

## Solution

Invalidate any active mining template by changing the watch value to `None`
before waiting to retry a failed template request. The existing solver
cancellation signal now immediately stops in-flight work when a template
becomes unavailable. Repeated failures while the channel is already `None` do
not send redundant notifications.

Factor the solver's cancellation check into a private helper and add focused
tests showing that:

- an unsynced `get_block_template` failure invalidates a previously active
  template; and
- changing the template from `Some(Block)` to `None` cancels current solver
  work, while an identical header does not.

## Testing

- `cargo test -p zakura --lib --features internal-miner components::miner::tests`
- `cargo clippy -p zakura --lib --features internal-miner -- -D warnings`
- `git diff --check`

The producer regression uses a mocked Mainnet RPC whose sync status changes the
template request into an immediate not-synced error, then verifies that the
watch channel no longer retains its previous block.

## Changelog

- Added `changelog-unreleased/333.md` with an operator-facing security entry.

---

> [!NOTE]
> **Medium Risk**
> Touches internal-miner template lifecycle and solver cancellation; behavior change is intentional but could affect mining during sync/transient RPC errors.
> 
> **Overview**
> Fixes internal mining continuing on **obsolete block templates** after `get_block_template` starts failing (e.g. while unsynced).
> 
> When template generation errors, **`generate_block_templates` now clears the watch channel to `None`** (only if a template was active), instead of leaving the last good `Some(Block)`. That update **cancels in-flight Equihash work** via the existing solver cancellation path. Solver cancellation logic is **extracted into `cancel_if_mining_template_changed`**, which also treats **`None` as invalidating** current work (not only header changes).
> 
> Adds **unit/integration tests** for `None` cancellation and for unsynced RPC failures clearing a previously published template, plus a **security changelog** entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9dc9426c381d75d985970bb8a4fea13f835b85b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/changelog-unreleased/333.md
+++ b/changelog-unreleased/333.md
@@ -1,0 +1,6 @@
+## Security
+
+- Stop internal mining when block-template generation fails, preventing stale
+  work and rejected submissions while the node is unsynced or unable to create
+  valid templates
+  ([#333](https://github.com/zakura-core/zakura/pull/333)).
--- a/zakurad/src/components/miner.rs
+++ b/zakurad/src/components/miner.rs
@@ -63,6 +63,35 @@ fn should_replace_mining_template(
     current_header != Some(new_header) && (current_header.is_none() || submit_old != Some(true))
 }
 
+/// Cancels mining when the current template changes or becomes unavailable.
+fn cancel_if_mining_template_changed(
+    template_receiver: &mut WatchReceiver<Option<Arc<Block>>>,
+    old_header: block::Header,
+) -> Result<(), SolverCancelled> {
+    match template_receiver.has_changed() {
+        // Guard against `get_block_template()` providing an identical header.
+        // This could happen if something irrelevant to the block data changes,
+        // the time was within 1 second, or there is a spurious channel change.
+        Ok(has_changed) => {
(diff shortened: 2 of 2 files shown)


changelog-unreleased/333.md
```diff
@@ -0,0 +1,6 @@
+## Security
```
ValarDragon, line 1: Oh this should be fixed

PR #336, opened 2026-07-21
fix(network): converge native peer discovery

evan-forbes:
## Motivation

Native Zakura nodes can settle at unexpectedly different peer and service counts even when a small fleet runs the same version and service set.

The audit reproduced several independent causes:

- valid third-party signed records were stored and regossiped but excluded from dial candidates, leaving nodes in a bootstrap-centered topology;
- an unspecified native listener produced an empty self-record even when `external_addr` supplied a reachable IP;
- bootstrap configuration could create self-dials and duplicate maintained tasks for one identity;
- discovery ran one exchange and then stopped, while live service summaries expire after 30 seconds;
- connection lifetime was inferred from negotiated capabilities rather than an exact service-owned `(peer, connection)` session;
- cleanup from an older discovery stream generation could remove or close its replacement; and
- signed connected-peer service membership survived the signed record's expiry.

Enabling third-party record dialing also exposed target-filtering, shared-IP backoff, and worker-panic cleanup gaps that could otherwise cause unsafe or permanently stuck dialing state.

## Solution

- Advertise reachable native addresses from the specific listen address, configured external IP, or a bootstrap entry for the local identity.
- Group bootstrap entries by remote identity, merge their addresses, and exclude malformed, duplicate, and local identities from tasks and discovery headroom.
- Admit valid signed gossip records as candidates only after public-target filtering; canonicalize mapped IPv4, enforce per-IP in-flight limits, and apply IP-scoped exponential backoff.
- Recover panicking dial workers as failed attempts while releasing their node and IP reservations.
- Track exact discovery stream generations and reject duplicate/stale admission, messages, cleanup, and connection teardown.
- Refresh discovery at least every 15 seconds while another exact legacy-gossip, header-sync, or block-sync session owns the connection.
- Use bounded, clock-skew-tolerant live-summary lifetimes without extending a peer's deliberately shorter future expiry.
- Expire connected-peer service membership with its signed record, while allowing independently fresh live summaries to remain.
- Serialize self-record refresh publication so concurrent refreshers cannot publish an older sequence after a newer one.

### Pre-fix reproduction

The following were run against `origin/main` or the pre-audit branch in temporary worktrees and passed only by asserting the buggy state:

- third-party public signed records were not dialable;
- an unspecified listener plus external address authored no usable address;
- local and duplicate bootstrap entries created three maintained tasks instead of one;
- negotiated block capability counted as connection ownership without actual service admission;
- old cleanup removed a replacement stream on the same connection;
- discovery stopped after one exchange while the shared header connection remained;
- the 10-minute refresh interval outlived the 30-second live-summary TTL;
- a real legacy service admitted ordered gossip and answered a legacy request, but discovery still disconnected its shared connection after settling;
- cancellation was reported as a completed discovery exchange; and
- an expired connected record retained active service membership.

Every confirmed defect above has a fixed-tree regression covering its state transition. The legacy-ownership integration regression uses the production `LegacyGossipSink`, verifies gossip and request-response before the pre-fix disconnect, and verifies both remain usable after discovery settles on the fixed tree. Hardening tests additionally inject a worker panic, exercise mapped/special-use targets and shared-IP backoff, and verify exact legacy/header/block ownership.

A literal three-node public-gossip dial cannot be made hermetic in CI: production intentionally rejects loopback and private addresses learned from untrusted gossip, and the test environment does not own routable public addresses. The signed-record-to-candidate path is tested deterministically, while a real local dial is covered by the trusted static-candidate integration test.

### Checks

- [x] `npx --yes markdownlint-cli changelog-unreleased/336.md --config .trunk/configs/.markdownlint.yaml`
- [x] `cargo clippy -p zakura-network --all-targets -- -D warnings`
- [x] `cargo test -p zakura-network` (1,038 passed; 4 pre-existing ignored)
- [x] pre-fix real-service legacy teardown reproduction
- [x] fixed-tree legacy gossip and request-response transport regression
- [x] `git diff --check origin/main...HEAD`

- Untrusted gossip remains restricted to public dial targets. Private/LAN clusters must use trusted static bootstrap configuration.
- `zakura.p2p.conn.active`, discovery admission, header-sync admission, and block-sync admission count different ownership layers; configured caps and headroom do not promise a full mesh or identical service counts at every instant.

### Follow-up Work

- Persist and reload `last_authored_sequence`. The current wall-clock seed handles normal same-key restarts, but a backward clock step can still produce a non-increasing sequence.
- Add a deployment smoke test with three routable public nodes to cover the one path a hermetic local test cannot exercise end to end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,10 +27,6 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
-- Let native Zakura peer sets converge beyond bootstrap nodes by dialing valid
-  gossiped records, advertising configured external addresses, refreshing
-  long-lived discovery sessions, and tracking actual service ownership
-  ([#336](https://github.com/zakura-core/zakura/pull/336)).
 - Treat IPv4 and IPv4-mapped IPv6 peer addresses as the same address when
   enforcing bans, preventing banned peers from reconnecting through the
   alternate representation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
+- Let native Zakura peer sets converge beyond bootstrap nodes by dialing valid
+  gossiped records, advertising configured external addresses, refreshing
+  long-lived discovery sessions, and tracking actual service ownership
+  ([#336](https://github.com/zakura-core/zakura/pull/336)).
 - Treat IPv4 and IPv4-mapped IPv6 peer addresses as the same address when
   enforcing bans, preventing banned peers from reconnecting through the
   alternate representation
--- a/changelog-unreleased/336.md
+++ b/changelog-unreleased/336.md
@@ -1,0 +1,6 @@
+## Fixed
+
+- Let native Zakura peer sets converge beyond bootstrap nodes by dialing valid
+  gossiped records, advertising configured external addresses, refreshing
+  long-lived discovery sessions, and tracking actual service ownership
+  ([#336](https://github.com/zakura-core/zakura/pull/336)).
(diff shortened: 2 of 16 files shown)


evan-forbes:
Superseded by the focused draft stack #373, #374, and #375. This PR and its branch are intentionally preserved as the pre-split audit backup.

PR #341, opened 2026-07-21
fix(mempool): retry failed transaction advertisements

p0mvn:
## Motivation

PR #64 made the mempool service the recovery source for transaction gossip channel lag, but its bounded take removed IDs before the peer-set advertisement completed. A readiness error or broadcast timeout therefore still dropped those IDs permanently.

## Solution

- Keep a retry copy of each bounded pending-gossip batch.
- Requeue only IDs that are still stored in the mempool after peer-set failures.
- Retry transient broadcast failures after the gossip delay while propagating terminal readiness failures after preserving their batch.
- Remove redundant post-query chunking and distinguish lag events from skipped channel messages in metrics.
- Update paused-time integration coverage so successful advertisements are acknowledged before their timeout.

Filtering requeued IDs through indexed current-mempool lookups prevents mined, expired, invalidated, or evicted transactions from being restored after a failed advertisement.

## Testing

- `npx --yes markdownlint-cli changelog-unreleased/341.md --config .trunk/configs/.markdownlint.yaml`
- `cargo clippy -p zakura -p zakura-node-services --all-targets -- -D warnings`
- `cargo test -p zakura --lib` — 310 passed
- `cargo test -p zakura-node-services` — passed, including doc tests

The primary paused-time regression test hangs the first peer-set call, advances through `TIPS_RESPONSE_TIMEOUT`, verifies the IDs are requeued, advances `PEER_GOSSIP_DELAY`, and verifies the same IDs are automatically advertised successfully without another channel wakeup. It then confirms pending gossip is empty and the task remains alive. Additional tests verify terminal readiness errors requeue IDs and propagate, stored IDs survive take/requeue/take, and IDs no longer present in storage are not requeued.

- Follow-up to #64.

### Follow-up Work

None.

### AI Disclosure

- [x] AI tools were used for implementation, review, and verification.

### DIFF
--- a/changelog-unreleased/341.md
+++ b/changelog-unreleased/341.md
@@ -1,0 +1,5 @@
+## Fixed
+
+- Retry transaction advertisements after transient peer-set failures instead
+  of dropping them
+  ([#341](https://github.com/zakura-core/zakura/pull/341)).
--- a/zakura-node-services/src/mempool.rs
+++ b/zakura-node-services/src/mempool.rs
@@ -92,6 +92,11 @@ pub enum Request {
         limit: usize,
     },
 
+    /// Restore transaction IDs after their peer-set advertisement failed.
+    ///
+    /// The mempool ignores IDs that are no longer stored.
+    RequeuePendingGossipTransactionIds(HashSet<UnminedTxId>),
+
     /// Query matching [`UnminedTx`] in the mempool,
     /// using a unique set of [`UnminedTxId`]s.
     TransactionsById(HashSet<UnminedTxId>),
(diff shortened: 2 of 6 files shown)


evan-forbes (commented):
fixing #64 makes sense

however my understanding of the existing mempool is that is wasn't ever supposed to guarantee gossip of a tx

can we enter a scenario where we continually retry again and again and again? is that problematic?

ai is saying that 

> tx_ids.len() > MAX returns Err, which kills the node. gossip.rs:180. The mempool honours limit (mempool.rs:908-917), so this is an internal-invariant check — but ? propagates it out of the loop, ending the gossip task and taking down zakurad. Same for requeue_pending_mempool_transaction_ids's response-variant check. For a non-consensus-critical best-effort gossip path, warn! + truncate would be the safer failure mode.

and

> 3. Possible starvation above 25k pending. TakePendingGossipTransactionIds uses pending_gossip_tx_ids.iter().take(limit) — HashSet iteration order is stable for a given set state. If the batch fails and is re-inserted, the same front-of-iteration subset can be taken again, so the IDs beyond 25k can be starved indefinitely while a slow peer keeps the head batch failing. Pre-PR this couldn't happen because a failed batch was dropped, which forced progress. Needs >25k pending, so low likelihood, but it's a new property.


much more minor but 

> 5. mempool.gossip.lagged.events.total silently changes meaning — was incremented by skip_count, now by 1, with the old semantics moved to a new ...messages.total. Any existing dashboard/alert reads differently after this deploys, and changelog-unreleased/341.md doesn't mention it.

p0mvn:
Expected this to be an easy change, but seems that there is a lot more nuance + Dev didn't want this. I think we can just close it for now

PR #349, opened 2026-07-22
ci: add manual mempool load test harness

p0mvn:
## Motivation

Mempool changes like #342 (expired-tx dependents), #341 (advertisement retries), and #64 (gossip txid drain on channel lag) are load- and backpressure-dependent. Nothing in the current test suite exercises them under sustained transaction load across peers, so "did this regress throughput or propagation?" has no answer today.

## Solution

A manually dispatched, ephemeral DigitalOcean testnet that blasts funded Orchard transactions at a small fleet and grades the result.

The chain is synthetic and isolated: local genesis, a fresh network magic per run, premine funds, no public peers. No Mainnet state or real value is involved.

This **extends the existing PR-node harness** rather than adding a second stack — same `doctl` provisioning, same baked image, same TTL reaper (now sweeping `zakura-mempool-load` too). Kresko's Rust binary is the load generator, built against this repo's crates via a vendored compat patch (`.github/workflows/scripts/kresko/`, which documents each divergence and is intended to be upstreamed).

No `environment:` gate, matching `zakura-pr-node.yml`. Spend is bounded by `workflow_dispatch`-only triggering, validated input caps, a `droplet_size` choice list, `timeout-minutes`, `teardown_after_run` defaulting to **true**, and the hourly reaper.

Set `baseline_ref` to run the same workload twice on one droplet and get a side-by-side regression table — same hardware sequentially is cheaper and lower-noise than two fleets.

Also wires the repo's three Python test suites into a gating `ci-python-tests` lint job. They existed but **no workflow ran them**; one had a pre-existing Python 3.12 failure, fixed here.

## Testing

Validated end-to-end locally: 3-node testnet seeded to height 128, peered, mining, 100+ Orchard transactions submitted with 0 errors, transactions observed propagating across all nodes, clean teardown, no key material in artifacts. Six consecutive green rehearsals.

102 unit tests (`python3 .github/workflows/scripts/test_mempool_load.py`), each pinned to a failure a real run actually produced. Notable ones:

- **Secret leak.** `kresko genesis` writes every funded key's `secret_key_hex` and the bootstrap treasury key into `config.json` — a filename no name-based check flags. Collection now sanitizes it, refuses any file whose *content* carries key material, and the workflow re-scans before upload.
- **Shared state DB.** `cache_dir` exists in both `[network]` and `[state]`; a bare-key rewrite hit only the first, leaving every node on one RocksDB. Config rewrites are now addressed by `section.key`.
- **Kresko hardcodes NU7**, whose consensus branch id is compiled out of release builds — no block could ever be mined. Made configurable, defaulting to Nu6_3.
- **Kresko's tx builder treated `Nu6_1` as inactive** while its config writer activated it on the chain — every transaction rejected *after* paying for Orchard proving.
- **Verdict could go green on a dead run.** A frozen chain converges trivially; grading now requires the chain to have advanced, every node to still answer RPC, and metrics to have been scraped.
- **Anchor races counted as node rejects.** One `unknown_orchard_anchor` in a 22-submission run is 4.35% against a 5% threshold — a live false-positive generator. Now split out and ungraded, with a minimum-sample floor.

### Follow-up Work

- **The droplet path is unexercised.** A/B leg logic, bootstrap wait, and Kresko patch-and-build have only been validated by inference. Smoke dispatch is next; expect iterations.
- Upstream the Kresko compat patch to valargroup/kresko, then drop the vendored copy and the `git apply` step.
- Optionally warm a Kresko cargo cache into the bake, so the first run doesn't pay a cold rocksdb+orchard build.
- Unrelated, but found while checking environments: `zakura-mainnet` and `zakura-testnet` have **no protection rules**, so those deploy workflows declare `environment:` while providing no approval gate.


evan-forbes (approved):
definitely not blocking as I think we'd need to update kresko again

but for the testnet, we might want to go ahead and active 6.3 and use ironwood

also we might be able to use some kresko sub commands or further modifications to handle more things here instead of hack around kresko with a lot of python scripts

p0mvn:
As a follow-up, added another column that would help interpret the data. 27 submitted vs 26 observed is because the monitor started later. 10 confirmed vs 27 submitted is due to block production

p0mvn:
Going to address recommendations in a follow-up

ValarDragon:
Amzing to get a load test harness. Want to go through it and see how we fair. Likely tons more to improve!

p0mvn:
Addressed Evan's comments Kresko side:
- https://github.com/valargroup/kresko/pull/6
- https://github.com/valargroup/kresko/pull/7

Will follor-up with script clean ups in Zakura

PR #348, opened 2026-07-22
feat(state): persist highest completed checkpoint

p0mvn:
## Motivation

Header-root authentication needs a durable state-owned fact proving that every canonical header in a configured checkpoint bracket has been stored. Network scheduling intent is not sufficient evidence, and reconstructing this fact from the entire chain on every write would be expensive.

## Solution

- Add a dedicated, versioned singleton `highest_completed_checkpoint` column family storing `HighestCompletedCheckpoint`.
- Reconstruct it during the 28.0.2 database migration, anchored at the verified body tip.
- Advance it atomically with bracket-closing header or body commits.
- Treat headers at or below the highest completed checkpoint as immutable.
- Rebase it during rollback and startup header-store repair.
- Keep root authentication requests, network publication, and root-row migration out of this PR.

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p zakura-state highest_completed_checkpoint --lib`
- `cargo test -p zakura-state header_store_coherence::startup_audit --lib`
- `cargo test -p zakura-state rollback --lib`
- `cargo test -p zakura-state disk_format::tests::snapshot::test_raw_rocksdb_column_families --lib`
- `cargo clippy -p zakura-state --all-targets -- -D warnings`

The tests cover checkpoint-bracket completion, interior gaps, restart restoration, malformed metadata, migration, startup repair, rollback, and raw database-format snapshots.

## Changelog

- Added `changelog-unreleased/348.md` with an explicit no-changelog marker because this is internal state infrastructure without operator-visible behavior.

### Specifications & References

- `docs/design/header-sync-vct-root-authentication.md`, Phase 2: state-owned `HighestCompletedCheckpoint`.
- Builds independently on `main`; it does not depend on PR #346.

### Follow-up Work

- Publish the state-owned highest completed checkpoint to header sync.
- Add authenticated-root promotion and network scheduling in separate PRs.

PR #323, opened 2026-07-21
feat(sync): authenticate VCT roots before persistence

p0mvn:
## Motivation

Header sync accepted peer-supplied commitment roots into the authoritative VCT index before authenticating them. Existing databases could therefore contain mixed-provenance rows, and restart recovery had no durable canonical authentication boundary.

## Solution

- seal successful supplied-root verification behind `VerifiedHeaderCommitmentRoots`
- migrate existing databases by deleting legacy header-ahead rows and initialize durable authenticated-root and completed-checkpoint frontiers
- enforce exact canonical, contiguous, append-only root promotion with body-derived precedence
- reconstruct and validate checkpoint brackets across restart, reorg, rollback, and body commits
- serialize `AuthenticateHeaderRoots` with state writes and publish compact durable frontier updates
- schedule checkpoint-covered overlapping root ranges with exact peer/session/request attribution
- reuse aligned forward responses only after their headers commit canonically
- apply bounded asynchronous state operations, stale/local retry handling, and peer blame only for invalid supplied data
- require every successor witness to be checkpoint-covered; an uncheckpointed successor cannot authenticate authoritative roots

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy -p zakura-chain -p zakura-state -p zakura-network -p zakura --all-targets -- -D warnings`
- `cargo test -p zakura-state --lib` (414 passed, 3 ignored before final focused review)
- `cargo test -p zakura-chain parallel::commitment_aux_verify::tests` (6 passed)
- header-sync focused suite (179 passed)
- zakurad integration suite (47 passed)
- full `zakura-network` suite reached 1022 passing tests; 3 macOS loopback-address tests were excluded because their requested local source addresses are unavailable on this host
- full `zakura` suite reached 304 passing tests; one unrelated tracing-global test passed individually
- `git diff --check`
- IDE diagnostics report no errors in changed files

Coverage includes migration cutover, malformed snapshot decoding, canonical append/stale/gap rejection, checkpoint reconstruction and interior gaps, witness bounds, body precedence and rebase, watch ordering, one-header overlap, short responses, response reuse, session replacement, delayed retries, cancellation, and exact peer attribution.

### Specifications & References

- Design: #230
- Phase 0 prerequisites: #227, #246, #298, #309, #307, #316

### Follow-up Work

- Complete VCT handoff integration and remove obsolete provisional-root repair behavior in focused follow-ups.

PR #351, opened 2026-07-22
feat(state): track completed checkpoints in memory

p0mvn:
## Motivation

Header-root authentication needs an efficient state-owned fact proving which configured checkpoint bracket is completely stored. Reconstructing the full header prefix on every write is wasteful, while persisting an independent checkpoint row creates another source of truth and a database migration surface.

## Solution

- reconstruct `HighestCompletedCheckpoint` from durable canonical headers and configured checkpoints when state opens
- cache the completed checkpoint and the contiguous in-progress cursor in the state write worker
- compute header-batch proposals without mutating memory, then publish them only after the RocksDB batch succeeds
- refresh progress after finalized-body and committed-header seed writes, clearing published progress if reconstruction fails
- expose a watch subscription for later header-root authentication consumers and keep read-only subscriptions open
- reject header reorgs at or below the completed checkpoint

## Testing

- `cargo test -p zakura-chain checkpoint_list_neighbor_lookups --locked`
- `cargo test -p zakura-state highest_completed_checkpoint --lib --locked` (6 passed)
- `cargo test -p zakura-state side_chain_commit_does_not_seed_zakura_headers --lib --locked`
- `cargo test -p zakura-state read_only_completed_checkpoint_subscription_stays_open --lib --locked`
- `cargo clippy -p zakura-state --all-targets --locked -- -D warnings`
- `git diff --check`

Coverage includes successful bracket completion, restart reconstruction, proposal purity on failed writes, interior gaps, immutable completed-checkpoint ancestors, fail-closed reconstruction errors, watch notifications, and read-only subscription lifetime.

## Changelog

Added `changelog-unreleased/351.md` with an explicit no-changelog marker because this adds library APIs for future integration without operator-visible behavior.

- `docs/design/header-sync-vct-root-authentication.md`, state-owned completed checkpoint frontier
- Follow-up integration is based on draft PR #323.
- Supersedes the dedicated durable checkpoint-row approach in PR #348.

### Follow-up Work

- Rebase a fresh successor branch from PR #323 onto this PR and make root authentication consume the in-memory tracker.

### DIFF
--- a/changelog-unreleased/351.md
+++ b/changelog-unreleased/351.md
@@ -1,0 +1,4 @@
+<!-- changelog: none -->
+
+This PR adds internal monitoring state for future header-root authentication;
+it does not change operator-visible behavior or published crate APIs.
--- a/changelog-unreleased/351.md
+++ b/changelog-unreleased/351.md
@@ -1,4 +1,4 @@
 <!-- changelog: none -->
 
-This PR adds internal monitoring state for future header-root authentication;
-it does not change operator-visible behavior or published crate APIs.
+This PR adds state and chain APIs for future header-root authentication;
+it does not change operator-visible behavior.
--- a/zakura-chain/src/parameters/checkpoint/list.rs
+++ b/zakura-chain/src/parameters/checkpoint/list.rs
@@ -7,7 +7,7 @@
 
 use std::{
     collections::{BTreeMap, HashSet},
-    ops::RangeBounds,
+    ops::{Bound, RangeBounds},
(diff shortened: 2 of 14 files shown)


p0mvn:
Addressed [F-103451](https://v12.sh/runs/5038/103451). [F-103452](https://v12.sh/runs/5038/103452) I think is an overkill, can be addressed later as needed

p0mvn:
@cursor review

evan-forbes (approved):

PR #352, opened 2026-07-22
feat(sync): authenticate peer-supplied VCT commitment roots

p0mvn:
## Motivation

Header sync downloads peer-supplied VCT commitment roots and ZIP-221 leaf inputs together with headers, but nothing below the last checkpoint with embedded frontier semantically verifies the header commitment field they feed into. A malicious peer could therefore poison the authoritative commitment-root index and, through it, VCT fast sync. Roots must be authenticated against the canonical header chain before they are persisted or served.

## Design

Full design: `docs/design/header-sync-vct-root-authentication.md`. Key details:

- **Central invariant**: every row visible through the authoritative commitment-root index has already been authenticated against the selected header chain. Trust is set at the write boundary — after promotion, peer-supplied and body-derived roots are indistinguishable, so the index stores no per-row provenance.
- **Witness rule**: a header commits to the history tree as of its parent, so roots at height `H` are authenticated by header `H + 1`. Adjacent root-carrying ranges overlap by one header, so every single-peer response contains its own successor witness and a cryptographic mismatch is attributable to exactly that response.
- **Checkpoint gating**: only state-published completed checkpoint brackets authorize promotion. A valid-PoW witness above the completed checkpoint is not pinned by any configured checkpoint hash and authenticates nothing, so every promoted root and its witness must be at or below `HighestCompletedCheckpoint.height`.
- **State owns the frontier**: `zakura-state` persists a single ascending `HeaderRootAuthFrontier` (height, hash, history tree). Authenticated roots and the frontier advance in one atomic database batch, and the sealed `VerifiedHeaderCommitmentRoots` type is the only path into the write helper — raw peer roots cannot reach persistence.
- **Retained-first supply**: the reactor keeps root-carrying committed forward responses in memory, with exact peer/session/request attribution, through the last checkpoint with embedded frontier. In the steady state every root is transferred once; dedicated `AuthenticateRoots` requests are fallback-only for gaps, invalidated or rejected payloads, and post-restart catch-up.
- **Fail-closed everywhere**: the state format upgrade deletes legacy unauthenticated root rows above the verified body tip and seeds the frontier from the body-derived history tree; serving reads only the authoritative index and reports interior gaps as local incoherence; body verification remains the final proof that authenticated counts and auth-data roots match downloaded transactions.

## Solution

State (`zakura-state`):

- add `Request::AuthenticateHeaderRoots`: canonical-hash and witness validation, `verify_supplied_roots_from_parts`, atomic confirmed-prefix promotion with the durable frontier, and typed peer/stale/local outcomes
- persist only the authenticated root frontier (height, hash, and history tree); stop `CommitHeaderRange` from writing raw peer roots
- add the one-time format upgrade that removes unauthenticated header-ahead roots and initializes the frontier from the body history tree
- source completed-checkpoint progress from PR #351's in-memory tracker, gate each authentication request and successor witness against the exact tracker snapshot, and publish combined root-auth/checkpoint state only after durable writes succeed
- restore the frontier at startup with coherence checks, rebase it when the verified body tip overtakes it, and reconstruct defensively from authenticated rows

Network and driver (`zakura-network`, `zakurad`):

- add a root-authentication lane to the header-sync reactor that follows the state-published frontier
- retain committed root-carrying forward responses keyed by contiguous start height, and consume them retained-first once their successor witness is checkpoint-covered
- overlap adjacent root-carrying forward requests by one header; on retained misses, prefetch overlapping fallback ranges across a bounded window; catch up after restart, where the retained store is intentionally not restored
- carry exact operation identity (peer/session/request) through buffering, driver actions, and state completion; attribute invalid roots and refetch from another peer; never score peers for stale or local errors
- expose root-auth work/retain/fallback metrics and `header_root_auth_diagnostics` JSONL traces

## Testing

~20 runs with regular restarts while monitoring header, auth and body tips. Lost in-memory roots recover as intended. No stalls observed. 

image


p0mvn:
@v12-auditor review

p0mvn:
@cursor review

p0mvn:
@cursor, please comprehensively review the PR and ensure that there are no unintended scripts or non-production code that was committed

p0mvn:
Looked into the Cursor Medium, which is the same as the V12 report. I think it is fine and better not to punish the peer in that case for now. Our producer does not have a guarantee to return >= 2 headers. This can be addressed separately if it becomes a concern

evan-forbes (commented):
great debugging all of this holy smokes

edit: ignore the below, we are only verifying VCT roots before checkpoint height! which makes a ton of sense

I'm currently pondering solutions (including not fixing) to a niche  issues. That is, serving VCT roots that are valid and verified, however are not on the canonical chain. Meaning a node downloads a header that it verifies the contiguity of, however there's fork or something and we end up serving these headers to peers.

this is only an issue after checkpoint sync ofc.

until we have DAG based headersync, my understanding is that currently here after checkpoint sync, we could end up overwriting the tree here with a fork of valid headers, and then have to just get lucky and re-request those headers from peers who are serving the correct ones.

fortunately, I'm a lot less worried about the integration into the reactor/state/storage or the bug above due the upcoming DAG based refactor, which removes our reliance on a purely height based schema for requests and storage.

still continuing through the code

zakura-chain/src/parallel/commitment_aux_verify.rs
```diff
@@ -108,10 +108,16 @@ where
             &roots.sapling_root,
         )
         .map_err(|error| (height, error))?;
+        verify_supplied_sapling_tx_below_sapling(network, height, roots.sapling_tx)
+            .map_err(|error| (height, error))?;
```
p0mvn, line 112: Note to reviewer: these have been added from AI-audits

evan-forbes (approved):
I can't find any obvious issues and we're about to more or less re-write headersync, therefore we should get this in alongside the upcoming #166 and test

PR #355, opened 2026-07-22
fix(sync): validate discovery hash responses

ValarDragon:
## Motivation

Malformed nonempty `FindBlocks` responses could poison legacy sync discovery.
After locating the first unknown hash, `obtain_tips` treated the rest of the
response as downloadable without validating that the suffix remained unknown.
A sequence such as `[unknown, known, random]` therefore reached the later
state-membership check and aborted the active sync attempt instead of discarding
only the malformed peer response.

This is a response-validation variant of GHSA-h9hm-m2xj-4rq9 and re-extracts
the focused fix from #151 onto current `main`.

## Solution

- Validate legacy discovery responses before mutating download state.
- Reject duplicate hashes, known hashes after the first unknown hash, locator
  echoes in advertised suffixes, and nonproductive response shapes.
- Apply equivalent validation to both tip discovery and extension building.
- Preserve the response-size guards added to `main` after the original port.
- Add regression coverage for malformed discovery and extension responses.

## Testing

- `cargo test -p zakura components::sync` (62 passed)
- `npx --yes markdownlint-cli changelog-unreleased/355.md --config .trunk/configs/.markdownlint.yaml`

The regression tests exercise known hashes embedded after the first unknown
hash, duplicate hashes, malformed extension responses, and permitted unrelated
trailing hashes.

## Changelog

Added `changelog-unreleased/355.md` under `Security`.

- Original Zakura port: #151
- Original Zebra port: https://github.com/valargroup/zebra/pull/318
- Advisory: GHSA-h9hm-m2xj-4rq9

### Follow-up Work

None.

---

> [!NOTE]
> **Medium Risk**
> Changes peer-response handling on the legacy sync hot path; behavior is stricter but aligned with existing size guards and should only drop bad responses, with regression tests covering the failure modes.
> 
> **Overview**
> **Hardens legacy block discovery** so malformed peer `FindBlocks` replies are dropped instead of aborting the active sync round.
> 
> `obtain_tips` now scans responses with `first_unknown_in_find_blocks_response`, rejecting duplicates and any **known hash after the first unknown** (the case that previously queued `[unknown, known, …]` and later tripped `queued download of hash behind our chain tip`). Tip extension moves its zcashd-tolerance and sanity checks into `validate_extend_find_blocks_response`, with a cloned state service passed into `build_extend` so validation can run off the main sync task. Extension validation also enforces the hash-count cap **before** state lookups, and rejects duplicate or known suffix hashes like obtain.
> 
> Vector tests expect per-hash `KnownBlock` checks and add coverage for malformed obtain/extend responses, oversized extend payloads without state amplification, and allowed unrelated trailing hashes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6496d4084c382bdd895a5ce64979956bbf2a35b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/changelog-unreleased/355.md
+++ b/changelog-unreleased/355.md
@@ -1,0 +1,5 @@
+## Security
+
+- Reject malformed legacy block-discovery responses instead of allowing them
+  to disrupt the active sync attempt
+  ([#355](https://github.com/zakura-core/zakura/pull/355)).
--- a/zakurad/src/components/sync.rs
+++ b/zakurad/src/components/sync.rs
@@ -1324,8 +1324,9 @@ where
                 );
 
                 let tip_network = self.tip_network.clone();
+                let state = self.state.clone();
                 let tips = std::mem::take(&mut self.prospective_tips);
-                extend = Some(Box::pin(Self::build_extend(tip_network, tips)));
+                extend = Some(Box::pin(Self::build_extend(tip_network, state, tips)));
             }
 
             // Dispatch from the reserve while we're below the lookahead limit.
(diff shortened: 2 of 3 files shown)


evan-forbes (approved):

ValarDragon:
@v12 pls audit

ValarDragon:
@v12-auditor audit this

PR #361, opened 2026-07-22
refactor(chain)!: remove unused public API

ValarDragon:
## Motivation

`zakura-chain` exposed several errors, constants, and helper methods that have
no callers anywhere in the workspace. Because the crate is published to
support `cargo install zakura`, those unused items still enlarge its declared
public API, rustdoc, and metadata and must be treated as compatibility surface.
This is API hygiene; the removed definitions are too small to expect a
meaningful or reliably measurable compile-time improvement.

## Solution

Remove the following unused public API:

- `error::{KeyError, NoteCommitmentError}`
- `parameters::testnet::MAX_HRP_LENGTH`
- `NetworkUpgrade::previous_upgrade`
- `DateTime32::checked_elapsed`
- `transparent::Input::value_from_ordered_utxos`
- `CodeTimer::ignore`

Every removed item had exactly one workspace occurrence: its definition. No
replacement is needed because the workspace has no callers.

This is a breaking source-level API change for any external consumer that used
one of these items despite `zakura-chain` being documented as an internal
crate.

Retain the PoW difficulty display helpers at reviewer request. Also retain
`GeneratedLocalTestnet::genesis_hex`: a downstream audit found two concrete
uses in [Kresko](https://github.com/evan-forbes/kresko), although Kresko
currently consumes them from a tagged `valargroup/zebra` fork.

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p zakura-chain --lib` (364 passed)
- `cargo clippy -p zakura-chain --all-targets --all-features -- -D warnings`
- `cargo check --workspace --all-targets`
- `./scripts/changelog.py check`
- `npx --yes markdownlint-cli changelog-unreleased/361.md --config
  .trunk/configs/.markdownlint.yaml`
- `cargo check --workspace --all-targets --all-features` reached downstream
  crates without removed-symbol errors, then stopped because optional
  `lightwalletd_grpc_tests` generation requires an unavailable `protoc` binary.

## Changelog

Added `changelog-unreleased/361.md` under Removed because this changes the
published crate's source-level API.

### Specifications & References

- Follow-up to the `zakura-chain` build-time and API-surface review.

### Follow-up Work

Move runtime and test-support utilities out of the normal `zakura-chain` API
in separately reviewed changes.

## AI disclosure

OpenAI Codex was used to inventory workspace references, remove the unused
items, run verification, and prepare this PR. The contributor remains
responsible for review and merge decisions.

---

> [!NOTE]
> **Low Risk**
[...]

PR #364, opened 2026-07-22
refactor(chain): narrow unreachable public items

ValarDragon:
## Motivation

`zakura-chain` contains declarations marked `pub` that are hidden behind
private parent modules. They cannot be reached by downstream crates, but their
visibility is broader than their actual callers require.

## Solution

- Narrow seven internal declarations to `pub(super)`.
- Narrow three `PrecomputedTxData` bundle accessors to `pub(crate)`.
- Enable `unreachable_pub` for non-test `zakura-chain` builds so accidental
  overexposure is caught in future changes.

The compiler identified every changed declaration as unreachable from outside
the crate. The public `Magic` re-export remains unchanged. A Kresko source
audit found no references to the affected internal names.

## Testing

- `cargo check -p zakura-chain --all-targets --all-features`
- `cargo clippy -p zakura-chain --all-targets --all-features -- -D warnings`
- `cargo test -p zakura-chain --lib` (364 passed)
- `npx --yes markdownlint-cli changelog-unreleased/364.md --config
- `git diff --check`
- IDE diagnostics were unavailable because the installed `rust-analyzer`
  launcher has no matching toolchain component.

## Changelog

Added `changelog-unreleased/364.md` with the internal-only no-changelog marker.

- Follow-up to #361 and the `zakura-chain` API-surface review.

### Follow-up Work

- Audit intentionally reachable APIs separately; changing those requires an
  explicit downstream compatibility decision.

## AI disclosure

OpenAI Codex was used to run the compiler visibility audit, apply and review
the visibility changes, run verification, and prepare this PR. The contributor
remains responsible for review and merge decisions.

### DIFF
--- a/changelog-unreleased/364.md
+++ b/changelog-unreleased/364.md
@@ -1,0 +1,4 @@
+<!-- changelog: none -->
+
+This PR only narrows declarations that were already unreachable outside
+`zakura-chain`, so it has no operator- or crate-consumer-visible effect.
--- a/zakura-chain/src/block/height.rs
+++ b/zakura-chain/src/block/height.rs
@@ -7,7 +7,7 @@ use zcash_protocol::consensus::BlockHeight;
 use crate::{serialization::SerializationError, BoxError};
 
 #[cfg(feature = "json-conversion")]
-pub mod json_conversion;
+pub(super) mod json_conversion;
 
 /// The length of the chain back to the genesis block.
 ///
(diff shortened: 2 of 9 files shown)


ValarDragon:
Closing because these visibility-only changes have negligible compile-time impact and do not justify the added maintenance surface. We will pursue the higher-impact feature-gating work separately.

PR #372, opened 2026-07-22
fix(sync): reject locator echoes in extensions

ValarDragon:
## Motivation

Legacy tip extension accepts a requested locator when a peer echoes it inside
the retained response suffix. Because extension runs before the locator is
necessarily committed, the state lookup can report it as unknown and allow a
backward prospective tip.

## Solution

Reject a locator found anywhere in the retained extension suffix before making
state queries. Keep the existing compatibility behavior for the discarded
leading and trailing hashes.

## Testing

The added regression test supplies an uncommitted locator in the response
suffix and asserts that the response is discarded without any state query.

- `cargo test -p zakura build_extend_rejects_locator_echo_before_state_queries`
- `cargo test -p zakura components::sync`
- `cargo clippy -p zakura --all-targets -- -D warnings`

GitHub CI is green, including the full unit-test shards and Zakura E2E checks.

## Changelog

Added `changelog-unreleased/372.md` under Security.

- Follow-up to #355.

### Follow-up Work

None.

---

> [!NOTE]
> **Medium Risk**
> Security-sensitive change to chain sync tip validation; scope is narrow (one early reject + test) but incorrect filtering could affect legitimate peer responses.
> 
> **Overview**
> Closes a **security gap** in legacy tip extension: a peer could echo the requested locator hash inside the retained `FindBlocks` suffix and trick the node into treating backward chain data as a valid prospective tip when that locator was not yet committed (follow-up to #355).
> 
> **`validate_extend_find_blocks_response`** now drops extension responses if the advertised unknown suffix contains the locator (`tip.tip`), **before** any state `KnownBlock` lookups—same ordering as the oversized-response guard.
> 
> A regression test asserts an echoed uncommitted locator yields an empty extend result with **no** state service calls. Unreleased changelog notes the fix under Security.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b05406ad1a7efe6a9ce7334522ee29e7944af467. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/changelog-unreleased/372.md
+++ b/changelog-unreleased/372.md
@@ -1,0 +1,4 @@
+## Security
+
+- Reject locator hashes echoed inside legacy tip-extension responses
+  ([#372](https://github.com/zakura-core/zakura/pull/372)).
--- a/zakurad/src/components/sync.rs
+++ b/zakurad/src/components/sync.rs
@@ -1614,6 +1614,14 @@ where
             return Ok(None);
         }
 
+        if unknown_hashes.contains(&locator) {
+            debug!(
+                ?locator,
+                "discarding FindBlocks extension response with locator hash in advertised suffix"
+            );
+            return Ok(None);
+        }
+
         for &hash in unknown_hashes {
             if Self::state_contains_service(state, hash).await? {
                 debug!(
--- a/zakurad/src/components/sync/tests/vectors.rs
+++ b/zakurad/src/components/sync/tests/vectors.rs
@@ -2033,6 +2033,65 @@ async fn build_extend_rejects_oversized_response_before_state_queries(
 }
 
 #[tokio::test]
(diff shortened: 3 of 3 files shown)


ValarDragon:
5.6 xhigh:

The conclusion is stronger for Zakura/Zebra responders.

  A valid FindBlocks([L]) response can never contain L:

  - If L remains on the best chain, the response starts strictly after it.
  - If a reorg removes L, the responder returns from its fallback intersection or returns empty; the existing expected_next check rejects the stale extension.
  - If the reorg happens during response construction, Zakura returns an empty or partial forward suffix, never the intersection itself.

  So I see no valid-response regression here, including during reorg catch-up. The leading/trailing exceptions are legacy compatibility baggage and don’t affect that conclusion.

PR #373, opened 2026-07-22
fix(network): make discovered peers safely dialable

evan-forbes:
## Motivation

Native Zakura nodes can remain centered on their configured bootstrap peers instead of converging on valid peers learned through signed discovery gossip.

The audit found independent causes along the same reachability path:

- an unspecified native listener authored no usable address even when `external_addr` supplied a reachable IP;
- bootstrap configuration could create self-dials and duplicate maintained tasks for one identity; and
- valid third-party signed records were stored and regossiped but excluded from dial candidates; and
- IPv4 transition encodings were normalized inconsistently, allowing 6to4 and Teredo targets to bypass IPv4 safety checks and occupy different dial and connection-limit buckets from the embedded IPv4 identity.

Making gossiped records dialable also exposes shared-IP backoff and worker-panic cleanup paths that must not leave unsafe or permanently reserved targets.

## Solution

- Advertise reachable native addresses from the specific listen address, configured external IP, or a bootstrap entry for the local identity.
- Group bootstrap entries by remote identity, merge their addresses, and exclude malformed, duplicate, and local identities.
- Admit valid signed gossip records as candidates only after public-target filtering.
- Canonicalize mapped, compatible, 6to4, Teredo, and well-known-prefix NAT64 addresses before discovery safety checks, dial reservations, backoff, admission, or supervisor registration.
- Enforce per-IP in-flight limits and apply IP-scoped exponential backoff.
- Recover panicking dial workers as failed attempts while releasing their node and IP reservations.

## Testing

Focused regressions cover external-address advertisement, matching local bootstrap addresses, bootstrap identity grouping, public third-party signed-record candidates, mapped and special-use targets, 6to4 and Teredo private-target rejection, transition-address dial identity, canonical supervisor buckets, shared-IP backoff, and panicking worker cleanup.

Quick pre-PR checks:

- [x] dialability, bootstrap grouping, shared-IP backoff, and panic-cleanup regressions through the full crate suite
- [x] `cargo test -p zakura-network` on this branch (1,030 passed; 4 ignored)
- [x] `cargo clippy -p zakura-network --all-targets -- -D warnings` on this branch
- [x] `git diff --check origin/main...HEAD`

## Changelog

Added `changelog-unreleased/373.md`.

Untrusted gossip remains restricted to public dial targets. Private or LAN clusters must use trusted static bootstrap configuration.

### Follow-up Work

This is the first PR in a three-PR discovery audit stack:

- #374 isolates exact stream and connection ownership.
- #375 isolates periodic refresh and service-state expiry.

### DIFF
--- a/changelog-unreleased/373.md
+++ b/changelog-unreleased/373.md
@@ -1,0 +1,5 @@
+## Fixed
+
+- Let native Zakura peer sets converge beyond bootstrap nodes by advertising
+  reachable addresses and safely dialing valid gossiped discovery records
+  ([#373](https://github.com/zakura-core/zakura/pull/373)).
--- a/changelog-unreleased/373.md
+++ b/changelog-unreleased/373.md
@@ -1,5 +1,7 @@
 ## Fixed
 
 - Let native Zakura peer sets converge beyond bootstrap nodes by advertising
-  reachable addresses and safely dialing valid gossiped discovery records
+  reachable addresses, safely dialing valid gossiped discovery records, and
+  applying discovery IP safety and connection limits consistently across IPv4
+  transition encodings
   ([#373](https://github.com/zakura-core/zakura/pull/373)).
--- a/changelog-unreleased/373.md
+++ b/changelog-unreleased/373.md
@@ -5,3 +5,13 @@
   applying discovery IP safety and connection limits consistently across IPv4
   transition encodings
   ([#373](https://github.com/zakura-core/zakura/pull/373)).
+
+## Security
+
+- Harden the native discovery candidate dialer against untrusted gossip: charge
+  each dialed connection to the network path iroh actually confirms (so a record
+  cannot escape the per-IP connection cap by listing a decoy address first),
+  scope dial-failure backoff to `(node_id, ip)` instead of IP alone (so a signed
+  record cannot back an honest peer's IP off for every candidate that shares it),
+  and reject RFC 5737 / RFC 3849 documentation ranges as dial targets
+  ([#373](https://github.com/zakura-core/zakura/pull/373)).
--- a/zakura-network/src/zakura.rs
+++ b/zakura-network/src/zakura.rs
@@ -21,6 +21,7 @@ mod exchange;
 mod handler;
 mod handshake;
 mod header_sync;
+mod ip;
 mod legacy_gossip;
 #[cfg(any(test, feature = "zakura-testkit"))]
 pub mod testkit;
@@ -33,6 +34,7 @@ pub use exchange::*;
 pub use handler::*;
 pub use handshake::*;
 pub use header_sync::*;
+pub(crate) use ip::canonical_ip;
 pub use legacy_gossip::*;
 pub use trace::{
     commit_state_trace, peer_label as zakura_trace_peer_label,
--- a/zakura-network/src/zakura/discovery/candidate_dialer.rs
+++ b/zakura-network/src/zakura/discovery/candidate_dialer.rs
@@ -13,7 +13,7 @@ use std::{
 };
 
 use iroh::{NodeAddr, NodeId};
-use tokio::task::JoinSet;
+use tokio::{task::JoinSet, time::Instant};
 use tracing::debug;
 
 use super::dialer::native_bootstrap_dial;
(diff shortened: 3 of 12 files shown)


evan-forbes:
fixing the v12 shortly

ValarDragon:
3 GPT 5.6 suggestions for follow-up:


• Review result: request changes. I found three actionable issues.

  - [P1] Charge connections to iroh’s selected path, not the first advertised address. The dialer passes multiple direct addresses to iroh, but serve_native_dial_connection
    (https://github.com/zakura-core/zakura/blob/81ac0beee86ada1d491622fdf16634fe445d9f8a/zakura-network/src/zakura/handler.rs#L3197-L3249) records direct_addresses().next() before
    connecting. If iroh succeeds through another address, the supervisor charges the wrong IP. A malicious multi-address record can use a unique first “decoy” address and a shared
    reachable address to bypass max_connections_per_ip once the in-flight reservations are released. Resolve the confirmed direct path after connection/handshake, and add a regression
    where only the second address is reachable.

  - [P1] Unverified address claims can poison IP-wide backoff. Third-party records are now dialable even though signatures do not prove address ownership. On failure,
    apply_discovery_ip_dial_result
    (https://github.com/zakura-core/zakura/blob/81ac0beee86ada1d491622fdf16634fe445d9f8a/zakura-network/src/zakura/discovery/candidate_dialer.rs#L324-L359) penalizes every claimed IP.
    An attacker can sign records using fresh node IDs but advertise an honest peer’s IP; the expected identity mismatch backs that IP off for all real candidates, exponentially up to
    one hour. Keep failure backoff scoped to (node_id, IP) or avoid promoting unauthenticated failures into IP-wide backoff.

  - [P2] The “globally routable” filter still accepts non-global ranges. is_discovery_dialable_addr
    (https://github.com/zakura-core/zakura/blob/81ac0beee86ada1d491622fdf16634fe445d9f8a/zakura-network/src/zakura/discovery/protocol.rs#L3425-L3457) accepts documentation ranges
    including 192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24, and 2001:db8::/32. These are explicitly non-global in the IANA IPv4
    (https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml) and IPv6
    (https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml) registries. Internally routed uses remain SSRF targets, contrary to the stated
    security boundary. One test even labels 192.0.2.x “genuinely globally routable.”

ValarDragon (approved):
New finding from codex that we can fix in follow-up:

  - [P1] Unverified address claims can poison IP-wide backoff. Third-party records are now dialable even though signatures do not prove address ownership. On failure,
    apply_discovery_ip_dial_result
    (https://github.com/zakura-core/zakura/blob/81ac0beee86ada1d491622fdf16634fe445d9f8a/zakura-network/src/zakura/discovery/candidate_dialer.rs#L324-L359) penalizes every claimed IP.
    An attacker can sign records using fresh node IDs but advertise an honest peer’s IP; the expected identity mismatch backs that IP off for all real candidates, exponentially up to
    one hour. Keep failure backoff scoped to (node_id, IP) or avoid promoting unauthenticated failures into IP-wide backoff.

evan-forbes:
All three findings from @ValarDragon's review (#pullrequestreview-4759382576) and the audit comment (#issuecomment-5052314848) are now fixed **in-place in this PR** (commit 8fd8ba677), rather than as a follow-up. They are all reachable only because #373 is the first PR to actually dial untrusted third-party gossip records, so the fixes belong here. Each has a regression test.

**[P1] Backoff poisoning** — dial-failure backoff was keyed by `IpAddr` alone. A signature proves node-key control, not IP ownership, so an attacker could sign throwaway node ids advertising an honest peer's IP and back that IP off for every legitimate candidate. Backoff is now keyed by `(node_id, ip)`. In-flight reservations stay IP-scoped (they bound a real local resource). Test: `failed_dials_back_off_only_the_failing_node_for_that_ip`.

**[P1] Per-IP cap bypass via decoy address** — `serve_native_dial_connection` charged the connection to `direct_addresses().next()` captured *before* connecting, so a record listing an unreachable decoy first could escape the per-IP cap while being served over a shared address. The connection is now charged to the path iroh actually confirms (reusing the inbound confirmed-path resolver); relay paths yield `None` and stay bounded by the global cap. Test: `outbound_dial_charges_confirmed_path_not_advertised_decoy`.

**[P2] Documentation-range SSRF** — the dialability filter still accepted RFC 5737 (192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24) and RFC 3849 (2001:db8::/32) documentation ranges. These are now rejected, and the test fixtures that used those ranges as routable targets were migrated to a globally routable block. Tests: added cases in `discovery_book_peer_import_rejects_private_and_internal_direct_addresses`.

The changelog fragment now carries a `## Security` entry. #374 and #375 have been rebased onto this update. Since this amends an approved PR, re-requesting review.

PR #376, opened 2026-07-22
fix(network): ban malformed getblocks responses

ValarDragon:
## Motivation

Malformed `getblocks` responses were previously scoped to a single sync
request. A peer could send too many block hashes, duplicate hashes, or echo a
requested locator without receiving a peer-level penalty, then repeat the same
structural violation on later requests.

## Solution

Validate block-hash `inv` responses at the peer connection boundary. Retain the
requested locator set until the response arrives, reject responses containing
more than the protocol maximum of 500 hashes, duplicates, or any requested
locator, assign the full banning score through the address-book updater, and
close the connection.

Direct peers are banned by IP. Connections without a safe address-book identity
(proxy or isolated connections) are still disconnected, but are not assigned a
ban to avoid penalizing a shared proxy address.

## Testing

Focused connection tests pass for all three ban triggers and for an accepted
response at the exact 500-hash limit:

- `cargo test -p zakura-network find_blocks_response`

Full risk-proportional verification will run after this draft is opened.

## Changelog

A changelog fragment will be added after this draft provides the PR number.

- Stacked on #372.
- Follow-up to #355.

### Follow-up Work

None.

---

> [!NOTE]
> **Medium Risk**
> Touches peer connection lifecycle and address-book banning on a security-sensitive P2P path; behavior is well-tested but misclassification could disconnect or ban legitimate peers.
> 
> **Overview**
> **Hardens P2P sync against abusive `getblocks` replies.** Peers could previously return oversized, duplicate, or locator-echoing block-hash `inv` responses with only per-request failure and no persistent penalty.
> 
> The connection layer now keeps the requested locator set on `FindBlocks`, rejects responses over **500** hashes (`MAX_FIND_BLOCKS_RESPONSE_HASHES`), duplicate hashes, or any hash that matches a sent locator, and surfaces `PeerError::InvalidFindBlocksResponse`. On that error the node fails the client request, applies **maximum misbehavior score** via the new per-connection `address_book_updater` when a safe address-book identity exists, and tears down the connection.
> 
> Tests cover acceptance at the 500-hash limit and ban paths for oversize, duplicate, and locator-echo cases. A security changelog entry documents the behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ba71c0e7190e862f8c680b7a732db79a786c9d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/changelog-unreleased/372.md
+++ b/changelog-unreleased/372.md
@@ -1,0 +1,4 @@
+## Security
+
+- Reject locator hashes echoed inside legacy tip-extension responses
+  ([#372](https://github.com/zakura-core/zakura/pull/372)).
--- a/changelog-unreleased/376.md
+++ b/changelog-unreleased/376.md
@@ -1,0 +1,5 @@
+## Security
+
+- Ban peers that return oversized, duplicate, or locator-echoing block hash
+  responses to `getblocks`
+  ([#376](https://github.com/zakura-core/zakura/pull/376)).
--- a/zakura-network/src/lib.rs
+++ b/zakura-network/src/lib.rs
@@ -194,7 +194,9 @@ pub use crate::{
     peer_set::{init, init_with_zakura_header_sync},
     policies::RetryLimit,
     protocol::{
-        external::{Version, VersionMessage, MAX_TX_INV_IN_SENT_MESSAGE},
+        external::{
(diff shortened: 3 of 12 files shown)


ValarDragon:
I want to do something like this, but I think this PR is too naive about it

PR #381, opened 2026-07-22
ci: speed up x86 release binary builds

ValarDragon:
## Motivation

The x86 release build spends about 15 minutes performing a cold optimized Rust
build on a standard four-core runner. The standalone release archive also builds
and links two maintenance binaries that it does not package.

## Solution

- Run all x86 release compilation jobs on the
  `zakura-release-x64-64-core` larger-runner pool.
- Add a `release-zakurad` Docker target that builds only `zakurad`.
- Keep the existing `release` target responsible for adding
  `zakura-rollback-state` and `zakura-prune-state` to runtime images.
- Add a manual `dry_run` mode and `checkout_ref` input for safely benchmarking
  the complete build matrix without uploading Actions artifacts, publishing
  GitHub release assets, or pushing Docker images/manifests.

## Testing

- Parsed `.github/workflows/release-binaries.yml` with Ruby YAML
- `cargo metadata --locked --no-deps --format-version 1` and verified all three
  binary targets
- GitHub Actions PR checks
- [Successful dry-run benchmark](https://github.com/zakura-core/zakura/actions/runs/29952894850)
  against commit `af2d104796b1142690b7c36d81a647f03084e076`
- [Repeat dry run](https://github.com/zakura-core/zakura/actions/runs/29960096538)
  against the same commit

The successful dry run completed the x86 standalone release build in 2m51s,
down from 14m52s on the four-core runner: an 81% reduction, or about 5.2x
faster. The complete x86 asset job took 3m12s. The regular and zcashd-compatible
x86 Docker build steps each took 2m30s. Every x86 job succeeded.

The same run verified dry-run isolation: packaging succeeded, Actions artifact
uploads were skipped, Docker Hub login and pushes were skipped, publish jobs
were skipped, and the workflow uploaded zero artifacts.

The repeat run completed both ARM builds and still has zero uploaded artifacts,
but its x86 jobs have spent more than 45 minutes waiting for GitHub to allocate
64-core VMs. No x86 build has started, so this delay is runner capacity rather
than compilation time.

A local Docker/BuildKit executable is unavailable, so the new release target
was not compiled locally. The PR remains in draft.

## Changelog

No changelog fragment is required because this PR only changes CI and Docker
build configuration. The PR is labeled `C-exclude-from-changelog`.

### Follow-up Work

- Evaluate 64-core larger-runner cold-start availability if repeat runs
  continue to wait substantially longer than the saved compilation time.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -197,7 +197,7 @@ jobs:
       matrix:
         include:
           - name: linux-x86_64
-            runner: ${{ vars.DOCKER_BUILD_RUNNER_AMD64 || 'ubuntu-latest' }}
+            runner: zakura-release-x64-64-core
             platform: linux/amd64
             target_triple: x86_64-unknown-linux-gnu
           - name: linux-aarch64
@@ -225,7 +225,7 @@ jobs:
           short_sha="$(git rev-parse --short=12 HEAD)"
           docker buildx build \
             --platform "${{ matrix.platform }}" \
-            --target release \
+            --target release-zakurad \
             --file ./docker/Dockerfile \
             --build-arg "FEATURES=${FEATURES}" \
             --build-arg "SHORT_SHA=${short_sha}" \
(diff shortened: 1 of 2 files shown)


ValarDragon:
Superceded by @czarcas7ic 's work

PR #382, opened 2026-07-22
fix(chain): reject NU7 activation without a branch ID

czarcas7ic:
## Motivation

A production build can accept an explicitly configured NU7 activation even though NU7 has no production branch ID. This is not a high-severity or remotely exploitable issue: stock Mainnet and Testnet never activate NU7, and only an operator configuring a custom Testnet or Regtest can reach it. It is still worth fixing as a correctness issue because the accepted configuration fails later in consensus code.

## Solution

Reject configured post-Overwinter upgrades without branch IDs and cover the production feature behavior with a regression test. Configured Regtest deserialization now returns validation errors instead of panicking.

## Testing

- `cargo test -p zakura-chain --no-default-features`
- `cargo test -p zakura-network config::tests::vectors`
- `cargo clippy -p zakura-chain --all-targets --no-default-features -- -D warnings`
- `cargo check -p zakura --features default-release-binaries --locked`

## Changelog

Added `changelog-unreleased/382.md`.

AI disclosure: Codex was used to investigate ZCA-800 and implement the focused guard and regression tests.

### DIFF
--- a/changelog-unreleased/382.md
+++ b/changelog-unreleased/382.md
@@ -1,0 +1,5 @@
+## Fixed
+
+- Reject custom Testnet and Regtest activation heights for network upgrades
+  without consensus branch IDs instead of accepting an unusable configuration
+  ([#382](https://github.com/zakura-core/zakura/pull/382)).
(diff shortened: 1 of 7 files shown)


czarcas7ic:
@codex review

PR #384, opened 2026-07-22
fix(rpc): reject mismatched indexer block hashes

czarcas7ic:
## Motivation

Indexer RPC responses carry a serialized block and an advertised hash independently. `BlockAndHash::decode` accepted both without checking that the hash identified the decoded block. This is correctness hardening for an explicitly trusted synchronization channel, not a high severity issue in the stock node configuration.

## Solution

Recompute the decoded block hash at the RPC boundary and reject responses whose advertised hash differs. Both the streamed-block and finalized-gap paths already use this decoder. This supersedes the ZCA-795 portion of #326 without taking on the broader synchronization and consensus revalidation changes in that draft.

## Changelog

Added `changelog-unreleased/384.md` under Fixed.

### Follow-up Work

The additional trusted-sync hardening proposed in #326 remains separate from this transport invariant.

### DIFF
--- a/zakura-rpc/src/indexer.rs
+++ b/zakura-rpc/src/indexer.rs
@@ -64,8 +64,11 @@ impl BlockAndHash {
     }
 
     /// Try to convert a [`BlockAndHash`] into a tuple of a decoded block and hash.
+    ///
+    /// Returns `None` if the advertised hash does not match the decoded block.
     pub fn decode(self) -> Option<(block::Block, block::Hash)> {
-        self.hash
+        let advertised_hash = self
+            .hash
             .try_into()
             .map(|bytes| block::Hash::from_bytes_in_display_order(&bytes))
             .map_err(|bytes: Vec<_>| {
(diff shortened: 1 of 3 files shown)


zakura-rpc/src/indexer.rs
```diff
@@ -74,13 +77,23 @@ impl BlockAndHash {
...
-                    .ok()
-                    .zip(Some(hash))
-            })
+            .ok()?;
+        let block: block::Block = self
+            .data
+            .zcash_deserialize_into()
+            .map_err(|err| tracing::warn!(?err, "failed to deserialize block"))
+            .ok()?;
+        let computed_hash = block.hash();
```
evan-forbes, line 86: do we really want to recompute the hash over the block every rpc call? if this is a trusted channel, then can we rely on the reader to check that?

evan-forbes (commented):
not blocking 

imo we should leave this to the reader to verfiy 

if we are seeing this then afaiu this means a deeper bug we need to fix

PR #386, opened 2026-07-22
fix(sync): reset successor work after header reanchors

czarcas7ic:
## Motivation

During a reorg, the node may already be processing block N from the old fork when the header chain switches to a different block N. If body progress arrives in the same update, block sync mistakes the fork switch for normal progress and leaves the old block marked as in progress. Since requests are tracked by height, it will not request the replacement block, and sync can remain stuck until another reset clears the stale claim.

## Solution

Only resets that explicitly preserve active successors can now be classified as growth. Header reanchors clear successor work above the verified tip, and a regression test covers a coalesced reanchor with a stale submitted successor.

## Testing

- `cargo test -p zakura-network zakura::block_sync::tests` (182 passed)
- `markdownlint changelog-unreleased/386.md`

The full `cargo test -p zakura-network` run passed 1,017 tests but had four unrelated failures. The handshake failure passed when rerun alone; three unchanged legacy listener tests fail alone on macOS with `AddrNotAvailable` while binding their selected loopback source addresses.

## Changelog

Added `changelog-unreleased/386.md` for the operator-visible sync-stall fix.

### DIFF
--- a/zakura-network/src/zakura/block_sync/sequencer_task.rs
+++ b/zakura-network/src/zakura/block_sync/sequencer_task.rs
@@ -473,12 +473,11 @@ impl SequencerTask {
         );
 
         // State can report a forward `Reset` while checkpoint commits advance
-        // under already-submitted or still-downloading successor bodies. Treat
-        // that as verified growth once it is inside our submitted/downloaded
-        // floor, or when we already have successor work in flight. Keep fork
-        // resets destructive when they are not anchored by active successor
-        // work.
-        if frontiers.verified_block_tip > self.sequencer.verified_tip()
+        // under active successor bodies. Preserve them only when the caller knows
+        // their anchor is unchanged; header reanchors can carry coalesced body
+        // growth while those successors belong to the old fork.
+        if preserve_active_successors
+            && frontiers.verified_block_tip > self.sequencer.verified_tip()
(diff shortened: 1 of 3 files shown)


ValarDragon:
Lets likely leave this one to Evan's refactor

czarcas7ic:
Evan is taking a look and will decide if needed or not.

zakura-network/src/zakura/block_sync/sequencer_task.rs
```diff
@@ -473,12 +473,11 @@ impl SequencerTask {
...
-        // that as verified growth once it is inside our submitted/downloaded
-        // floor, or when we already have successor work in flight. Keep fork
-        // resets destructive when they are not anchored by active successor
-        // work.
-        if frontiers.verified_block_tip > self.sequencer.verified_tip()
+        // under active successor bodies. Preserve them only when the caller knows
+        // their anchor is unchanged; header reanchors can carry coalesced body
+        // growth while those successors belong to the old fork.
+        if preserve_active_successors
+            && frontiers.verified_block_tip > self.sequencer.verified_tip()
```
evan-forbes, line 480: per sync:

we can keep as a draft for now

need to chew on this more

PR #393, opened 2026-07-23
fix(sync): keep reserve dispatch flowing during registry-miss backoff

czarcas7ic:
Fixes: Zellic 3.5

## Motivation

While a required block waits out its registry-miss backoff, legacy chain sync pauses all new reserve dispatch, so one unavailable hash stalls otherwise-available sync work for up to the full retry budget (roughly two minutes) before the round restarts.

## Solution

While any registry-miss retry is pending, limit speculative dispatch to one hash per loop step instead of pausing it. Unrelated reserve hashes keep flowing, and the loop still reaches the `select!` retry timer promptly: dispatching the whole reserve can block on peer-set readiness until in-flight downloads drain, which would starve the timer.

No routing-cause split on `SharedPeerError` is needed: `PeerSet::poll_ready` only returns `Ready` with at least one ready peer, the peer set is only reachable through its `tower::Buffer` worker (which pairs `poll_ready` and `call` atomically per request), and `route_inv`'s failure branch takes no ready service, so a synthetic `NotFoundRegistry` always means every ready peer is marked missing the hash. Peer-set saturation surfaces as readiness backpressure, not as this error, and bounded dispatch is safe under saturation anyway. The pre-existing `pool.route_inv.notfound.no_ready.count` metric stays as a tripwire on that invariant.

## Changelog

Added `changelog-unreleased/393.md` under `Security`.

- [ZCA-831](https://linear.app/zcale/issue/ZCA-831/zellic-35-registry-miss-globally-stalls-sync-dispatch)

### DIFF
--- a/zakurad/src/components/sync.rs
+++ b/zakurad/src/components/sync.rs
@@ -794,11 +794,12 @@ where
     /// Required blocks that registry-missed and are waiting for their backoff to elapse before being
     /// retried, keyed by hash with the [`tokio::time::Instant`] each backoff expires.
     ///
-    /// While this is non-empty, new speculative downloads are gated in [`Self::sync_round`] so the
-    /// in-flight pipeline drains and frees ready-peer slots for the critical head-of-line block.
-    /// The retries themselves are never gated: they fire from the `select!` timer arm so draining
-    /// and tip extension continue concurrently during the backoff. A map so that a second block missing while the first is still
-    /// backing off isn't dropped: every registry-missed required block stays scheduled.
+    /// While this is non-empty, [`Self::sync_round`] limits new speculative dispatch to one hash per
+    /// loop step: unrelated reserve hashes keep flowing (one unavailable hash must not stall the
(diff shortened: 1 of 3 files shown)


zakurad/src/components/sync.rs
```diff
@@ -1339,13 +1340,15 @@ where
                 || (self.downloads.in_flight() >= lookahead_limit / 2
                     && self.past_lookahead_limit_receiver.cloned_watch_data());
 
-            // Head-of-line priority: while a required block is missing from all current peers and
-            // waiting on its registry-miss backoff, pause *new* speculative dispatch so in-flight
-            // downloads drain and free up ready-peer slots. Otherwise lookahead work can keep every
-            // peer busy and starve the critical retry. This is inert in healthy sync.
-            let head_of_line_starved = !self.registry_miss_retry.is_empty();
```
czarcas7ic, line 1346: Tracking why we originally made this change, which is why the new change limits unrelated work to a single dispatch per loop step instead of resuming the full reserve.

PR #394, opened 2026-07-23
fix: check best-tip context before async verification

czarcas7ic:
Fixes: OS-ZCA-ADV-06

## Motivation

Mempool best-tip anchor and nullifier validation currently runs in the same unordered future set as script and shielded authorization checks. Contextual validation can reject the transaction after verification jobs have been submitted. Dropping their waiters does not cancel those jobs, so queued or running verification continues to completion and its results are discarded, consuming shared CPU capacity.

## Solution

Await the mempool-only best-tip check before polling the lazy asynchronous authorization checks. Block verification is unchanged. The regression test now delays a contextual rejection alongside an invalid script and verifies that the contextual error is returned first.

## Changelog

Added `changelog-unreleased/394.md` with the explicit no-changelog marker because this changes internal scheduling without changing accepted transactions or operator configuration.

### DIFF
--- a/changelog-unreleased/394.md
+++ b/changelog-unreleased/394.md
@@ -1,0 +1,4 @@
+<!-- changelog: none -->
+
+This PR hardens internal verification scheduling without changing accepted transactions or
+operator configuration.
--- a/zakura-consensus/src/transaction.rs
+++ b/zakura-consensus/src/transaction.rs
@@ -570,7 +570,7 @@ where
 
             tracing::trace!(?tx_id, "got state UTXOs");
 
-            let mut async_checks = match tx.as_ref() {
+            let async_checks = match tx.as_ref() {
                 Transaction::V1 { .. } | Transaction::V2 { .. } | Transaction::V3 { .. } => {
                     tracing::debug!(?tx, "got transaction with wrong version");
                     return Err(TransactionError::WrongVersion);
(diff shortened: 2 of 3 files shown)


czarcas7ic:
Neither of the above V12 reports should be actioned.

czarcas7ic:
I had AI benchmark this PR, these were the results

---

The performance tradeoff is approximately **15 µs of added latency for accepted shielded mempool transactions**, in exchange for avoiding cryptographic verification after a transaction has already failed the state check.

Release-build measurements used Zakura’s production `StateService` and RocksDB path, with the required anchor present and one million unrelated nullifiers in each relevant index.

| Accepted workload | Before | Added latency | Relative change |
|---|---:|---:|---:|
| Minimal transparent transaction | 35.8 µs | 6.7 µs | +18.7% |
| Sapling, 1 spend | 120.661 ms | ~15.3 µs typical, 20.5 µs p95 | +0.013% |
| Orchard, 2 actions | 110.665 ms | ~15.1 µs typical, 20.4 µs p95 | +0.014% |

The shielded overhead is the state-check duration that can no longer overlap proof verification. It is below the noise floor of the roughly 110–121 ms end-to-end shielded verification, so it was measured directly rather than inferred from an artificial delay.

For transactions rejected by the state check, the old code already returned the error without waiting for Rayon. The benefit is avoiding abandoned proof and signature work:

| Workload | Before | After | Change |
|---|---:|---:|---:|
| Return 16 duplicate-nullifier Orchard rejections | 2.05 ms | 2.94 ms | +0.89 ms |
| Next valid Orchard transaction after those rejections | 123.37 ms | 114.00 ms | 9.37 ms faster |
| CPU across 336 rejected Orchard transactions | 11.56 s | 8.43 s | 3.13 CPU-seconds saved |

That is about **9.3 ms of CPU avoided per rejected two-action Orchard transaction**. The next valid Orchard transaction was 7.45–11.29 ms faster at 95% confidence because it no longer waited behind abandoned proof work.

PR #398, opened 2026-07-23
DRAFT: fix(state): handle empty state after root invalidation

czarcas7ic:
Fixes: OS-ZCA-ADV-08

## Motivation

Invalidating the root of the non-finalized state can legitimately remove every live non-finalized chain. The state writer then unconditionally called a channel update helper that requires a best chain and panicked. Zakura unwinds the panic, but the sole state writer exits and later state writes fail until the node restarts.

## Solution

Use an invalidate/reconsider-specific channel update path. If a non-finalized chain remains, it uses the existing update helper. If the state is empty, it publishes that empty state and switches the shared chain tip back to the finalized database tip.

Normal block commits still use the original helper unchanged. The finalized database is read only after an invalidate or reconsider request leaves no non-finalized chain.

## Changelog

Added `changelog-unreleased/398.md` under `Security`.

- [ZCA-809](https://linear.app/zcale/issue/ZCA-809/os-zca-adv-08-root-invalidation-panic-in-state-writer)

### DIFF
--- a/changelog-unreleased/398.md
+++ b/changelog-unreleased/398.md
@@ -1,0 +1,4 @@
+## Security
+
+- Prevented root invalidation from stopping state writes when the non-finalized chain becomes empty
+  ([#398](https://github.com/zakura-core/zakura/pull/398)).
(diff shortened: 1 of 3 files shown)


czarcas7ic:
Do not review yet, in progress

PR #399, opened 2026-07-23
ci: move mempool-load config localization and seeding into Kresko

p0mvn:
## What

Follow-up to #349 (Evan's review point): move the ~350 lines of Python that rewrite Kresko's generated node configs out of the mempool-load harness and into Kresko itself.

The harness used to rewrite each generated node config in Python (`prepare_node_dirs` + the `set_toml_values`/`set_toml_arrays` line rewriters) and submit the genesis/premine blocks itself (`seed_node`/`submit_block`). That logic now lives in Kresko as two subcommands (valargroup/kresko#9):

- `kresko localize-fleet` — per-node binds, directories, metrics endpoint, internal-miner flag, emptied public seed peers, regenerated loopback peer list.
- `kresko seed-local` — the `submitblock` genesis/premine loop with duplicate/rejected retry and height idempotence.

The lab still owns node process lifecycle (spawn/`up`/`down`, port preflight, peering and convergence gates).

## Root cause / why this is safe

Pure refactor, no behavior change. The rendered per-node configs are **byte-identical** to the former Python output: Kresko keeps the rewrite line-oriented (not a `toml::Value` round-trip) precisely so the bytes match. Verified locally by generating a genesis payload, running the pre-refactor `prepare_node_dirs` and the new `kresko localize-fleet` against it, and diffing — all 6 configs (running + bootstrap × 3 nodes) identical.

## Tests

- Removed the now-dead `ConfigRewriting` tests and the `GENERATED_CONFIG` fixture; the equivalent assertions moved to Kresko's `zebra_config::local_fleet_tests`. Retained tests (verdict, propagation, secret safety, teardown, peering) unchanged — `ci-python-tests` green (99 tests).
- `kresko_ref` stays `main` (kept in step with the bake), so this rides Kresko main once valargroup/kresko#9 merges.

Depends on: valargroup/kresko#9.


evan-forbes (approved):
lfg net negative

PR #410, opened 2026-07-23
fix(chain): reject noncanonical Orchard and Ironwood proofs

ValarDragon:
## Motivation

Canonical Orchard proof sizes were never unenforced on deployed mainnet.
Keeping a pre-NU6.2 exception in the parser and semantic state machine therefore
added an unnecessary consensus branch and left malformed transactions alive
longer than needed.

Strict parsing moves the rejection to the wire boundary. Because that happens
before consensus and mempool attribution, the network path must preserve the
existing peer-ban outcome for this fault.

## Solution

- Reject noncanonical Orchard proof sizes while parsing whole V5 transactions.
- Reject noncanonical Orchard and Ironwood proof sizes while parsing V6
  transactions.
- Remove the NU6.2/NU6.3 activation conditional and enforce canonical sizes
  unconditionally in semantic verification for defense in depth and
  transactions constructed in memory.
- Classify this serialization fault and assign peers the existing ban-threshold
  misbehavior score when they send it; unrelated parser errors remain ordinary
  connection failures.
- Preserve the official ZIP-244 digest coverage while identifying synthetic
  vectors with noncanonical proof lengths.
- Keep generated state-test transactions canonical after their action lists are
  changed.
- Bump `zakura-chain` to `4.0.0-rc1` and `zakura-consensus` to `6.0.0-rc1`,
  updating workspace dependency requirements and the lockfile.

## API changes

This PR intentionally:

- removes `Network::orchard_canonical_proof_size_rule_active`;
- narrows `shielded_proof_size_is_canonical` to `pub(super)` and removes its
  height/network parameters; and
- adds `SerializationError::NonCanonicalShieldedProofSize` so the network
  layer can preserve peer attribution after strict parsing.

## Testing

- [x] `cargo metadata --locked --format-version 1`
- [x] `./scripts/changelog.py check`
- [x] `git diff --check`
- [ ] Build, lint, and test suites (delegated to CI)

Regression coverage includes canonical, one-byte-short, and one-byte-padded
proofs for V5 Orchard and V6 Orchard/Ironwood transactions, plus a network test
that checks this serialization fault receives the ban-threshold score while an
ordinary parse error does not.

## Changelog

Updated `docs/changelog/unreleased/410.md` under `Fixed`.

### Specification

- [ZIP 257](https://zips.z.cash/zip-0257)

AI disclosure: Codex was used for investigation, implementation, tests, and PR
preparation.


ValarDragon:
Look into why this is so many loc addition, and needs checkpoint sync test to ensure no old oversized proofs

ebfull:
Heads-up: changelog fragments moved from the root `changelog-unreleased/` directory to `docs/changelog/unreleased/` in #520, and this PR still writes its fragment to the old location (`changelog-unreleased/410.md`). When the PR merges, that recreates the deleted root directory and the fragment gets stranded where release assembly won't pick it up (this already happened once with #513, fixed in #527).

ValarDragon:
Evan tested this is safe:

- Heights scanned: 1,744,600 (1,687,104–3,431,703)
  - Orchard bundles: 1,211,332
  - Orchard actions: 50,278,060
  - Canonical proofs: 1,211,332
  - Oversized/undersized proofs: 0 / 0
  - Gaps, final-shard overlaps, RPC errors, and parse errors: 0

ValarDragon:
@v12sec pls audit

evan-forbes (approved):
tested manually and we're clear

PR #417, opened 2026-07-23
perf(consensus): flush crypto batches at block boundary

czarcas7ic:
## Motivation

Semantic block verification can queue signature and proof work, then wait for
the global `MAX_BATCH_LATENCY` of 100 ms before a partial crypto batch starts.
Lowering that timer globally would also change mempool batching, so block
execution needs a targeted boundary signal.

## Solution

Add a bounded, non-blocking explicit flush command, `Batch::try_flush`, and
expose the primary service through `Fallback`. Each semantic block registers
its transaction count, and the final transaction to reach its
asynchronous-check boundary queues best-effort flushes for the initialized
Ed25519, Sapling, and Orchard batch services. Verification remains unchanged.
Every transaction future is still awaited, and the existing size and latency
triggers remain as fallback paths.

Flush commands use the existing semaphore capacity. Batch bookkeeping treats
every queued item as having a weight of at least one, so a zero-weight item
cannot be mistaken for an empty batch. Cancellation removes the block
registration before its pointer identity can be reused.

## Benchmarks

Semantic-verification A/B on the bench runner (this branch `12794b6e0` vs merge
base `ad49535c8`), syncing real mainnet blocks from height 1707210 with
`consensus.checkpoint_sync = false` (everything past the mandatory checkpoints
gets full script+proof verification), DNS-seeder peers:

- `full_verify_concurrency_limit = 1` (approximates near-tip block-at-a-time
  verification), 1000 blocks: 1.04 → 1.24 blocks/s overall (**1.19×**); steady
  state 681 → 591 ms/block, recovering ~90 of the up-to-100 ms
  `MAX_BATCH_LATENCY` wait per block. Run:
  https://github.com/zakura-core/zakura/actions/runs/30039304446
- `full_verify_concurrency_limit = 20` (concurrent catch-up), 3000 blocks:
  4.22 → 4.19 blocks/s post-first-commit (−0.7%, well inside the lane's 8.7%
  noise band) — no regression from per-block flushes shortening batches shared
  across concurrently verified blocks. Run:
  https://github.com/zakura-core/zakura/actions/runs/30038568453

Both runs' red status is a pre-existing bench teardown bug (`set -e` in the
EXIT trap; fix drafted separately) — all four legs completed cleanly and the
data is complete.

## Changelog

Added `changelog-unreleased/417.md`.

No consensus rule changes. This ports the previously explored block-boundary
batch-flush design onto current Zakura `main`.

### Follow-up Work

Concurrent semantic catch-up is benchmarked above (neutral at concurrency 20).
Remaining: a tip-following canary A/B on the fleet to measure real
block-arrival-to-commit latency, and optionally a Prometheus counter for
explicit flushes, since release builds compile out the trace-level log line
that would otherwise confirm the path in production.

### AI Disclosure

OpenAI Codex was used to port and harden the implementation, add tests, run
validation, and prepare this description.

### DIFF
--- a/changelog-unreleased/417.md
+++ b/changelog-unreleased/417.md
@@ -1,0 +1,6 @@
+## Changed
+
+- Flush partial cryptographic verification batches at semantic block boundaries,
+  avoiding the global 100 ms batch delay during block execution when the
+  boundary signal succeeds
+  ([#417](https://github.com/zakura-core/zakura/pull/417)).
--- a/tower-batch-control/src/lib.rs
+++ b/tower-batch-control/src/lib.rs
@@ -125,6 +125,9 @@ pub use self::service::Batch;
 pub trait RequestWeight {
     /// Returns the weight of a request relative to the maximum threshold for flushing
     /// requests to the underlying service.
+    ///
+    /// Batch bookkeeping treats every queued item as having a weight of at least 1,
+    /// so zero-weight items still start the batch timer and are flushed.
     fn request_weight(&self) -> usize {
         1
     }
--- a/tower-batch-control/src/message.rs
+++ b/tower-batch-control/src/message.rs
@@ -4,13 +4,22 @@ use tokio::sync::{oneshot, OwnedSemaphorePermit};
 
 use super::error::ServiceError;
 
-/// Message sent to the batch worker
+/// Message sent to the batch worker.
 #[derive(Debug)]
-pub(crate) struct Message<Request, Fut> {
-    pub(crate) request: Request,
-    pub(crate) tx: Tx<Fut>,
-    pub(crate) span: tracing::Span,
-    pub(super) _permit: OwnedSemaphorePermit,
+pub(crate) enum Message<Request, Fut> {
+    /// A batch item request.
+    Item {
+        request: Request,
+        tx: Tx<Fut>,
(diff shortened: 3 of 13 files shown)


czarcas7ic:
This branch vs merge base `ad49535c8`: near-tip block-at-a-time verification (set concurrency to 1) improves from 681 to 591 ms per block (1.19x overall), recovering about 90 of the up-to-100 ms `MAX_BATCH_LATENCY` wait.

This was done with checkpoint sync set to false so each block gets full script + proof verification.

```mermaid
---
config:
  themeVariables:
    xyChart:
      plotColorPalette: "#3080de, #888888"
---
xychart-beta
  title "Near-tip, concurrency 1 (lower is better)"
  x-axis ["baseline", "PR 417"]
  y-axis "ms per block (steady state)" 0 --> 750
  bar [681, 591]
  line [581, 581]
```

Runs:

https://github.com/zakura-core/zakura/actions/runs/30039304446
https://github.com/zakura-core/zakura/actions/runs/30038568453

czarcas7ic:
@codex review

ValarDragon (approved):

ValarDragon:
@v12-auditor review pls

PR #419, opened 2026-07-23
ci(perf): add ephemeral mainnet profiling workflow

czarcas7ic:
## Motivation

We need repeatable CPU profiles for historical sync and live mainnet head without a shared benchmark host.

## Solution

Adds ephemeral DigitalOcean historical A/B and live-head profiles. The existing fixed-host bench also gains optional semantic verification and configurable full-verification concurrency.

## Testing

Historical A/A, semantic A/B, and 5-, 10-, and 60-minute live-head runs passed. All CI checks and 36 script tests pass.

## Changelog

No fragment; CI, scripts, and docs only.


czarcas7ic:
@codex review

.github/workflows/zakura-perf-bench.yml
```diff
@@ -0,0 +1,397 @@
...
+          path: legs/
+
+      - name: Compare legs
+        run: |
+          set -euo pipefail
+          P="legs/zakura-perf-bench-${GITHUB_RUN_ID}-primary"
+          B="legs/zakura-perf-bench-${GITHUB_RUN_ID}-baseline"
+          [ -f "$P/meta.json" ] && [ -f "$B/meta.json" ] \
+            || { echo "one or both legs produced no meta.json; nothing to compare" >> "$GITHUB_STEP_SUMMARY"; exit 0; }
+          FAILED_LEGS=$(python3 - "$P/meta.json" "$B/meta.json" <<'PY'
```
evan-forbes, line 352: [not blocking]

I've never seen python be executed like this within a yaml file lmao

I am both impressed and horrified

do we need to hear or could we clean this up?

.github/workflows/zakura-perf-bench.yml
```diff
@@ -0,0 +1,397 @@
...
+        env:
+          LEG: ${{ matrix.leg.leg }}
+          DROPLET_SIZE: ${{ inputs.droplet_size || 'c-16' }}
+          SSH_FINGERPRINT: ${{ vars.DO_SSH_KEY_FINGERPRINT }}
+          IMAGE_ID: ${{ steps.baked.outputs.image_id }}
+          VOL_SNAP_ID: ${{ steps.baked.outputs.vol_snap_id }}
+        run: |
+          set -euo pipefail
+          VOLUME_NAME="zakura-pr-perf-vol-${GITHUB_RUN_ID}-${LEG}"
+          SIZE=$(doctl compute snapshot get "${VOL_SNAP_ID}" --format MinDiskSize --no-header)
```
evan-forbes, line 200: Not blocking

Don't we already have code to spin up droplets somewhere from ci? Would it make sense to unify those? Presumably something nice we can easily reuse

Same with the above code

czarcas7ic: So there are four other workflows that copy this similar set up but do different things. It 100 percent can be abstracted out but given this PR has been tested extensively and this is just isolated CI, I would prefer not to do this atm. If strong feelings though, I can for sure get it done here.

evan-forbes (approved):

PR #425, opened 2026-07-24
perf(chain): defer Orchard and Ironwood point decompression

czarcas7ic:
## Motivation

Checkpoint verification does not use Orchard or Ironwood `cv` and `epk` points, so decompressing them during parsing wastes CPU.

## Solution

Keep their 32-byte encodings and validate them before state access on the semantic path. Wire bytes and consensus acceptance stay unchanged. This is a breaking public `zakura-chain` representation change covered by the current 4.0.0 prerelease line.

## Testing

- `cargo test -p zakura-chain --locked`
- `cargo test -p zakura-consensus orchard_and_ironwood_invalid_points_are_rejected_early --locked -- --nocapture`
- `cargo clippy -p zakura-chain -p zakura-consensus --all-targets --locked -- -D warnings`

## Changelog

`docs/changelog/unreleased/425.md`

AI disclosure: Codex was used for implementation, tests, conflict resolution, and PR drafting.


ebfull:
Heads-up: changelog fragments moved from the root `changelog-unreleased/` directory to `docs/changelog/unreleased/` in #520, and this PR still writes its fragment to the old location (`changelog-unreleased/425.md`). When the PR merges, that recreates the deleted root directory and the fragment gets stranded where release assembly won't pick it up (this already happened once with #513, fixed in #527).

ValarDragon:
Closing for now, lets re-open if it comes up in sync or when @czarcas7ic is free enough to take to completion. We keep assigning him not-Zakura things 😂

PR #427, opened 2026-07-24
fix(sync): require durable header anchors

evan-forbes:
## Motivation

A dual-stack mainnet node stalled for ten minutes at height 3422947. Legacy
gossip committed a full block into the non-finalized chain, and header sync
advanced its in-memory frontier to that block. Every forward header range it
then built was anchored at that non-finalized hash — but `CommitHeaderRange`
resolves anchors only against durable storage, so the anchor came back
`UnknownAnchor`. Six honest peers absorbed 2,764 of those local rejections and
were misclassified as supplying invalid ranges.

The same mismatch appears after restart, when restored non-finalized bodies are
ahead of the durable header view.

## Why a downloaded block is not automatically a usable anchor

Having a full block means we have its header, but a `CommitHeaderRange` anchor
needs more than possession of one header: it must be a row in a **contiguous,
linked, difficulty-validatable** durable header chain. The commit walks backward
from the anchor to assemble the PoW difficulty context (`recent_header_context`)
and rejects any gap or broken linkage, so a valid anchor at height H requires
H-1, H-2, … to also be present and linked in durable storage.

The dual-stack node already tries to make committed blocks anchorable:
`seed_zakura_header_from_committed_block` writes a committed block's header into
the header store. But seeding **deliberately refuses to seed across a gap** — a
block whose parent is not the stored row below it — to keep the store linked, and
relies on header-range sync to later deliver the missing rows. That recovery only
works if header sync anchors at the durable tip, *below* the gap, and fetches
forward to fill it.

The stall was the opposite ordering: header sync anchored *above* the gap, at the
non-finalized body tip, so the only component that could deliver the linking rows
was itself stuck failing `UnknownAnchor` on the far side of the gap. The
non-finalized tip leaked into the frontier two ways — the `BestHeaderTip` read
took the already-durable `db.best_header_tip()` and folded the non-finalized
best-chain tip back on top, and the reactor advanced its frontier to each
committed block's own height regardless of whether that height was anchorable.

## Solution

- Make the best-header-tip read durable-only: return `db.best_header_tip()`
  (finalized tip ∨ header frontier) without folding in the non-finalized
  best-chain tip. Header sync now anchors only where `CommitHeaderRange` can
  resolve, restoring the assumption the seeding path already relies on.
- Keep verified-body progress separate from durable header progress. A committed
  full block is still published for dedup/body-frontier purposes, but it advances
  header work only after a durable tip read confirms the block is anchorable — it
  advances to the durable tip, not to the raw committed height. In the no-gap case
  that durable tip *is* the committed height (seeding succeeded); when a gap
  exists, header sync stays below it and fetches the missing rows.
- Initialize and refresh header sync from that same durable tip.
- Preserve in-flight v2 repair across the legacy/v2 race, allowing a response
  containing H and H+1 to fill a missing durable H anchor.
- Give `UnknownAnchor` a dedicated local, non-scoring recovery path: it is never
  attributed to the peer, because our own frontier chose the anchor. Reload and
  reanchor to the durable frontier, clear stale work, and retry with bounded
  exponential backoff.
- Use the same durable reload after verified-chain resets and multi-peer stale
  anchor evidence. Other local commit failures use delayed retries instead of a
  hot loop.
- Bound durable state reads with a timeout and stop mirror/driver delivery
  cleanly when the header reactor has exited.

This makes the frontier used to construct every forward range agree with the
state view that resolves its anchor. Verified body progress can run ahead without
turning a not-yet-linked body hash into a speculative header anchor.

## Testing

- `git diff --check`
- `CXXFLAGS="-include cstdint" cargo check -p zakura-state -p zakura-network -p zakura --all-targets`
- `CXXFLAGS="-include cstdint" cargo clippy -p zakura-state -p zakura-network -p zakura --all-targets -- -D warnings`
- `cargo test -p zakura-network zakura::header_sync::tests` (176 passed)
- `CXXFLAGS="-include cstdint" cargo test -p zakura zakura_header_sync_driver_tests` (47 passed)
- `CXXFLAGS="-include cstdint" cargo test -p zakura-state header_range_reads_include_non_finalized_best_chain_blocks`

The regressions cover:

- a real `ReadStateService` where ordinary header reads see restored
  non-finalized bodies while durable-only reads do not;
- startup with the verified body frontier ahead of the durable header frontier;
- a full block at H racing an in-flight v2 H..H+1 response without cancelling
  the repair or changing its H-1 anchor;
- `UnknownAnchor` causing no peer action, an authoritative durable reload, and a
  delayed retry;
- stale-link recovery reanchoring to durable state rather than the verified body
  tip.

## Changelog

Added `changelog-unreleased/427.md`.

Based on a production dual-stack stall at height 3422947 in which six peers
received 2,764 local `UnknownAnchor` rejections until legacy fallback recovered.


evan-forbes:
v12 was fixed as well already

ValarDragon (approved):

PR #428, opened 2026-07-24
fix(network): shorten misbehavior batching interval

p0mvn:
## Motivation

The original Zebra misbehavior-scoring PR ([ZcashFoundation/zebra#9201](https://github.com/ZcashFoundation/zebra/pull/9201)) states that Zebra should drop connections to, and avoid reconnecting with, peers that advertise blocks or mempool transactions which fail semantic validation.

That PR batched score updates to stop repeated invalid payloads from turning every report into address-book queue and mutex work. However, the interval was not part of the enforcement policy: it was initially one minute, later changed to 30 seconds, and the author noted during review that the exact interval was not important unless finer-grained scores were added.

Current ban-eligible validation failures contribute the full threshold score. The 30-second interval therefore does not provide meaningful score accumulation or tolerance; it intentionally delays when the threshold score reaches the address book and becomes visible through the shared ban list.

## Solution

Reduce the batch interval to one second while preserving per-peer coalescing. This keeps the original protection against address-book queue and mutex pressure while making threshold bans visible to the inbound listener and peer set within one second.

Extract the batch task into a testable helper and use a delayed first tick so every report receives a predictable one-second maximum batching window.

This PR does not add a ban notification or independent disconnection deadline. Existing connections are still dropped when `PeerSet` is next polled; this change only bounds the intentional score-application delay.

## Testing

- `cargo test -p zakura-network --lib misbehavior_reports_are_batched_for_one_second -- --nocapture` — verifies reports stay buffered before one second and are coalesced into one threshold update at the tick.
- `cargo test -p zakura-network --lib misbehavior_ban -- --nocapture` — verifies a threshold update bans the IP and safely removes its addresses.
- `cargo test -p zakura --test acceptance disconnects_from_misbehaving_peers -- --ignored --nocapture` — verifies an invalid-PoW peer is eventually disconnected through the existing peer-set polling path.
- `cargo clippy -p zakura --test acceptance -- -D warnings`

## Changelog

- Added `changelog-unreleased/428.md`.
- Recorded `MISBEHAVIOR_BATCH_INTERVAL` in `CHANGELOG_PARAMS.md`.

- Original misbehavior scoring and ban design: [ZcashFoundation/zebra#9201](https://github.com/ZcashFoundation/zebra/pull/9201)

### Follow-up Work

- Make disconnection event-driven by waking `PeerSet` when a new ban is applied and actively cancelling matching ready and in-flight peer services.


evan-forbes:
Pushed 4f623cc3e to change the batching interval from 1s to 5s, per the Slack feedback.

evan-forbes (approved):

ValarDragon (approved):
Awesome, happy to merge @5s once we fix conflict

ValarDragon:
Fixing conflicts

PR #431, opened 2026-07-24
perf(network): index address book updates

ValarDragon:
## Motivation

Address-book updates currently use `ordered-map`, which maintains a sorted
`Vec` alongside a `HashMap`. Inserting, replacing, and removing peers therefore
shift up to the full address-book capacity. Removing every address for a banned
IP also relies on scanning the priority order.

## Solution

Replace `ordered-map` with a private address-book index that maintains:

- a `HashMap` for direct address lookup,
- a `BTreeSet` preserving the existing `MetaAddr` connection-attempt order, and
- a per-IP map for targeted removal when an IP is banned.

Property tests check that all indexes remain consistent after generated update
sequences. The existing ban regression test now verifies that every port on a
banned IP is removed while unrelated IPs remain.

## Benchmarks

Criterion benchmarks exercise the real public `AddressBook` paths with 16 and
5,000 entries. Each dataset contains 16 addresses on the IP used by the ban
case. The update case applies a valid below-threshold misbehavior update; the
5,000-entry insertion also exercises capacity eviction.

### 16 entries

| Operation | `ordered-map` | Indexed | Change |
| --- | ---: | ---: | ---: |
| Clone and drop | 257 ns | 333 ns | 1.30x slower |
| Iterate in priority order | 138 ns | 16.2 ns | 8.56x faster |
| Existing-peer lookup | 156 ns | 26.0 ns | 5.98x faster |
| Existing-peer update | 1.40 us | 595 ns | 2.35x faster |
| Insert peer | 1.24 us | 477 ns | 2.60x faster |
| Ban IP with 16 addresses | 1.19 us | 1.52 us | 1.28x slower |

### 5,000 entries

| Operation | `ordered-map` | Indexed | Change |
| --- | ---: | ---: | ---: |
| Clone and drop | 51.0 us | 200 us | 3.92x slower |
| Iterate in priority order | 53.5 us | 9.92 us | 5.40x faster |
| Existing-peer lookup | 35.8 us | 26.2 ns | 1,366x faster |
| Existing-peer update | 365 us | 90.5 us | 4.03x faster |
| Insert peer and evict | 341 us | 91.0 us | 3.74x faster |
| Ban IP with 16 addresses | 112 us | 6.18 us | 18.05x faster |

Measurements used optimized builds, 50 samples, and three-second measurement
windows, with `origin/main` and this branch run back-to-back on the same Apple
Silicon host (macOS 26.2, rustc 1.93.0). Run the added benchmark with:

```console
cargo bench -p zakura-network --features internal-bench --bench address_book -- --noplot
```

## Testing

- `cargo clippy -p zakura-network --features internal-bench --bench address_book
  -- -D warnings`
- `cargo test -p zakura-network --lib address_book` (11 passed)
- `cargo test -p zakura-network --lib` (1,130 passed, 3 host-environment
  failures, 4 ignored)

The three full-suite failures are unrelated listener tests that fail before
exercising network logic because macOS returns `AddrNotAvailable` when they
bind their requested loopback source addresses.

## Changelog

`changelog-unreleased/431.md` records this as an internal-only change.

No protocol or production API behavior changes. The `internal-bench` feature
adds a doc-hidden public clone helper that gives benchmark samples independent
ban lists.

### Follow-up Work

None.

---

> [!NOTE]
> **Medium Risk**
> Touches peer selection and ban cleanup in networking code with no intended API change, but incorrect index invariants could affect outbound peer choice or leave banned addresses behind.
> 
> **Overview**
> Replaces the **`ordered-map`**-backed peer store with a private **`AddressBookIndex`** that keeps a **`HashMap`** for lookups, a **`BTreeSet<MetaAddr>`** for the same connection-attempt priority order, and a **per-IP index** so banning an IP drops every listener on that IP without scanning the full book.
> 
> **`AddressBook::get`** no longer remove/reinserts on every lookup; updates, iteration, surplus eviction, and clone paths all go through the new index. Property and vector tests call **`assert_consistent()`** after updates; the misbehavior-ban regression now expects **all ports on the banned IP** removed.
> 
> The **`ordered-map`** workspace dependency is dropped. A Criterion **`address_book`** bench ( **`internal-bench`** ) and a doc-hidden **`clone_with_fresh_bans_for_benchmark`** helper exercise clone, iterate, lookup, update, insert, and ban at 16 and max capacity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aec659c1f33dc044c8ac66e63c43c088baa19d94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


czarcas7ic:
@codex review

PR #432, opened 2026-07-24
fix(state): retain VCT checkpoint header witness

evan-forbes:
## Motivation

VCT checkpoint sync can authenticate roots through the block immediately below the final checkpoint while using the checkpoint header as the terminal one-block-lag witness. The authenticated-root frontier persisted only the confirmed prefix and discarded that witness. Committing the final pre-checkpoint block then could not recover the checkpoint block `auth_data_root`, so the state writer retried forever while later blocks queued behind it.

This was reproduced on `temp-zakura-sync-test-2`: checkpoint `3,418,406`, stalled commit `3,418,405`.

Databases that had already advanced to the affected frontier also needed an upgrade path; simply changing fresh writes would leave them stalled after restart.

## Solution

Retain a compact `HeaderWitness { height, hash, auth_data_root }` in the durable header-root authentication frontier. The witness is written atomically with the confirmed roots and history tree. Its note-commitment roots remain unconfirmed, are not stored in the authoritative root index, and are not served to peers.

VCT successor lookup uses the canonical witness only when an authoritative successor root row is absent. Existing root gossip and checkpoint cutoffs remain unchanged; no `C+1` data is requested or retained.

Database format 28.0.3 now runs a real migration. It rebases header-root authentication to the finalized body tip and removes header-ahead root rows while preserving canonical headers. Normal root authentication then refetches the required range and durably reconstructs the terminal witness. For the incident shape, body tip `C-2` refetches `[C-1, C]`, promotes only `C-1`, and retains `HeaderWitness(C)`.

## Testing

- `cargo check --workspace --locked`
- `cargo clippy -p zakura-chain -p zakura-state --all-targets --locked -- -D warnings`
- `cargo test -p zakura-chain --locked`: 372 passed plus 9 doc tests
- `cargo test -p zakura-state --locked`: 451 passed, 4 existing ignored, plus binary, integration, and doc tests
- `./scripts/changelog.py check`: 33 fragments validated
- `markdownlint changelog-unreleased/432.md --config .trunk/configs/.markdownlint.yaml`

Regression coverage verifies compact v2 encoding and malformed data rejection, legacy v1 decoding, persistence across database reopen, canonical header-only successor lookup without the checkpoint body or authoritative checkpoint root row, atomic storage failure behavior, and recovery of an affected `C-2`/`C-1`/`C` database through the 28.0.3 migration.

Local state builds used `CXX=clang++ CXXFLAGS=-include cstdint` because the host C++ standard library no longer supplies the RocksDB includes transitively.

## Changelog

- Added `changelog-unreleased/432.md`.

- Incident artifacts: `/home/evan/src/valar/art/debug/v2-fails-2/DIAGNOSIS.md`


zakura-chain/src/parallel/commitment_aux_verify.rs
```diff
@@ -23,9 +23,44 @@ use crate::{
 pub struct VerifiedHeaderCommitmentRoots {
     confirmed_roots: Vec<BlockCommitmentRoots>,
     confirmed_hashes: Vec<block::Hash>,
+    header_witness: Option<HeaderWitness>,
```
ValarDragon, line 30: Lets add a comment here for when this option is expected to be Some(), its pretty unclear coming in why it'd be needed ever!

evan-forbes: damn codex didn't push the commit and I didn't check apologies https://github.com/zakura-core/zakura/pull/432/commits/55c7507394aaba504981231ede94f0ec67ff98c3

ValarDragon (approved):
LGTM! Way more edge cases at this boundary than I expected 🙏

PR #433, opened 2026-07-24
build: bump zcash_primitives to 0.30.0

czarcas7ic:
Update Zakura's librustzcash dependency family to the published releases aligned with zcash_primitives 0.30.0. zcash_keys 0.16.0 now uses zcash_transparent 0.10.0, so this removes the temporary git patch and its cargo-vet and cargo-deny accommodations while retaining a single TransparentAddress type.

This is dependency-only and has no operator-visible effect.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1904,8 +1904,7 @@ dependencies = [
 [[package]]
 name = "equihash"
 version = "0.3.0"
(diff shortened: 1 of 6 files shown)


czarcas7ic:
Need upstream to tag zcash_keys

czarcas7ic:
Done

evan-forbes (approved):

PR #439, opened 2026-07-25
fix(network): Fix tests for removing all addresses for banned IPs

czarcas7ic:
## Motivation

Banning an IP must remove all of its addresses, even when connection priority
separates those entries in the address book.

## Solution

The indexed address book on `main` now groups addresses by IP and removes them
through its per-IP index. This PR preserves regression coverage that separates
same-IP entries by peer priority and verifies that the ban removes both while
leaving an unrelated address intact.

The former direct metrics-publication commit was dropped during the rebase
because #451 and #453 already provide that behavior and regression coverage.

## Testing

- `npx --yes markdownlint-cli changelog-unreleased/439.md --config

The rebased diff only changes the regression test and changelog fragment.
Rust tests were not rerun locally; the force-push restarted CI.

## Changelog

`changelog-unreleased/439.md` records the banned-address fix.

- ZcashFoundation/zebra#10580
- #431
- #451
- #453

### Follow-up Work

None.

### DIFF
--- a/changelog-unreleased/439.md
+++ b/changelog-unreleased/439.md
@@ -1,0 +1,4 @@
+## Fixed
+
+- Prevent banned peer addresses from blocking outbound reconnections
+  ([#439](https://github.com/zakura-core/zakura/pull/439)).
--- a/zakura-network/src/address_book/tests/vectors.rs
+++ b/zakura-network/src/address_book/tests/vectors.rs
@@ -69,11 +69,10 @@ fn gossiped_change(
 /// Regression test for https://github.com/ZcashFoundation/zebra/issues/10580.
 ///
 /// Applying a ban-threshold misbehavior update with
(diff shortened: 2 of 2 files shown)


evan-forbes (approved):
we've discussed this in the past and I think we actually explicitly enabled this in v2 lol

Totally fine with legacy blocking it, I don't buy this makes attacks _that_ much more difficult tbh

or put a cap on the number > 1

ie what if I want to run two zcash nodes from my house? Or many nodes behind the same dns for horizontal scaling?

czarcas7ic:
@evan-forbes and I had an offline convo, his comment above was in reference to a different feature that he mistook this one for

ValarDragon:
Test failure is an underlying race conditon unrelated to this code, fixed in #450

ValarDragon:
After the other data structure work, this became a test-only PR

PR #440, opened 2026-07-25
fix(network): avoid pruned peers during historical block sync

ValarDragon:
## Motivation

During initial sync, the peer set can speculatively route historical block-body requests to connected peers that do not advertise `NODE_NETWORK`. Pruned peers then return repeated `notfound` responses for blocks outside their retention window, delaying bootstrap despite usable full nodes joining later.

## Solution

- retain the handshake service flags on each load-tracked client as a routing capability
- require `NODE_NETWORK` for generic block requests while the local tip is more than 5,000 blocks behind the estimated network tip
- treat pruned peers equally within the 5,000-block live-sync window
- continue routing explicitly advertised blocks to their advertiser regardless of `NODE_NETWORK`
- request more peers when no eligible historical block peer is ready

This adds the internal `pub(crate)` method `LoadTrackedClient::advertises_node_network`; there are no downstream-public API changes.

## Testing

Passed:

- `cargo test -p zakura-network peer_set::set::tests --locked` (29 passed)
- `npx --yes markdownlint-cli@0.48.0 CHANGELOG_PARAMS.md changelog-unreleased/440.md --config .trunk/configs/.markdownlint.yaml`

The full `cargo test -p zakura-network --locked` run completed with 1,143 passed, 4 ignored, and 3 unrelated macOS-host failures. All three failures are listener tests that attempt to bind unavailable loopback source addresses and fail with `AddrNotAvailable`:

- `listener_bans_zcashd_compat_peer_before_reserved_slot`
- `listener_reserves_one_zcashd_compat_inbound_slot`
- `listener_zcashd_compat_reconnect_bypasses_recent_ip_limit`

## Changelog

- Added `changelog-unreleased/440.md`.
- Recorded the new 5,000-block routing threshold in `CHANGELOG_PARAMS.md`.

### Follow-up Work

Consider applying the same capability policy to historical `FindBlocks` requests after validating interoperability with third-party pruned peers.

---

> [!NOTE]
> **Medium Risk**
> Changes live P2P block routing during sync; mis-tuned tip distance or service-flag handling could slow downloads or over-restrict peers, but scope is limited and covered by new tests.
> 
> **Overview**
> **Fixes initial-sync stalls** from speculative `BlocksByHash` traffic to peers that do not advertise `NODE_NETWORK`, which often answer with `notfound` for blocks outside their prune window.
> 
> Peers now expose handshake **`NODE_NETWORK`** via `LoadTrackedClient::advertises_node_network`. **`PeerSet`** uses that when picking peers for generic block downloads: while the local tip is more than **5,000 blocks** behind the estimated network tip (`PRUNED_PEER_BLOCK_ROUTING_MAX_TIP_DISTANCE`), only full-node peers are eligible for P2C and inventory “maybe” routing. Inside that window, pruned peers are treated like any other peer. **Explicit inventory advertisements** still route to the advertising peer even if pruned. When no ready full-node peer exists for a historical block request, the set signals **`MorePeers`** and returns **`NoReadyPeers`** instead of wasting attempts on pruned connections.
> 
> Changelog and parameter docs record the new threshold; unit tests cover historical vs near-tip vs advertised-block cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99b8821530637c8745f0d15438b81e6555dfd8bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


evan-forbes (approved):
makes sense to me

the current mechanism however still feels very broken however

for example, peer exchange seems to be filtering out pruned nodes? surely that's a bug?

even if the mechanism was working correctly, that would mean no one dials pruned nodes. so any two pruned nodes should never talk to each other. That doesn't happen currently but also doesn't seem like the desired behavior.

ValarDragon:
Evan identified another bug with how NODE_NETWORK=off affects legacy peer exchange, which can also contribute to the startup problem. We'll fix that in a diff follow-up

PR #441, opened 2026-07-25
ci: share droplet provisioning and extract the perf-bench compare script

evan-forbes:
## Motivation

Follow-up to the review on #419, which flagged two things that PR made
obvious but deliberately left alone:

- DigitalOcean droplet provisioning is copy-pasted across five workflows
  ("there are four other workflows that copy this similar set up ... It 100
  percent can be abstracted out").
- The A/B step summary was generated by Python heredocs embedded in YAML
  ("I've never seen python be executed like this within a yaml file ... could
  we clean this up?").

#419 was tested extensively as a unit, so unifying provisioning there would
have invalidated that testing. This PR does it separately, with no behavior
change intended.

## Solution

Provisioning moves into four composite actions in `.github/actions/`:

| action | responsibility |
| --- | --- |
| `do-cli` | pinned `doctl` install + auth, optional fleet SSH key |
| `do-droplet` | resolve the newest baked image, optionally clone the newest `zakura-pr-state-<network>-*` snapshot, create the tagged droplet |
| `do-wait-ssh` | poll a new droplet until it accepts SSH |
| `do-teardown` | best-effort droplet + volume delete, retried while the detach settles |

`zakura-pr-node.yml`, `zakura-perf-bench.yml`, `zakura-mempool-load.yml`,
`zakura-pr-node-bake.yml`, and `zakura-pr-node-reaper.yml` call these instead
of carrying their own copies.

On lines of code, the honest total is **+658 / -377, net +281**, and it is
worth being explicit about where that goes rather than quoting the workflow
subset:

| bucket | net |
| --- | --- |
| workflows (the dedup itself) | **-246** |
| composite actions | +242, of which 98 is shell and 144 is input/output schema |
| new compare script | +93, replacing ~30 lines of heredoc |
| new tests | +183, replacing none |

Five copies of ~50 lines of provisioning shell collapse into 98. The rest is
the composite-action metadata tax plus a test suite that did not exist before,
which is what makes the extraction verifiable rather than hopeful. The actions
are deliberately not parameterized beyond what callers use -- `image`,
`region`, `ssh_key_path`, `version`, `attempts`, and `delay` were all dropped
after the audits pointed out no caller sets them (and that `ssh_key_path` could
not be set without also editing every caller's hardcoded `/tmp/do_ssh`).

`zakura-pr-node-bake.yml` still creates its droplet inline and uses only the
other three actions: it starts from a stock Ubuntu image (it is what produces
the baked one) and attaches two blank volumes it later snapshots per network.

Resource names, tags, regions, and the teardown-vs-keep conditions are
unchanged, so the hourly reaper keeps sweeping exactly what it did before.
The reaper itself gains an `actions/checkout` only so it can reference the
local action.

The two heredocs become `.github/workflows/scripts/perf-bench-compare.py`. It
writes the leg table to the step summary and sets a `compare` step output, so
the CPU profile diff is now a separate step gated on both legs having run to
completion, rather than on shell control flow.

The second commit closes latent hazards found while auditing the first:

- `do-droplet` publishes `volume_id` as soon as the volume exists, so a
  droplet create that fails afterwards no longer strands a volume the
  caller's teardown never learns about (it was left to the reaper's 2h
  detached sweep, on `main` too).
- `do-teardown` regains the two diagnostics the extraction dropped: the kept
  resources are named in the skip message, and an undeletable volume says so
  on stderr. Those strings are what a leak audit greps for.
- `zakura-pr-node` / `zakura-mempool-load` string-compare `teardown_after_run`
  the way `zakura-perf-bench` already did. Behavior-identical today (both are
  dispatch-only), but `null || ...` will not silently resolve to "keep" if
  either gains a `schedule:` or `workflow_call:` trigger.
- `perf-bench-compare.py` treats an unreadable `meta.json` like an absent one
  rather than failing the step with a traceback: `perf-bench-run.sh` tolerates
  a failed meta write and `json.dump` truncates before writing, so a
  zero-length file is a real outcome. The `compare` flag is now written before
  the summary is printed.

## Testing

- `actionlint -shellcheck=shellcheck` over all workflows: clean for every
  touched file (the only output is pre-existing `SC2317` info in
  `test-docker.yml`, untouched here). Note `actionlint` does **not** parse
  `action.yml` files, so the composite actions' shell was extracted and
  shellchecked separately — also clean.
- `python3 .github/workflows/scripts/test_perf_bench_compare.py`: 12 unit
  tests pass, run in CI by the `fast lints` job (confirmed in this PR's run).
  The suite includes a golden-output case asserting the exact markdown the
  removed heredoc produced; the extracted script's output was separately
  diffed byte-for-byte against `main`'s heredoc on the same input.
- Mutation-checked the suite: a truncated sha, a dropped table column, an
  ASCII arrow, a truncating `GITHUB_OUTPUT` open, a changed speedup
  precision, and a flipped row order are each caught.
- `markdownlint --config .trunk/configs/.markdownlint.yaml`: clean.
- `./scripts/changelog.py check`: passes.
- Every `uses: ./.github/actions/...` reference was machine-checked against
  its `action.yml` for unknown and missing-required inputs.
- Two independent audits of the diff (behavior-equivalence and CI-wiring)
  found no breaking issues; their actionable items are in the second commit.
- Not run locally: `codespell` (CI covers it) and the droplet workflows
  themselves, which need the DO secrets and a dispatch. A `zakura-pr-node`
  dispatch on this branch is the cheapest end-to-end proof before this leaves
  draft.

## Changelog

No fragment: CI, scripts, and docs only.

- Review comments on #419.

### Follow-up Work

- **`status-checks.patch.yml` mirror drift.** That file says its `paths-ignore`
  "MUST be an exact inverse of paths in lint.yml", but it only lists the
  Rust/config subset — `scripts/**.sh`, `.github/workflows/scripts/**`,
  `deploy/**`, and `.github/scripts/**` are already absent on `main`, and this
  PR adds a fifth (`.github/actions/**`). The effect is that script-only PRs
  (#419 included, and this one) get two checks named `lint success`: the real
  one and the stub. I deliberately did **not** "fix" this by adding the new
  path to `paths-ignore`: the stub serves four required checks from one shared
  path list, so ignoring an actions-only path there would leave `test success`,
  `test crate build success`, and `semver-checks` permanently "Expected" and
  block the PR. The real fix is a per-check inverse path set, which is its own
  change.
- Four other workflows still generate output from Python heredocs
  (`create-release.yml`, `release-binaries.yml`, `update-release-state.yml`,
  `zakura-continuous-sync.yml`). They are release-pipeline code and were left
  out to keep this diff reviewable.
- No linter covers shell inside `action.yml`. This is unchanged by this PR —
  that shell lived in workflow `run:` blocks before, which CI does not lint
  either (there is no `actionlint` job) — but the code is now shared by five
  workflows, so it is worth closing. `actionlint` would cover the workflow
  half; the action half needs a small extraction step like the one used to
  verify this PR.

### DIFF
--- a/.github/actions/do-cli/action.yml
+++ b/.github/actions/do-cli/action.yml
@@ -1,22 +1,14 @@
 name: "DigitalOcean CLI"
-description: "Install and authenticate doctl, and optionally write the fleet SSH key."
+description: "Install and authenticate doctl, and optionally write the fleet SSH key to /tmp/do_ssh."
 
 inputs:
   token:
     description: "DigitalOcean API token (secrets.DIGITALOCEAN_ACCESS_TOKEN)"
     required: true
   ssh_private_key:
-    description: "Optional SSH private key to write (secrets.DO_SSH_PRIVATE_KEY). Omit for API-only jobs like the reaper."
+    description: "Optional SSH private key (secrets.DO_SSH_PRIVATE_KEY). Omit for API-only jobs like the reaper."
     required: false
     default: ""
-  ssh_key_path:
-    description: "Where to write the SSH private key. Calling workflows hardcode /tmp/do_ssh in their own scp/ssh steps, so overriding this means overriding do-wait-ssh and those steps too."
-    required: false
-    default: "/tmp/do_ssh"
-  version:
-    description: "doctl release to install"
-    required: false
-    default: "1.120.0"
 
 runs:
   using: composite
@@ -28,20 +20,21 @@ runs:
       shell: bash
       env:
         DIGITALOCEAN_ACCESS_TOKEN: ${{ inputs.token }}
-        DOCTL_VERSION: ${{ inputs.version }}
       run: |
         set -euo pipefail
+        DOCTL_VERSION=1.120.0
         curl -sL "https://github.com/digitalocean/doctl/releases/download/v${DOCTL_VERSION}/doctl-${DOCTL_VERSION}-linux-amd64.tar.gz" | tar -xz -C /tmp
         sudo install /tmp/doctl /usr/local/bin/doctl
         doctl auth init -t "${DIGITALOCEAN_ACCESS_TOKEN}"
 
+    # /tmp/do_ssh is hardcoded here and in do-wait-ssh because every calling
+    # workflow's own scp/ssh steps use that path too.
     - name: Write SSH key
       if: inputs.ssh_private_key != ''
       shell: bash
       env:
         DO_SSH_PRIVATE_KEY: ${{ inputs.ssh_private_key }}
-        SSH_KEY_PATH: ${{ inputs.ssh_key_path }}
       run: |
         set -euo pipefail
-        install -m 600 /dev/null "${SSH_KEY_PATH}"
-        printf '%s\n' "${DO_SSH_PRIVATE_KEY}" > "${SSH_KEY_PATH}"
+        install -m 600 /dev/null /tmp/do_ssh
+        printf '%s\n' "${DO_SSH_PRIVATE_KEY}" > /tmp/do_ssh
--- a/.github/actions/do-cli/action.yml
+++ b/.github/actions/do-cli/action.yml
@@ -10,7 +10,7 @@ inputs:
     required: false
     default: ""
   ssh_key_path:
-    description: "Where to write the SSH private key"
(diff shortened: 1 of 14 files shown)


ValarDragon (approved):
TY!

PR #442, opened 2026-07-25
fix(network): avoid pruned peers for historical FindBlocks

ValarDragon:
## Motivation

Pruned peers retain chain hashes but do not advertise pruned block bodies in `FindBlocks` responses. During historical sync, routing `FindBlocks` to peers without `NODE_NETWORK` can therefore produce empty responses, retry delays, and disconnect churn.

## Solution

Apply the existing 5,000-block pruned-peer routing boundary to `FindBlocks` requests. Historical requests are routed only to peers advertising `NODE_NETWORK`; pruned peers remain eligible near the estimated network tip. `FindHeaders` routing is unchanged.

The zcashd-compat stall-tracker fixture now explicitly advertises `NODE_NETWORK`, keeping that test focused on its intended full-node sidecar scenario.

## Testing

- `cargo test -p zakura-network find_blocks` — 9 passed
- `npx --yes markdownlint-cli changelog-unreleased/442.md --config .trunk/configs/.markdownlint.yaml`

The routing tests verify that historical `FindBlocks` skips a pruned peer and selects a full node, while near-tip `FindBlocks` remains eligible for a pruned peer.

## Changelog

Added `changelog-unreleased/442.md`.

- [BIP 159](https://github.com/bitcoin/bips/blob/master/bip-0159.mediawiki)
- Builds on the capability-aware block routing introduced by #440.

### Follow-up Work

None currently identified.

### DIFF
--- a/changelog-unreleased/442.md
+++ b/changelog-unreleased/442.md
@@ -1,0 +1,4 @@
+## Fixed
+
+- Avoided querying pruned peers for historical block discovery
+  ([#442](https://github.com/zakura-core/zakura/pull/442)).
--- a/zakura-network/src/peer_set/set.rs
+++ b/zakura-network/src/peer_set/set.rs
@@ -913,17 +913,17 @@ where
         self.select_p2c_peer_from_list(&self.ready_services.keys().copied().collect())
     }
 
-    /// Returns true if pruned peers are eligible for generic block downloads.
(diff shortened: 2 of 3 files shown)


ValarDragon:
Lets concentrate work around Evan's #447  for now

PR #447, opened 2026-07-26
fix(network): prefer archive peers during initial sync

evan-forbes:
## Motivation

Historical legacy sync stalled after #440 because peer-set readiness counted
pruned peers, while block routing rejected them. When all archive peers were
busy, a ready pruned peer made `poll_ready` succeed and the subsequent block
request returned `NoReadyPeers`.

Restarting a sync round is intentional for structural failures such as a bad
hash chain or a block that remains unavailable. `NoReadyPeers`, however, only
means that every usable peer is busy at that instant. Classifying that transient
condition as fatal discarded the active checkpoint pipeline and made saturation
repeat as a non-converging restart cycle.

## Solution

While the local node is more than the shared at-or-near-tip threshold behind:

- disconnect peers that do not advertise `NODE_NETWORK`, including peers
  currently handling work;
- signal the crawler to find replacement peers;
- use ordinary peer-set routing and global readiness for the remaining archive
  peers, so `ready().await` parks when they are busy;
- classify `NoReadyPeers` with registry-level unavailability, using the existing
  two-second backoff and 60-retry budget while preserving downloaded blocks.

If the retry budget is exhausted, the normal fatal path still restarts the round
and obtains fresh tips. This preserves the bad-tip and incomplete-checkpoint
escape hatch.

At or near the network tip, the archive preference stops and pruned and archive
peers are retained and routed equally.

## Service-bit audit

The version decoder reads the remote service bits and the peer client checks
`NODE_NETWORK` directly. Production archive and zcashd-compatible Zakura nodes
advertise `NODE_NETWORK`; plain pruned Zakura nodes intentionally advertise no
service bits.

The managed continuous-sync and benchmark templates configure pruned nodes, so
seeing many Zakura peers classified as pruned is expected for that fleet rather
than evidence of a decoder bug. Test sidecar fixtures were updated to match the
production zcashd-compatible service advertisement.

## Testing

- `git diff --check`
- `npx --yes markdownlint-cli@0.44.0 changelog-unreleased/447.md --config .trunk/configs/.markdownlint.yaml`
- `cargo test -p zakura-network` (1,160 passed, 4 ignored; acceptance test
  passed)
- `env 'CXXFLAGS=-include cstdint' cargo test -p zakura no_ready_peers_schedules_backoff_retry`
- `env 'CXXFLAGS=-include cstdint' cargo clippy -p zakura --all-targets -- -D warnings`

The `CXXFLAGS` override works around the host RocksDB C++ header environment;
the targeted Zakura test and lint both passed.

Added coverage for:

- disconnecting pruned peers during historical sync and routing to an archive
  peer;
- waiting without a hot loop while archive peers are busy or only pruned peers
  are connected;
- scheduling `NoReadyPeers` through the backed-off retry path without restarting
  the round;
- retaining and routing to pruned peers at or near tip, including advertised
  inventory;
- matching sidecar fixtures to production archive service bits.

## Changelog

Added `changelog-unreleased/447.md`.

Regression introduced by #440
(`fix(network): avoid pruned peers during historical block sync`).

### Follow-up Work

Run a controlled legacy genesis-sync deployment to validate archive-peer
discovery and first-block progress under real mainnet peering.

### DIFF
--- a/changelog-unreleased/447.md
+++ b/changelog-unreleased/447.md
@@ -1,0 +1,5 @@
+## Fixed
+
+- Fixed legacy initial sync stalls by replacing pruned peers with archive peers
+  until the node reaches the network tip
+  ([#447](https://github.com/zakura-core/zakura/pull/447)).
--- a/changelog-unreleased/447.md
+++ b/changelog-unreleased/447.md
@@ -1,5 +1,6 @@
 ## Fixed
 
 - Fixed legacy initial sync stalls by replacing pruned peers with archive peers
-  until the node reaches the network tip
+  until the node reaches the network tip, and retrying temporary peer
+  unavailability without discarding the sync round
   ([#447](https://github.com/zakura-core/zakura/pull/447)).
--- a/zakura-network/src/peer/error.rs
+++ b/zakura-network/src/peer/error.rs
@@ -28,6 +28,9 @@ pub enum NotFoundClass {
     /// [`PeerError::NotFoundResponse`] — a specific peer lacked the item; retry on another peer.
     Response,
     /// [`PeerError::NotFoundRegistry`] — no ready peer has it; needs fresh tips/peers.
+    ///
+    /// Also includes [`PeerError::NoReadyPeers`], which needs a peer to become ready or connect
+    /// before the request can be retried.
     Registry,
     /// Any other error (not a `notfound`-style failure).
     Other,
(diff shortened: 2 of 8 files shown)


ValarDragon:
Lets also add FindBlocks to hitting this codepath

ebfull:
Heads-up: changelog fragments moved from the root `changelog-unreleased/` directory to `docs/changelog/unreleased/` in #520, and this PR still writes its fragment to the old location (`changelog-unreleased/447.md`). When the PR merges, that recreates the deleted root directory and the fragment gets stranded where release assembly won't pick it up (this already happened once with #513, fixed in #527).

evan-forbes:
stale

PR #450, opened 2026-07-26
fix(sync): prioritize mined block gossip events

ValarDragon:
## Motivation

PR #439 intermittently failed
`mined_block_marks_tip_after_successful_broadcast` because the gossip task
received `AdvertiseBlock` before the expected `AdvertiseBlockToAll`.

The mined-block channel and committed-tip fallback can both become ready after
the short submission delay. Tokio's default `select!` ordering chooses among
ready branches pseudo-randomly, so the fallback can win even though a mined
block notification is queued. The node then advertises the block to a sampled
fraction of ready peers before processing the queued mined-block notification
and advertising it to all ready peers. This produces a duplicate advertisement
and makes the paused-time regression test flaky under CI load.

## Solution

Use biased selection and order completed mined-block marks and new mined-block
submissions before the committed-tip fallback. The fallback still runs when
neither mined-block branch is ready.

Extract the selection into a private helper and exercise all three priorities
with every event ready. The vector repeats 64 times so removing `biased;`
reliably exposes Tokio's randomized ready-branch selection.

## Testing

- `cargo test -p zakura gossip --lib` (13 passed)
- `cargo test -p zakura components::sync --lib` (64 passed)
- 20 concurrent post-fix runs of
  `mined_block_marks_tip_after_successful_broadcast`
- `npx --yes markdownlint-cli changelog-unreleased/450.md --config

## Changelog

Added `changelog-unreleased/450.md` under Fixed because this changes runtime
gossip behavior, not only the test.

- Observed in PR #439 unit-test shard 3:
  https://github.com/zakura-core/zakura/actions/runs/30207940060/job/89809435976

### DIFF
--- a/changelog-unreleased/450.md
+++ b/changelog-unreleased/450.md
@@ -1,0 +1,5 @@
+## Fixed
+
+- Avoid duplicate block advertisements when a mined-block notification and the
+  committed-tip fallback become ready together
+  ([#450](https://github.com/zakura-core/zakura/pull/450)).
--- a/zakurad/src/components/sync/gossip.rs
+++ b/zakurad/src/components/sync/gossip.rs
@@ -2,7 +2,7 @@
 //!
 //! [`block::Hash`]: zakura_chain::block::Hash
 
-use std::time::Duration;
+use std::{future::Future, time::Duration};
 
 use futures::TryFutureExt;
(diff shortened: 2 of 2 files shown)


ValarDragon:
@v12-auditor review now

PR #452, opened 2026-07-26
fix(sync): retain complete findblocks responses

czarcas7ic:
## Motivation

Legacy sync discards the final `FindBlocks` hash and accepts the expected continuation in the second position to tolerate retired zcashd inventory mixing. With current peers, that can turn a valid short response at the chain head into no usable continuation and stall block discovery.

## Solution

Treat `FindBlocks` responses as ordered protocol data. Retain every returned hash, require extension responses to begin with the expected hash, enforce the 500-hash limit on the raw response, and queue a single new hash even when there are not yet two hashes for a follow-on prospective tip. Duplicate, locator-echo, and known-suffix checks remain enforced.

This replaces the earlier watchdog-based draft with a direct removal of the retired compatibility assumptions.

## Testing

- `cargo test -p zakura --lib --locked` (324 passed)

The sync tests cover one-hash initial discovery, a one-hash extension, retention of the final response hash, strict continuation ordering, and rejection of oversized, duplicate, locator-echo, and known-suffix responses.

In a mixed four-Zakura/four-Zebra partition and reorg rehearsal, all four nodes running this PR reached the winning H+1 tip through P2P sync. The only node still one block behind after the 300-second convergence gate was an unmodified Zebra node. All eight nodes remained active and accepted the 1.5 MB activation block; the run stopped at the expected all-node convergence failure.

## Changelog

Added `changelog-unreleased/452.md` under Fixed.

### DIFF
--- a/changelog-unreleased/452.md
+++ b/changelog-unreleased/452.md
@@ -1,0 +1,5 @@
+## Fixed
+
+- Fixed legacy block sync stalling when a peer advertises only one new block
+  at the chain head
+  ([#452](https://github.com/zakura-core/zakura/pull/452)).
--- a/zakurad/src/components/sync.rs
+++ b/zakurad/src/components/sync.rs
@@ -165,12 +165,6 @@ pub const MAX_TIPS_RESPONSE_HASH_COUNT: usize = 500;
 
 /// Returns true if a `FindBlocks` hash list has the protocol maximum number of
 /// block hashes or fewer.
-///
-/// Callers that discard zcashd's extra trailing hash must apply this check to
-/// the hashes that remain after stripping. That way a 501-hash zcashd response
-/// (500 chain hashes plus one appended hash) is accepted as 500 usable hashes,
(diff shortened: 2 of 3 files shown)


czarcas7ic:
@codex review

ValarDragon:
This makes sense to me! Its actually what I assumed your other PR was 😅

czarcas7ic:
v12 is non issue, this is expected given we aren't supporting zcashd

ValarDragon:
I'm reviewing rn - it changes more than I expected

ValarDragon (approved):
Reviewed! Was more LOC of sync code changed, but overall the invariants make sense.

The zcashd compat peer should not be delayed. It could _slightly_ degrade its zakura counterpart now with a random wasted hash, but it already was degrading it in block discovery. So its not really that much different.

It won't impact miners with zcashd compat in practice, as you always do obtain tips to multiple peers, so the other peers have the right behavior.

PR #455, opened 2026-07-26
fix(network): separate findblocks responses from block gossip

czarcas7ic:
## Motivation

Legacy TCP uses `inv` for both `getblocks` responses and unsolicited block announcements. Because those messages carry no request ID, a tip announcement can arrive first and incorrectly complete an in-flight `FindBlocks` request; the real response then arrives after the handler has moved on.

## Solution

Send outbound legacy TCP `FindBlocks` requests as `getheaders`, then convert the distinct `headers` response into ordered block hashes. Unsolicited `inv` messages continue through the normal block-advertisement path without completing the request. Native Zakura transport and inbound `getblocks` handling are unchanged.

This deliberately changes the legacy discovery batch limit from 500 hashes to 160 headers. The sync loop already requests another batch whenever its reserve is below 500, so discovery continues immediately rather than stalling at 160.

## Testing

- `cargo test -p zakura-network find_blocks_ignores_block_gossip_before_response --locked`
- Serialized `zakura-network` library suite: 1,154 passed, with only three macOS-incompatible source-address binding tests skipped
- `cargo test -p zakura --lib --locked`: 325 passed

The regression test reproduces the ordering race by delivering singleton block gossip before the actual response. It verifies that gossip is advertised normally, the discovery request remains pending, and the later headers response completes it with the expected hash.

## Changelog

Added `changelog-unreleased/455.md` under `Fixed`.

### DIFF
--- a/changelog-unreleased/455.md
+++ b/changelog-unreleased/455.md
@@ -1,0 +1,4 @@
+## Fixed
+
+- Fixed legacy peer block discovery so unsolicited block announcements cannot be mistaken for discovery responses
+  ([#455](https://github.com/zakura-core/zakura/pull/455)).
--- a/zakura-network/src/peer/connection.rs
+++ b/zakura-network/src/peer/connection.rs
@@ -397,15 +397,12 @@ impl Handler {
                 }
             }
 
-            // TODO:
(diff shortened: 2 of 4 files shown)


czarcas7ic:
@codex review

PR #456, opened 2026-07-26
fix(network): gate singleton findblocks replies near tip

czarcas7ic:
## Motivation

Legacy getblocks uses untagged inv messages for both solicited responses and unsolicited tip gossip. While Zakura is syncing, a one-block tip announcement can arrive before the real multi-hash response and be mistaken for that response. Discovery then discards the useful reply and stalls until it retries.

## Solution

Use the existing committed-chain timestamp estimate when a legacy getblocks request is sent. That estimate treats the node as near tip when its estimated distance is at most five blocks. Near the estimated network tip, retain singleton response handling. While far from tip or when tip proximity is unknown, route a singleton inv through the normal block-gossip path and keep waiting for a multi-hash response. Native Zakura connections are unchanged.

This is the heuristic alternative to changing discovery to getheaders. It avoids the additional header download and deserialization cost. The intentional tradeoff is that a genuine singleton reply from a stale peer may wait for the request timeout while the local estimate says far from tip or unknown.

## Testing

- cargo test -p zakura-network find_blocks_ -- --nocapture
- cargo clippy -p zakura-network --all-targets -- -D warnings
- cargo fmt --all -- --check
- cargo test -p zakura-network ran 1,162 library tests: 1,155 passed, 3 failed, and 4 were ignored. The three failures are listener tests whose configured loopback source addresses could not be bound on this macOS host.

## Changelog

Added changelog-unreleased/456.md.

### DIFF
--- a/changelog-unreleased/456.md
+++ b/changelog-unreleased/456.md
@@ -1,0 +1,4 @@
+## Fixed
+
+- Prevented unsolicited single-block announcements from interrupting legacy block discovery while syncing
+  ([#456](https://github.com/zakura-core/zakura/pull/456)).
--- a/zakura-network/src/peer/connection.rs
+++ b/zakura-network/src/peer/connection.rs
@@ -47,6 +47,8 @@ mod peer_tx;
 #[cfg(test)]
 mod tests;
 
+type IsAtOrNearNetworkTip = Box<dyn Fn() -> bool + Send>;
+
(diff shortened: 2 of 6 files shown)


czarcas7ic:
@codex review

czarcas7ic:
Responses to v12
Screenshot 2026-07-26 at 9 18 11 PM

PR #463, opened 2026-07-27
Expand the legacy peer cache

ValarDragon:
## Summary

- increase the legacy peer disk cache from 75 to 125 addresses
- prefer recently active peers that have responded locally over peers known only through gossip
- select the preferred subset before converting it to a `HashSet`
- retain the final lexicographic file sort so cache ordering does not expose selection priority or connection timing
- correct stale rustdoc that described the default hard outbound limit as 75 rather than 300

## Why this is safe

The default legacy peer target is 100 and the hard outbound multiplier is 3, so the current default hard outbound limit is 300. A 125-address disk cache remains well below that limit.

Loading the cache does not dial every cached address. Startup connection attempts remain capped by the existing target, and all later connections remain subject to the independently enforced outbound limit. The existing three-hour activity filter is unchanged.

Prioritizing peers with `last_response.is_some()` uses stronger local evidence than an unverified gossiped timestamp. A peer remains in that preferred group if it responded previously and a later connection attempt failed.

## Tests

- `cargo test -p zakura-network cacheable_prefers_locally_observed_peers`
- `cargo test -p zakura-network peer_cache`
- `git diff --check`

## API surface

- changes the public `MAX_PEER_DISK_CACHE_SIZE` constant value from 75 to 125
- no function signatures, visibility, traits, enum variants, constructors, or type aliases change

### DIFF
--- a/zakura-network/src/address_book.rs
+++ b/zakura-network/src/address_book.rs
@@ -541,8 +541,8 @@ impl AddressBook {
 
         let peers = self.peers.clone();
 
-        // Get peers in preferred order, then keep the recently active ones
-        peers
+        // Get peers in preferred order, then keep the recently active ones.
+        let mut peers: Vec<_> = peers
             .ordered_values()
             // # Security
             //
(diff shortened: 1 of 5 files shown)


evan-forbes:
closing as completed since picked up by the other params pr

PR #464, opened 2026-07-27
fix(state): serve embedded VCT frontier roots at handoff

p0mvn:
## Motivation

Genesis VCT sync fail-closed at Mainnet handoff height `C = 3418406`, leaving the tip stuck at `3418405`. Observed on continuous-sync node `temp-zakura-sync-test-7` (dual-stack): after ~10 minutes of

```text
VCT: no verifiable supplied root for a frozen-frontier height; refusing to recompute (retryable) height=Height(3418406)
```

the controller halted the run for stall, and the Slack monitor later posted a DOWN alert because `zakura.service` was already stopped.

### How we regressed

This worked before header-root authentication ([#352](https://github.com/zakura-core/zakura/pull/352)) and broke when that PR landed.

**Before #352**, header sync wrote peer `tree_aux` roots into the DB as provisional rows ahead of body commit, including at handoff height `C`. `PeerSource::vct_root` was a plain DB lookup, so commit of `C` found a row and finished. Handoff “worked” because a provisional peer write papered over `C` — not because handoff was specially handled.

**#352** correctly stopped promoting unauthenticated peer roots past `C - 1`: there is no later configured checkpoint to witness a successor at `C + 1`, so the authenticated index never gets a row at `C` (design §13.2). `PeerSource` still only read DB rows and was never taught to fall back to the embedded frontier at `C`.

So the same commit path that used to see a provisional row at `C` now sees `None`, fail-closes, and retries forever. Last successful continuous-sync on this host was Jul 22; #352 merged Jul 23; this stall reproduced Jul 26.

## Solution

At the VCT handoff height, `PeerSource::vct_root` returns the embedded frontier roots instead of requiring a DB row that header-root auth never writes at `C`. Heights below `C` still require authenticated DB rows; heights above `C` are unchanged.

## Testing

- Unit test `peer_source_serves_embedded_frontier_roots_at_handoff` in `commitment_aux.rs`
- Property test `vct_peer_source_handoff_without_db_root_at_c` in `prop.rs` (authenticated roots through `C - 1` only; handoff commit must still succeed and match legacy)
- Manual validation: fixed binary on `temp-zakura-sync-test-7` advanced tip past `3418406`

## Changelog

- `changelog-unreleased/464.md` — user-visible Fixed entry

- `docs/design/header-sync-vct-root-authentication.md` §13.2 (handoff at the last checkpoint with embedded frontier)
- Regression introduced by [#352](https://github.com/zakura-core/zakura/pull/352)

### Follow-up Work

- CI gate that syncs through the VCT handoff (e.g. revive pre-checkpoint PR-node mode from [#310](https://github.com/zakura-core/zakura/pull/310)) so this seam cannot land again without a real crossing

### DIFF
--- a/changelog-unreleased/464.md
+++ b/changelog-unreleased/464.md
@@ -1,0 +1,6 @@
+## Fixed
+
+- Fixed Mainnet genesis VCT sync failing closed at the handoff height by serving
+  embedded frontier roots from `PeerSource` when no authenticated DB root exists
+  at that height
+  ([#464](https://github.com/zakura-core/zakura/pull/464)).
--- a/zakura-state/src/service/finalized_state/commitment_aux.rs
+++ b/zakura-state/src/service/finalized_state/commitment_aux.rs
@@ -439,6 +439,12 @@ impl CommitmentRootSource for FixtureSource {
 /// immutably here and never fetched over the network — a peer source always has one,
 /// because peer mode is only selected on networks with an embedded frontier. Committed
 /// rows are cleaned up by the database's own retention, not through this seam.
+///
+/// At the handoff height itself the header-root lane never promotes a peer row: checkpoint
+/// header `C` is only the final pinned witness for `C - 1`, and no later configured
+/// checkpoint can authenticate a successor at `C + 1`. [`Self::vct_root`] therefore returns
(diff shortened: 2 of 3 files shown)


p0mvn:
@cursor review

evan-forbes (approved):

PR #467, opened 2026-07-27
fix(network): stop connection-state metrics from leaking per-peer series

p0mvn:
## Motivation

Continuous-sync node `temp-zakura-sync-test-6` was flapping Slack "Zakura down" / recovered alerts while `zakura.service` stayed active and kept syncing.

Root cause: `zakura.net.connection.state` retained a Prometheus gauge series for every historical `addr` × `command` pair. Setting gauges to `0` on disconnect does not remove them from the exporter, so a long genesis sync grew `/metrics` to ~22 MB (~116k connection-state series). The alert helper's short scrape then timed out and the monitor treated that as node-down.

## Solution

- Aggregate `zakura.net.connection.state` by `command` only, using increment/decrement for live connections instead of per-`addr` zeroed gauges.
- Harden continuous-sync monitoring:
  - scrape `/metrics` with a 15s timeout and read the full body
  - when `zakura.service` is active, log metrics scrape failures as `metrics-degraded` instead of paging "down"

## Testing

- `python3 -m unittest deploy.continuous-sync.tests.test_continuous_sync`
- `cargo check -p zakura-network --locked`
- Deployed the infra scripts to `temp-zakura-sync-test-6` with `--no-start`; status now reports height correctly and services stayed active

## Changelog

- `changelog-unreleased/467.md`

### DIFF
--- a/changelog-unreleased/467.md
+++ b/changelog-unreleased/467.md
@@ -1,0 +1,6 @@
+## Fixed
+
+- Fixed Prometheus cardinality growth for `zakura.net.connection.state` by
+  aggregating live connections by command instead of retaining historical
+  per-peer address series
+  ([#467](https://github.com/zakura-core/zakura/pull/467)).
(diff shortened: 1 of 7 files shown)


p0mvn:
@cursor review

p0mvn:
@cursor push 4491424890

p0mvn:
@cursor review

ValarDragon (approved):

PR #470, opened 2026-07-27
fix(rpc): report pruned storage mode and add pruneheight to getblockchaininfo

emersonian:
RPC consumers like Zaino and Lightwalletd should be able to know if Zakura is running pruned mode.

## Motivation

`getblockchaininfo.pruned` was sourced from `ZakuraDb::is_pruned()`, which reports
whether the durable `pruning_metadata` marker has been written. That marker is only
written by a commit that actually deletes raw transaction data, so a node configured
with `storage_mode.pruned` reported `pruned: false` until its tip rose above the
retention window — at least 10,000 blocks on Mainnet and Testnet, and longer when
pruning is first enabled on an existing archive database.

zcashd reports `fPruneMode`: whether blocks are *subject to* pruning, not whether any
block has been deleted yet.

```c
obj.pushKV("pruned", fPruneMode);
```

zcashd also emits `pruneheight` whenever `pruned` is true, so a client can learn where
the servable range starts. Zakura omitted that field entirely, leaving `pruned: true`
with no way to tell which bodies are still available.

Reopening an already-pruned database was never affected: the marker is durable, so it
reports pruned before any block is committed. `FinalizedState::new` relies on exactly
that to refuse an archive-mode open of a pruned database.

## Solution

Add `ZakuraDb::prunes_historical_data()` — true when pruned storage mode is configured
or the database has already pruned — and serve the RPC from it.

Add `ZakuraDb::prune_height()`: the lowest height at and above which every block body is
retained. It is `Height::MIN` for a pruning node that has not deleted anything yet,
because such a node still retains every body, and `None` for an archive node. It is
serialized as the optional `pruneheight` field, omitted when `pruned` is false.

`is_pruned()` keeps its "has actually pruned" meaning for the one-way archive reopen
guard and the subtree upgrade check, which need the on-disk fact rather than the
configured mode.

Zakura prunes raw transaction data rather than whole blocks, so blocks below
`pruneheight` keep their headers, transaction IDs, and consensus state, and the genesis
block is never pruned. Both departures from zcashd's "lowest-height complete block
stored" are documented on the field.

`ReadRequest::IsPruned` becomes `ReadRequest::PruningInfo` and its response carries both
values, so `getblockchaininfo` still makes one state call.

## Testing

`zakura-state/src/service/finalized_state/zakura_db/block/tests/prune.rs`:

- `pruned_mode_reports_pruned_before_anything_is_pruned` — asserts both `!is_pruned()`
  and `prunes_historical_data()` on a pruned-mode database whose tip is far below the
  retention window, pinning the old bug and the new behavior in one test, plus
  `prune_height() == Some(Height::MIN)`.
- `archive_mode_never_reports_pruned` — archive nodes report neither.
- `reopened_pruned_database_reports_pruned_before_committing_a_block` — a reopened
  pruned database reports pruned and `prune_height() == Some(Height(4))` before any
  block is committed, covering the case that was already correct so it stays that way.

`zakura-rpc/src/methods/tests/prop.rs` covers the pass-through of both fields to the RPC
response.

Run locally: `cargo test -p zakura-state --lib` (459 passed), `cargo test -p zakura-rpc`
(90 + 33 passed), `cargo clippy --all-targets -- -D warnings` on both crates,
`cargo fmt --all -- --check`, `cargo doc -p zakura-state`.

## Changelog

Fragment added as `changelog-unreleased/<PR-number>.md`.

Two compatibility breaks are called out there:

- `GetBlockchainInfoResponse::new()` gained a `prune_height` argument. It is a public fn
  in a published crate; the in-workspace caller is updated, but downstream callers break.
- The read-request metric label changed from `is_pruned` to `pruning_info`, which
  affects any dashboard or alert keyed on the old name.

- zcashd `getblockchaininfo`, `pruned` and `pruneheight`:
  https://github.com/zcash/zcash/blob/v6.3.0/src/rpc/blockchain.cpp

### Follow-up Work

- **`getblock` pruned-vs-unknown errors** — `getblock` maps every missing body to
  "Block not found" (-8), which a client cannot tell apart from an unknown hash. zcashd
  reports `RPC_INTERNAL_ERROR` with "Block not available (pruned data)". Branch is ready
  and rebased onto `main`; it is a distinct change to RPC error semantics, so it gets its
  own PR and its own review.
- **`getrawtransaction`** — same pruned-vs-unknown gap for transactions. Pruning retains
  `tx_loc_by_hash`, so the information exists, but no read request exposes it; fixing
  this needs a new one.
- **`z_gettreestate` on VCT fast-synced nodes** — for a height in the `[U, H)` absent
  band the RPC returns 200 with null tree fields, indistinguishable from "this pool was
  not active yet". `ZakuraDb::vct_historical_tree_unavailable()` exists for exactly this
  check and currently has no production caller.

  Worth deciding with that change: `pruneheight` tells a client which *bodies* are
  missing, but nothing tells it that historical *treestates* are missing. Those are two
  different limits on what a node can serve, so the VCT work may want its own field
  rather than overloading this one.

### DIFF
--- a/changelog-unreleased/470.md
+++ b/changelog-unreleased/470.md
@@ -1,0 +1,29 @@
+## Added
+
+- Added the `pruneheight` field to `getblockchaininfo`, reporting the lowest
+  height at and above which every block body is retained. As in `zcashd`, it is
+  only present when `pruned` is true. Zakura prunes raw transaction data rather
+  than whole blocks, so blocks below this height keep their headers, transaction
+  IDs, and consensus state, and the genesis block is never pruned
+  ([#470](https://github.com/zakura-core/zakura/pull/470)).
+
+## Changed
+
+- Changed the read-request metric label `is_pruned` to `pruning_info`, because
+  the request now reports the prune height alongside the pruned flag. Dashboards
+  and alerts keyed on the old label need updating
+  ([#470](https://github.com/zakura-core/zakura/pull/470)).
+- Changed the public `GetBlockchainInfoResponse::new()` in `zakura-rpc` to take
+  a `prune_height` argument, so that downstream callers can set the new field.
+  This breaks existing callers of that constructor
+  ([#470](https://github.com/zakura-core/zakura/pull/470)).
+
+## Fixed
+
+- Fixed `getblockchaininfo` reporting `pruned: false` on a node configured with
+  `storage_mode.pruned` until it first deleted block data, which took at least
+  10,000 blocks on Mainnet and Testnet and longer when pruning was enabled on an
+  existing archive database. As in `zcashd`, the field now reports whether blocks
+  are subject to pruning, not whether any block has been deleted yet. Databases
+  that had already pruned were unaffected
+  ([#470](https://github.com/zakura-core/zakura/pull/470)).
--- a/zakura-rpc/src/methods.rs
+++ b/zakura-rpc/src/methods.rs
@@ -1018,26 +1018,37 @@ where
         let debug_force_finished_sync = self.debug_force_finished_sync;
         let network = &self.network;
 
-        let (usage_info_rsp, is_pruned_rsp, tip_pool_values_rsp, chain_tip_difficulty) = {
+        let (usage_info_rsp, pruning_info_rsp, tip_pool_values_rsp, chain_tip_difficulty) = {
             use zakura_state::ReadRequest::*;
             let state_call = |request| self.read_state.clone().oneshot(request);
             tokio::join!(
                 state_call(UsageInfo),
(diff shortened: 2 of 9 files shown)


ValarDragon:
Sweet, gonna review this post upgrade

ebfull (changes requested):
ACK modulo bumping `zakura-rpc` to `6.0.0` (and rebasing / merging with `main`)

ValarDragon:
We can take over the version bump side

emersonian:
Rebased, confirming that you're handling the version bump. Thanks.

PR #477, opened 2026-07-28
feat(deploy): track fleet tip agreement and redesign the cluster dashboard

p0mvn:
## Motivation

The mainnet cluster dashboard reported per-node height, commit, and liveness,
but nothing about whether the fleet agreed on a chain tip. A split or a reorg
was only visible by comparing heights by eye across thirteen rows, and a tip
switch at an unchanged height was invisible. The page itself had no hierarchy:
ten identical summary cards in a ragged four-column grid, and a table whose
64-character node ids wrapped over three lines while the Ironwood countdown sat
in a card the same size as everything else.

## Solution

**Tip agreement.** Each poll groups nodes by `(height, tip hash)`. One group
means agreement; several groups at the leading height mean a split. Nodes are
labelled `majority`, `fork`, `ahead`, or `behind`. Fork depth between two tips
at the same height is estimated from best-chain ancestor hashes sampled at 1,
2, 5, 10, and 32 blocks back, so an unresolved fork reports `> 32 or unknown`
rather than a wrong number.

**Orphan pairs.** A height drop or a tip switch at the same height records the
discarded hash, the new canonical hash, and the depth. `--state-file` persists
that history across restarts. Both deploy workflows now pass the flag, pointed
at `/var/lib/zakura-<network>-dashboard/orphan-pairs.json`; without it the
history was in-memory only and lost on every restart, which would have made the
feature inert in production.

**Page rebuild.** An Ironwood activation panel (countdown, ETA, progress over
the final 10,000 blocks) and a fleet health panel (healthy count, health
segments, leading height, slowest advance, max header lag) lead the page. The
node table is filterable and sortable, colours rows by severity, shows each
node's delta against the majority tip, and expands every row into the full
probe output: ssh target, client and version, service state, RPC chain, node
id, tip and parent hashes, sampled ancestors, and any RPC or pool errors.
Badge tone is now mapped explicitly instead of interpolating server-supplied
values into CSS class names.

`/data`, `/healthz`, and `/ironwood-status.json` are unchanged. The public
Ironwood response shape and its verification rules are untouched.

## Testing

- `python3 -m unittest discover -s deploy/runner -p test_zakura_cluster_status.py`
  — 18 tests pass. New cases cover tip event classification, agreed/lagging/
  split summaries, chain role enrichment, header lag, fork depth from ancestor
  samples, and orphan pairs surviving a collector restart through the state
  file.
- `npx markdownlint-cli2 deploy/runner/README.md` — clean.
- `./scripts/changelog.py check` — clean.
- Rendered the page against a captured snapshot of the live 13-node mainnet
  fleet and against a synthetic degraded fleet (down, stale, rpc_error, and
  ahead nodes plus a two-way tip split) to exercise the failure styling, the
  split tip group, and the unreachable-endpoint banner. Also checked the
  narrow-viewport layout and confirmed the document does not overflow.
- Deployed to `us-east-0` and verified live: 13/13 healthy, tip groups and row
  drawers populated, and a real one-block propagation lag correctly surfaced as
  `Lagging` with a second tip group labelled `1 behind`.

## Changelog

No fragment: this PR changes no Rust source and no `Cargo.toml`. It touches
only the deploy runner dashboard, its tests, and the two deploy workflows.

### Follow-up Work

- The `--state-file` addition means the next mainnet deploy will create
  `/var/lib/zakura-mainnet-dashboard/orphan-pairs.json` on the runner. The
  directory is created by the writer, so no provisioning step is needed.
- The public DNS record for `status.mainnet.zakura.valargroup.dev` resolves to
  `18.204.152.241`, which does not answer for that name; the working hostname is
  `status-mainnet.valargroup.dev`. Unrelated to this PR, but worth fixing.

### DIFF
--- a/.github/workflows/zakura-mainnet-deploy.yml
+++ b/.github/workflows/zakura-mainnet-deploy.yml
@@ -373,7 +373,7 @@ jobs:
           [Service]
           Type=simple
           WorkingDirectory=/opt/zakura-mainnet-dashboard
-          ExecStart=/usr/bin/python3 /opt/zakura-mainnet-dashboard/zakura-cluster-status.py --config /opt/zakura-mainnet-dashboard/nodes.toml --host 0.0.0.0 --port 8090 --network mainnet --interval 10 --stale-after 300 --upgrade-height 3428143 --target-spacing 75
+          ExecStart=/usr/bin/python3 /opt/zakura-mainnet-dashboard/zakura-cluster-status.py --config /opt/zakura-mainnet-dashboard/nodes.toml --host 0.0.0.0 --port 8090 --network mainnet --interval 10 --stale-after 300 --upgrade-height 3428143 --target-spacing 75 --state-file /var/lib/zakura-mainnet-dashboard/orphan-pairs.json
           Restart=always
           RestartSec=5
 
--- a/.github/workflows/zakura-testnet-deploy.yml
+++ b/.github/workflows/zakura-testnet-deploy.yml
@@ -302,7 +302,7 @@ jobs:
           [Service]
           Type=simple
           WorkingDirectory=/opt/zakura-testnet-dashboard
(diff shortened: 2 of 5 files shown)


ValarDragon (approved):
Wow looks wonderful

PR #478, opened 2026-07-28
fix(network): allow more inbound than outbound connections

evan-forbes:
## Motivation

Two recent tuning decisions in this fork are now working against these nodes:

- `INBOUND_PEER_LIMIT_MULTIPLIER` was dropped to `1` (v5.0.0-rc.1 sync defaults),
  capping inbound connections at the peer set target size while outbound stayed
  at three times the target — 100 inbound and 300 outbound by default. The
  justification in the comment above that constant was that "inbound peers do not
  help us download the chain", so the cap traded inbound-serving capacity for our
  own sync speed. That premise does not hold in this codebase (see the correction
  below), so the cap gave up reachability without buying the sync advantage it
  assumed: peers behind a firewall, or with a misconfigured listener port, can
  only reach us on an inbound slot, and a node that mostly dials out contributes
  little to peer reachability or to the address book coverage other nodes build
  from us.

> **Correction.** The claim that inbound peers do not help us download the chain
> is wrong, and it should not be repeated as a rationale in future retunes.
> `accept_inbound_connections` and `crawl_and_dial` both send `DiscoveredPeer`
> values over the same `peerset_tx` channel (`peer_set/initialize.rs:321`, `:336`),
> `PeerSet::ready_services` records no connection direction, and `route_p2c` →
> `select_ready_p2c_peer` picks the lower-`PeakEwma`-load of two random ready
> services (`peer_set/set.rs:908`, `:1027`) with no direction, `PeerServices`, or
> `start_height` filter — `remote_start_height` is read only for trace output.
> Inbound peers are therefore fully eligible to serve `FindBlocks` and block
> downloads, and in practice they do.
>
> What direction actually selects for is *provenance*, not capability: outbound
> peers are ones we chose from the address book, while inbound peers are
> self-selected and include crawlers, monitors, and pruned nodes that advertise
> `NODE_NETWORK` but serve no historical blocks. Any future argument for capping
> inbound connections has to be made on peer *quality* grounds and backed by
> measurements, not on the incorrect premise that inbound peers are not asked for
> blocks at all.
- The peer disk cache is capped at 75 addresses, well below the number of
  connections these nodes keep, so a restarted node has few dialable candidates
  before the DNS seeders and the address crawler catch up.

## Solution

- `INBOUND_PEER_LIMIT_MULTIPLIER`: `1` → `3` (inbound limit 100 → 300).
- Outbound limit: add `OUTBOUND_PEER_LIMIT_DIVISOR = 2` alongside the existing
  multiplier of `3`, so the outbound limit is one and a half times the target
  size (300 → 150). That still leaves 50 connections of headroom above the
  initial outbound target of 100 to replace stalled or slow sync peers.
- `MAX_PEER_DISK_CACHE_SIZE`: `75` → `300`, matching the new inbound limit.
- Doc comments and the `Security` note now describe the inbound-favoring policy
  again, including the tradeoff it accepts (an attacker can supply a larger
  fraction of inbound peers; outbound peers still come from our address book).

## Testing

- `ensure_peer_connection_limits_consistent` now asserts the inverse invariant —
  the inbound limit is at least the outbound limit — with a matching
  `const_assert!` over the multipliers and divisor, so a future retune cannot
  silently flip the policy back.
- The listener multi-peer tests (`listener_peer_limit_many_handshake_*`, renamed
  from `*_default_*`) now pass an explicit small peer set target size. Going over
  the real default inbound limit would need more than 600 real TCP connections;
  the connections queue behind the listener's accept pacing and the tests time
  out. The default limit arithmetic is covered by the config tests instead.
- `listener_reserves_one_zcashd_compat_inbound_slot` now derives its public slot
  count from the configured inbound limit instead of assuming the limit equals
  the target size, and fills every public slot before offering one extra
  connection of each kind.
- `cargo test -p zakura-network`: 1170 passed, 0 failed.
- `cargo clippy --workspace --all-targets`: clean. `cargo fmt --all -- --check`: clean.

### Follow-up Work

Operators who tuned `peerset_initial_target_size` down for bandwidth reasons now
get proportionally fewer outbound connections than before (1.5x rather than 3x
the target); worth watching on the constrained fleet nodes after rollout.

Post-rollout observation: on the mainnet fleet the outbound limit is not the
binding constraint at all. Nodes carrying ~300 inbound connections hold only
9–12 outbound ones — an order of magnitude below either the old 300 ceiling or
the new 150 one. Outbound count is limited by the supply of dialable address book
candidates, not by `peerset_outbound_connection_limit`, which is what the
`MAX_PEER_DISK_CACHE_SIZE` half of this PR was aimed at. Lowering the outbound
ceiling here was therefore a no-op in practice; the peer set composition shift
these nodes actually saw (roughly 96% inbound) comes entirely from the inbound
limit going 100 → 300.


p0mvn (approved):
The rationale behind the params makes sense to me.

One concern is the magnitude of the change, especially relative to Zebra that uses 125 inbound / 75 outbound and 25 initial peer. But I don't think this is blocking, just to keep track of.

PR #482, opened 2026-07-29
fix(consensus): grace early NU6.3 branch IDs before activation

ebfull:
## Motivation

Backport of the early-direction half of upstream Zebra [PR #11113](https://github.com/ZcashFoundation/zebra/pull/11113). Time-sensitive ahead of mainnet Ironwood (NU6.3) activation. edit: lol!

## Root cause

Zakura's NU6.3 mempool branch-ID grace period (`is_nu6_3_branch_id_misbehavior_grace_period`) only suppressed the peer misbehavior score for **stale NU6.2** transactions relayed *after* our node crosses the activation height. The reverse direction was unhandled: when a peer is already past activation and relays **NU6.3** transactions while our node is still up to 40 blocks *behind* activation, `check::consensus_branch_id` returns `WrongConsensusBranchId`, which scores 100 and bans the peer — even though the peer is honest and its chain tip is simply a few blocks ahead of ours.

This is the direction that fires *first* during an activation transition, because a node one step behind sees NU6.3 traffic before it crosses the boundary itself.

## Solution

Extend `is_nu6_3_branch_id_misbehavior_grace_period` to grace both directions within `NU6_3_BRANCH_ID_MISBEHAVIOR_GRACE_BLOCKS` (40) of the NU6.3 activation height, keyed on the signed height offset from activation (`Height - Height` yields an exact `i64` `HeightDiff`):

- `(Nu6_2 tx, current Nu6_3)` with offset `0..40` — unchanged (stale-after-activation), and
- `(Nu6_3 tx, current Nu6_2)` with offset `-40..0` — new (early-before-activation).

The existing call site maps a graced `WrongConsensusBranchId` to `WrongConsensusBranchIdNu6_3GracePeriod` (score 0), so no call-site change is needed — the transaction is still rejected, only the ban score is suppressed. This is not consensus-critical: it affects mempool peer-scoring policy only.

This keeps Zakura's error-variant approach rather than adopting upstream's `transaction_upgrade`/`verification_height` fields on the download-verify error.

## Testing

Extended `public_nu6_3_consensus_branch_id_boundary` with early-direction assertions: an NU6.3 transaction is graced at `activation - 1` and `activation - 40`, and not graced at `activation - 41`. Verified the NU6.2→NU6.3 gap on both public networks (63,543 and 82,000 blocks) is far wider than the 40-block window, so the test heights land in the NU6.2 epoch. Local test evidence added below once run.

### DIFF
--- a/changelog-unreleased/482.md
+++ b/changelog-unreleased/482.md
@@ -1,0 +1,6 @@
+## Fixed
+
+- Extended the NU6.3 mempool branch-ID grace period to cover honest peers that
+  relay NU6.3 transactions in the 40 blocks before activation, so they are no
+  longer scored as misbehaving and banned during the activation transition
+  ([#482](https://github.com/zakura-core/zakura/pull/482)).
--- a/zakura-consensus/src/error.rs
+++ b/zakura-consensus/src/error.rs
@@ -265,7 +265,7 @@ pub enum TransactionError {
     WrongConsensusBranchId,
 
     #[error(
-        "mempool transaction uses the NU6.2 consensus branch id during the NU6.3 grace period"
+        "mempool transaction uses a mismatched NU6.2/NU6.3 consensus branch id during the NU6.3 grace period"
     )]
     WrongConsensusBranchIdNu6_3GracePeriod,
 
--- a/zakura-consensus/src/transaction.rs
+++ b/zakura-consensus/src/transaction.rs
@@ -78,25 +78,49 @@ const MEMPOOL_OUTPUT_LOOKUP_TIMEOUT: std::time::Duration = std::time::Duration::
 /// response from the transaction verifier.
 const POLL_MEMPOOL_DELAY: std::time::Duration = Duration::from_millis(50);
 
-/// Number of blocks after NU6.3 activation during which a NU6.2 branch ID
-/// does not count as peer misbehavior.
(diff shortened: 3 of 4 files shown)


ebfull:
Local verification (release build, `CXXFLAGS="-include cstdint"` RocksDB/GCC16 workaround):

```
cargo test --release -p zakura-consensus --lib public_nu6_3_consensus_branch_id_boundary
test result: ok. 1 passed; 0 failed; 0 ignored
```

The extended boundary test exercises the new early direction (NU6.3 tx graced at `activation - 1` and `activation - 40`, not graced at `activation - 41`) on both public networks.

ebfull:
Human update: This seems useful only to align with an upstream fix (if others think it's warranted) to assist in future backports and encourage similar fixes in the next network upgrades. It's not really necessary now.

ebfull:
Heads-up: changelog fragments moved from the root `changelog-unreleased/` directory to `docs/changelog/unreleased/` in #520, and this PR still writes its fragment to the old location (`changelog-unreleased/482.md`). When the PR merges, that recreates the deleted root directory and the fragment gets stranded where release assembly won't pick it up (this already happened once with #513, fixed in #527).

ValarDragon:
lol woops, teaches us more ways to get our automated testing suite improved. Lets close for now

PR #483, opened 2026-07-29
fix(state): correct dequeue_children by_height index handling

ebfull:
## Motivation

Backport of upstream Zebra [PR #10604](https://github.com/ZcashFoundation/zebra/pull/10604).

## Root cause

`QueuedBlocks::dequeue_children` removed the entire `by_height` entry for a dequeued child's height (`self.by_height.remove(&queued.0.height)`) instead of removing just that child's hash. `by_height` is a `BTreeMap<Height, HashSet<Hash>>`, so when two queued blocks share a height — fork siblings with different parents — dequeuing one un-indexed the other. Because `prune_by_height` walks `by_height` to clean up, the un-indexed sibling could never be pruned, leaking entries in `blocks`, `by_parent`, and `known_utxos`. Leaked `known_utxos` are worse than a memory leak: a stale UTXO can answer a pending-UTXO (`AwaitUtxo`) request.

## Solution

Remove only the specific hash from the height bucket, and drop the height entry only once its set becomes empty.

## Testing

Added `dequeue_children_preserves_same_height_siblings`: queues two same-height fork grandchildren, dequeues one branch, and asserts the other branch's block survives in both `blocks` and `by_height`. Local test evidence added below.

### DIFF
--- a/changelog-unreleased/483.md
+++ b/changelog-unreleased/483.md
@@ -1,0 +1,6 @@
+## Fixed
+
+- Fixed a queued-block index leak where dequeuing one fork sibling un-indexed
+  another block at the same height, preventing it from being pruned and leaking
+  entries in the block queue and its known-UTXO cache
+  ([#483](https://github.com/zakura-core/zakura/pull/483)).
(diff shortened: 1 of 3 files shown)


ebfull:

```
cargo test --release -p zakura-state --lib dequeue_children_preserves_same_height_siblings
test result: ok. 1 passed; 0 failed; 0 ignored
```

ValarDragon:
@v12sec audit this

ValarDragon (approved):
LGTM, but waiting on the v12 report back

PR #484, opened 2026-07-29
fix(network): retry the first peer disk-cache write

ebfull:
## Motivation

Backport of upstream Zebra [PR #11073](https://github.com/ZcashFoundation/zebra/pull/11073).

## Root cause

After a cold start, the address book is usually still empty when the peer-cache updater makes its first write attempt, and the cache is only written once there are cacheable peers. The updater then slept the full `PEER_DISK_CACHE_UPDATE_INTERVAL` (5 minutes) before retrying, so a node killed within that window restarted with no cached peers and had to rediscover the network from scratch.

## Solution

Retry every `DNS_LOOKUP_TIMEOUT * 4` (20s) until the first successful non-empty write, then fall back to the normal 5-minute interval. `update_peer_cache_once` now returns `io::Result<bool>` indicating whether the cache was written (`false` when the cacheable peer list was empty, keeping any previous cache).

## Testing

Compilation checked (`cargo build --release -p zakura-network`), including the two existing `update_peer_cache_once` test callers, which discard the return value via `.expect(...)` and are unaffected by the `()` → `bool` change. Local build evidence added below.

### DIFF
--- a/changelog-unreleased/484.md
+++ b/changelog-unreleased/484.md
@@ -1,0 +1,6 @@
+## Changed
+
+- The first peer disk-cache write is now retried every 20 seconds until it
+  succeeds, instead of waiting the full 5-minute update interval, so a
+  cold-started node caches its peers soon after finding them
+  ([#484](https://github.com/zakura-core/zakura/pull/484)).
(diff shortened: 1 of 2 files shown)


ebfull:

```
cargo build --release -p zakura-network --tests
Finished `release` profile [optimized + debuginfo] target(s)
```

Compiles cleanly including the existing `update_peer_cache_once` test callers (return type changed `()` → `bool`; both callers discard via `.expect(...)`).

ValarDragon:
Approving + merging for short term problem. Re-flagging, this addrbook cache method is silly / needs to be re-designed

ValarDragon (approved):

PR #486, opened 2026-07-29
fix(rpc): apply backpressure on indexer gRPC streams instead of dropping subscriptions

ebfull:
## Motivation

Backport of upstream Zebra [PR #10932](https://github.com/ZcashFoundation/zebra/pull/10932).

## Root cause

The indexer gRPC `chain_tip_change` and `mempool_change` streams used `try_send` into a 64-message buffer and dropped the **entire subscription** the instant that buffer filled. A momentarily slow consumer (e.g. a wallet rescan) therefore triggered rapid re-subscribe cycles and a flood of WARN-level "slow consumer" log lines. Only the `non_finalized_state_change` stream already applied timeout-bounded backpressure.

## Solution

- Switch `chain_tip_change` and `mempool_change` to `send().await` bounded by a 60s timeout, matching the non-finalized stream, so a slow consumer gets grace instead of an instant drop. The shared timeout constant is renamed `NON_FINALIZED_SEND_TIMEOUT` → `SEND_TIMEOUT`.
- Add a completeness guard to `non_finalized_state_change`: when the listener buffer is full (`capacity() == 0`), the state-side task may already have missed non-finalized updates, so the stream can no longer guarantee completeness and drops the subscription rather than silently skipping blocks.

Zakura's `NonFinalizedBlocksListener::unwrap()` yields a plain `tokio::sync::mpsc::Receiver`, which exposes `capacity()` directly, so the guard ports without touching the listener type.

## Testing

Compilation of `zakura-rpc` checked (evidence below). No API change; slow consumers now get 60s of backpressure grace instead of an immediate drop.

### DIFF
--- a/crates/zakura-rpc/src/indexer/methods.rs
+++ b/crates/zakura-rpc/src/indexer/methods.rs
@@ -23,13 +23,13 @@ use super::{
 /// The maximum number of messages that can be queued to be streamed to a client.
 const RESPONSE_BUFFER_SIZE: usize = 64;
 
-/// How long to wait for a backpressured send to the non-finalized stream before
-/// treating the consumer as hung and dropping the subscription.
+/// How long to wait for a backpressured send before treating the consumer as hung
+/// and dropping the subscription.
 ///
-/// The non-finalized stream applies backpressure rather than dropping blocks,
-/// so a slow consumer doesn't miss blocks. Without a bound, a half-open
-/// connection could block the listener task indefinitely.
-const NON_FINALIZED_SEND_TIMEOUT: Duration = Duration::from_secs(60);
+/// All three indexer streams apply backpressure so a slow consumer doesn't miss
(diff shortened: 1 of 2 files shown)


ebfull:

```
cargo build --release -p zakura-rpc
Finished `release` profile [optimized + debuginfo] target(s)
```

Compiles cleanly; the completeness guard uses `capacity()` on the plain `tokio::sync::mpsc::Receiver` from `NonFinalizedBlocksListener::unwrap()`.

ValarDragon:
@v12-auditor pls review

evan-forbes (approved):

ebfull:
@v12sec pls review

ValarDragon:
@v12 pls review

ValarDragon:
@v12sec pls review

ValarDragon:
We need to do some more rewrites for this area of the code

PR #487, opened 2026-07-29
perf(consensus): Arc-wrap the halo2 batch item bundle

ebfull:
## Motivation

Backport of the halo2 hunk of upstream Zebra [PR #10886](https://github.com/ZcashFoundation/zebra/pull/10886).

## Root cause

`tower-fallback` clones every request eagerly (so it can retry against the fallback service on batch failure). The halo2 batch `Item` held its Orchard/Ironwood bundle by value, so each clone deep-copied all of the bundle's actions plus its multi-kilobyte proof — on the proof-verification hot path.

## Solution

Wrap the bundle in an `Arc` so cloning an `Item` shares it instead of deep-copying. `BatchValidator::add_bundle` only needs `&Bundle` (now `bundle.as_ref()`), and `RequestWeight::request_weight` still reads `actions()` through the `Arc` via `Deref`. No semantic change.

## Testing

Built and ran the `zakura-consensus` halo2 unit tests (evidence below).

### DIFF
--- a/changelog-unreleased/487.md
+++ b/changelog-unreleased/487.md
@@ -1,0 +1,5 @@
+<!-- changelog: none -->
+
+Internal performance change only: the halo2 batch item now shares its bundle
+via `Arc` instead of deep-copying it on every eager `tower-fallback` clone.
+No operator- or crate-consumer-visible behavior changes.
--- a/changelog-unreleased/487.md
+++ b/changelog-unreleased/487.md
@@ -1,5 +1,6 @@
 <!-- changelog: none -->
 
 Internal performance change only: the halo2 batch item now shares its bundle
-via `Arc` instead of deep-copying it on every eager `tower-fallback` clone.
+via `Arc` instead of deep-copying it on every eager `tower-fallback` clone,
+saving about 1.5 microseconds per Halo2 item in a representative benchmark.
 No operator- or crate-consumer-visible behavior changes.
--- a/zakura-consensus/benches/halo2.rs
+++ b/zakura-consensus/benches/halo2.rs
@@ -29,16 +29,26 @@
 
 mod common;
 
-use std::sync::Arc;
+use std::{collections::BTreeMap, sync::Arc};
 
-use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use criterion::{
+    black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion, Throughput,
+};
+use orchard::bundle::{Authorized, Bundle};
 
 use zakura_chain::{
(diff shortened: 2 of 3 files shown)


ebfull:

```
cargo test --release -p zakura-consensus --lib primitives::halo2
test result: ok. 4 passed; 0 failed; 0 ignored
```

changelog-unreleased/487.md
```diff
@@ -0,0 +1,5 @@
+<!-- changelog: none -->
+
+Internal performance change only: the halo2 batch item now shares its bundle
+via `Arc` instead of deep-copying it on every eager `tower-fallback` clone.
+No operator- or crate-consumer-visible behavior changes.
```
ValarDragon, line 6: something's wrong with our changelog AI, perf improvements should appear in changelog

ValarDragon:
Gonna add a bench to this - i have a somewhat hard time imagining this matters. It could potentially slow things down if we had to add things to heap

ValarDragon:
• The valid saved distributions are decisive: mean clone/drop cost fell from 1,271 ns to 30.2 ns, a 42.0× speedup (97.6% less time). Their 95% confidence intervals are 1,235–1,309 ns
  before and 29.6–30.8 ns after, with no overlap. I’m cleaning up only the temporary parent worktree; the benchmark remains as the sole uncommitted change on backport/halo2-item-arc.
  
Ok so a real savings for this part but negligible overall. (1 microsecond saved per action)

ValarDragon:
## Halo2 `Item` lifecycle benchmark

Added a focused Criterion benchmark in commit [`175cb5692`](https://github.com/zakura-core/zakura/commit/175cb5692). The identical benchmark was run against the parent of the `Arc` change and against this PR, using a real two-action mainnet test-vector bundle.

| operation | before | after | result |
|---|---:|---:|---:|
| `Item::clone` + cloned-item drop | 1,192 ns | 21.3 ns | **98.34% faster** |
| `Item::new` + final item drop | 799 ns | 206 ns | **about 70.7% faster** |
| approximate combined lifecycle | 1,991 ns | 227 ns | **about 8.8x faster** |

Criterion configuration: 50 samples, 2-second warmup, 5-second measurement. Both direct comparisons were statistically significant (`p < 0.05`).

The clone benchmark includes the relevant memory accounting: before this PR it includes deep-copy allocation/copy plus freeing the copied bundle and proof; after this PR it includes the `Arc` atomic increment/decrement. The construction benchmark excludes preparation of the raw input bundle, but includes `Arc::new` allocation and the final drop/free.

So the improvement is not just moving cost from clone into construction. Even accounting for allocation and destruction, the representative item lifecycle saves about 1.76 microseconds. The changelog rounds this conservatively to about 1.5 microseconds per Halo2 item.

ValarDragon (approved):

PR #497, opened 2026-07-31
fix(network): finish broadcasts after peer disconnects

evan-forbes:
## Motivation

`PeerSet::broadcast_all` snapshots peers that are unready when a mined-block advertisement is sent. If one of those peers disconnects before becoming ready, its address disappears from the live peer maps but remains in `queued_broadcast_all`. The retained sender prevents the broadcast response channel from closing, so the response waits until its outer timeout instead of completing.

The bounded queued-future channel had a second liveness edge: after three readiness waves, `try_reserve()` could report a full channel without registering a wakeup for newly available capacity. Canceled response futures also left closed senders queued because reservation failures were treated as temporary backpressure.

Three value-summation arithmetic errors were also assigned a peer-misbehavior score even though they can be produced by local bookkeeping faults. That could ban an honest block supplier for a node-local failure.

## Solution

- Prune banned and disconnected addresses from queued broadcasts using the current ready and unready peer maps.
- Run queued-broadcast cleanup before the no-ready-peers early return.
- Drop queued work when its response receiver has closed.
- Avoid producing empty delivery futures while every remaining peer is still unready.
- Use an unbounded future handoff whose total contents are bounded by the original peer snapshot, avoiding unwakeable channel-capacity stalls.
- Add regression coverage that local header-root frontier faults retain a zero peer-misbehavior score.
- Keep residual value-summation arithmetic faults unscored while preserving scores for block-proven consensus violations.

## Testing

- `git diff --check`
- `npx --yes markdownlint-cli docs/changelog/unreleased/497.md --config .trunk/configs/.markdownlint.yaml`
- `cargo test -p zakura-network broadcast_all_queued --lib --locked` (5 passed)
- `cargo test -p zakura-network --lib --locked` (1,173 passed, 4 ignored)
- `cargo test -p zakura-state commit_block_error_misbehavior_scores --locked` (1 passed)
- `cargo test -p zakura-state --lib --locked` (462 passed, 4 ignored)
- `cargo test -p zakura-consensus --lib transaction::tests::v4_and_v5_with_sapling_spends --locked -- --exact` (1 passed)

The regressions cover an only peer disconnecting while unready, caller cancellation, four readiness waves without polling the response receiver, and both local-state misbehavior classifications. Existing delivery tests cover both ready and originally-unready peers, including keeping queued response futures alive until the connection consumes each request.

The full local consensus test binary intermittently cancels `v4_and_v5_with_sapling_spends` when run concurrently with the other tests; the test passes when isolated. The unrelated `inbound_pruned_block_is_not_advertised_and_getdata_logs_error` log-capture assertion also fails locally, including when isolated. GitHub checks are the authoritative clean-environment result for those broader suites.

## Changelog

Added `docs/changelog/unreleased/497.md` under `Fixed`.

### DIFF
--- a/crates/zakura-state/src/error.rs
+++ b/crates/zakura-state/src/error.rs
@@ -893,6 +893,8 @@ impl ValidateContextError {
     // peer-attribution decision.
     fn misbehavior_score(&self) -> u32 {
         match self {
+            // Consensus violations the block itself proves: the supplier sent a
+            // block that no honest peer could have produced.
             ValidateContextError::NonSequentialBlock { .. }
             | ValidateContextError::TimeTooEarly { .. }
             | ValidateContextError::TimeTooLate { .. }
@@ -907,9 +909,6 @@ impl ValidateContextError {
             | ValidateContextError::DuplicateOrchardNullifier { .. }
             | ValidateContextError::DuplicateIronwoodNullifier { .. }
             | ValidateContextError::NegativeRemainingTransactionValue { .. }
-            | ValidateContextError::CalculateRemainingTransactionValue { .. }
-            | ValidateContextError::CalculateTransactionValueBalances { .. }
-            | ValidateContextError::CalculateBlockChainValueChange { .. }
             | ValidateContextError::AddValuePool { .. }
             | ValidateContextError::InvalidBlockCommitment(_)
             | ValidateContextError::UnknownSproutAnchor { .. }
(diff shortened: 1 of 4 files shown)


evan-forbes:
@v12-auditor

evan-forbes:
@v12-auditor

crates/zakura-state/src/error.rs
```diff
@@ -907,17 +909,31 @@ impl ValidateContextError {
             | ValidateContextError::DuplicateOrchardNullifier { .. }
             | ValidateContextError::DuplicateIronwoodNullifier { .. }
             | ValidateContextError::NegativeRemainingTransactionValue { .. }
-            | ValidateContextError::CalculateRemainingTransactionValue { .. }
-            | ValidateContextError::CalculateTransactionValueBalances { .. }
-            | ValidateContextError::CalculateBlockChainValueChange { .. }
```
p0mvn, line 912: This is the questionable reclassification: malformed block contents can directly cause this error, so changing it to score 0 may allow a peer supplying an invalid block to avoid a penalty.

This variant currently combines two different situations:

- Invalid block data caused the arithmetic failure.
- Local state or an implementation invariant caused it.

Treating both as local (0) is too permissive. The best long-term fix is to split the error into peer-attributable and local-invariant variants. For this PR, the safest minimal change is:

- CalculateRemainingTransactionValue → 0
- CalculateTransactionValueBalances → 0
- CalculateBlockChainValueChange → keep 100

That preserves existing enforcement until the code can distinguish the two block-level causes reliably.

p0mvn (approved):

ValarDragon:
@v12-auditor pls review

ValarDragon:
@v12-auditor pls review

ValarDragon:
@v12sec pls review

crates/zakura-state/src/error.rs
evan-forbes: > malformed block contents can directly cause this error, so changing it to score 0 may allow a peer supplying an invalid block to avoid a penalty.

I added a comment to explain more but these were originally not added as they can't get hit unless there is a bug since all the other errors are caught before hand. Therefore is this statement fully true? perhaps I'm missing how we actually trigger those errors in this scenario

still though we likely should disconnect and assume there isn't a bug. however I think we actually want to do the opposite of

```
CalculateRemainingTransactionValue → 0
CalculateTransactionValueBalances → 0
CalculateBlockChainValueChange → keep 100
```

in order to get that outcome we want no?

```
CalculateRemainingTransactionValue → 100
CalculateTransactionValueBalances → 100
CalculateBlockChainValueChange → 0
```

evan-forbes: > in order to get that outcome we want no?

this being defensive and assuming that a peer is being malicious and found a way to get past the preliminary checks

PR #418, opened 2026-07-23
ci: add automated release-prep draft PR workflow

p0mvn:
## Motivation

The first step of every release — preparing the release PR (crate version bumps, `zakura` version, lockfile, stored config fixture, end-of-support height, changelog assembly; see #337) — is manual today and varies release-by-release. This adds a CI job that performs the mechanical parts for a given tag, verifies the result with `make pre-release RELEASE_TAG=… BASE_TAG=…`, and opens a draft PR through the release bot against freshly-rebased `main` (same bot-PR pattern as `update-release-state.yml` / #317).

## Solution

- **`scripts/prepare-release.sh`** — all mutation logic, equally runnable locally (`make prepare-release RELEASE_TAG=… [BASE_TAG=…] [NO_CRATES=1] [DRY_RUN=1]`):
  - Computes changed publishable crates since `BASE_TAG` (default: most recent `v*` tag; loop lifted from `check-crate-version-bumps.sh`).
  - Chooses bump levels with `cargo-semver-checks --baseline-rev <BASE_TAG>`: breaking → major, anything else → minor (patch is never auto-selected; the git baseline is deliberate — a base-tag crate version may never have been published to crates.io). Crates already bumped on `main` keep their level and only get the prerelease suffix aligned (handles rc→rc and rc→stable uniformly); new crates are reported for manual versioning.
  - Applies bumps with `cargo release version` (explicit target versions; `dependent-version = "fix"` rewrites internal requirements), bumps `zakura` to the tag version, refreshes `Cargo.lock`, regenerates the stored config fixture, updates the README install tag via `cargo release replace` (stable only), floors `ESTIMATED_RELEASE_HEIGHT` from the committed checkpoint list, and runs `make prepare-release-changelog`.
  - Analysis is fully separated from mutation, so `--dry-run` is exact.
[...]

PR #500, opened 2026-08-01
ci: prepare and merge the release PR from create-release

p0mvn:
## Motivation

PR #418 added `prepare-release-pr.yml`, which mechanically prepares a release and opens a draft PR. `create-release.yml` never looked at that PR: it validated whatever was committed at the dispatched ref and failed if preparation had not landed. Every release was therefore two dispatches with a human merge in between.

This makes `Create release` a single dispatch when that is what you want, without taking away the faster manual path.

## Solution

`create-release.yml` resolves preparation before anything else:

- **`prepare-check`** compares the `zakura` package version at the dispatched ref to `release_tag`. Equal → preparation is skipped and the workflow behaves exactly as it always has. Otherwise it fails fast off `main`, under `source_first_release`, with `auto_prepare` unchecked, when an open `release/<tag>` PR already claims the version, or when the tag exists.
- **`prepare`** calls `prepare-release-pr.yml` with the new `auto_merge` input.
- **`validate`** takes the merge commit as `target_sha` (falling back to the dispatch commit when nothing was prepared) and re-runs the full `make pre-release` against it. `handoff-canary`, `build-assets`, and `publish` needed no changes — they already consume `validate.outputs.target_sha`.

`prepare-release-pr.yml` gains a `workflow_call` trigger, `merged_sha`/`head_sha`/`pull_request_number` outputs, and `auto_merge` (also on `workflow_dispatch`, so the merge can be rehearsed without running a release). Under `auto_merge` the three verification steps stop being advisory: the PR opens ready for review only when all three pass, and any failure leaves it as a draft and fails the run without merging. The merge itself waits for GitHub to compute mergeability, squash-merges with `--match-head-commit` pinned to the commit the run pushed, and asserts `main` is still at the merge commit before handing it back.

### Prerequisite

The `main` ruleset requires one approving review and the App cannot approve a PR it authored, so `zakura-release-bot` must be added to that ruleset's bypass list with **Pull requests only** — it can then merge an unapproved PR but still cannot push to `main` directly. Without it the auto path fails at the merge with HTTP 405 and nothing else is affected.

### Trade-off

On the automated path the release diff (version bumps, assembled changelog, end-of-support floor) merges with no human reading it; the PR body says so in place of the usual judgment checklist. The human gate for the release moves to the required reviewers on the protected `release` environment, which approve after the assets are built and before the tag exists. Preparing the release explicitly and reviewing the PR remains the documented path when those judgment items need a pass.

## Testing

Not exercised against real infrastructure — the auto path cannot run until the ruleset bypass entry exists, and rehearsing it consumes a release tag. Ran locally:

- `actionlint` on both changed workflows — clean (the one repo-wide failure, `test-docker.yml` SC2329, predates this branch)
- `markdownlint` with `.trunk/configs/.markdownlint.yaml` on the changed Markdown — clean
- `./scripts/changelog.py check` — 11 fragments validated

Planned verification, in order:

1. Dispatch **Prepare release PR** with `dry_run` on an rc tag — confirms the `workflow_call` refactor did not break the dispatch path.
2. After the bypass entry exists, dispatch it with `auto_merge` on a throwaway rc tag — exercises `gh pr ready`, the mergeability poll, `--match-head-commit`, branch deletion, and `merged_sha` in isolation from the release.
3. Dispatch **Create release** for a tag already prepared on `main` — `prepare` must be `skipped` and `target_sha` must equal the dispatch commit, proving the manual path is untouched.
4. Full end-to-end on the next rc, plus the negative cases (open `release/<tag>` PR present; `auto_prepare` off with preparation missing).

## Changelog

No fragment: CI and docs only, no Rust source or `Cargo.toml` changes (same as #418).

- #418 — the release-prep draft PR workflow this builds on
- `docs/release-tag-protection.md` — updated with the `main` bypass, the `release-automation` environment, and the corrected App permission list (it still claimed Contents-only, but #418 already mints `pull-requests: write`)

### Follow-up Work

- The bypass entry must be added by a repo admin before the auto path works.
- `scripts/release-t0.sh` still drives the human-prepared path only: without `--pr` it requires `main` to already be at `--head-sha`, which is never true before an automated preparation has merged. The skill documents the raw dispatch for the auto path; teaching the orchestrator about it is separate work.


p0mvn:
### Required repository-admin setup

Before the automated release-preparation path can merge, add the release App to the `main` ruleset bypass list:

1. Open the [`main` ruleset (ID 17939034)](https://github.com/zakura-core/zakura/settings/rules/17939034).
2. In **Bypass list**, select **Add bypass**.
3. Find and add the `zakura-release-bot` GitHub App.
4. Set its bypass mode to **Pull requests only**.
5. Save the ruleset.

This allows the App to merge its release PR without the otherwise-required approving review, while still preventing it from pushing directly to `main`. The App cannot approve its own PR or add itself to the bypass list, so a repository admin must perform this one-time setup. Until then, the auto path reaches the merge and fails with HTTP 405; the existing manually prepared release path is unaffected.

p0mvn:
I think this automation should wait for the preparation workflow to get more optimized

PR #506, opened 2026-08-01
fix(network): key outbound replenishment to the peer set target size

evan-forbes:
## Motivation

Nodes on the CI fleet hold roughly a third of their configured peer set size
outbound: `peerset_initial_target_size` is 100, but `temp-zakura-sync-test-4`
through `-7` sit at 25-37 outbound connections. The outbound connection limit is
150, so the cap is not binding, and dialing is nowhere near the
`MIN_OUTBOUND_PEER_CONNECTION_INTERVAL` rate limit — the observed rate is about
0.06 dials/second against an allowed 10/second. There are 694 never-attempted
gossiped addresses sitting in the address book. The node simply is not converting
candidates into connections.

The root cause is in `outbound_peer_replenishment_demand()`. The timer-based
crawler refreshes its dial budget from that function every
`crawl_new_peer_interval` (61 s), and the function targets **27% of the outbound
connection limit** rather than 27% of the configured peer set size.

The outbound connection limit is a safety ceiling derived from
`peerset_initial_target_size`. #478 halved it (`OUTBOUND_PEER_LIMIT_DIVISOR`,
300 → 150 at the default target size) to reduce connections this node does not
need. Because the replenishment target was keyed to that ceiling, halving it also
halved the steady-state outbound peer count:

| | outbound limit | replenishment target | budget at 34 active |
| --- | ---: | ---: | ---: |
| before #478 | 300 | 81 | **47** |
| after #478 | 150 | 41 | **7** |

Measured on `temp-zakura-sync-test-6`: 34 outbound connections and **exactly 7
dial attempts** in a 122-second window, matching the predicted budget.

A safety ceiling and a steady-state target should not share one number. The
arithmetic makes the original intent visible: 27% of the old limit, where the old
limit was `peerset_initial_target_size * 3`, is 81% of
`peerset_initial_target_size`. #462 meant "keep outbound near the configured peer
set size", expressed against the wrong denominator.

## Solution

`outbound_peer_replenishment_demand()` takes the peer set target size and keys
the replenishment target to it, at 80%. The outbound connection limit stays in
the signature as a bound only, so an operator who lowers the ceiling below the
target does not get a crawler dialing forever against the cap.

At the defaults this restores a target of 80 and a budget of 46 at 34 active
connections — the pre-#478 dialing pressure — without reverting #478's
inbound/outbound rebalance, which addressed a different problem.

80% rather than the incidental 81% keeps the constant a round share of the
operator's configured size. Say so in review if you would rather the restoration
be bit-identical to the pre-#478 value.

### The budget is now logged at info

A second commit promotes the per-crawl accounting from `debug!` to `info!`.

This is not a convenience. Release builds enable the `max_level_info` tracing
feature, which compiles debug records out entirely, so this record could not be
read on a deployed node whatever the configured filter said — and this function
is the *only* place the crawler's view of its own outbound connections is
visible. Without it, a node that has stopped replenishing looks exactly like a
node that has enough peers. The original version of this PR shipped a
verification signature that could not be verified.

One record per `crawl_new_peer_interval` bounds the volume. On a fleet node:

```text
timer-based crawl finished active_outbound_connections=39 peerset_target_size=100
outbound_connection_limit=150 remaining_replenishment_demand=41
```

## Testing

- `cargo test -p zakura-network --lib` — 1193 passed, 0 failed.

The unit table in
`crawler_replenishes_outbound_peers_below_eighty_percent_of_target_size` now
varies the peer set target size and the connection limit independently, where it
previously could only vary the limit.

Two new tests exercise the root cause directly:

- `crawler_replenishment_ignores_outbound_connection_limit_changes` holds the
  peer set target size fixed and changes only the outbound connection limit
  between the pre-#478 and post-#478 values, asserting the demand does not move.
  This is precisely the coupling that caused the regression: under the old rule
  the same two limits produced 81 and 41.
- `crawler_replenishment_is_bounded_by_the_outbound_connection_limit` covers the
  retained clamp, including a `usize::MAX` target size.

The existing async `crawler_*` race tests are unchanged apart from
`CRAWLER_REPLENISHMENT_CONNECTION_TARGET_FOR_TESTS`, which is now derived from
the test peer set target size (10) instead of the test outbound limit (15). Its
value moves 5 → 8 and flows into the dependent assertions.
`crawler_replenishment_respects_hard_outbound_limit` asserts against the outbound
limit, which this PR does not touch, and passes unchanged.

## Fleet verification

Deployed as `test/combined-506-507` on two mainnet nodes for three days. The
budget behaves exactly as designed — asia-0 logs a crawl every 61 s with
`remaining_replenishment_demand=41` against `active_outbound_connections=39`,
where the old rule would have computed 2.

**The outbound peer count did not recover.** It is flat at 32-39 against a target
of 80, i.e. roughly 59k dial attempts per day for a net gain of 4 connections
over 24 h.

That is not this PR failing to do what it claims — the dial budget is restored
and the crawler is spending it. The binding constraint moved: the recent
handshake failure mix on asia-0 is 138 timeout / 90 peer-closed / 43 refused /
36 reset against 77 obsolete-version, so what limits the peer count now is the
supply of reachable, willing peers rather than dialing pressure. See the
follow-up note below.

### Correcting an earlier claim in this PR

An earlier revision of this description gave the verification signature as "dial
attempts per crawl rise from ~7 to ~46 and outbound climbs toward 80". The first
half holds; **the second does not**, and three days of fleet data say so. The
honest signature for this PR is the logged budget, not the peer count.

## Changelog

Parameter row added to `docs/changelog/params.md`; fragment at
`docs/changelog/unreleased/506.md`.

- Introduced in #462, regressed by #478.
- Diagnosis and measurements: outbound peer depletion investigation on the CI
  fleet, 2026-08-01 through 2026-08-04.

### Follow-up Work

1. **Post-NU6.3 outbound supply.** Nodes that restarted after Ironwood activation
   (2026-07-28T14:07:23Z) converge to ~35-40 outbound, while long-lived nodes
   that came up before it hold 66-115. Established connections are never
   re-version-checked, so the older nodes are holding peers they could not
   re-establish today. Their peer counts are not a reproducible target, and the
   gap is not a dialing problem. This needs its own investigation.
2. **Peer discovery coverage.** `GET_ADDR_FANOUT` is 1 and `sanitized()` excludes
   inbound entries, so each crawl yields very few genuinely new addresses. A
   restored dial budget cannot help if the candidate pool does not grow.

Related: #507 fixes address book and disk cache pollution, an independent cause
of the same symptom.


ebfull:
Heads-up: changelog fragments moved from the root `changelog-unreleased/` directory to `docs/changelog/unreleased/` in #520, and this PR still writes its fragment to the old location (`changelog-unreleased/506.md`). When the PR merges, that recreates the deleted root directory and the fragment gets stranded where release assembly won't pick it up (this already happened once with #513, fixed in #527).

ValarDragon (approved):
Checked out this makes sense. 

TBH I think we can remove this 80% cap as well. Maybe what we do in a later pass is just consider 80% as what to hit fast, and then the last 20% we fill over time?

ValarDragon:
Great job on this one!

PR #507, opened 2026-08-01
fix(network): stop dialing and caching inbound peer addresses

evan-forbes:
## Motivation

Nodes on the CI fleet hold a fraction of their configured outbound peer set, and
the peer disk cache explains most of it. Of the 206 entries in
`/var/lib/zakura/network/mainnet.peers` on `temp-zakura-sync-test-6`, only
**43 (20.9%) are on port 8233**. The rest look like this:

```text
100.36.20.73:60642
104.199.175.115:34448
104.199.175.115:35822
104.199.175.115:38290
```

One IP with several different high ports is unmistakable: those are the *source*
ports of inbound TCP connections, stored as though they were dialable listen
addresses. That node's lifetime handshake success rate was 20.7%.

They enter the address book because `ConnectedAddr::get_address_book_addr()`
returns the inbound socket's remote address, which becomes a `MetaAddr` with
`is_inbound = true`. The flag is set correctly; it is just not honoured anywhere
that matters. This is inherited from upstream, not a recent regression —
`get_address_book_addr()` is byte-identical in `zebra-upstream/main`.

Three separate mechanisms turn that into undialable peers, and fixing only the
first is not enough unfortunatley

## Solution

Four commits, each with its own failure mode and its own tests.

### 1. Honour `is_inbound` on the dial and cache paths

`is_ready_for_connection_attempt()` and `AddressBook::cacheable()` both ignored
the flag, while `MetaAddr::sanitize()` honoured it. Inbound peers are connected
*right now*, so they score as maximally active and dominate cache ordering.

### 2. Only cache addresses that answered one of our own connections

**This is what actually cleaned the cache, and the reason commit 1 alone was not
enough.** The `is_inbound` flag lives on the address book entry, and *nothing on
that entry survives the disk cache* — the file holds bare socket addresses.
`Config::initial_peers()` chains them with the DNS seeds, so a cached address is
re-added through `MetaAddr::new_initial_peer()` with no inbound flag, no
services, and no history, then dialed directly at startup.

Worse, the same junk arrives by **gossip** with `is_inbound == false`, so neither
guard from commit 1 ever sees it. Measured on a node running commit 1 only: the
cache was wiped to 0 and refilled to **82% ephemeral (88/107) within one
minute**.

`cacheable()` now also requires `MetaAddr::has_ever_responded()`, and
`load_peer_cache()` drops entries in the ephemeral port range, using a new
`EPHEMERAL_PORT_FLOOR` of 32768.

#### Why the floor, and not the network's default port

An earlier revision of this branch filtered the cache on
`port == network.default_port()`. That is wrong, and the fleet says so. Counting
`zakurad`'s own established outbound connections — peers it dialed and
handshaked, so unambiguously real listeners:

| node | on :8233 | on other ports | non-default share |
| --- | ---: | ---: | ---: |
| canada-0 | 95 | 16 | 14.4% |
| us-east-0 | 81 | 16 | 16.5% |
| us-0 | 73 | 17 | ~16% |

Ports observed carrying real peers: 7952, 8131, 8231, 8234, 8235, 8236, 8243,
8433, 9058, 9222, 9533, 18233, 21099, 28233. A default-port filter would drop
about one in six good peers from the bootstrap path — and worse, it would be
asymmetric with `cacheable()`, which has no port check, so the node would keep
writing those peers to the cache for the next startup to discard.

The ephemeral floor separates the two populations cleanly on live data: every
port above carries a real listener and is below 32768, while every inbound source
port found in a cache (57266, 55224, 44330, 43546, 37336, …) is above it. Linux
defaults `net.ipv4.ip_local_port_range` to `32768 60999` and other platforms
allocate from `49152-65535`, so 32768 is the inclusive lower bound across both.
Port 0 is rejected separately.

### 3. Stop inbound connections from blocking dials to the same IP

`AddressBook::most_recent_by_ip` records the responded address with the newest
`last_response` per IP, and `reconnection_peers()` refuses to dial any address
whose IP has a recent entry there. The entry does not record direction.

An inbound peer's entry is refreshed by every message it sends, so it holds its
IP's slot for the whole life of the connection — and the address it blocks is a
*different* address on that IP: the peer's listener, which is exactly what we
want to dial. Peers that connect to us are the most active on the network, so
this removes the best candidates, and removes more of them the more inbound
connections we accept. #478 raised `INBOUND_PEER_LIMIT_MULTIPLIER` from 1 to 3,
tripling the number of IPs a node bars itself from.

The cache exists to space out *our own* outbound connections to one IP, and
`max_connections_per_ip` is only enforced in `accept_inbound_connections()`,
never on the dial path — so an inbound connection was suppressing dials to
enforce a limit that would not have rejected them.

Measured before this commit: 300 inbound connections, 9 outbound, a
replenishment demand of 70, and **zero connection attempts over 20 minutes**.
The crawler was not failing to dial; it had no candidate left to dial.

## Correcting an earlier claim in this PR

An earlier revision of this description said existing polluted entries need no
migration because "the cache is rewritten without them within
`PEER_DISK_CACHE_UPDATE_INTERVAL`". **That was wrong**, and the fleet showed it:
because `load_peer_cache()` returns bare addresses with no flags, a polluted
cache launders itself clean across every restart. Commit 2's port filter on load
is what makes the claim true.

### Reviewer's attention: the flag is sticky

`MetaAddrChange` merges OR `is_inbound` with the previous value, so an address
that was ever inbound is never dialed again on that entry. A peer that dialed us
*from its own listener port* would not be dialed back on that entry. Outbound
connections use ephemeral source ports — zcashd and Zebra both do — and the same
IP on its real listener port is a *separate* address book key. Commit 3 relies on
this stickiness deliberately, and both behaviours are pinned by tests.

`maybe_connected_peers()` is defined as the complement of
`is_ready_for_connection_attempt()`, so inbound peers now fall into it. That
reads correctly — they are connected — and the method has no callers in the
workspace.

## Testing

- `cargo test -p zakura-network --lib` — 1197 passed, 0 failed.

New tests, each verified to fail with its own source change reverted:

- `inbound_addresses_are_not_ready_for_connection_attempts` — two peers differing
  only in the `is_inbound` argument, checked past `MIN_PEER_RECONNECTION_DELAY`
  so `was_recently_updated` cannot be what separates them.
- `inbound_flag_keeps_an_address_out_of_dial_candidates` — applies a later
  `new_responded` change, the update most likely to look like fresh evidence the
  address is good, and asserts it stays out of the candidate set.
- `inbound_peers_are_not_reconnection_or_cache_candidates` — asserts both peers
  are `is_active_for_gossip` first, so the existing filter is not what does the
  work.
- `inbound_connections_do_not_block_dialing_the_same_ip` — puts a live inbound
  connection and its peer's listener on one IP, plus an address sharing an IP
  with a peer we dialed ourselves, and asserts only the listener is a
  reconnection candidate.
- `cached_peers_on_ephemeral_ports_are_not_loaded` seeds a cache with listeners
  on 8233, 8235, 21099 and 28233, a port 0 entry, and three ephemeral source
  ports from one IP, then asserts only the last four are dropped. It fails under
  the default-port predicate, so it pins the choice above rather than just the
  ephemeral filtering.

`address_book/tests/prop.rs` asserts every peer from `reconnection_peers` is
`is_probably_reachable`; this change only narrows that set.

## Fleet verification

Deployed as `test/combined-506-507` on two mainnet nodes (`asia-0`,
`europe-central-0`) against seven controls on `1.0.5`, sustained over three days:

| | cache entries | on :8233 | ephemeral |
| --- | ---: | ---: | ---: |
| patched (asia-0) | 38 | 28 | **0** |
| patched (europe-central-0) | 33 | 26 | **0** |
| controls | 74 | 22-27 | **33-44** |

The cache hygiene goal is met. Outbound peer count did **not** recover — it sits
at 32-39 against a target of 80 — but that is now bound by the supply of
reachable peers rather than by address hygiene: the recent handshake failure mix
on asia-0 is 138 timeout / 90 peer-closed / 43 refused / 36 reset against 77
obsolete-version. See the follow-up note below.

## Changelog

Fragment at `docs/changelog/unreleased/507.md`. No tunable parameter changed, so
`docs/changelog/params.md` is untouched.

### Follow-up Work

Deliberately out of scope:

1. **Stop recording the inbound socket address at all** — returning `None` for
   `InboundDirect` in `get_address_book_addr()`, or recording the peer's
   advertised listener from its `version` message. That is the root fix, but
   `book_addr` also keys `new_responded` / `new_errored` / `new_shutdown` and
   misbehavior tracking.
2. **`is_probably_reachable()` keeping failed addresses alive.** The same pattern
   can still hold a repeatedly-failing gossiped address in the retry pool.
3. **Post-NU6.3 outbound supply.** Nodes that restarted after Ironwood activation
   (2026-07-28T14:07:23Z) converge to ~35-40 outbound because established
   connections are never re-version-checked, so long-lived nodes keep peers they
   could not re-establish today. This is not an address-hygiene problem and needs
   its own investigation.
4. **Address book dial-candidate tracing**, which produced the measurements
   above, is split into its own PR stacked on this one.

Related: #506 fixes the crawler's replenishment budget, an independent cause of
the same symptom.

### DIFF
--- a/crates/zakura-network/src/address_book.rs
+++ b/crates/zakura-network/src/address_book.rs
@@ -601,6 +601,21 @@ impl AddressBook {
             return false;
         };
 
+        // # Correctness
+        //
+        // This cache exists to space out our own outbound connections to one IP, and
+        // `Config::max_connections_per_ip` is only enforced when we accept inbound
+        // connections, never when we dial. An inbound peer refreshes its entry with
+        // every message it sends, so it would hold its IP's slot for as long as it
+        // stays connected, and the address we would refuse to dial is a different
+        // address on that IP: the peer's own listener, which is the address we want.
+        // Recording inbound entries here stops us dialing exactly the peers that are
+        // most active on the network, and the more inbound connections we accept, the
+        // fewer peers we are able to reach.
+        if updated.is_inbound() {
+            return false;
+        }
+
         if let Some(previous) = most_recent_by_ip.get(&updated.addr.ip()) {
             updated.last_connection_state == PeerAddrState::Responded
                 && updated.last_response() > previous.last_response()
@@ -869,6 +884,12 @@ impl AddressBook {
             // If there's no entry for this IP, any connection is allowed
             return true;
         };
+        if same_ip_peer.is_inbound() {
+            // An address can become inbound after it is recorded here, because
+            // `is_inbound` is sticky. Ignore it once it does, for the reasons in
+            // `should_update_most_recent_by_ip`.
+            return true;
+        }
         !same_ip_peer.has_connection_recently_responded(chrono_now)
     }
 
--- a/crates/zakura-network/src/address_book.rs
+++ b/crates/zakura-network/src/address_book.rs
@@ -553,6 +553,14 @@ impl AddressBook {
             // This prevents Zebra from caching nodes that are likely unreachable,
             // which improves startup time and reliability.
             .filter(|addr| addr.is_active_for_gossip(now))
+            // # Security
+            //
+            // Remove addresses learned from inbound connections, whose ports are
+            // the peer's ephemeral source port rather than its listener. They
+            // are connected right now, so they rank as maximally active and
+            // would otherwise dominate the cache, leaving a restarted node with
+            // mostly undialable candidates.
+            .filter(|addr| !addr.is_inbound())
             .cloned()
             .collect()
     }
--- a/crates/zakura-network/src/address_book.rs
+++ b/crates/zakura-network/src/address_book.rs
@@ -561,6 +561,17 @@ impl AddressBook {
             // would otherwise dominate the cache, leaving a restarted node with
             // mostly undialable candidates.
             .filter(|addr| !addr.is_inbound())
+            // # Security
+            //
+            // Only cache addresses that have answered one of our own connections.
+            // The cache file stores bare socket addresses, so nothing else in this
+            // entry survives a restart: an address we cache on someone else's word
+            // is re-read as an ordinary initial peer, with no record of where it
+            // came from. Requiring a response of our own means an undialable
+            // address can never enter the cache, however it reached the address
+            // book — gossip, a hand-edited cache file, or an earlier cache written
+            // by a release without this filter.
+            .filter(|addr| addr.has_ever_responded())
             .cloned()
             .collect()
     }
--- a/crates/zakura-network/src/address_book/tests/vectors.rs
+++ b/crates/zakura-network/src/address_book/tests/vectors.rs
@@ -11,7 +11,10 @@ use zakura_chain::{
 };
 
 use crate::{
-    constants::{DEFAULT_MAX_CONNS_PER_IP, MAX_ADDRS_IN_ADDRESS_BOOK, MAX_PEER_MISBEHAVIOR_SCORE},
+    constants::{
+        DEFAULT_MAX_CONNS_PER_IP, MAX_ADDRS_IN_ADDRESS_BOOK, MAX_PEER_MISBEHAVIOR_SCORE,
+        MIN_PEER_RECONNECTION_DELAY,
+    },
     meta_addr::{MetaAddr, MetaAddrChange},
     protocol::external::types::PeerServices,
     AddressBook,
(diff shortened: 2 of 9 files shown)


ebfull:
Heads-up: changelog fragments moved from the root `changelog-unreleased/` directory to `docs/changelog/unreleased/` in #520, and this PR still writes its fragment to the old location (`changelog-unreleased/507.md`). When the PR merges, that recreates the deleted root directory and the fragment gets stranded where release assembly won't pick it up (this already happened once with #513, fixed in #527).

ValarDragon:
AI slop reply, (1) sounds legitimate

  ### 1. The same-IP dial will still be rejected downstream

  The PR claims max_connections_per_ip is only enforced while accepting inbound connections. That is false.

  After an outbound handshake, the shared PeerSet counts all existing connections with that IP and drops the new service when the limit is already reached: set.rs
  (https://github.com/zakura-core/zakura/blob/0d20e427201b0047f9822172a5fe8e585a9a09e9/crates/zakura-network/src/peer_set/set.rs#L784-L790).

  With the default limit of one:

  1. An inbound connection occupies the IP.
  2. This PR allows dialing that IP’s listener.
  3. The outbound handshake succeeds.
  4. PeerSet immediately drops it because the inbound peer already occupies the IP.

  Therefore fix 3 does not produce the additional outbound connection it claims. It merely permits expensive handshakes that the next layer rejects. The new test only checks candidate
  selection, so it misses this integration behavior.

  The PR must either:

  - retain same-IP suppression while the total per-IP limit is one; or
  - deliberately redefine the limit per direction, update PeerSet, assess the Sybil/security tradeoff, and add an end-to-end test.

  ### 2. Port ≥32768 does not prove an address is ephemeral

  The load filter discards every cached address at port 32768 or higher: config.rs
  (https://github.com/zakura-core/zakura/blob/0d20e427201b0047f9822172a5fe8e585a9a09e9/crates/zakura-network/src/config.rs#L637-L641).

  But a service may listen on any valid nonzero port. Zakura’s own validation permits that: priority.rs
  (https://github.com/zakura-core/zakura/blob/0d20e427201b0047f9822172a5fe8e585a9a09e9/crates/zakura-network/src/peer/priority.rs#L95-L112). OS ephemeral ranges are configurable as
  well.

  This creates a direct inconsistency:

  - Zakura successfully connects to a legitimate listener at :40000.
  - The new cacheable() rule writes it to disk.
  - The next restart discards it as “ephemeral.”

  The live fleet sample shows a useful correlation, but it cannot establish a protocol invariant. A safer migration would version the cache format or discard old unversioned caches
  once, then allow the new “successfully contacted outbound” invariant to keep future caches clean.

evan-forbes:
converting to a draft and will likely close as I've lost context here, and we might have already fixed in a different way.

PR #512, opened 2026-08-01
fix(release): catch and prevent unpublishable crates.io publish graphs

ebfull:
## Motivation

`cargo publish --workspace --dry-run` failed on main after the v1.1.0-rc0 preparation, and none of our release checks had caught it: **the v1.1.0-rc0 crate graph cannot be published to crates.io.**

Root cause: under changed-crates-only publishing, a crate already published at its workspace version (`zakura-node-services` 3.0.0, `zakura-script` 3.0.0) is skipped at publish time, so dependents resolve its manifest **from the index** — which still pins `zakura-chain ^3.0.0`. A requirement without a pre-release tag can never select a pre-release, so once the graph moved to `zakura-chain 3.1.0-rc0`, every crate depending on both stopped resolving. Publishing would have shipped unresolvable crates.

Two tooling behaviors conspired to hide this:

- `release.toml`'s `dependent-version = "fix"` rewrote the chain requirement *inside* node-services and script (required just to keep the path-dep workspace building) without bumping them — so their local 3.0.0 manifests silently diverged from the published 3.0.0.
- `check-crate-packaging.sh` packages the whole workspace, so the local (rewritten) manifests shadow the index copies and everything resolves. It structurally cannot see this failure class.

The failure class is wider than the incident, and cargo reports only half of it:

- **Unresolvable** (the incident): the stale index pin is in the same semver-compatible class as the version being published — the dry-run fails loudly.
- **Duplicated** (found in review): the stale pin is a *different* class (an old major). That never fails resolution: cargo selects both majors side by side, the dry-run exits 0, `--no-verify` compiles nothing, and consumers get two copies of the crate with mismatched types. Simulated concretely: `zakura-chain` 3.1.0-rc0 → 4.0.0 with node-services/script left at 3.0.0 passes the dry-run while the `Cargo.lock` cargo writes into `zakura-state-6.0.0-rc0.crate` carries both `zakura-chain 3.0.1` and `4.0.0`.

The general rule the tooling was missing: **a prerelease bump (or a new major) of a crate forces a republish of every published crate that depends on it, directly or transitively.**

## Solution

**Catch** — `scripts/check-crate-publish-graph.sh` (new): computes the real publish set by querying the sparse index for each publishable crate's exact version, warns when a published crate's local requirements have diverged from the index copy, then runs `cargo publish --workspace --dry-run --no-verify --locked` excluding the already-published crates — so resolution happens against the live index, exactly as it will for consumers. A dry-run failure catches the unresolvable shape. For the duplicated shape the check does not model the resolver; it reads the answer cargo already computed: after a passing dry-run, **every workspace crate appearing in any packaged archive's `Cargo.lock` must resolve at its workspace version**. That single assertion catches duplicates, stale selections, and the original prerelease case with no semver reasoning. The dry-run packages into a dedicated target directory so the assertion only ever reads archives it wrote, and it fails closed on a missing archive or lockfile — a future cargo layout change becomes a loud failure, not a silent pass.

Wired in as step 5/6 of `make pre-release`, which runs in `prepare-release-pr.yml` and in `create-release.yml`'s validate job (no compilation: `--no-verify` stays). Documented emergency override for a deliberately GitHub-only release: `ZAKURA_ALLOW_UNPUBLISHABLE_CRATE_GRAPH=1` / the new `allow_unpublishable_crate_graph` workflow input (plumbed through `release-t0.sh --allow-unpublishable-crate-graph`); it covers both failure shapes.

**Prevent** — `scripts/prepare-release.sh` now:

- plans a **cascade** row (patch bump + release suffix) for the transitive dependent closure of every crate taking a prerelease or a new semver-compatible class (judged against the live index; never-published crates are exempt, since no index manifest can pin them). This is plain dependency-graph reachability over workspace normal/build edges — no index requirement parsing, no resolver model; the lockfile gate above verifies the outcome. Cascade targets skip versions already taken on the index (e.g. by a hotfix line), for both the suffixed and future stable form;
- seeds the publish set with crates already at unpublished versions (bumps from an earlier GitHub-only rc), so triggers and cascades see the full set — this self-heals preparations from an already-diverged main (like today's);
- after applying bumps, **fails closed** if `cargo-release` rewrote a manifest outside the bump plan for a published crate (dev-dependency-only rewrites are reported but exempt — dependents never see published dev requirements).

Sparse-index access lives in `scripts/lib/crates-index.sh`. Cascade rows surface in the bump plan, the prep summary JSON, and the generated release PR body with a "do not drop" note. Policy text updated in the release checklist template, the release skill, and the headers of `check-crate-packaging.sh` / `check-crate-version-bumps.sh` (which documents that the existing stable-only `check-release-requirements.sh` debris check is complemented by this prerelease-side gate).

## Testing

- **Catch, unresolvable (incident) case**: `./scripts/check-crate-publish-graph.sh` on current main → correctly computes the 7-crate publish set, warns on both diverged manifests, reproduces the exact resolution failure in ~15s without compiling, exits 1 with remediation guidance. Override env → warning + exit 0.
- **Catch, duplicated case**: in a worktree simulating `zakura-chain` 3.1.0-rc0 → 4.0.0 with node-services/script left at published 3.0.0, the dry-run exits 0 — and the lockfile assertion flags `zakura-chain 3.0.1` next to 4.0.0 in five packaged archives (zakura, consensus, state, rpc, utils; utils via its versioned dev-dependency on node-services), exit 1. `zakura-network`, whose lock is genuinely clean, is not flagged. Override env → warning + exit 0.
- **Catch, positive case**: in a worktree with node-services/script cascade-bumped to 3.0.1-rc0 (the remediation this PR automates), the full 9-crate rc graph dry-run publishes cleanly and every packaged lockfile resolves at workspace versions → exit 0. Also confirms cargo 1.97 writes a `Cargo.lock` into library archives, which the fail-closed guards depend on.
- **Prevent**: driving the cascade stage verbatim against the live index on main's diverged graph plans exactly the incident's two missing rows — `zakura-node-services`/`zakura-script` 3.0.0 → 3.0.1-rcN, reason "depends on zakura-chain, which takes the prerelease 3.1.0-rc0" — and nothing else (leaf and dev-only support crates stay out). The new-major trigger branch verified on the 4.0.0 simulation ("takes the new major 4.0.0", same closure); `semver_class` spot-checked for 1.x / 0.x / 0.0.x / prerelease shapes.
- **Prevent, incident replay** (ran on this PR's initial revision, before the closure rework): on the actual pre-#502 commit (`65a3c213d^`), `prepare-release.sh --dry-run --release-tag v1.0.6-rc0 --base-tag v1.0.5` reproduced #502's bump plan exactly plus the two cascade rows whose absence caused the incident, and the full non-dry-run apply completed end-to-end — cascade bumps applied, "No requirement rewrites outside the bump plan", and the prepared tree passed the graph check against the live index. The reworked planner emits the same two rows for that graph (previous bullet), and the apply path is unchanged: plan rows in, cargo-release bumps out.
- Dev-only rewrite classifier: 6 synthetic diff cases (dev table, normal table, mixed, target dev table, deletion-only, clean) all classify as designed.
- `shellcheck` 0.11.0 clean on `scripts/*.sh` and `scripts/lib/*.sh`; `bash -n` on all touched scripts; YAML parse of both workflows; `make -n pre-release-publish-graph`; `markdownlint` findings on changed Markdown identical to main's; `changelog.py check` passes; `scripts/tests` unittest discovery still green after deleting `test_semver_req.py` along with the `semver-req.sh` library it covered.
- **Post-#511 rebase**: every live-index and static check above re-run on the rebased branch, plus one extra healthy variant — chain at 4.0.0 with the full cascade published (the fixed counterpart of the duplicated case) → exit 0. #511's retarget-unpublished "kept" rows compose with the planner unchanged: they suppress duplicate plan rows while the seeding step still carries their unpublished versions into trigger evaluation.

## Changelog

No fragment: no Rust source or `Cargo.toml` changes (release tooling, workflows, and docs only); `changelog.py check-pr` passes without one.

- Incident: v1.1.0-rc0 prepared in #502/#505; `cargo publish --workspace --dry-run` resolution failure on main.
- Cargo version semantics: requirements without a pre-release tag never match pre-release versions; requirements in different semver-compatible classes resolve as duplicate side-by-side copies, not failures.
- Complements `check-release-requirements.sh` (stable-tag rc-debris gate, added after the v1.0.3 hand-fix).

### Follow-up Work

- Main currently carries the unpublishable v1.1.0-rc0 graph. No action needed for the GitHub-only rc; the next crates-publishing preparation will self-heal it via cascade rows (verified above on main's actual diverged graph). Until then, `create-release.yml` dispatches for GitHub-only tags from this state need the `allow_unpublishable_crate_graph` input.
- The stable fold leaves matching-but-stale `-rc0` requirement strings in published manifests (`dependent-version = "fix"` only rewrites non-matching requirements); the existing `check-release-requirements.sh` already forces normalization at stable tags.


scripts/check-crate-publish-graph.sh
```diff
@@ -0,0 +1,152 @@
...
+if [ "${#publish_set[@]}" -eq 0 ]; then
+  echo "Every publishable crate is already on the index; nothing to publish, graph check passes."
+  exit 0
+fi
+
+echo "Dry-run publishing ${#publish_set[@]} crate(s) against the live index..."
+publish_status=0
+cargo publish --workspace --dry-run --no-verify --locked \
+  "${exclude_args[@]}" || publish_status=$?
+
```
evan-forbes, line 169: claude noticed that there could be an issue here for major upgrades

```
Under a simulated zakura-chain 3.1.0-rc0 → 4.0.0 with node-services/script left at published 3.0.0, the dry-run exits 0, and the Cargo.lock cargo itself wrote into zakura-state-6.0.0-rc0.crate says:

[[package]] name = "zakura-chain"  version = "3.0.1"   source = "registry+…"
[[package]] name = "zakura-chain"  version = "4.0.0"   source = "registry+…"

Two copies of zakura-chain in a graph that "passed". 3.0.1 is dragged in by the skipped index copy of zakura-node-services, whose service trait aliases are built out of chain types — so any consumer touching both gets type mismatches. --no-verify never compiles, so nothing sees it.
```

more details:

```
The oracle is "did resolution succeed", but that isn't the invariant the PR is actually defending. The invariant is "the published graph contains exactly the crate versions this workspace intends." 
Those two coincide for the prerelease case only by accident: a prerelease and a stable of the same major are semver-compatible, so cargo must unify them, so it fails. Across a major boundary cargo isn't required to unify — it just installs both, silently.
```

suggested solution
```
Don't add a second oracle; read the answer cargo already computed. Every packaged archive the dry-run leaves in target/package/tmp-crate/ carries the Cargo.lock cargo resolved for it against the live index plus the overlay of crates being published. Assert on it:

▎ every workspace-internal package appearing in any packaged lock must be at that crate's workspace version

That single rule catches duplicates, stale selections, and the original prerelease case, with no semver reasoning at all. Roughly, after the existing dry-run:

declare -A workspace_version=()
while IFS=$'\t' read -r crate version; do
  workspace_version["$crate"]="$version"
done <<<"$publishable"

archives=(target/package/tmp-crate/*.crate)
if [ "${#archives[@]}" -ne "${#publish_set[@]}" ]; then
  echo "ERROR: expected ${#publish_set[@]} packaged archives, found ${#archives[@]}." >&2
  exit 1                       # fail closed if cargo's layout ever moves
fi

stale=0
for archive in "${archives[@]}"; do
  member="$(basename "$archive" .crate)"
  while IFS=$'\t' read -r crate version; do
    want="${workspace_version[$crate]:-}"
    [ -n "$want" ] && [ "$version" != "$want" ] || continue
    printf 'ERROR: %s resolves %s %s; this workspace publishes %s\n' \
      "$member" "$crate" "$version" "$want" >&2
    stale=1
  done < <(tar -xOzf "$archive" "${member}/Cargo.lock" | awk '
    /^\[\[package\]\]/ { n = "" }
    /^name = /    { gsub(/"/, ""); n = $3 }
    /^version = / { gsub(/"/, ""); if (n != "") print n "\t" $3 }')
done
[ "$stale" = 0 ] || exit 1

Why this is the bulletproof version rather than the alternatives:

- It is cargo's own resolution, not a model of it. No semver-req.sh, no index requirement parsing, no assumptions about which requirement forms exist.
- It costs nothing — the archives are already produced; --no-verify stays, runtime stays ~30s.
- Dropping --no-verify instead is both slower and unsound: it only catches a duplicate crate if the duplicated types happen to cross the API surface being compiled. Absence of a compile error is not evidence of a sound graph.
- The archive-count guard means a future cargo layout change turns into a loud failure, not a silent pass.

Two things go with it: the --exclude set is the thing that makes any of this possible, so the check must stay excluded-set-driven (it already is); and the "caret requirements never cross a major" line should come out of the error message, the header, and SKILL.md — it should say that a skipped crate pinning a different major yields a duplicate, not a resolution failure.

If you want the failure class gone rather than detected

Make the publish set total: on any release that publishes crates, republish every publishable crate. No crate is skipped, so no index manifest ever participates, and both failure modes cease to exist — check-crate-publish-graph.sh collapses to a dry-run with no excludes, and the whole cascade planner plus semver-req.sh can go. The cost is a patch version number on unchanged crates, and it contradicts the standing policy in release.toml ("a crate is bumped and published only when it has changes to publish"). The middle ground preserving that policy: apply totality only when some crate takes a prerelease or a new major — i.e. bump the transitive dependent closure — which is exactly when the class exists and is pure cargo metadata graph reachability.

I'd ship both: the lock assertion is the gate you want regardless, and it's ~20 lines that subsume the part of the PR that's currently 200.
```

evan-forbes (approved):
likely worth the ai's taking a look at the comment claude found

but could also be a follow up so approving

evan-forbes (approved):

ebfull:
Rebased over #511. What it took, at a high level:

- Conflicts were confined to `prepare-release.sh` and `release-t0.sh`, and only where both PRs add flags, header steps, or dispatch inputs in the same spots — resolved as straight unions, no behavior change to either side.
- #511's retarget-unpublished and readiness machinery needed no planner changes: retarget "kept" rows already suppress duplicate plan rows, and the publish-set seeding still carries those unpublished versions into cascade trigger evaluation.
- One drive-by while merging: the `--help` line ranges in both scripts had drifted from their actual header lengths (#511's `prepare-release.sh` range was cutting off the option list); recounted to the real extents.
- Full re-verification on the rebased branch: the resolution-failure, duplicated-major, and healthy-graph simulations against the live index, the cascade planner on both trigger shapes, and the static checks (shellcheck, unittest discovery, YAML, markdownlint vs main).

evan-forbes (approved):

PR #519, opened 2026-08-02
fix(ci): install nightly for release API diffs

p0mvn:
## Motivation

The release preparation workflow added `cargo public-api diff latest` in #511, but only provisioned the stable Rust toolchain. `cargo-public-api` requires nightly to build rustdoc JSON, so preparation failed after all other verification passed in [run 30734596608](https://github.com/zakura-core/zakura/actions/runs/30734596608).

## Solution

Install the minimal nightly toolchain immediately before public API verification. Stable remains the workspace's active toolchain for preparation, packaging, tests, and formatting; nightly is available only when `cargo-public-api` requests it.

## Testing

- `actionlint .github/workflows/prepare-release-pr.yml`
- `git diff --check`
- IDE diagnostics: no errors

## Changelog

No fragment: this changes CI release tooling only and does not modify Rust source or `Cargo.toml`.

### Specifications & References

- #511
- [Failed release preparation run](https://github.com/zakura-core/zakura/actions/runs/30734596608)

PR #521, opened 2026-08-02
fix(ci): make release API diff verification reliable

p0mvn:
## Motivation

Follow up #519 and the subsequent [release preparation failure](https://github.com/zakura-core/zakura/actions/runs/30736190306). Public API verification had two reliability problems:

- a nightly installation outage would abort before creating the diagnostic draft release PR;
- the workflow-wide `RUSTFLAGS=-D warnings` turned nightly-only deprecations into build errors while generating rustdoc JSON.

## Solution

- Move nightly provisioning into the existing fallible `Record public API diffs` step, preserving draft creation when setup fails and skipping installation when no API diffs are needed.
- Clear `RUSTFLAGS` only for `cargo public-api diff`. API comparison must build both current and previously published crates and should not double as a nightly lint gate.

Stable remains the active toolchain and warning policy for the workflow's actual release verification.

## Testing

- `actionlint .github/workflows/prepare-release-pr.yml`
- `git diff --check`
- IDE diagnostics: no errors

## Changelog

No fragment: this changes CI release tooling only and does not modify Rust source or `Cargo.toml`.

- #519
- [Missing-nightly failure](https://github.com/zakura-core/zakura/actions/runs/30734596608)
- [Nightly-deprecation failure](https://github.com/zakura-core/zakura/actions/runs/30736190306)

### DIFF
--- a/.github/workflows/prepare-release-pr.yml
+++ b/.github/workflows/prepare-release-pr.yml
@@ -241,11 +241,6 @@ jobs:
         continue-on-error: true
         run: cargo fmt --all -- --check
 
-      # cargo-public-api builds rustdoc JSON with nightly even when the
-      # workspace itself is built and verified with stable.
-      - name: Install nightly for public API diffs
-        run: rustup toolchain install nightly --profile minimal
-
       - name: Record public API diffs
         id: verify-public-api
         continue-on-error: true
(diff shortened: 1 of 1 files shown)


p0mvn:
Continuing to fix the release prep CI per discussion on Slack: https://github.com/zakura-core/zakura/actions/runs/30736190306

PR #520, opened 2026-08-02
chore(repo): clean up the repository root

ebfull:
## Motivation

The repository root had accumulated developer-focused configs and process artifacts alongside the human-surfaced deliverables. The root should carry what users and contributors reach for first (`README.md`, `CHANGELOG.md`, `CONTRIBUTING.md`, `SECURITY.md`, licenses, `AGENTS.md`); tool configuration and interstitial development artifacts belong in `.github/` and `docs/`.

## Solution

One focused commit per cleanup, in branch order:

- **`chore(ci): move codespell config out of repo root into .github/`** — codespell is CI tooling; `.codespellrc` → `.github/codespellrc`, with the `docs-check.yml` trigger paths and run step updated.
- **`chore(ci): move lychee config out of repo root into .github/`** — same treatment for the link checker: `.lychee.toml` → `.github/lychee.toml`.
- **`chore(ci): ignore exact fragment renames in changelog check-pr`** — groundwork for the next commit. `changelog.py check-pr` treated any fragment path in a PR diff as a fragment change owned by that PR, so a PR relocating the fragment directory would fail its own gate. Exact (100%-similarity) renames are now recognized as repository moves. Rename destinations still count toward the "Rust change requires a fragment" rule, and a renamed-and-edited fragment surfaces as add+delete and is still rejected, so the gate loses no strictness over content.
- **`chore(repo): move changelog dev docs into docs/changelog/`** — `CHANGELOG.md` stays in root as the operator-facing deliverable; the developer artifacts get one grouped home: `CHANGELOG_GUIDELINES.md` → `docs/changelog/guidelines.md`, `CHANGELOG_PARAMS.md` → `docs/changelog/params.md`, `changelog-unreleased/` → `docs/changelog/unreleased/`. A new `docs/changelog/README.md` explains the directory records and tracks changes between and across releases for development purposes and points users at the root changelog. All tooling paths (`scripts/changelog.py`, `prepare-release.sh`, `release-readiness.py`, the script tests) and every prose/link reference (AGENTS.md, CONTRIBUTING.md, PR templates, copilot instructions, semver-checks help text, release skill) updated.
- **`chore(repo): remove .gitattributes so Cargo.lock diffs collapse again`** — its single line existed only to un-hide `Cargo.lock` in GitHub PR diffs. We prefer the default collapsed rendering; lockfile changes are audited at release time and by automation that flags suspicious changes.
[...]

PR #525, opened 2026-08-02
chore(repo): move workspace crates into crates/

ebfull:
## Motivation

The workspace's 14 member crates sat at the repository root, crowding it and mixing
crate directories with docs, scripts, and CI configuration. Following the repository
root cleanup (#520), this PR moves them all into the conventional `crates/` directory.
`deploy/zakura-watchdog` stays where it is (it has no path dependencies — everything
is `workspace = true`).

## Solution

Three commits, reviewable independently:

1. **Pure `git mv`** of the 14 crates into `crates/` — 100% renames, no content changes.
2. **Functional path fixes**: workspace `members`, the nine `readme = "../README.md"`
   fields plus the cargo-release replacement target, `xtask::repo_root()` (now two
   levels up), the two tests that escape their crate (`check_no_git_dependencies`,
   the regtest e2e runner path), release/compat scripts and `make/zcashd-compat.mk`,
   CI path filters and change-detection greps (`tests-unit`, `test-docker`,
   `zakura-e2e`, `create-release`, `update-release-state`, pr-node workflows and
   helper scripts), codespell/lychee excludes, `.dockerignore` re-includes
   (`!crates` + `!crates/**`), all 14 Dockerfile bind mounts (`source=` and
   `target=`), and the three anchored `.gitignore` patterns.
3. **Docs and templates**: AGENTS.md, the release skill, design/spec docs, PR/issue
   templates, commented-out CODEOWNERS entries, in-repo GitHub URLs, relative links,
   and comment-only source pointers.

Inter-crate `path = "../<crate>"` dependencies are sibling-relative and unchanged;
`Cargo.lock` is byte-identical.

**Release-tooling transition fallback**: `prepare-release.sh` and
`create-release.yml` read `zakurad/Cargo.toml` at the release base tag via
`git show`. Pre-move tags keep that file at the old path, so both now try
`crates/zakurad/Cargo.toml` first and fall back to the old path. The fallback is
removable once a post-move tag exists.

**Upstream backport automation**: the subsystem is layout-agnostic (commit-graph
discovery, checkout-driven adaptation), so only two touches were needed — the Codex
prompt now states the `zebra-*` → `crates/zakura-*` path mapping explicitly, and the
offline test fixture uses a representative post-move path.

Notes for maintainers:

- The first post-move release will see every crate as "changed"
  (`git diff <pre-move-tag> -- <crate_dir>`), so expect a one-time version
  bump/republish of all publishable crates.
- External deep links into `tree/main/<crate>/…` will 404 (GitHub does not redirect
  moved directories); docs.rs and rustdoc URLs are package-name based and survive.
- Manual backports now map upstream `zebra-X/…` to `crates/zakura-X/…` (e.g.
  `git apply --directory=crates` after the rename rewrite).

## Testing

- `cargo build --workspace --locked` with `Cargo.lock` verified byte-identical to
  the pre-move baseline; `cargo metadata` confirms all 15 members resolve (14 under
  `crates/`, watchdog under `deploy/`) and every manifest `readme` path resolves.
  and `cargo fmt --all -- --check` are clean.
- `cargo test -p zakura --test acceptance -- check_no_git_dependencies --exact --ignored`
  passes, exercising the corrected `../../Cargo.lock` path. Workspace `--lib` tests
  pass except four failures verified byte-identical on `main` (machine-local:
  one consensus concurrency flake, three network listener tests) — unrelated to
  this change.
- Release tooling executed against the new layout: `scripts/check-release-state.sh`
  (reads the moved checkpoint/frontier/provenance files),
  `scripts/resolve-zcashd-compat-manifest.sh` (resolves the moved manifest),
  `scripts/changelog.py check`, and the base-tag fallback tested against the
  pre-move tag `v1.1.0-rc1` (old path) and post-move `HEAD` (new path).
- `cargo run -p xtask -- package ubuntu` prints the correct repo root and
  Dockerfile path (docker unavailable locally, so it stops at spawn as expected).
- `actionlint` and `shellcheck` report no new findings vs `main`; the offline
  upstream-sync fixture test passes.
- Repo-wide sweeps for stale crate paths return empty (the only remaining old-path
  references are the intentional transition fallback lines).
- Docker builds are exercised by this PR's own CI: the renamed files match the
  updated `paths:` filters, so `tests-unit`, `test-docker`, and `zakura-e2e` all
  trigger and live-test the new filters, `.dockerignore`, and bind mounts.

## Changelog

`docs/changelog/unreleased/525.md` with the explicit no-changelog marker
(internal repository-layout change; no operator- or crate-consumer-visible effect) —
added after opening this draft PR.

### Follow-up Work

- Remove the base-tag `git show` transition fallback and the temporary lychee
  exclusion for `/main/crates/` URLs after the first post-move tag
  exists.
- A tiny follow-up PR touching only
  `crates/zakura-chain/src/parameters/checkpoint/main-checkpoints.txt` to confirm
  the updated data-file CI filters still trigger the unit-test job (nothing else
  catches that silently).


ValarDragon:
Need to just ask AI that the build exclusions for testing in CI still apply (we remove xtask and tower from every PR build iirc)

ebfull:
> Need to just ask AI that the build exclusions for testing in CI still apply (we remove xtask and tower from every PR build iirc)

Checked — summary: **every mechanism that scopes CI builds is package-name or cargo-metadata based, never path-based, so the `crates/` move cannot detach any of it.** Details:

- **The only explicit workspace exclusion in CI** is `lint.yml:119` — `cargo check --workspace --bins --exclude zakura-watchdog`. That's a package *name*, unaffected by the move (job is green on this PR).
- **xtask** is kept out of test runs by its own manifest, not by CI config: its `[[bin]]` sets `test = false` (`crates/xtask/Cargo.toml:15`), so it contributes no test targets to any `--workspace` run. `semver-checks.yml` also skips it automatically as a `publish = false` member via `cargo metadata` (see its header comment) — note this PR's checks have no `semver (xtask)` job, as expected.
- **tower-batch-control / tower-fallback are not excluded from PR builds today**: `test-crates.yml` runs `cargo hack check --workspace` (lines 95–96) over every member, the unit-test shards run from a `cargo nextest archive` of the whole workspace (`tests-unit.yml:182-186`), and the CI profiles in `.config/nextest.toml` filter by *test name* only — there is no `package(...)` exclusion for them. They are also semver-checked as `zakura-tower-*` (both green here). The remembered exclusion likely belongs to the previous per-crate build matrix that the single cargo-hack pass replaced (see the comment at `test-crates.yml:53-60`) — I found no xtask/tower exclusion anywhere in this repo's workflow history (`git log -S`).
- Direct confirmation: this PR's own run on the moved layout is fully green, including "Build crates (default and no-default features)", all three unit-test shards, and all 13 `semver (…)` jobs.

*Verification run by Claude Code at Sean's request.*

ValarDragon (approved):

PR #531, opened 2026-08-03
fix(deploy): require real height progress to clear a stalled-node alert

evan-forbes:
## Motivation

On 2026-08-03 `zakura-testnet-3` sat pinned at height **4,129,396** for 43 minutes
(19:39:40 → 20:22:58, confirmed by its own log: `chain updates have stalled ...
is_syncer_stopped=true`). The watchdog posted:

```
20:03  :white_check_mark: zakura-testnet-3 recovered from stalled
       health: healthy - height: 4129396      <-- the height it was stuck at
20:13  :rotating_light: zakura-testnet-3 stalled for 10m 23s
       health: stale - height: 4129396        <-- still not moved
```

The recovery was false. It reconciles to the second with a dashboard restart:
`zakura-testnet-dashboard.service` restarted at **20:02:37**, and the re-alert at
20:13 read "stalled for 10m 23s" — 20:13:00 − 623s = 20:02:37. The dashboard
restarted five times that day, producing one spurious recovery each.

Two defects combine:

- **`zakura-cluster-status.py`** keeps `last_advanced_at` in memory only. `--state-file`
  persists reorg pairs and nothing else, and the first poll after a start seeds the
  timer to `now` for every node regardless of height. A restart reports the whole
  fleet as freshly advanced and silently clears in-flight stall alerts — the blast
  radius is every node, not just the stalled one.
- **`zakura-cluster-watchdog.py`** cleared an alert on `condition == "ok"` alone.
  There is no height comparison anywhere on the recovery path, which is why the
  recovery message printed the stuck height.

Since nothing compares heights across a `stalled → ok` transition, a node pinned at
one height oscillates `stalled → recovered → stalled` indefinitely.

There is a second, independent height-blind path: `node_condition` returns `ok` for
`starting_grace` seconds after any dashboard bounce, and the dashboard's initial rows
are literally `health: "starting"`.

## Solution

**Watchdog (primary).** Record the height a stall alert fires at (`alert_height`) and
require the node to be strictly higher before retiring it. A reset timer, a missing
height (the `starting` window) or a height going backwards all keep the alert latched.
Only stalls are height-gated — a restarted node still recovers at whatever height it
comes back on. This closes both paths and holds even if the timer resets again, so it
does not depend on the dashboard change.

**Dashboard (complementary).** Persist per-node height and last-advanced timestamps
into the existing state file and restore them on start, so the stall clock survives a
restart instead of zeroing. Load is tolerant of a missing, corrupt, or pre-existing
file written before `progress` existed.

## Testing

- New `deploy/runner/test_zakura_cluster_watchdog.py` (8 tests): stall records its
  firing height; unchanged height does not clear it; five consecutive timer resets
  never clear it; forward progress does; missing height and backwards height stay
  latched; `down` alerts still recover without height progress; the anchor survives
  intermediate stalled cycles.
- `deploy/runner/test_zakura_cluster_status.py` 18 → 20: the stall timer survives a
  collector restart and still reports `health: stale` with the correct age, plus
  missing/corrupt/legacy state-file handling.
- Replaying the actual incident (five restarts, node never moving):

  ```
  origin/main : 6 stall alerts, 5 false recoveries
  with fix    : 1 stall alert,  0 false recoveries
  both        : genuine recovery still posted once height advances
  ```

**These tests were not running in CI at all** — `lint.yml` referenced neither runner
suite. It now runs both.

## Changelog

Deploy/monitoring tooling only; no Rust or `Cargo.toml` change, so no fragment.

### Follow-up Work

The underlying node stall was a header-sync livelock on the unmerged
`evan/header-chain-squashed` branch (staged headers hit `MAX_STAGED_HEADERS_V1` and the
target was discarded before ever committing), already fixed there by `7c283bb82`.
Neither the defect nor its fix is on `main`. This PR is only about the alerting being
truthful about it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -197,6 +197,12 @@ jobs:
         if: ${{ !cancelled() }}
         run: python3 deploy/continuous-sync/tests/test_continuous_sync.py
 
+      - name: Cluster dashboard and watchdog tests
+        if: ${{ !cancelled() }}
+        run: |
+          python3 deploy/runner/test_zakura_cluster_status.py
+          python3 deploy/runner/test_zakura_cluster_watchdog.py
+
       - name: Ensure no arbitrary or proptest dependency on default build
         if: ${{ !cancelled() }}
         run: cargo +stable tree --package zakura -e=features,no-dev | grep -Eq "arbitrary|proptest" && exit 1 || exit 0
--- a/deploy/runner/test_zakura_cluster_status.py
+++ b/deploy/runner/test_zakura_cluster_status.py
@@ -361,6 +361,65 @@ class TipAgreementTests(unittest.TestCase):
             self.assertEqual(event["discarded_hash"], "a" * 64)
             self.assertEqual(event["canonical_hash"], "b" * 64)
 
+    def test_stall_timer_survives_a_collector_restart(self):
+        # A restart used to reseed last_advanced_at to "now", reporting every
+        # node as freshly advanced and clearing in-flight stall alerts.
+        with tempfile.TemporaryDirectory() as tmp:
+            state_file = Path(tmp) / "state.json"
+            first = status.ClusterCollector(
+                [node()],
+                interval=10,
+                stale_after=300,
+                upgrade_height=0,
+                target_spacing=7.5,
+                network="testnet",
+                state_file=state_file,
+            )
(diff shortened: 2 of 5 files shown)


ValarDragon (approved):
Good catch - didn't realize we restart the dashboard this much

PR #533, opened 2026-08-04
refactor(state)!: remove the VCT Sprout-history repair artifact

p0mvn:
## Motivation

The VCT Sprout-history repair artifact was always documented as temporary. It fixed Mainnet databases damaged by the original verified-commitment-trees fast path, which advanced Sprout locally but never persisted the historical `Sprout root -> frontier` anchors.

The forward fix shipped long ago: the fast commit path now updates the Sprout tree for every block, so no database built by a current release needs the repair. What remained was pure carrying cost:

- a hard-pinned 71.7 MB dependency, `zakura-vct-sprout-history = "=1.0.0"`, split across 10 crates.io part packages and embedded unconditionally into every `zakura-state` build
- ~2,700 lines of format, replay, migration, and test code
- an offline generator binary, a `zakurad` subcommand, and a release-state CI gate

## Solution

Remove all of it. Format version 28.0.1 becomes a `NoMigration` no-op, so the disk format does not move (current is 28.0.3).

The one safety property worth keeping is preserved. A database still missing those anchors silently rejects JoinSplits that spend a historical Sprout root — a consensus divergence — so it must never be used. With no repair path left, such databases now **fail closed** rather than being repaired in place: a Mainnet database that is VCT-synced and below format version 28.0.1 is rejected at startup with the new `StateInitError::VctSproutHistoryUnrepairable`, in **every** open mode including writable. Operators discard and resync, or restore a current snapshot.

Incidental cleanups that fall out of the removal:

- `DbOpenMode` collapses back to a plain `read_only` bool now that the Sprout-validation open mode is gone
- `DiskDb::write_sync` and `ZakuraDb::write_batch_sync` are removed with their only caller
- fixes a pre-existing doctest failure: the C++ snippet in `prunes_historical_data` was an indented block, so rustdoc tried to compile it as Rust

### Versioning

`ZakuradCmd` is the clap subcommand enum for the binary, not a downstream consumer contract, so removing a subcommand variant should not force a major version. This PR downgrades `enum_variant_missing` to a warning for the `zakura` package, following the precedent already in the same manifest for `feature_missing`. The crate stays at **1.1.0-rc1**. The lint is `warn`, not `allow`, so the break stays visible in the job output.

`zakura-state`'s public API removals need no action: its baseline is 5.0.1 and it is already at 6.0.0-rc1.

### Breaking changes

- The `zakurad validate-vct-sprout-history` subcommand is removed.
- Public `zakura-state` API removed: `generate_mainnet_from_archive`, `GeneratorError`, `validate_vct_sprout_history`, `VctSproutHistoryValidationSummary`, `VctSproutHistoryValidationError`.
- `StateInitError::VctSproutHistoryRepairRequired` and `::VctSproutHistoryRepairInvalid` are replaced by `::VctSproutHistoryUnrepairable`.

## Testing

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo test -p zakura-state` — 443 pass (441 lib + 2 integration), doctests pass
- `cargo test -p zakura --lib` — 348 pass
- `./scripts/changelog.py check` validates
- `cargo tree -p zakura-state` shows no VCT dependency; `Cargo.lock` drops all 11 packages
- verified `zakurad validate-vct-sprout-history` now reports `unrecognized subcommand`

Two new tests in `zakura_db.rs` replace the deleted repair suite and cover the guard in both directions:

- `unrepairable_vct_database_is_rejected_in_every_open_mode` — a seeded Mainnet VCT database below 28.0.1 is refused on both writable and read-only opens
- `databases_without_missing_sprout_history_open_normally` — the guard does not over-fire on a Mainnet database that was never VCT-synced, one already at 28.0.1, or a non-Mainnet network

## Changelog

`docs/changelog/unreleased/533.md` — one `Removed` entry for the subcommand and artifact, one `Changed` entry for the new fail-closed startup behavior.

### Follow-up Work

Dropping the `embedded_mainnet_artifact_matches_frozen_identity_and_current_sprout_root` step from `update-release-state.yml` also drops the gate that blocked a checkpoint bump when a new Sprout-changing JoinSplit appeared above the artifact handoff height. That gate only ever protected artifact staleness, so it is meaningless with no artifact — but it is a deliberate reduction in release-state checking and deserves an explicit sign-off rather than passing unnoticed.


p0mvn:
@cursor review

czarcas7ic:
The normal startup path still calls `expect`, which formats this error using `Debug`:

https://github.com/zakura-core/zakura/blob/86287cb86c99a5f59543c05a6cf05e0798a0ca4c/crates/zakura-state/src/service.rs#L348-L353

An affected node will therefore show `VctSproutHistoryUnrepairable` with the unrelated writable/lock/disk-space hint, instead of the actionable discard/resync message:

https://github.com/zakura-core/zakura/blob/86287cb86c99a5f59543c05a6cf05e0798a0ca4c/crates/zakura-state/src/error.rs#L109-L117

Could we surface the latter during startup?

czarcas7ic (approved):
Minor comment above, also I think this will require a major release bump when it comes time to tag?

p0mvn:
> Minor comment above, also I think this will require a major release bump when it comes time to tag?

Per Slack msg, I do not think removing a debug command justifies a major release bump, so I intentionally disabled a lint to enable minor bump instead

p0mvn:
Good catch, fixed in `4473cc97e`.

You are right on both counts: `expect` formats with `Debug`, so a unit variant printed as bare `VctSproutHistoryUnrepairable`, and the generic hint pointed at permissions, locks, and disk space — none of which apply, all of which send the operator looking in the wrong place.

Switched to `unwrap_or_else` so the panic carries the error's `Display` text. The unrepairable case reports its guidance on its own, without the misleading hint; every other `StateInitError` keeps that hint, now appended to its own `Display` message rather than replacing it.

Before:
```
VctSproutHistoryUnrepairable
```

After:
```
cannot open state: this database was written by an early verified-commitment-trees fast sync
and is missing historical Sprout anchors, so it cannot verify JoinSplits that spend a
historical Sprout anchor. Hint: discard the database and resync, or restore a snapshot taken
with a current release
```

I checked the other state-init paths while I was in there: `init_read_only` propagates with `?`, so its callers already format with `Display`. This `expect` was the only place losing the message.

PR #532, opened 2026-08-04
chore(headersync): add release and rollout support

evan-forbes:
> Stack layer **8/8**. Parent: #572. Review bottom-to-top beginning with #586.

## Motivation

The feature needs CI enforcement, packaging/release awareness, rollout support, and one user-visible changelog entry without mixing repository automation into production review.

## Stack map

1. #586 — `2be164ceb97ae51b2de91cb5ac02ab1315f50d5c` — engine implementation, deterministic tests, replay adapter, and bottom-boundary Docker mount
2. #587 — `91379561b76eff5a8967225bb5dd5ed8e6b35b4e` — typed service and authority contracts
3. #563 — `768941d527ce9fe1fe92a6d093e4e6062b613f0d` — durable state production
4. #588 — `53aa22af61e19f037fd86f4f5bd7b8e102107b75` — state durability and recovery tests
5. #569 — `f9fbde47d2c2a6ee125020a39d203ab91ab08866` — network, consensus/RPC, and node runtime integration
6. #571 — `5221e94155ab43f0113461d897670206d26383e9` — fuzzing
7. #572 — `6a4f48d7fd61dfda6eda61114b061f37830cc13d` — conformance and migration documentation
8. #532 — `e59614be62320b871befe8788e876b4f7577338c` — repository/release support

The stack contains exactly 16 focused commits over frozen `main` `87479333cd670791eca801b8083fbc1b19a835e2`. The final rebase onto that main was conflict-free, and `git range-diff` marks every commit patch-identical to the fully validated corrected series. Subjects, production ownership, and all eight commit boundaries are unchanged. The cumulative published tree is `746e87878caf3379f2c983b4bfbfeb46789382c9`. The previously published top remains archived at `evan/header-chain-v12-publish-pre-remaining-20260807`.

## V12 audit corrections

All 28 V12 findings outside #586 have final dispositions: the 16 state findings from run 6095 were already corrected, and this publication adds the accepted fixes and regressions for the remaining 12 findings across #587, #563/#588, #569, and #572. Run 6102 on #571 produced no findings; #532 had no V12 audit comment.

## Testing

Passed at the cumulative head:

- formatting, locked all-feature/all-target workspace check, and strict workspace clippy
- header-chain tests — 126 passed, 1 ignored
- state tests — 417 passed, 3 ignored
- network library tests — 1,048 passed, 3 ignored
- focused `zakurad` sync tests — 80 passed
- consensus routing tests — 6 passed; RPC tests — 101 passed, 1 ignored
[...]

PR #534, opened 2026-08-04
fix(state): report a typed error for absent historical treestates

p0mvn:
## Motivation

A verified-commitment-trees fast-synced node has no per-height note commitment trees below its checkpoint handoff. `z_gettreestate` maps that to a JSON `null` and `z_getsubtreesbyindex` to an empty list — both of which a lightwalletd-contract client reads as real data. An absent treestate becomes the *empty* tree, so a wallet derives a birthday anchor asserting an empty commitment tree deep in the chain, with no error at the point of corruption.

Implements §4.5 of `docs/design/historical-treestate-serving.md`, which says to wire this up regardless of whether the rest of the design proceeds.

## Solution

Two typed errors, returned by the tree and subtree read APIs instead of an absent tree or an empty list. Because availability checks are part of those APIs' `Result` values, callers cannot accidentally omit them. The checks run only after a read comes back empty, so nothing present is ever rejected.

The subtree gate needs a bound or it fires on ordinary queries: it compares the requested index against the pool's leaf count at the handoff, so only subtrees the fast path actually skipped are reported, and an index past the tip still returns today's empty list. The served run is checked to its end rather than only its start, because `z_getsubtreesbyindex` returns one contiguous run and a gap anywhere truncates the response.

Verbose `getblockheader` previously replaced the state error with a bare "missing Sapling tree" error, while verbose `getblock` could replace it with "missing Orchard tree". Both now preserve the typed state error and explain why the historical tree is unavailable.

## Testing

Unit tests cover the completion bound and gap finder, including bounded requests, gaps, and the `u16::MAX` edge. State tests cover ordinary mid-sync unavailability and the fail-closed missing-handoff-tree branch. RPC tests verify that `z_gettreestate`, `z_getsubtreesbyindex`, verbose `getblock`, and verbose `getblockheader` surface the typed errors rather than null, empty, or generic missing-tree responses. Property-test assertions verify that the gate matches the read guard.

## Note for reviewers

This will look like a regression to anyone whose client "works" today by treating `null` as the empty tree. That client was silently corrupting; it now fails loudly. **This PR does not unblock a wallet** — it makes the existing breakage diagnosable. Serving is restored in the follow-ups.

### Follow-up Work

Stacked: replay engine (#531), subtree artifact (#532), then treestate serving and the frontier grid.

### DIFF
--- a/crates/zakura-state/src/service/read/tree.rs
+++ b/crates/zakura-state/src/service/read/tree.rs
@@ -14,11 +14,12 @@
 use std::{collections::BTreeMap, sync::Arc};
 
 use zakura_chain::{
-    ironwood, orchard, sapling,
-    subtree::{NoteCommitmentSubtreeData, NoteCommitmentSubtreeIndex},
+    block, ironwood, orchard, sapling,
+    subtree::{NoteCommitmentSubtreeData, NoteCommitmentSubtreeIndex, TRACKED_SUBTREE_HEIGHT},
 };
 
 use crate::{
+    error::{HistoricalSubtreeUnavailable, HistoricalTreeUnavailable},
     service::{finalized_state::ZakuraDb, non_finalized_state::Chain},
     HashOrHeight,
 };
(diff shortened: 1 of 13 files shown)


p0mvn:
@cursor review

crates/zakura-state/src/service/read/tree.rs
```diff
@@ -27,6 +28,190 @@ use crate::{
...
+/// `None` when the run covers the whole request.
+///
+/// `subtrees` is contiguous from `start_index` by the time this runs, so the first index it fails
+/// to cover is one past its last key.
+pub(crate) fn first_missing_subtree_index<Node>(
+    subtrees: &BTreeMap<NoteCommitmentSubtreeIndex, NoteCommitmentSubtreeData<Node>>,
+    start_index: NoteCommitmentSubtreeIndex,
+    end_index: Option<NoteCommitmentSubtreeIndex>,
+) -> Option<NoteCommitmentSubtreeIndex> {
+    let first_missing = match subtrees.keys().next_back() {
```
czarcas7ic: This assumes `subtrees` is contiguous, but `subtrees()` only verifies that the requested start key exists. On an upgraded database, index 0 can predate U, index 1 can be skipped during VCT fast sync, and index 2 can be written after semantic sync resumes at H. For keys `{0, 2}`, this reports 3 instead of 1. The handoff check can then accept the response, while `z_getsubtreesbyindex` drops the keys with `.values()` and presents the index-2 root as index 1. That silently misindexes wallet subtree data. Please scan or trim to the contiguous prefix and add an internal-gap regression covering pre-U and post-H rows.

p0mvn: Addressed [39497a3](https://github.com/zakura-core/zakura/pull/534/commits/39497a3c83460000c50c244fb41513491f440f6a)

p0mvn:
@v12this review

p0mvn:
@cursor review

p0mvn:
@cursor review

crates/zakura-state/src/service/read/tree.rs
```diff
@@ -27,14 +32,262 @@ use crate::{
...
+
+    let Some(handoff) = db.vct_synced_below() else {
+        return Ok(());
+    };
+
+    // The durable handoff marker is written by the first fast-path commit, so it exists throughout
+    // an ordinary fast sync even though the handoff height is still above the finalized tip. Keep
+    // this guard before `handoff_leaves`: its backward search has no tip guard and can return a row
+    // below the unreached handoff.
+    if is_syncing_below_last_checkpoint(db.finalized_tip_height(), handoff) {
```
czarcas7ic: AI comment

---

This returns `Indeterminate` for every missing subtree while the finalized tip is below H, before consulting any handoff bound.

But fast mode has already parsed and retained the authenticated final frontier before synchronization begins.

https://github.com/zakura-core/zakura/blob/857007674aeff2c0aca3d9143c08723d303c8f62/crates/zakura-state/src/service/finalized_state/vct.rs#L139-L177

This produces a concrete false error on the current Mainnet base. Its H=3,434,971 artifact decodes to 20,554 Ironwood leaves, so with 2^16 leaves per tracked subtree, subtree 0 cannot exist anywhere through H.

https://github.com/zakura-core/zakura/blob/7496c686f41279f3b1b9d81ba2019d7c8ea1f105/crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.json#L1-L10
https://github.com/zakura-core/zakura/blob/7496c686f41279f3b1b9d81ba2019d7c8ea1f105/crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.bin
https://github.com/zakura-core/zakura/blob/857007674aeff2c0aca3d9143c08723d303c8f62/crates/zakura-chain/src/subtree.rs#L12-L13

`z_getsubtreesbyindex("ironwood", 0, ...)` nevertheless reaches this branch and errors for the entire pre-H sync window, although the correct response is known to be an empty list.

https://github.com/zakura-core/zakura/blob/857007674aeff2c0aca3d9143c08723d303c8f62/crates/zakura-rpc/src/methods.rs#L2202-L2208

Please make the authenticated handoff leaf counts available to this guard before the tip check, and add a current-Mainnet or equivalent regression that expects empty Ironwood index 0 while tip < H.

p0mvn: [39c7919](https://github.com/zakura-core/zakura/pull/534/commits/39c79193174641f88e7d09fa6623ea3e0bf65652)

czarcas7ic (approved):

p0mvn:
@cursor review

PR #535, opened 2026-08-04
feat(state): rebuild historical note commitment trees by verified replay

p0mvn:
Stacked on #534.

## Motivation

A fast-synced node cannot serve historical treestates because it never wrote the per-height trees. It can, however, rebuild them: it retains the block bodies, and it already stores authenticated per-height roots.

Implements the verification property in §4.2 and the replay in §4.3 of `docs/design/historical-treestate-serving.md`, plus the phase A tooling in §8.1.

## Solution

Replay the retained block bodies forward from the nearest frontier the node already holds, and accept the result only if it reproduces the authenticated root in `commitment_roots_by_height`. A note commitment root is a binding commitment to its frontier, so a frontier that reproduces the authenticated root *is* the frontier — which is what makes a rebuilt tree trustworthy without trusting its source.

**No RPC behaviour changes in this PR.** The engine is exercised only through `audit-historical-treestates`, which reports whether a database has the inputs replay needs (gap-free root index, retained bodies) and can walk the range checking and timing every rebuilt tree. Landing the engine with zero behaviour change lets the risky part be reviewed on its own.

Read-only opens no longer demand column families the database lacks. Without that, this tooling cannot open an older database at all — a pre-existing limitation that only surfaces once you try to inspect one.

## Testing

Unit tests for the streaming gap scan and body scan, both covering interior holes rather than just range endpoints; property tests that derive every height in a synthetic absent band and compare each pool against a legacy node's own per-height trees.

**Measured on a Mainnet fast-synced archive snapshot** (band `[0, 3,358,006)`):

- Inputs present: root index gap-free across all 3,358,006 heights, every block body retained.
- **3,358,006 heights rebuilt and root-checked, zero mismatches**, across real Sapling and Orchard replay.
- ~1.9 ms/block mean, varying ~50x by height.

`scripts/differential-treestate-check.py` compares rebuilt trees against an independently legacy-synced node's `z_gettreestate`: **421 roots across both pools, zero mismatches**.

Ironwood is structurally unexercised — NU6_3 activates above the snapshot's handoff, so no Ironwood commitment exists in the band.

### DIFF
--- a/crates/zakura-rpc/src/methods/tests/vectors.rs
+++ b/crates/zakura-rpc/src/methods/tests/vectors.rs
@@ -1154,6 +1154,150 @@ async fn rpc_getblock_missing_error() {
     assert!(rpc_tx_queue_task_result.is_none());
 }
 
+/// A `z_gettreestate` inside a fast-synced node's absent band must fail loudly.
+///
+/// Before this, an absent per-height tree became a JSON `null` treestate. Clients following
+/// the lightwalletd contract read that as the *empty* tree and derive a wallet birthday anchor
+/// asserting an empty commitment tree deep in the chain, with no error at the point of
+/// corruption. The read handler now returns [`zakura_state::HistoricalTreeUnavailable`], which
+/// the RPC surfaces as an error.
+#[tokio::test(flavor = "multi_thread")]
+async fn rpc_z_get_treestate_absent_band_is_an_error() {
+    let _init_guard = zakura_test::init();
+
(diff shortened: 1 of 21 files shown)


p0mvn:
Superseded by #552, which recreates this against updated `main` (`04ebb2703`, v1.1.0-rc2) with no conflicts.

The first commit on this branch — "fix(state): report a typed error for absent historical treestates" — was dropped there, because it merged as #534 in an evolved form and main restructured that API afterwards. Only the verified-replay commit carries over. Full rebase notes are on #552.

PR #536, opened 2026-08-04

p0mvn:
Stacked on #535.

## Motivation

A fast-synced node records no completed subtree roots below its handoff — they are a by-product of the per-height tree maintenance the fast path skips, so they cannot be recorded during fast sync at all. Implements §4.6 of `docs/design/historical-treestate-serving.md`.

## Solution

Publish them as a small artifact (71,879 bytes for Mainnet: 1,127 Sapling, 763 Orchard) and serve them, with the node's own rows always taking precedence.

- **Published records never displace the node's own rows.** A correct artifact holds only subtrees completed at or below the handoff and so never overlaps — which makes an overlap exactly the corrupt-or-hostile case where the node's own verified value must win. The artifact's digest is self-contained, not a signature.

Generation replays the band and checks itself against the authenticated roots, so it needs an archive node but **not** a legacy-synced one — which §5 of the design assumed. Subtree roots fall out of that replay as interior nodes, pinned between two root-checked endpoints.

## Testing

**Verified against ground truth**: above a fast-synced node's handoff the database *does* store subtree rows, so replaying `(3,358,006, 3,432,538]` and comparing gives **5 matched, 0 mismatched**.

### DIFF
--- a/.github/workflows/update-release-state.yml
+++ b/.github/workflows/update-release-state.yml
@@ -64,6 +64,7 @@ jobs:
           CHECKPOINTS: crates/zakura-chain/src/parameters/checkpoint/main-checkpoints.txt
           FRONTIER: crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.bin
           PROVENANCE: crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.json
+          SUBTREES: crates/zakura-state/src/service/finalized_state/vct/mainnet-subtrees.bin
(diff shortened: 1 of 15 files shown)


p0mvn:
Superseded by #553, which recreates this on top of #552 against updated `main`, with no conflicts.

One resolution there is worth review rather than being purely mechanical: main's #534 turned the subtree reads into `Result`-returning calls that apply the continuity contract themselves, so the published-artifact fallback was adapted to that shape. Notes are on #553.

PR #544, opened 2026-08-05
fix(release): preserve crate identity across moves

p0mvn:
## Motivation

The automated v1.1.0-rc2 preparation treated every crate moved under `crates/` after v1.1.0-rc1 as a new package. Because the planner matched base-tag crates by their current manifest paths, it emitted manual rows instead of the required version and cascade bumps, and the publish-graph verification failed.

## Solution

Resolve packages at the base tag by Cargo package name, version, and historical manifest path. The release planner now compares both the historical and current crate directories, preserving package identity across repository-only moves while retaining the manual gate for genuinely new crates.

Add a focused git-fixture regression test that tags a package before moving its manifest and verifies that its original identity and version are still discovered. Run that test in the existing scripts test job.

## Testing

- `bash scripts/tests/test_release_packages.sh`
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`
- `bash -n scripts/prepare-release.sh scripts/lib/release-packages.sh scripts/tests/test_release_packages.sh`
- `shellcheck scripts/prepare-release.sh scripts/lib/release-packages.sh scripts/tests/test_release_packages.sh`
- `actionlint .github/workflows/changelog.yml`
- Verified `list_release_packages_at v1.1.0-rc1` discovers the pre-move manifests and versions.
- Full `prepare-release.sh` dry run not run locally: macOS ships Bash 3, while this existing script requires Bash 4 associative arrays. The draft CI runs on Linux.

## Changelog

No fragment required: this changes internal release tooling without modifying Rust sources or Cargo manifests.

- Failed preparation: https://github.com/zakura-core/zakura/actions/runs/30957674595
- Generated release PR: https://github.com/zakura-core/zakura/pull/540
- Crate layout move: https://github.com/zakura-core/zakura/pull/525

### Follow-up Work

After this fix merges, rerun the v1.1.0-rc2 preparation workflow so it replaces the incomplete draft release branch with a correctly bumped crate graph.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -63,5 +63,7 @@ jobs:
         run: ./scripts/changelog.py check
 
       - name: Run scripts/tests unit tests
-        # covers the changelog tool and the bench digest helper (both stdlib-only)
-        run: python3 -m unittest discover -s scripts/tests -p 'test_*.py'
+        # These tests use only Python's stdlib and standard shell/git tools.
+        run: |
+          python3 -m unittest discover -s scripts/tests -p 'test_*.py'
+          bash scripts/tests/test_release_packages.sh
--- a/scripts/lib/release-packages.sh
+++ b/scripts/lib/release-packages.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# Print package name, version, and manifest path for every package at a git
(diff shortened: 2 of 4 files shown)


p0mvn:
This PR is blocking further iterations on RC CI; going to merge to keep fixing

PR #548, opened 2026-08-05
ci(release): parallelize public API verification

p0mvn:
## Motivation

The successful v1.1.0-rc2 preparation took 35m47s. Its independent public-API diffs ran serially for 11m15s, making them the largest safely parallelizable stage: `zakura-rpc` took 3m19s, `zakura-network` 1m57s, and `zakura-consensus` 1m16s.

## Solution

Move `cargo public-api diff latest` from the serial preparation job into the existing per-package release-PR matrix. Each changed published library now runs on its own runner after the draft opens, and the required `semver-checks` aggregate fails if any API diff fails.

Preserve the existing scope by excluding the separately versioned `zakura` package. Ordinary PRs and main pushes do not run the new matrix; it is limited to `release/v*` pull requests.

The generated release PR records that API evidence is delegated to parallel PR CI rather than claiming it already passed.

## Testing

- `actionlint .github/workflows/prepare-release-pr.yml .github/workflows/semver-checks.yml`
- `python3 .github/workflows/scripts/test_affected_semver_packages.py`
- `git diff --check`

The matrix itself requires a generated `release/v*` PR and therefore cannot execute on this non-release draft. After merge, rerun release preparation and compare the preparation duration plus the parallel API job durations against the baseline run.

## Changelog

No fragment required: this changes internal release CI without modifying Rust sources or Cargo manifests.

- Timing baseline: https://github.com/zakura-core/zakura/actions/runs/30966042771
- Generated release PR: https://github.com/zakura-core/zakura/pull/547

### Follow-up Work

Measure the first post-merge release run. The serial 11m15s stage should be replaced by the slowest per-package matrix path plus runner setup, while preserving all API-diff evidence.

### DIFF
--- a/.github/workflows/prepare-release-pr.yml
+++ b/.github/workflows/prepare-release-pr.yml
@@ -162,11 +162,10 @@ jobs:
           restore-keys: prepare-release-baselines-${{ steps.toolchain.outputs.cachekey }}-
 
       # Deliberately unpinned (like semver-checks.yml): cargo-semver-checks
-      # must track new stable rustdoc formats. cargo-public-api supplies
-      # independent API-diff evidence for each changed published library.
+      # must track new stable rustdoc formats.
       - uses: taiki-e/install-action@c44f6b046f1c29ae5918b1e0bfdbb2f1813836fd #v2.84.1
         with:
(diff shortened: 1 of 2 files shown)


p0mvn:
Merging to immediately test the speed-up

PR #552, opened 2026-08-05
feat(state): rebuild historical note commitment trees by verified replay

p0mvn:
## Motivation

A fast-synced node cannot serve historical treestates because it never wrote the per-height trees. It can, however, rebuild them: it retains the block bodies, and it already stores authenticated per-height roots.

## Solution

Replay the retained block bodies forward from the nearest frontier the node already holds, and accept the result only if it reproduces the authenticated root in `commitment_roots_by_height`. A note commitment root is a binding commitment to its frontier, so a frontier that reproduces the authenticated root *is* the frontier — which is what makes a rebuilt tree trustworthy without trusting its source.

**No RPC behaviour changes in this PR.** The engine is exercised only through `audit-historical-treestates`, which reports whether a database has the inputs replay needs (gap-free root index, retained bodies) and can walk the range checking and timing every rebuilt tree. Landing the engine with zero behaviour change lets the risky part be reviewed on its own.

Read-only opens no longer demand column families the database lacks. Without that, this tooling cannot open an older database at all — a pre-existing limitation that only surfaces once you try to inspect one.

## Testing

Verified after the rebase and the version bump: `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets` (clean), `cargo fmt --all`, `cargo test -p zakura-state --lib` (457 passed), `cargo test -p zakura-rpc --lib` (99 passed), `./scripts/changelog.py check`, `zakurad generate` diffed against the stored config fixture, and the differential property test `vct_fast_sync_handoff_marks_database_and_resumes` specifically.

The original evidence below stands unchanged — the rebase dropped a superseded commit and reconciled imports and export lists; it did not touch the replay engine.

Unit tests for the streaming gap scan and body scan, both covering interior holes rather than just range endpoints; property tests that derive every height in a synthetic absent band and compare each pool against a legacy node's own per-height trees.

**Measured on a Mainnet fast-synced archive snapshot** (band `[0, 3,358,006)`):

- Inputs present: root index gap-free across all 3,358,006 heights, every block body retained.
- **3,358,006 heights rebuilt and root-checked, zero mismatches**, across real Sapling and Orchard replay.
- ~1.9 ms/block mean, varying ~50x by height.

`scripts/differential-treestate-check.py` compares rebuilt trees against an independently legacy-synced node's `z_gettreestate`: **421 roots across both pools, zero mismatches**.

Ironwood is structurally unexercised — NU6_3 activates above the snapshot's handoff, so no Ironwood commitment exists in the band.

### DIFF
--- a/scripts/make/treestate-audit.mk
+++ b/scripts/make/treestate-audit.mk
@@ -29,9 +29,9 @@ TREESTATE_RANGE_ARGS = \
 	$(if $(TREESTATE_TO),--to "$(TREESTATE_TO)") \
 	--step "$(TREESTATE_STEP)"
 
-# Report the database's historical-treestate inventory without deriving any trees.
+# Report the database inventory and explicitly verify full-band block-body retention.
 treestate-audit-inventory:
-	$(TREESTATE_AUDIT) $(TREESTATE_EXTRA_ARGS)
+	$(TREESTATE_AUDIT) --scan-block-bodies $(TREESTATE_EXTRA_ARGS)
 
 # Replay the post-checkpoint band and compare the resulting subtree roots with stored rows.
 treestate-audit-subtrees:
--- a/scripts/make/treestate-audit.mk
+++ b/scripts/make/treestate-audit.mk
@@ -1,0 +1,59 @@
+# Historical treestate audit helpers.
+#
+# Override variables on the command line, for example:
+#   make treestate-audit-walk TREESTATE_CACHE_DIR=/path/to/cache TREESTATE_FROM=419200 TREESTATE_TO=435000
+
+.PHONY: \
+	treestate-audit-inventory \
+	treestate-audit-subtrees \
+	treestate-audit-walk \
+	treestate-audit-samples \
+	treestate-audit-roots \
+	treestate-audit-differential
+
+TREESTATE_ZAKURAD_BIN ?= $(CURDIR)/target/release/zakurad
+TREESTATE_CHECK_SCRIPT ?= $(CURDIR)/scripts/differential-treestate-check.py
+TREESTATE_CACHE_DIR ?= $(HOME)/.local/zakura-dev/cache
+TREESTATE_NETWORK ?= Mainnet
+TREESTATE_FROM ?=
+TREESTATE_TO ?=
+TREESTATE_STEP ?= 1
+TREESTATE_RPC_URL ?=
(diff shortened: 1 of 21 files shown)


p0mvn:
Heads-up on the one red check, carried over rather than introduced by the rebase: `semver (zakura)` fails with `enum_variant_added` on `ZakuradCmd::AuditHistoricalTreestates`. Adding a variant to an exhaustive public enum needs a major bump of the `zakura` crate (currently `1.1.0-rc2`), per the job's own remedy:

```
cargo release version --verbose --execute --allow-branch '*' -p zakura major
```

#535 and #536 failed this identically before being recreated, so nothing here changed it. Left alone deliberately — the release-line decision is yours, not something a rebase should make.

p0mvn:
@v12this review

p0mvn:
@cursor review

p0mvn:
[F-172112](https://v12.sh/runs/6086/172112) is fine not to do, I think. Intrusive on core paths for a CLI change

p0mvn:
@v12this review

p0mvn:
@cursor review

scripts/make/treestate-audit.mk
```diff
@@ -0,0 +1,65 @@
...
+TREESTATE_AUDIT = "$(TREESTATE_ZAKURAD_BIN)" audit-historical-treestates \
+	--cache-dir "$(TREESTATE_CACHE_DIR)" \
+	--network "$(TREESTATE_NETWORK)"
+TREESTATE_RANGE_ARGS = \
+	$(if $(TREESTATE_FROM),--from "$(TREESTATE_FROM)") \
+	$(if $(TREESTATE_TO),--to "$(TREESTATE_TO)") \
+	--step "$(TREESTATE_STEP)"
+
+# Report the database inventory and explicitly verify full-band block-body retention.
+treestate-audit-inventory:
```
evan-forbes, line 33: generally I don't mind switching to make for our scripts if that's helpful

currently we have cargo xtask for one or two things. originally just used that to avoid make

evan-forbes (approved):
utACK

scripts/make/treestate-audit.mk
p0mvn: Most of our infra is set up with make:

https://github.com/zakura-core/zakura/blob/main/Makefile

PR #553, opened 2026-08-05

p0mvn:
> Recreated from #536 against updated `main`. #536 was conflicting through its base and is now closed. Stacked on #552 (the recreated #535), same shape as before.
>
> **What changed in the rebase.** Main's #534 turned the subtree reads into `Result`-returning calls that apply the continuity contract and report the absent band themselves, where this branch was written against the older "return empty, then check separately" API. The published-artifact fallback in `ReadRequest::{Sapling,Orchard,Ironwood}Subtrees` was adapted to that shape rather than reimplemented: the node's own rows still answer whenever they cover `start_index`, the artifact union is consulted only when they do not, and if that union still does not reach `start_index` the original result — including its typed absent-band error — is propagated unchanged. **This is the one resolution that is not purely mechanical and is worth a look.**
>
> Other conflicts, all mechanical: export lists in `lib.rs` / `finalized_state.rs` / `read.rs` reconciled against main (`generate_mainnet_from_archive`, `GeneratorError` and `ValidateVctSproutHistory` are gone via #533); `tree.rs` kept main's `subtree_completed_by_last_checkpoint` naming rather than the branch's `subtree_completed_by_handoff`, with the two new helpers inserted ahead of it.
>
> **Version bump.** This branch adds `ZakuradCmd::ExportHistoricalTreestates`, a second variant on the same exhaustive public enum, so it is covered by the `zakura` `1.1.0-rc2` -> `2.0.0-rc2` bump that #552 carries. Nothing extra is needed here. The stored-config fixture rename that bump requires (`v2.0.0-rc2.toml`) also comes from #552; `zakurad generate` on this branch reproduces it byte for byte, because the `historical_subtree_artifact` field this PR adds is an unset `Option` and so does not appear in the generated file.
>
> Verified after rebase: `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets` (clean), `cargo fmt --all`, `cargo test -p zakura-state --lib` (457 passed), `cargo test -p zakura-rpc --lib` (99 passed), `./scripts/changelog.py check`.

## Motivation

A fast-synced node records no completed subtree roots below its handoff — they are a by-product of the per-height tree maintenance the fast path skips, so they cannot be recorded during fast sync at all. Implements §4.6 of `docs/design/historical-treestate-serving.md`.

## Solution

Publish them as a small artifact (71,879 bytes for Mainnet: 1,127 Sapling, 763 Orchard) and serve them, with the node's own rows always taking precedence.

- **Published records never displace the node's own rows.** A correct artifact holds only subtrees completed at or below the handoff and so never overlaps — which makes an overlap exactly the corrupt-or-hostile case where the node's own verified value must win. The artifact's digest is self-contained, not a signature.

Generation replays the band and checks itself against the authenticated roots, so it needs an archive node but **not** a legacy-synced one — which §5 of the design assumed. Subtree roots fall out of that replay as interior nodes, pinned between two root-checked endpoints.

## Testing

**Verified against ground truth**: above a fast-synced node's handoff the database *does* store subtree rows, so replaying `(3,358,006, 3,432,538]` and comparing gives **5 matched, 0 mismatched**.

### DIFF
--- a/.github/scripts/import-release-state.py
+++ b/.github/scripts/import-release-state.py
@@ -7,6 +7,7 @@ import argparse
 import hashlib
 import json
 import re
+import struct
 import sys
 import tempfile
 import unittest
@@ -16,8 +17,14 @@ from typing import Any
 CHECKPOINTS = Path("crates/zakura-chain/src/parameters/checkpoint/main-checkpoints.txt")
 FRONTIER = Path("crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.bin")
 PROVENANCE = Path("crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.json")
+SUBTREES = Path("crates/zakura-state/src/service/finalized_state/vct/mainnet-subtrees.bin")
 EOS_FILE = Path("crates/zakurad/src/components/sync/end_of_support.rs")
 EOS_PATTERN = re.compile(r"(ESTIMATED_RELEASE_HEIGHT: u32 = )([0-9_]+)")
+SUBTREE_BUNDLE_NAME = "mainnet-treestate-subtrees.bin"
+SUBTREE_HEADER_PREFIX = struct.Struct("<8sHBIIII")
+SUBTREE_HEADER_LEN = SUBTREE_HEADER_PREFIX.size + 32
+SUBTREE_RECORD_LEN = 2 + 4 + 32
+MAX_SUBTREE_RECORDS = 2**16
 RESOLUTION_KEYS = {
     "height",
     "block_hash",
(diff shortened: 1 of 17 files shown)


p0mvn:
Heads-up on the one red check, carried over rather than introduced by the rebase: `semver (zakura)` fails with `enum_variant_added` on `ZakuradCmd::AuditHistoricalTreestates`. Adding a variant to an exhaustive public enum needs a major bump of the `zakura` crate (currently `1.1.0-rc2`), per the job's own remedy:

```
cargo release version --verbose --execute --allow-branch '*' -p zakura major
```

#535 and #536 failed this identically before being recreated, so nothing here changed it. Left alone deliberately — the release-line decision is yours, not something a rebase should make.

p0mvn:
Rebased onto current main after #552 merged. Closing this draft so it can be recreated from the rewritten branch.

p0mvn:
Replacement draft: https://github.com/zakura-core/zakura/pull/593

PR #557, opened 2026-08-05
fix(network): let an unproven block-sync peer re-probe after a failed probe

### LINKED ISSUE #526: Investigate transient native block-sync stall caught by E2E trace oracle
## Summary

PR #524 hit a transient pure-Zakura block-sync stall in the required regtest E2E lane. The functional catch-up eventually recovered, but the trace oracle correctly retained and reported the earlier progress violation.

- PR: https://github.com/zakura-core/zakura/pull/524
- Failed workflow: https://github.com/zakura-core/zakura/actions/runs/30760807405
- Failed job: https://github.com/zakura-core/zakura/actions/runs/30760807405/job/91530908311
- Trace artifact: https://github.com/zakura-core/zakura/actions/runs/30760807405/artifacts/8837465314
- Successful run of the exact base commit: https://github.com/zakura-core/zakura/actions/runs/30756972335

## Observed behavior

During the from-scratch kind-6 catch-up, pure-Zakura node2 connected to one servable outbound peer and sent a block-sync request, but `sync_block_body_received` remained at zero for about 208 seconds.

The trace oracle uses a 180-second persistent-lag window and failed this state:

```text
invariant: block_sync_queue_lag_makes_progress
best_header_tip: 163
verified_block_tip: 0
needed_count: 163
queue_len: 1
assigned_len: 0
outstanding: 0
floor_gap_state: queued
floor_gap_servable_peers: 1
outbound_peers: 1
```

The node then recovered:

- received all 163 requested bodies;
- caught up to height 163;
- completed the non-finalized reorg to height 164;
- drained the block-sync budget, reorder, applying, and outstanding counters.

The test failed only when the trace oracle examined the earlier stalled interval.

## Expected behavior

A queued body-sync gap with a connected servable peer should produce request assignment and real body progress within the oracle window.

## Scope

This appears unrelated to PR #524. That PR only stops rewriting the legacy BIP14 user-agent string. The stalled node was the pure-Zakura node2 using native kind-6 block sync, and its native connection and initial request were already established.

The exact base-main E2E run passed; in that run node2 received all 163 bodies immediately after its catch-up request.

## Reproduction

Run the PR-gate E2E lane:

```console
ZAKURA_REGTEST_E2E=1 ZAKURA_E2E_MODE=pr-gate \
  cargo test -p zakura --test zakura_regtest_e2e -- --ignored --nocapture
```

The failure is currently intermittent. A rerun of the failed PR workflow has been requested.

megabyte0x:
Closes #526

## Motivation

An unproven block-sync peer could wedge permanently the first time its cold-start probe ended without a body — stalling body sync outright when it was the only servable carrier.

An unproven peer's single cold-start probe is by construction a *floor* request, so it carries the short floor-rescue leash rather than the base request timeout. But `try_fill` charged the probe-first cap against the cumulative `requests_without_block_progress` streak, which only an accepted body or a destructive view reset clears. So when that probe ended without a body — floor leash expiring, request timeout, or `RangeUnavailable` — the streak stayed at the unproven cap of 1 and the peer could never issue another request: clearing the charge required a body, and obtaining a body required a request.

Same defect class as the already-fixed destructive-reset trigger (`view_reset_reclears_probe_streak_so_unproven_peer_can_reprobe`); the floor-rescue trigger was missed.

The #526 trace is this wedge — a servable peer and queued work with zero requests in flight (`outstanding: 0`, `queue_len: 1`, `floor_gap_servable_peers: 1`), 163 blocks pending and `sync_block_body_received` at zero for ~208 s.

## Solution

Gate an unproven peer on its **in-flight** probes instead of the cumulative streak, via `DownloadWindow::no_progress_at_cap`. That is what the cap is for, and what `FillStop::InitialBlockProbeRequestCap` already documents: "an unproven peer's single cold-start probe is in flight … until it serves (or fails) a body".

The proven path stays cumulative. Burst protection is unchanged — still exactly one probe in flight. A re-probe buys no extra time: `arm_liveness` arms `block_liveness_deadline` only when unset, so a peer that truly serves nothing is still parked on the schedule its first probe started.

## Testing

`unproven_peer_reprobes_after_its_probe_ends_without_a_body` drives the real routine over a single peer, lets the probe expire unanswered, and requires a second `GetBlocks` for the queued floor gap before the liveness deadline — it times out against the old code. `probe_cap_reopens_once_an_unproven_peers_probe_ends_without_a_body` pins the window contract (at cap while in flight, reopened once drained, deadline not extended by a re-probe); `proven_peer_no_progress_cap_stays_cumulative` pins that the proven cap is unaffected. `fuzz_silent_peer_request_cap` now asserts peak *concurrent* unproven request depth, which is the invariant that actually encodes the burst bound.

The second commit repairs a test this fix invalidated. `remote_stream_eof_after_probe_timeout_still_parks_session` pins the EOF branch where `outstanding` is empty but liveness is still armed, and it reached that state by relying on the cap wedging the peer. The peer now re-probes every ~170 ms, so the EOF landed with a probe outstanding and the test had drifted onto the branch already covered by `remote_stream_eof_parks_session`. Moving the peer's servable range past the queued height restores the intended state; verified re-pinned by reverting `handle_remote_stream_closed` to key clean exit off `outstanding.is_empty()` alone and confirming the test fails again.

```console
cargo fmt --all -- --check
cargo clippy -p zakura-network --all-targets -- -D warnings
cargo test -p zakura-network
```

Clean, clean, and 1190 passed / 4 ignored. Three `peer_set::initialize` listener tests fail locally with `AddrNotAvailable` — they bind arbitrary `127.x.x.x` source addresses, which macOS does not auto-assign; untouched by this branch and covered by Linux CI.

## Changelog

Fragment added under `## Fixed`.

### DIFF
--- a/crates/zakura-network/src/zakura/block_sync/peer_routine.rs
+++ b/crates/zakura-network/src/zakura/block_sync/peer_routine.rs
@@ -713,8 +713,7 @@ impl PeerRoutine {
             if !self.received_status {
                 break FillStop::NoStatus;
             }
-            if self.window.requests_without_block_progress >= self.window.no_progress_request_cap()
-            {
+            if self.window.no_progress_at_cap() {
                 break if self.window.has_block_progress() {
                     FillStop::NoBlockProgressRequestCap
                 } else {
--- a/crates/zakura-network/src/zakura/block_sync/state.rs
+++ b/crates/zakura-network/src/zakura/block_sync/state.rs
@@ -690,6 +690,34 @@ impl DownloadWindow {
         }
     }
 
+    /// Whether the no-progress gate currently withholds further work from this peer.
+    ///
+    /// A **proven** peer is charged cumulatively: it has served a body, so a long run
+    /// of requests with no further accepted body is the signal, and once it reaches
+    /// the cap the block-liveness deadline governs.
+    ///
+    /// An **unproven** peer is charged on its *in-flight* probes instead. The
+    /// probe-first cap exists so a peer that accepts requests but never serves bodies
+    /// cannot spend a full BBR cold-start burst — a bound on concurrency, not on
+    /// lifetime attempts. Charging it cumulatively wedged the peer permanently the
+    /// first time a probe ended without a body (a floor request rescued at the short
+    /// floor leash, a request timeout, or a `RangeUnavailable`), because only an
+    /// accepted body — which needs a request the peer may no longer issue — or a
+    /// destructive view reset clears the streak. With a single servable peer that
+    /// stalled body sync outright until the liveness park and its cooldown elapsed.
+    ///
+    /// A re-probe never buys extra time: [`arm_liveness`](Self::arm_liveness) only
+    /// arms `block_liveness_deadline` when it is unset, so a peer that truly serves
+    /// nothing is still parked on the schedule its first probe started.
+    pub(super) fn no_progress_at_cap(&self) -> bool {
+        let charged = if self.has_block_progress() {
+            self.requests_without_block_progress
+        } else {
+            u32::try_from(self.outstanding.len()).unwrap_or(u32::MAX)
+        };
+        charged >= self.no_progress_request_cap()
+    }
+
     pub(super) fn arm_liveness(&mut self, now: Instant, timeout: Duration) {
         self.last_request_at = Some(now);
         self.requests_without_block_progress =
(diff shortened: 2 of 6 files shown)


ValarDragon:
Closing for now, were doing other refactors to this part of the code and will evaluate if the problem still appears. Please ask if taking on an issue is helpful! (Unless its a net new one you found)

Appreciate the initiative

megabyte0x:
no worries :)

PR #563, opened 2026-08-05
feat(state)!: persist and recover fork-aware header chains

evan-forbes:
Stack layer 3/8. Parent: #587. Next: #588.

This PR retains the durable fork-aware header-chain state store. It now contains one state-store performance commit: perf(state): slide header validation context incrementally. The writer recognizes only an authenticated one-block finalized-parent advance, inserts the previous finalized header as context, and removes exactly the outgoing context row after the predecessor window fills. Every other shape keeps the existing reconstruction path. Header-validation context encoding now serializes directly into the durable encoder while preserving compatible bytes.

Tests at exact head 23f630d9a:
- cargo fmt --all -- --check
- cargo test -p zakura-state --lib service::finalized_state::header_chain: passed
- cargo clippy -p zakura-header-chain -p zakura-node-services -p zakura-state --all-targets -- -D warnings

Changelog: docs/changelog/unreleased/563.md remains the existing explicit no-changelog fragment.


evan-forbes:
Superseded by #586, which now contains the final combined stack.

PR #460, opened 2026-07-27
fix(network): tag block gossip to avoid findblocks races

evan-forbes:
## Motivation

Legacy `FindBlocks` responses and unsolicited block gossip both use block-only
`inv` messages. If gossip arrives while `FindBlocks` is pending, the legacy
connection state machine can consume that gossip as the response and attribute
sync-only lookahead penalties to its advertiser.

## Solution

Append a small versioned `ZINV` trailing extension only to block advertisements
sent through `AdvertiseBlock` and `AdvertiseBlockToAll`. Legacy decoders
continue to read the standard inventory prefix and ignore the extension.

A valid tag is private connection metadata, not a public Rust API change. There
is no request ID, service bit, handshake capability, protocol threshold, or
`VersionMessage` field. `FindBlocks` response inventory and transaction
inventory remain untagged.

| Message during `FindBlocks` | Handling | Ban/disconnect behavior |
| --- | --- | --- |
| Untagged block `INV` | Existing `FindBlocks` response path | Existing scoring and bans remain possible |
| Tagged block-gossip `INV` | Existing inbound gossip path; `FindBlocks` remains pending | No mistaken sync lookahead score; no disconnect merely for sending the gossip |
| Tagged gossip containing a consensus-invalid block | Existing gossip verification | Existing consensus misbehavior score and ban remain possible |
| Malformed protocol traffic or transport failure | Existing connection error paths | Existing disconnect behavior remains unchanged |

A tagged far-ahead gossip block follows the existing inbound path, which drops
out-of-range blocks without attributing a sync misbehavior score. The tag itself
does not disable banning: consensus-invalid advertised blocks retain normal
gossip verification and scoring.

Untagged responses preserve terminal-hash stripping and #459 behavior, the
50,000-block lookahead limit, `AboveLookaheadHeightLimit`, `InvalidHeight`,
consensus-invalid and missing-height scoring, and
`Response::BlockHashes(Vec<Hash>)`.

For an attributed untagged sync violation, sync emits 100 points. Updates are
batched for up to 30 seconds; the address book bans at a cumulative score of
100; the peer set then drops active services and rejects future connections.
The sync handler does not directly close the socket.

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p zakura-network --lib`
  - 1,163 passed; 4 ignored
- `CXXFLAGS='-include cstdint' cargo test -p zakura components::sync::tests`
  - 54 passed
- focused codec tests for tagged/untagged round trips, malformed extensions,
[...]

PR #574, opened 2026-08-05
fix(sync): stop scoring peers that serve far-ahead blocks

evan-forbes:
## Motivation

`handle_block_response` scored 100 misbehavior points — an immediate ban at the
address book's threshold — whenever a downloaded block came back above the
lookahead height limit:

```rust
Err(BlockDownloadVerifyError::AboveLookaheadHeightLimit {
    advertiser_addr: Some(advertiser_addr),
    ..
}) => {
    let _ = self.misbehavior_sender.try_send((advertiser_addr, 100));
}
```

`advertiser_addr` there is filled in by the block download task from the peer
that answered our `BlocksByHash` request. But a far-ahead height does not
originate with that peer: the hash was queued into `download_set` by
`obtain_tips` / `extend_tips` from some peer's `FindBlocks` response, and
`request_blocks` carries only hashes forward. The two peers are usually
different, so the ban lands on a peer whose only action was to honestly serve a
block this node asked for.

This is OtterSec finding OS-ZCA-ADV-15, *Far-Ahead Hash Rejection is
Misattributed* (Low).

## Solution

Stop scoring `AboveLookaheadHeightLimit`, and document why at the match arm.

The peer actually responsible cannot be identified at this point:
`zn::Response::BlockHashes(Vec<Hash>)` carries no peer address, so there is
nothing to attribute the far-ahead height to. Attributing it correctly means
plumbing the responding peer through the `FindBlocks` response type, which is a
larger network-layer change than this fix.

Everything protective about the current path is unchanged:

- the far-ahead block is still dropped by `handle_response`;
- sync still does not restart on this error (GHSA-gvjc-3w7c-92jx);
- `advertiser_addr` is still carried on the error, so the drop log still names
  the serving peer;
- `InvalidHeight` and consensus-invalid blocks — where the serving peer *is* at
  fault, because the fault is in the bytes it sent — keep their 100-point
  scores.

The same hunk was previously written as part of #457, which was closed because
of the near-tip singleton gate it was bundled with. This PR takes only the
attribution fix, with none of the gating.

## Testing

- `CXXFLAGS='-include cstdint' cargo test -p zakura --lib components::sync::tests`
  - 55 passed; 0 failed
- `CXXFLAGS='-include cstdint' cargo clippy -p zakura --all-targets --locked -- -D warnings`
- `npx markdownlint-cli docs/changelog/unreleased/574.md --config .github/.markdownlint.yaml`

New regression test `far_ahead_block_does_not_score_serving_peer` drives
`handle_block_response` with an `AboveLookaheadHeightLimit` error carrying a
serving-peer address and asserts the misbehavior channel stays empty, then
drives the *same* peer through an `InvalidHeight` error and asserts it is still
scored 100 — pinning the boundary rather than just the removal. Both calls
return `Ok(())`, which is the existing no-restart behavior for these errors;
the block itself is discarded earlier, in the download task, before it ever
reaches the verifier.

`above_lookahead_has_peer_attribution` still asserts the error carries
`advertiser_addr`; its doc comment now says the address is for the drop logs
rather than for scoring.

## Changelog

`docs/changelog/unreleased/<PR>.md` under `Fixed`.

- OtterSec, *Zcash — Security Assessment* (July 18, 2026), finding
  `OS-ZCA-ADV-15`.
- Prior attempt: #457 (closed).
- Related, still open: #460, which removes one *source* of bad attribution
  (block gossip mistaken for a `FindBlocks` response) but does not change this
  scoring arm.

### Follow-up Work

Attributing far-ahead hashes to the peer that supplied them requires carrying
the responding peer address on `FindBlocks` responses. Not attempted here.

### DIFF
--- a/crates/zakurad/src/components/sync.rs
+++ b/crates/zakurad/src/components/sync.rs
@@ -2139,9 +2139,9 @@ where
             // requested, but the height came from whichever peer supplied the
             // hash in a `FindBlocks` response. Those are usually different
             // peers, and `Response::BlockHashes` carries no address, so the
-            // responsible peer is unknown here. The block is still dropped by
-            // `handle_response`; scoring the server would ban a peer for
-            // honestly answering our own request.
+            // responsible peer is unknown here. The download task has already
+            // discarded the block without verifying it; scoring the server as
+            // well would ban a peer for honestly answering our own request.
             Err(_) => {}
         };
 
--- a/crates/zakurad/src/components/sync.rs
+++ b/crates/zakurad/src/components/sync.rs
@@ -2127,13 +2127,6 @@ where
                     .try_send((advertiser_addr, error.misbehavior_score()));
             }
 
-            Err(BlockDownloadVerifyError::AboveLookaheadHeightLimit {
-                advertiser_addr: Some(advertiser_addr),
-                ..
-            }) => {
-                let _ = self.misbehavior_sender.try_send((advertiser_addr, 100));
-            }
-
             Err(BlockDownloadVerifyError::InvalidHeight {
                 advertiser_addr: Some(advertiser_addr),
                 ..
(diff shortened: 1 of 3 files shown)


evan-forbes:
Moving this to the private repo: zakura-core/zakura-private#17. Branch `fix/lookahead-misattribution` is left in place here.

PR #579, opened 2026-08-05
ci(release): import fresh state during release preparation

p0mvn:
## Motivation

Release preparation compares the committed Mainnet release state with a pointer that advances daily. Even after merging a release-state update, the next publisher bundle can make preparation fail again before the release PR is opened.

## Solution

- Extract the existing append-only release-state import into a shared, self-tested Python helper.
- Import and validate a newer verified bundle directly into the draft release PR before running release preparation.
- Re-run the checkpoint/frontier Rust proofs when an import occurs and include bundle provenance plus an explicit review item in the PR body.
- Keep the standalone update workflow on the same importer so both paths enforce identical behavior.
- Preserve the urgent-RC waiver path by skipping automatic import when a waiver is supplied.

## Testing

- `python3 .github/scripts/import-release-state.py --self-test`
- `python3 .github/scripts/fetch-release-state.py --self-test`
- `python3 scripts/release-readiness.py self-test`
- `actionlint .github/workflows/prepare-release-pr.yml .github/workflows/update-release-state.yml`
- `git diff --check`
- IDE diagnostics: no errors

The importer tests exercise successful append-only import, a current-state no-op, end-of-support flooring, provenance generation, and rejection of rewritten checkpoint history.

## Changelog

Internal release automation only; no changelog fragment required because no Rust source or `Cargo.toml` changed.

The imported checkpoints and frontier remain subject to the existing consensus-state validation and manual review requirements.

### DIFF
--- a/.github/scripts/import-release-state.py
+++ b/.github/scripts/import-release-state.py
@@ -1,0 +1,223 @@
+#!/usr/bin/env python3
+"""Import a verified Mainnet release-state bundle into a Zakura checkout."""
+
+from __future__ import annotations
+
+import argparse
+import hashlib
+import json
+import re
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from typing import Any
+
+CHECKPOINTS = Path("crates/zakura-chain/src/parameters/checkpoint/main-checkpoints.txt")
+FRONTIER = Path("crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.bin")
+PROVENANCE = Path("crates/zakura-state/src/service/finalized_state/vct/mainnet-frontier.json")
+EOS_FILE = Path("crates/zakurad/src/components/sync/end_of_support.rs")
+EOS_PATTERN = re.compile(r"(ESTIMATED_RELEASE_HEIGHT: u32 = )([0-9_]+)")
+RESOLUTION_KEYS = {
+    "height",
+    "block_hash",
+    "generated_at",
+    "meta_url",
+    "meta_sha256",
+}
+
+
+class BundleImportError(RuntimeError):
+    """The verified bundle cannot safely extend the committed release state."""
+
+
+def _load_resolution(path: Path) -> dict[str, Any]:
+    try:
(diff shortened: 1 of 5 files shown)


p0mvn:
@cursor review

p0mvn:
@cursor push fe2092b4d4

p0mvn:
@cursor review

czarcas7ic:
AI comment:

---

the importer and `prepare-release.sh` both write `ESTIMATED_RELEASE_HEIGHT`, so the second reads its "old" value after the first changed it, and on the default `expected_tag_delay_days: 0` path the PR body prints "kept at 3438827" while the diff shows `3_438_427 -> 3_438_827` (suggest a `--no-eos-floor` flag so `prepare-release.sh` stays the sole owner); minor: `sort_keys=True` reorders `mainnet-frontier.json` into a full-file diff, `docs/design/verified-commitment-trees.md` section 16.2 still documents `update-release-state.yml` as the only import route, and the height-mismatch guard is untested.

https://github.com/zakura-core/zakura/blob/96a0b486aa9b05ee6eecf9a0ebddf2f41dafb93c/scripts/prepare-release.sh#L679-L691

https://github.com/zakura-core/zakura/blob/96a0b486aa9b05ee6eecf9a0ebddf2f41dafb93c/.github/scripts/import-release-state.py#L125-L142

czarcas7ic (approved):

p0mvn:
Addressed in 7ee327679:

- added `--no-eos-floor` and use it during release preparation so `prepare-release.sh` remains the sole owner of the EOS projection and summary;
- stopped sorting provenance JSON keys, avoiding a full-file reorder;
- documented both release-state import routes and their EOS responsibilities;
- added self-tests for the resolution/checkpoint height mismatch and the no-EOS-floor path.

Verified with:
- `python3 .github/scripts/import-release-state.py --self-test`
- `actionlint .github/workflows/prepare-release-pr.yml .github/workflows/update-release-state.yml`
- `markdownlint docs/design/verified-commitment-trees.md --config .github/.markdownlint.yaml`
- `git diff --check`

PR #569, opened 2026-08-05
feat(headersync)!: integrate network and lifecycle coordination

evan-forbes:
> Stack layer **5/8**. Parent: #588. Next: #571.

## Motivation

The legacy requester could not preserve exact request identity, bounded work, fork choice, or terminal ownership across retries, cancellation, fallback, and restart.

## Solution

- add the bounded, versioned, fork-aware header protocol with exact request/session identity;
- add deadlines, retries, retained-path leases, VCT repair, and persistent alarms;
- route normal and repair completions through one registered-attempt gate;
- integrate selected-header work with block sync, consensus, RPC, and node coordination;
- preserve acknowledgements through cancellation, fallback, handoff, and buffered applies;
- retain completion-gate and runtime integration behavior at the cumulative top; and
- bump `zakura-network` and `zakura-consensus` to 7.0.0 at their breaking boundaries.

## V12 audit corrections

All three findings from V12 run 6100 were reproduced and fixed in their owning commits: deferred actions remain bounded and preserve channel backpressure; dropping an accepted apply publishes terminal failure, removes its operation record, and wakes waiters; and unknown-height downloads reserve capacity without falsely claiming checkpoint completion. All 375 zakurad library tests plus the focused driver, coordinator, acknowledgement, handoff, fallback, and restart suites passed.

## Testing

Passed at exact rebased head `f9fbde47d2c2a6ee125020a39d203ab91ab08866`:

- `cargo check --workspace --locked --all-features --all-targets -j 2`
- `cargo clippy --workspace --all-targets --features default-release-binaries -j 2 -- -D warnings`
- `cargo test -p zakura-network --lib -j 2` — 1,048 passed, 3 ignored
- `cargo test -p zakura --lib components::sync -j 2` — 80 passed
- `cargo test -p zakura-consensus router::tests -j 2` — 6 passed
- `cargo test -p zakura-rpc --lib -j 2` — 101 passed, 1 ignored
- `cargo test -p zakura --test acceptance zakura_state_conflict -j 2 -- --exact`

A broad combined consensus/RPC invocation saw one unrelated Sapling batch task cancellation; the exact failed test passed immediately in isolation with `-j 1`, and the header-facing consensus and full RPC suites passed. No Docker, node, SSH, release, or deployment operation was run.

## Changelog

`docs/changelog/unreleased/569.md` is an explicit no-changelog fragment.

### Specifications & References

[...]

PR #586, opened 2026-08-06
feat(header-chain)!: add fork-aware engine

evan-forbes:
> **Scope note (updated).** This PR is no longer the "layer 1 of 8" leaf-crate
> change its earlier description promised. `main` has since absorbed several
> layers of the stack, and this branch now carries the **complete** fork-aware
> header-sync change: the engine crate, its state persistence, its network
> reactor, and the `zakurad` integration. Read this description, not the stack
> table that used to be here.

## What this is

Replaces the legacy linear header-sync model with a fork-aware header chain: a
synchronous header DAG with an atomic transition engine, durable crash-atomic
persistence, a bounded network reactor, and a lifecycle coordinator that owns
the native/legacy handoff.

The legacy model assumes one linear header path and narrows cumulative work. It
cannot represent competing valid forks, independent header/body frontiers, or
bounded retention.

## Size and shape

294 files, ~68.3k insertions, ~51.8k deletions, 32 non-merge commits.

| Area | + | − | What lands here |
| --- | ---: | ---: | --- |
| `crates/zakura-state` | 25.9k | 13.7k | header-chain persistence, recovery, migrations, VCT authentication sweep |
| `crates/zakura-network` | 24.6k | 31.2k | header-sync reactor, scheduler, wire codec; removes the legacy exchange/requester/range modules |
| `fuzz/` | 6.3k | 0 | header-chain fuzz targets |
| `crates/zakurad` | 5.9k | 4.6k | sync lifecycle coordinator, header/block sync drivers |
| `crates/zakura-header-chain` | 1.8k | 1.9k | graph, transition engine, validation (net of in-place refactors) |
| `crates/xtask` | 1.6k | 8 | `header-conformance` and `header-fuzz` tooling |
| everything else | 1.3k | 0.4k | consensus, rpc, chain, docs, CI, docker |

**This diff has no useful linear reading order.** The commit history is the
intended entry point — see "Suggested review order" below.

## Breaking changes

**Library API.** Three crates take a major bump for removed/renamed public
items (`cargo-semver-checks` reports 7 major-level failures for `zakura-state`
alone: removed enums, structs, enum variants, inherent methods and
module-level consts, a changed enum discriminant, and a variant added to an
exhaustive enum):

| crate | before | after |
| --- | --- | --- |
| `zakura-state` | 6.2.0 | **7.0.0** |
| `zakura-network` | 6.1.0 | **7.0.0** |
| `zakura-consensus` | 6.1.0 | **7.0.0** |

`zakura-chain` (5.0.0), `zakura-node-services` (3.2.0) and `zakura-rpc` (7.0.0)
are unchanged; their semver checks pass as-is. `zakura-header-chain` ships at
`1.0.0-rc0` and is excluded from registry-baseline semver checks because it is
not yet published.

**Database.** Format `28.0.3` → `28.1.5`. No wire-format change. Three
registry entries are added: a reserved no-op at `28.0.2` preserving
append-only history, column-family creation at `28.1.3`, then
`drop_header_root_auth_frontier` (`28.1.4`) and
`unauthenticated_commitment_roots` (`28.1.5`).

**Operator impact — read `docs/header-chain-v1.4-migration.md` before
deploying.** A database containing predecessor header-overlay rows is
**rejected** at startup; those rows are not decoded, imported, or deleted.
Affected nodes need a fresh state database and a resync.
`network.zakura.header_sync.accept_new_blocks` is removed and config parsing
now rejects it.

## Consensus-adjacent changes

These are the parts that alter validation behaviour, and the ones worth the
closest reading:

- **256-bit cumulative work.** Best-chain work comparison uses full-width
  `U256` via `WorkCoordinate`/`SuffixWork`, removing a previously reachable
  overflow. `Chain::cmp` now scores through `ChainScore`; the tie-break
  (work, then raw tip-hash byte order) is unchanged. `Chain::push` computes
  the checked work sum *before* mutating `height_by_hash`.
- **Overflow-safe difficulty.** `AdjustedDifficulty` moves to
  `zakura-header-chain` and its constructors become fallible: the context must
  now contain exactly `min(candidate_height, 28)` predecessors, and a short
  context fails closed as `NotReadyToBeCommitted` instead of panicking in
  `mean_target_difficulty`'s slice. `threshold_bits` avoids a `U256`
  multiplication overflow via `scaled_mean > target_limit / bounded_timespan
  ⇒ target_limit`, which is exactly equivalent to the previous
  `min(limit, scaled_mean * b)`. `mean_target_difficulty` sums per-item
  quotients plus a remainder correction, which is exact.
- **Testnet max-block-time activation** becomes a network parameter.
  `ParametersBuilder::finish` now defaults to the public Testnet soft-fork
  height instead of `Height(2)`; Regtest sets `Height(2)` explicitly.
- **VCT roots are authenticated at header time** (#617). The ahead-of-body
  fast path that persisted peer-supplied roots before authentication is gone.
  The only remaining unauthenticated writer is gated
  `#[cfg(any(test, feature = "proptest-impl"))]`; every production row comes
  from the body-commit path. The `28.1.5` migration removes rows any existing
  database persisted through the old path.

## Suggested review order

1. `04ed5f077` `feat(header-chain)!: add fork-aware graph model` — the engine model
2. `dd5b79dd1` `refactor(node-services)!: add typed header-sync contracts` — the boundary types
3. `4fefdd266` `feat(state)!: persist and recover fork-aware header chains` — durability
4. `ebb60ab2d` `feat(network)!: add bounded fork-aware header sync` — the reactor
5. `c6dc0e016` `feat(state)!: authenticate VCT roots at header time (#617)`
6. the `fix(...)`/`refactor(...)` commits, which are audit-response deltas on the above

## Testing

Verified locally at `58477bcf7`:

- `cargo check --workspace --locked --all-features --all-targets`
- `cargo xtask header-conformance` — 114 rules validated, 0 unimplemented

No regtest, node, SSH, release, or deployment operation was run by hand; the
regtest e2e and testkit cluster jobs run in CI.

## Changelog

`docs/changelog/unreleased/586.md` carries the user-visible Added / Changed /
Removed / Fixed entries for this work.


evan-forbes:
for reviewers, fixed all the v12 stuff but likely worth running again

ci is broken but only due to pretty long actions outage https://www.githubstatus.com/

evan-forbes:
going through each of the issues above and then rebasing

evan-forbes:
## Fourth-audit consolidated disposition

This single follow-up maps all 14 findings from the [fourth-audit report](https://github.com/zakura-core/zakura/pull/586#issuecomment-5209288192) to their final disposition and owning review layer after the corrected stack was rebased onto `main` `db9fa5b8cbeeb221c00038229b43e3b317084d70`.

| Finding | Disposition | Owning PR(s) |
| --- | --- | --- |
| F-183209 | Already corrected: validation leases bind exact network, PoW policy, trust anchor, and predecessor context. | #586 engine checks; #563 state issuance |
| F-183210 | Already corrected at cumulative top: only the registered source/session/request/target/generation attempt can complete. | #569 completion gate/runtime integration |
| F-183211 | New defect corrected: full-state authority binds the complete event payload, and absent supplied headers receive full observable/contextual consensus validation. | #586 enforcement/regressions; #563 state authority/context |
| F-183212 | Already corrected and strengthened: operator actions require exact full-state authority, now bound to the complete payload. | #586 authority; #563 serialized adapter |
| F-183214 | New defect corrected: auxiliary delivery IDs are globally unique across the bounded retained set in admission and recovery. | #586 |
| F-183217 | New defect corrected: panicking cumulative-work addition is removed and production consumers use checked addition. | #586 boundary regression; #563 checked consumers/removal |
| F-183218 | New defect corrected: the 17-target custom mean uses exact quotient/remainder accumulation without U256 overflow. | #586 |
| F-183219 | New defect corrected: retention references are bounded before traversal and eviction uses one deterministic candidate ordering. | #586 |
| F-183220 | New defect corrected: auxiliary puts/deletes are finalized before generation derivation, so same-transition eviction has no phantom generation effect. | #586 |
| F-183223 | Valid generic limitation, production-safe: startup reconciles the verified projection from authoritative full state before first publication; no duplicate provenance table is added. | #563 |
| F-183225 | New defect corrected: reconsideration names the exact installed invalidation episode, preventing old evidence from removing a newer reason. | #586 event/removal semantics; #563 serialized episode lookup |
| F-183226 | New defect corrected: every fallible delta precondition is validated before commit; committed apply returns typed errors and writer failures propagate fatally. | #586 prevalidation/error; #563 writer propagation |
| F-183227 | Already corrected: recovery requires anchored contiguous finality history plus independent canonical root/pin evidence. | #586 recovery checks; #563 durable reconstruction |
| F-183228 | New defect corrected: persisted `Valid` flags no longer bypass complete header consensus validation during recovery. | #586 recovery validation/regression; #563 archived context/fixtures |

Summary: nine new defects are fixed, four findings were already covered at the cumulative top, and the one generic provenance limitation remains production-safe at the state startup boundary. All nine new defects retain red-to-green regressions. The final published owning heads are #586 `acd47ff34b0820c91dfe5ca1f764bcd37a0992f8`, #563 `5c8b89e00dc9078b0590a3610a18546a3b745dd0`, and #569 `504895cfd1cc87379161dffb2c3d9f738f001b0d`.

evan-forbes:
Profiling found that each transition copied about 500 KiB of checkpoint-pin data. The follow-up perf(headersync): reuse immutable transition data intentionally lives in #587, so this reviewed engine branch and its commits remain unchanged.

crates/zakura-header-chain/src/graph/overlay.rs
```diff
@@ -0,0 +1,933 @@
...
+    pub(crate) finalized: Option<Frontier>,
+    pub(crate) put_nodes: Vec<HeaderNode>,
+    pub(crate) delete_nodes: Vec<block::Hash>,
+    pub(crate) add_children: Vec<(block::Hash, block::Hash)>,
+    pub(crate) remove_children: Vec<(block::Hash, block::Hash)>,
+    pub(crate) add_eligible_tips: Vec<block::Hash>,
+    pub(crate) remove_eligible_tips: Vec<block::Hash>,
+}
+
+pub(crate) struct GraphOverlay<'a> {
```
p0mvn, line 100: Please consider doing a pass on human readability and abstractions.

For example, I would never consider searching for a graph in an `overlay.rs` file.

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1754 @@
...
+    fn view_ancestor(
+        &self,
+        descendant: block::Hash,
+        height: block::Height,
+    ) -> Result<Option<Frontier>, GraphError>;
+}
+
+pub(crate) trait HeaderGraphEdit: HeaderGraphView {
+    fn edit_node_mut(&mut self, hash: block::Hash) -> Result<&mut HeaderNode, GraphError>;
+    fn edit_insert(
```
p0mvn: Could you please do a pass on adding comments with the human audience as the goal?

For example, neither the trait nor the implementations specify what these edit variants are.

Can be a follow-up, but I would really appreciate it.

I find that with AI-generated code, our codebase becomes extremely hard to model and grok

And having such a baseline means the AI will only keep extending it via the same patterns

crates/zakura-header-chain/src/graph/overlay.rs
```diff
@@ -0,0 +1,1030 @@
...
+                    .values()
+                    .filter(|node| !self.deletes.contains(&node.hash)),
+            )
+    }
+
+    pub(crate) fn retained_hashes(&self) -> impl Iterator<Item = block::Hash> + '_ {
+        self.nodes().map(|node| node.hash)
+    }
+
+    pub(crate) fn hashes_at_height(&self, height: block::Height) -> Vec<block::Hash> {
```
p0mvn: I think we should have method specs for most methods in this file and add comments to separate code into logical groups.

p0mvn (commented):
I will keep flushing comments as I review in case any could be addressed in this PR.

Will leave a clear comment and Slack message once done.

So, far only requests on human readability

crates/zakura-header-chain/src/transition/planner.rs
```diff
@@ -0,0 +1,1753 @@
...
+        selected,
+        verified,
+        aux_changes,
+        finality_append,
+        retention,
+        request.event.fingerprint(),
+        cause,
+        invariant_pins(context),
+        context.config.limits,
+    )?;
```
p0mvn: Should we have this take a struct that encapsulates this many params?

p0mvn:
I notice the general pattern of 

```rust
  Ok(EngineTransition {
      plan,
      #[cfg(test)]
      projected,
  })
```

where we embed `#[cfg(test)]` for testing convinience.

Please consider trying to avoid this, as it hinders readability by keeping test-specific embeds in production code paths.

We should aim to instead abstract the system such that we can set up black-box-style testing

crates/zakura-chain/src/parameters/constants.rs
```diff
@@ -29,6 +29,17 @@ pub const SLOW_START_SHIFT: Height = Height(SLOW_START_INTERVAL.0 / 2);
 // TODO: change to HeightDiff
 pub const MAX_BLOCK_REORG_HEIGHT: u32 = 1000;
 
+/// The maximum number of non-finalized chain forks Zakura will track.
+/// When this limit is reached, we drop the chain with the lowest work.
+///
+/// When the network is under heavy transaction load, there are around 5 active forks in the last
+/// 100 blocks. (1 fork per 20 blocks.) When block propagation is efficient, there is around
+/// 1 fork per 300 blocks.
```
p0mvn: Curious, how was this measured?

crates/zakura-chain/src/parameters/constants.rs
```diff
@@ -29,6 +29,17 @@ pub const SLOW_START_SHIFT: Height = Height(SLOW_START_INTERVAL.0 / 2);
...
+///
+/// This limits non-finalized chain memory, in the worst case, to around:
+/// `10 forks * 1000 blocks * 2 MB per block = 20 GB`
+pub const MAX_NON_FINALIZED_CHAIN_FORKS: usize = 10;
```
p0mvn: This worst case, combined with block sync is already beyond the recommended specs.

Would it make sense to have a rule that if there is at least one for over 100 blocks, then we support a max of 5 forks?

crates/zakura-header-chain/src/graph/mod.rs
```diff
@@ -0,0 +1,1764 @@
...
+        children
+    }
+
+    /// Insert one admitted header after the graph retains its exact parent.
+    pub(crate) fn insert(
+        &mut self,
+        header: Arc<block::Header>,
+        block_work: Work,
+        validation: HeaderValidationState,
+        direct_reasons: impl IntoIterator<Item = EligibilityReason>,
```
p0mvn, line 491: When reading the code, it was unclear to me why we have these `direct_reasons` that represent failure modes in a data structure where we are storing candidate header chains.

A cleaner way: define an invariant that only validated headers get into the store and prevalidate before.

Alternatively, we should have a `HeaderCandidate` abstraction that wraps the header and the reasons having a `validate()` method. This way, we can cleanly check it inside the function while keeping the model clear.

I think this may be an example of a leaky abstraction we discussed yesterday.

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+            .is_eligible()
+        {
+            self.eligible_tips.remove(&parent_hash);
+            self.eligible_tips.insert(hash);
+        }
+        Ok(InsertResult::Inserted(Frontier::new(height, hash)))
+    }
+
+    /// Add one independent direct reason, then recompute the affected subtree cache.
+    pub(crate) fn add_reason(
```
p0mvn: Let me start by explaining how I've been attempting to review:

- Start from the lowest level of abstraction (i.e., the store)
- Go layer-by-layer higher

From reviewing `MemHeaderStore`, I cannot tell what the " reason" is and why we are adding it to the store.

I think such key parts of the architecture should be well-formulated. Otherwise, it will become untenable to model humanly over time.

I can, of course, ask AI what it is doing, and it makes sense afterward. But I think that, if we keep taking these shortcuts, the codebase will be impossible to make a change to without talking to AI for hours about various footguns that no human tracks.

The way I would change this is:
```rust
    /// Marks an already-retained header as newly ineligible without deleting it.
    /// Its descendants also become ineligible, so the method recomputes ancestry
    /// eligibility and eligible-tip indexes.
    pub(crate) fn add_eligibility_reason(
        &mut self,
        hash: block::Hash,
        reason: EligibilityReason,
    ) -> Result<bool, GraphError> {
   /// 
   } 
```

This way, I can understand what this is doing as a black box without reading the code or talking to AI, which saves meaningful cognitive overhead, unblocking more time to review faster.

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+            .direct_reasons
+            .insert(reason);
+        if changed {
+            self.recompute_descendant_eligibility(hash)?;
+        }
+        Ok(changed)
+    }
+
+    /// Remove exactly one operator invalidation, preserving every unrelated reason.
+    pub(crate) fn remove_operator_invalidation(
```
p0mvn: This is another example of a "leaky" abstraction:

The method is called only from tests and models a manual invalidation removal.

However, why does a low-level `MemHeaderStore` need to codify the awareness of some "operator".

At this layer, it should be a basic CRUD API (add, delete, update...) without higher-level context.

crates/zakura-header-chain/src/graph/mod.rs
```diff
@@ -0,0 +1,1764 @@
...
+            hashes.remove(&hash);
+            if hashes.is_empty() {
+                self.heights.remove(&height);
+            }
+        }
+        self.refresh_eligible_tip(parent_hash);
+        Ok(())
+    }
+
+    fn recompute_descendant_eligibility(&mut self, root: block::Hash) -> Result<(), GraphError> {
```
p0mvn, line 1240: It's probably fine since we bound the number of chains and depth but there are several loops in this function that look like they could save CPU and be a single loop

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+        });
+        let changed = reasons.len() != before;
+        if changed {
+            self.recompute_descendant_eligibility(hash)?;
+        }
+        Ok(changed)
+    }
+
+    /// Atomically record one commitment-matching deterministic body failure.
+    pub(crate) fn set_consensus_body_invalid(
```
p0mvn: This marks a header as invalid due to a body being invalid.

`add_reason` is another API that serves a similar function.

Why can't we have a single API that achieves this function?

It seems that we are defining too many specific behaviors on an API that should be broad.

p0mvn: AI seems to agree:

The real issue is that one fact—body invalidity—is stored twice:

- in BodyValidationState,
- in EligibilityReason::ConsensusBodyInvalid.
That duplication creates BodyEligibilityMismatch and forces a special setter.

A generic add_reason() cannot safely handle it because it updates only one copy. Better designs would be:

1. Derive eligibility from body state
Remove ConsensusBodyInvalid from direct_reasons. Let is_eligible() check whether body is consensus-invalid. This removes the duplicated fact and special consistency checks.

3. Use one typed transition API
For example, apply_eligibility_change(hash, EligibilityChange), with variants for body invalidation, operator invalidation, and reconsideration. Internally, each variant still performs its required coupled updates.

I prefer option 1: direct_reasons should contain independent exclusion facts, while body invalidity should remain authoritative in BodyValidationState. Eligibility can be derived from both. That reduces stored state and makes contradictory states unrepresentable.

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+        node.body = body;
+        node.eligibility.direct_reasons.insert(reason);
+        if changed {
+            self.recompute_descendant_eligibility(hash)?;
+        }
+        Ok(changed)
+    }
+
+    /// Update body availability or verification without changing fork choice eligibility.
+    pub(crate) fn set_body_state(
```
p0mvn: I think that `set_consensus_body_invalid` and `set_body_state` could be a single method `set_body_state`

That takes a struct, encapsulating valid and invalid states. Unwrap internally and be processed correctly.

This would reduce the API complexity on the `MemHeaderStore`

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+        &mut self,
+        hash: block::Hash,
+        body: BodyValidationState,
```
p0mvn: The var name `body` representing `body_validation_state` is semantically insufficient, making it difficult to understand what is happening downstream in the function

crates/zakura-header-chain/src/graph/mod.rs
```diff
@@ -0,0 +1,1764 @@
...
+                pending.extend(self.children(hash));
+            }
+        }
+        let mut deleted: Vec<_> = self
+            .nodes
+            .keys()
+            .copied()
+            .filter(|hash| !retained.contains(hash))
+            .collect();
+        deleted.sort_unstable_by_key(|hash| hash.0);
```
p0mvn, line 1171: Feels wrong that we need to traverse all nodes again just so that we can disconnect a branch that got obsoleted by a frontier move.

Not sure if there is a deeper reason why this could not be a linked list we disconnect.

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+    fn view_ancestor(
+        &self,
+        descendant: block::Hash,
+        height: block::Height,
+    ) -> Result<Option<Frontier>, GraphError> {
+        self.ancestor(descendant, height)
+    }
+}
+
+impl HeaderGraphEdit for GraphOverlay<'_> {
```
p0mvn: It is unclear to me why this is here and not in overlay.rs. I think it should be there.

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+            .values()
+            .filter(|node| node.is_eligible() && !self.has_eligible_child(node.hash))
+            .map(|node| node.hash)
+            .collect();
+    }
+
+    pub(crate) fn apply_delta(&mut self, delta: &GraphDelta) -> Result<(), GraphError> {
+        self.validate_delta(delta)?;
+
+        for hash in &delta.delete_nodes {
```
p0mvn: I think this function has a lot of duplication with others.

For example, there is also `remove_leaf()` that applies deletions. We could probably develop a shared helper.

It also does not seem right to me that `GraphDelta` encapsulates parent and child deletions separately,

If the parent is removed, the child must be removed as well. So, it feels more like we need a single hash set of node deletions

crates/zakura-header-chain/src/graph.rs
```diff
@@ -0,0 +1,1764 @@
...
+                children.remove(child);
+                if children.is_empty() {
+                    self.children.remove(parent);
+                }
+            }
+        }
+        for (parent, child) in &delta.add_children {
+            self.children.entry(*parent).or_default().insert(*child);
+        }
+        for hash in &delta.remove_eligible_tips {
```
p0mvn: Since we also apply eligible tip removals separately, I am worried about data inconsistency: remove the node but forget the tip, or vice versa.

p0mvn (commented):
Flushing review of `graph.rs`

Please consider merging this PR with comments and rename improvements (no logic changes)
https://github.com/zakura-core/zakura/pull/610

crates/zakura-header-chain/src/graph/overlay.rs
```diff
@@ -0,0 +1,1030 @@
...
+    add_children: HashMap<block::Hash, HashSet<block::Hash>>,
+    remove_children: HashMap<block::Hash, HashSet<block::Hash>>,
+    eligible_tips: HashSet<block::Hash>,
+    node_count: usize,
+    #[cfg(test)]
+    base_nodes_cloned: Cell<usize>,
+    #[cfg(test)]
+    eligibility_nodes_visited: Cell<usize>,
+    #[cfg(test)]
+    finality_nodes_visited: Cell<usize>,
```
p0mvn: Understood that this is for testing but not having these test-specific embeds (e.g. `#[cfg(test)]`) on the core logic would be a major step for readability

crates/zakura-header-chain/src/graph/overlay.rs
```diff
@@ -0,0 +1,1030 @@
...
+    pub(crate) fn node(&self, hash: block::Hash) -> Option<&HeaderNode> {
+        if self.deletes.contains(&hash) {
+            return None;
+        }
+        self.puts.get(&hash).or_else(|| self.base.nodes.get(&hash))
+    }
+
+    pub(crate) fn node_mut(&mut self, hash: block::Hash) -> Result<&mut HeaderNode, GraphError> {
+        if self.deletes.contains(&hash) {
+            return Err(GraphError::UnknownNode(hash));
```
p0mvn: Is `UnknownNode` the right error for a hash that's pending deletion?

crates/zakura-header-chain/src/graph/overlay.rs
```diff
@@ -0,0 +1,1030 @@
...
+        }
+        if !self.puts.contains_key(&hash) {
+            let node = self
+                .base
+                .nodes
+                .get(&hash)
+                .cloned()
+                .ok_or(GraphError::UnknownNode(hash))?;
```
p0mvn: Here, we are using the same error to represent a different state which may become confusing during debugging

p0mvn (commented):
Flushing `overlay.rs`

Went faster this cycle because APIs are similar to `graph.rs`, just the implementation details differ.

crates/zakura-header-chain/src/transition/engine.rs
```diff
@@ -0,0 +1,342 @@
...
+    if let Some(height) = delta.remove_before {
+        projection.retain(|frontier| frontier.height >= height);
+    }
+    if let Some(height) = delta.remove_from {
+        projection.retain(|frontier| frontier.height < height);
+    }
+    projection.extend(delta.put.iter().copied());
+}
+
+fn apply_aux_delta(aux: &mut HashMap<block::Hash, Vec<AuxDelivery>>, changes: &[AuxDelta]) {
```
p0mvn: This and other helpers in this file may benefit from unit tests

crates/zakura-header-chain/src/transition/types.rs
```diff
@@ -0,0 +1,1701 @@
...
+    /// New or replaced nodes.
+    pub put_nodes: Vec<HeaderNode>,
+    /// Evicted or finalized-away nodes.
+    pub delete_nodes: Vec<block::Hash>,
+    /// Reconstructible indexes changed with the nodes.
+    pub index_changes: IndexChanges,
+    /// Selected-header height projection change.
+    pub selected_projection: ProjectionDelta,
+    /// Full-state verified height projection change.
+    pub verified_projection: ProjectionDelta,
```
p0mvn: Something that comes to mind is memory backpressure. We have so many duplications of header chains in various flavors

crates/zakura-header-chain/src/transition/engine.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// The exact preserved migration pin, if durable finality history contains the requested pin.
+    MigratedFinalityPin(Option<Frontier>),
+}
+
+/// One coherent in-memory owner for transition planning state.
+#[derive(Clone, Debug)]
+pub struct HeaderChainEngine {
+    graph: MemHeaderStore,
+    metadata: EngineMetadata,
+    selected: Vec<Frontier>,
```
p0mvn: would it make sense to all it `best_header_chain`?

crates/zakura-header-chain/src/transition/engine.rs
```diff
@@ -0,0 +1,342 @@
...
+    require_verified_bodies: bool,
+) -> Result<(), EngineHydrationError> {
+    if projection.first().copied() != Some(graph.finalized())
+        || projection.last().copied() != Some(tip)
+    {
+        return Err(EngineHydrationError::Incoherent(
+            "projection endpoints disagree with metadata",
+        ));
+    }
+    for frontier in projection {
```
p0mvn: I find it odd that each entry in a projection is called a "frontier"

p0mvn (commented):
Flushing engine.rs

crates/zakura-header-chain/src/graph/header_node.rs
```diff
@@ -0,0 +1,343 @@
+//! Admitted header records and eligibility state.
```
p0mvn, line 1: The name of this file should be either `header_node.rs` or `header_node_types.rs`.

`node.rs` is too abstract for what this contains

crates/zakura-header-chain/src/validation/prepare.rs
```diff
@@ -0,0 +1,631 @@
...
+    rules: &HeaderRules,
+    clock: &dyn Clock,
+) -> Result<PreparedHeaderBatch, HeaderFailure> {
+    if !lease.is_coherent(&rules.network, rules.trust_anchor_digest) {
+        return Err(HeaderFailure::InvalidLease);
+    }
+    prepare_headers_inner(input, lease.parent, Some(lease), rules, clock)
+}
+
+fn prepare_headers_inner(
```
p0mvn: It would be valuable to clearly document which validations this performs and what invariants can be expected after this is complete

p0mvn (commented):
Completed boilerplate: `frontier.rs`, `node.rs`, `transition/types.rs`, `config.rs`, `validation/prepare.rs `, `validation/contextual.rs`, `ids.rs`

Next steps:
- `retention.rs`, `transition/invariants.rs`, `transition/planner.rs`

evan-forbes:
Removed `EngineTransition.projected` and `into_projected_engine`. Tests now clone the source engine and call `apply_committed`. I also added black-box production-verifier coverage in `032ab571bfe3ca21a5ee986f85e361510ae519ec test(header-chain): cover production transition application`.

crates/zakura-header-chain/src/graph/overlay.rs
evan-forbes: expanded the staged-deletion semantics in `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`, will continue on this PR and the other stacks tonight/tmrw.

crates/zakura-header-chain/src/graph.rs
evan-forbes: alongside #610 the comments now describe committed and staged graph views, index maintenance, and the caller’s authority and invariant responsibilities.

crates/zakura-header-chain/src/graph/overlay.rs
evan-forbes: alongside #610 I added the missing staged-deletion and DAG-finality semantics in `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`.

crates/zakura-header-chain/src/transition/planner.rs
evan-forbes: Added `DerivePlanInputs` and changed `derive_plan` to accept it in `5abd463a4764f78eb49f8a059d001c443443bc87 refactor(header-chain): clarify graph transition interfaces`.

crates/zakura-chain/src/parameters/constants.rs
evan-forbes: These were originally from zebra but removed the frequency and memory claims in `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`.

crates/zakura-chain/src/parameters/constants.rs
evan-forbes: documented that the header engine inherits the existing full-state fork cap in `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`, changing these params imo should be separate unless we have a good reason to change specifically for headersync.

crates/zakura-header-chain/src/graph/mod.rs
evan-forbes: clarified that `MemHeaderStore` is the domain graph for admitted headers and that reasons are selection exclusions. Body invalidity is no longer duplicated as a reason. These changes are in `3abc5a23b86e6bb187e12bfa39a9752ece7b3f28 refactor(header-chain): derive body invalidity from body state` and `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`.

crates/zakura-header-chain/src/graph.rs
evan-forbes: Renamed the operation and trait method to `add_eligibility_reason`. I also preserved the black-box contract for descendant propagation and eligible-tip refresh in `5abd463a4764f78eb49f8a059d001c443443bc87 refactor(header-chain): clarify graph transition interfaces`.

crates/zakura-header-chain/src/graph.rs
evan-forbes: generic reason removal would allow removal of permanent exclusions. `remove_operator_invalidation` can remove only the exact reversible operator reason and evidence.

crates/zakura-header-chain/src/graph/mod.rs
evan-forbes: Removed the redundant parent-collection pass in `5abd463a4764f78eb49f8a059d001c443443bc87 refactor(header-chain): clarify graph transition interfaces`. I kept propagation and tip refresh as separate phases so every tip check observes the final inherited state.

crates/zakura-header-chain/src/graph.rs
evan-forbes: Removed `EligibilityReason::ConsensusBodyInvalid`. `HeaderNode::is_eligible` now derives exclusion directly from `BodyValidationState::ConsensusInvalid` in `3abc5a23b86e6bb187e12bfa39a9752ece7b3f28 refactor(header-chain): derive body invalidity from body state`.

evan-forbes: Implemented option 1 in `3abc5a23b86e6bb187e12bfa39a9752ece7b3f28 refactor(header-chain): derive body invalidity from body state`. Body state is authoritative, invalid evidence is permanent, identical evidence is idempotent, and conflicting or restoring updates fail closed.

crates/zakura-header-chain/src/graph.rs
evan-forbes: Replaced the specialized consensus-invalid setter with one `set_body_state` operation in `3abc5a23b86e6bb187e12bfa39a9752ece7b3f28 refactor(header-chain): derive body invalidity from body state`.

crates/zakura-header-chain/src/graph.rs
evan-forbes: alongside #610 the unified setter now names the value `body_validation_state` in `3abc5a23b86e6bb187e12bfa39a9752ece7b3f28 refactor(header-chain): derive body invalidity from body state`.

crates/zakura-header-chain/src/graph/mod.rs
evan-forbes: This cannot be a single linked-list disconnect because each path ancestor can have competing child subtrees. I documented the exact traversal and its retained-subtree bound in `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`.

crates/zakura-header-chain/src/graph.rs
evan-forbes: Moved the `GraphOverlay` trait implementations into `graph/overlay.rs` in `5abd463a4764f78eb49f8a059d001c443443bc87 refactor(header-chain): clarify graph transition interfaces`.

crates/zakura-header-chain/src/graph.rs
evan-forbes: Reduced `GraphDelta` to canonical finality, node puts, and node deletes. Both overlay reconstruction and committed application derive their indexes through one validated routine in `d43a3ecf68e3654d2033adf004ecab78db4890c8 refactor(header-chain): derive graph indexes from node deltas`.

crates/zakura-header-chain/src/graph.rs
evan-forbes: Eligible-tip and child-edge changes are now derived from canonical node deltas. Added regression coverage for stale-index prevention and pre-mutation rejection in `d43a3ecf68e3654d2033adf004ecab78db4890c8 refactor(header-chain): derive graph indexes from node deltas`.

crates/zakura-header-chain/src/graph/overlay.rs
evan-forbes: Consolidated the overlay counters behind `OverlayTestStatistics` in `032ab571bfe3ca21a5ee986f85e361510ae519ec test(header-chain): cover production transition application`. I retained the narrow counters because their tests prove clone and traversal bounds.

crates/zakura-header-chain/src/graph/overlay.rs
evan-forbes: A pending deletion means the node is absent from the projected overlay, originally the loose reasoning was to keep the graph callers agnostic of the source of the absence. It would basically leak the overlay into the callers of the graph. not the end of the world so happy to change still

crates/zakura-header-chain/src/graph/overlay.rs
evan-forbes: afaiu from the caller's point of view the state is the same, which is that node is not in the graph. differentiating means leaking overlay staging details, but that's not the end of the world so happy to change still.

crates/zakura-header-chain/src/transition/engine.rs
evan-forbes: Added direct unit tests for projection validation, projection-delta application, and auxiliary upsert and deletion behavior in `032ab571bfe3ca21a5ee986f85e361510ae519ec test(header-chain): cover production transition application`.

crates/zakura-header-chain/src/transition/types.rs
evan-forbes: yeah makes sense. definitely something to keep an eye on. I've kept some profiles on the smoke tests and didn't notic anything too egrigious at least in those. Ported the immutable-data changes in `a30662334af82ae3b42670122ed2ffa1b5775871 perf(header-chain): reuse immutable transition projections`. complete projections remain borrowed and share storage.

crates/zakura-header-chain/src/transition/engine.rs
evan-forbes: Renamed the private fields to `selected_projection`, `verified_projection`, and `aux_deliveries` in `5abd463a4764f78eb49f8a059d001c443443bc87 refactor(header-chain): clarify graph transition interfaces`. I used projection-specific names because these vectors contain height/hash projection entries rather than header objects.

crates/zakura-header-chain/src/transition/engine.rs
evan-forbes: Kept `Frontier` and documented the reason in `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`. Each projection entry is the frontier of the selected or verified chain prefix at that height.

crates/zakura-header-chain/src/graph/header_node.rs
evan-forbes: Renamed `node.rs` to `header_node.rs` and updated the module wiring in `5abd463a4764f78eb49f8a059d001c443443bc87 refactor(header-chain): clarify graph transition interfaces`.

crates/zakura-header-chain/src/validation/prepare.rs
evan-forbes: Documented the complete context-free check list, successful batch guarantees, and remaining planner checks in `094b5ff8fbb8f06aa1f6ccb5ac6d3c1a4766d470 docs(header-chain): document engine review contracts`.

evan-forbes:
ahh fack just realized I rebased and lost track of commits to feedback

let me see if I can reconstruct the old commits, then we can just squash later

crates/zakura-node-services/src/header_chain.rs
```diff
@@ -0,0 +1,693 @@
+//! Typed boundary between header-sync policy and header-chain state.
```
p0mvn, line 1: Do we want another crate for this? Could it exist in another crate?

p0mvn (approved):
After a few days of review, I'm  becoming ready to approve. Great work @evan-forbes 

Tracking 3 follow-ups from the recent pass:

- Improving header chain readability
- Restoring sync speed
- Restoring peer scoring on vct mmr verification failure

We also have open:
- Several of Evan's PRs / v12 follow-ups
- Daybreak AI report follow-ups
- Iterations on merging in design docs 


I've attempted splitting consensus-critical paths into their own PR. Reviewing them individually has given me confidence to merge as-is instead of getting blocked on separate PRs.

PR #593, opened 2026-08-06

p0mvn:
## Motivation

A fast-synced node records no completed subtree roots below its last checkpoint — they are a by-product of the per-height tree maintenance the fast path skips, so they cannot be recorded during fast sync at all. See §16 of `docs/design/verified-commitment-trees.md`.

## Solution

Publish them as a small artifact and serve them, with the node's own rows always taking precedence. Mainnet roots ship embedded beside the checkpoint frontier and are authenticated by `mainnet-vct-manifest.json`.

- **Published records never displace the node's own rows.** A correct artifact holds only subtrees completed below the last checkpoint and so never overlaps — which makes an overlap exactly the corrupt-or-hostile case where the node's own verified value must win. The artifact's digest is self-contained, not a signature.

### Sync-independent generation

`export-historical-treestates` is bound to the network last checkpoint `H`, not to how the source database was synced:

- **Legacy direct export.** When Sapling, Orchard, and Ironwood frontiers are present at `H - 1`, each pool's expected completed-subtree count is `frontier.count() >> TRACKED_SUBTREE_HEIGHT`. Stored `{pool}_note_commitment_subtree` rows must be exactly the contiguous indexes `0..expected` with `end_height < H`. No VCT marker is required. Direct exports report zero replayed blocks.
- **VCT replay fallback.** When those frontiers are absent, generation requires a matching VCT absent-band marker. Replay is authenticated against `commitment_roots_by_height`. For upgrade height `U > 0`, validated stored pre-`U` rows are prepended and replay seeds from the `U - 1` frontiers; `U == 0` starts from empty frontiers.
- Fail closed on tip below `H - 1`, incomplete/gapped stored rows, mixed frontiers, mismatched VCT last-checkpoint marker, and replay failures.

## Testing

Exporter unit tests cover markerless legacy success, checkpoint binding / exclusion of a subtree completing at `H`, gapped rows and tip-below-checkpoint rejection, VCT marker mismatch, and replay-path selection for `U == 0` and anchored `U > 0`.

**Verified against ground truth**: above a fast-synced node's last checkpoint the database *does* store subtree rows, so replaying `(3,358,006, 3,432,538]` and comparing gives **5 matched, 0 mismatched**.


p0mvn:
@v12this review

p0mvn:
I've addressed v12 that is worthwhile

p0mvn:
```
roman@Romans-MacBook-Pro-2 zakura % ssh root@162.243.185.175 "curl -sS -H 'content-type: application/json' --data '{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"z_getsubtreesbyindex\",\"params\":[\"sapling\",0,3]}' http://127.0.0.1:8232 | jq"
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "pool": "sapling",
    "start_index": 0,
    "subtrees": [
      {
        "root": "754bb593ea42d231a7ddf367640f09bbf59dc00f2c1d2003cc340e0c016b5b13",
        "end_height": 558822
      },
      {
        "root": "03654c3eacbb9b93e122cf6d77b606eae29610f4f38a477985368197fd68e02d",
        "end_height": 670209
      },
      {
        "root": "e2bf698f5ac10b44da560d11a5e1d5c191a82a968a2be0a6948aa8748b545160",
        "end_height": 780364
      }
    ]
  }
}
```

p0mvn:
The artifact got outdated, resulting in the CI break. Updating now.

crates/zakura-state/src/service.rs
```diff
@@ -1232,6 +1243,17 @@ impl ReadStateService {
...
 
+    /// Returns the embedded subtree artifact only when it describes this database's fast-sync
+    /// last checkpoint.
+    ///
+    /// This check is made when serving rather than at construction because the durable last
+    /// checkpoint marker can be written after the read service starts.
+    fn historical_subtrees_at_last_checkpoint(&self) -> Option<&finalized_state::SubtreeArtifact> {
+        let artifact = self.historical_subtrees.as_deref()?;
+
+        (self.db.vct_synced_below() == Some(artifact.last_checkpoint)).then_some(artifact)
```
czarcas7ic: Could this also allow a newer artifact for an older database marker?

If a database fast-synced at H1 and is already past H2 when it upgrades, its marker remains H1. The H2 artifact still contains the missing H1 roots, but this exact equality disables it.

Could an H1 database use a newer H2 artifact only to fill the history it skipped at H1, while continuing to use its local rows after H1? Then we could add a H1 database → H2 artifact regression test.

crates/zakura-state/src/service/read/tree.rs
```diff
@@ -77,6 +77,61 @@ fn resolve_historical_tree<Tree: Default>(
...
+/// reaches a client. The tip bound prevents the artifact from granting authority over blocks this
+/// node has not verified yet.
+pub fn merge_published_subtrees<Node>(
+    stored: &mut BTreeMap<NoteCommitmentSubtreeIndex, NoteCommitmentSubtreeData<Node>>,
+    published: impl IntoIterator<Item = (NoteCommitmentSubtreeIndex, NoteCommitmentSubtreeData<Node>)>,
+    verified_tip: block::Height,
+) {
+    for (index, data) in published
+        .into_iter()
+        .filter(|(_, data)| data.end_height <= verified_tip)
```
czarcas7ic: Noting that this returns a bad error in the following case. Imagine:

```
Artifact contains:
subtree 0 → block 100
subtree 1 → block 200

Syncing node is at block 150.
```

If someone were to query this node for both subtrees, instead of either getting just subtree 0 back or an error saying its still syncing, it instead gets back a NotStored error:

```
continuing synchronization will not restore it; use another node
```

Which might cause confusion since continuing to sync will in fact "restore" it.

czarcas7ic (commented):
Likely tracking already, but for the sake of being explicit, this PR makes `mainnet-treestate-subtrees.bin` a required bundle key, so after merge we need to redeploy the publisher and publish one bundle containing it before `update-release-state.yml` or `prepare-release-pr.yml` runs (both hard-fail on the missing key).

https://github.com/zakura-core/zakura/blob/d57f932bb3f3359ad1e9efa49f6fd74abd3ab573/.github/scripts/fetch-release-state.py#L237

crates/zakura-state/src/service.rs
p0mvn: [0961d8e](https://github.com/zakura-core/zakura/pull/593/commits/0961d8e88a2bbe96d4d8dc86cf122ab9ee0bf55a)

crates/zakura-state/src/service/read/tree.rs
p0mvn: Addressed such that, in the example, it would return subtree 0. This seems to be consistent with the prior API requirements

czarcas7ic (approved):

p0mvn:
Merged, then redeployed the snapshot-host publisher at `1dcbb0682` and published a three-file bundle so `update-release-state.yml` / `prepare-release-pr.yml` will not fail closed on the new required key.

- Publisher revision: `1dcbb0682187f02390f2b4a6b364ab154975c665`
- Pointer: height `3449371` at https://zakura-release.valargroup.dev/release-state/latest.json
- `meta.files` now includes `mainnet-treestate-subtrees.bin` (72107 bytes). Checkpoint list and frontier digests are unchanged from the previous two-file bundle at this height.

Same-height republish was required: the pruned exporter tip was still below the next 400-block checkpoint, and the Monday `update-release-state` cron would have fetched the old two-file pointer.

PR #592, opened 2026-08-06
fix(network): respecting stream frame cap

distractedm1nd:
## Motivation

Was playing around with registering custom services and ran into this. Currently, `Stream.frame_cap` is documented as the stream's "Maximum application frame bytes", but it gets discarded. This was acknowledged in the code.

## Solution

- Added the per-stream frame caps to `StreamWorkerContext`
- Passes around either `Stream` or the cap directly now instead of just stream_kind.

Having the field already exist as an "advisory field" made this diff a lot easier, nice foresight :)

## Testing

The diff touches a lot of existing tests from the signature changes of a few helpers. It replaces also `stream_specific_application_frame_caps_keep_gossip_and_discovery_tight` with `declared_stream_frame_caps_are_authoritative`.

## Changelog

~~I dont see a docs/changelog/unreleased/<PR-number>.md so going to ignore this section? Or is it just located elsewhere~~

 I was blind, updated

PR #594, opened 2026-08-07
feat(zakurad): support embedded nodes and custom p2p services

distractedm1nd:
## Motivation

Related to my #592, but I didn't stack cause I don't know your guys' preference

I wanted to be able to add custom p2p services and the abstractions kinda look like this should be allowed. Originally this PR only passed a vec through `zakura-network` but that didn't do enough to be able to actually use it in a running node. 

## Solution

Expose a helper that essentially runs the abscissa command and lets you pass a config + custom services.

`CustomService` is a new struct that just encapsulates an arced `Service` and a list of discovery IDs. Had to add this since discovery info is a bit disjoint from the service definition.

The embedding callstack is not my favorite flow, but I think it may be the best we can do with abscissa without introducing too much bloat or slop abstractions

## Testing

adds a pretty self explanatory `custom_service_is_registered_and_advertised` and tested manually in a local network that is serving spendability-pir stubs

also added two other slop tests that ensure that:
1.  an embedded node /actually/ starts a node and shuts it down with its cancellation token, 
2. and that a passed custom service can actually be reached between 2 nodes that speak it

## Changelog

```
## Added

- Applications can now embed `zakurad` to register custom Zakura p2p services and advertise them through discovery
  ([#594](https://github.com/zakura-core/zakura/pull/594)).
```

### DIFF
--- a/crates/zakurad/src/commands/start.rs
+++ b/crates/zakurad/src/commands/start.rs
@@ -487,7 +487,7 @@ impl StartCmd {
         let advertised_services = Self::advertised_services(&config);
 
         let (peer_set, address_book, misbehavior_sender, zakura_endpoint) =
-            zakura_network::init_with_zakura_custom_services(
+            zakura_network::init_with_zakura(
                 config.network.clone(),
                 inbound,
                 latest_chain_tip.clone(),
--- a/crates/zakurad/src/commands/start.rs
+++ b/crates/zakurad/src/commands/start.rs
@@ -87,6 +87,7 @@ use futures::FutureExt;
 use tokio::{
     pin, select,
     sync::{oneshot, watch},
+    task::{AbortHandle, JoinHandle},
 };
 use tokio_util::sync::CancellationToken;
 use tower::{builder::ServiceBuilder, util::BoxService, ServiceExt};
@@ -172,6 +173,22 @@ fn check_tcp_slow_start_after_idle() {
 
 fn use_zakura_block_sync(config: &zakura_network::Config) -> bool {
     config.v2_p2p()
+}
+
+#[derive(Default)]
+// Keeps track of `JoinHandle`s, and when `Drop`ped, `aborts` all of them.
+struct NodeTasks(Vec<AbortHandle>);
+
+impl NodeTasks {
+    fn track<T>(&mut self, task: &JoinHandle<T>) {
+        self.0.push(task.abort_handle());
+    }
+}
+
+impl Drop for NodeTasks {
+    fn drop(&mut self) {
+        self.0.iter().for_each(AbortHandle::abort);
+    }
 }
 
 #[cfg(not(target_os = "linux"))]
@@ -288,8 +305,10 @@ impl StartCmd {
         config: Arc<ZakuradConfig>,
         custom_services: Vec<zakura_network::zakura::CustomService>,
         shutdown: CancellationToken,
-        cleanup_ready: CancellationToken,
+        shutdown_cleanup_required: CancellationToken,
     ) -> Result<(), Report> {
+        let _shutdown_on_drop = shutdown.clone().drop_guard();
+        let mut node_tasks = NodeTasks::default();
         check_tcp_slow_start_after_idle();
 
         let is_regtest = config.network.network.is_regtest();
@@ -501,6 +520,20 @@ impl StartCmd {
             .await
             .map_err(|error| eyre!(error))?;
 
+        // Not added to node_tasks, because it must outlive start() being dropped to shutdown the
+        // endpoint
+        let zakura_endpoint_shutdown_task = if let Some(endpoint) = zakura_endpoint.clone() {
+            shutdown_cleanup_required.cancel();
+
+            let shutdown = shutdown.clone();
+            Some(tokio::spawn(async move {
+                shutdown.cancelled().await;
+                endpoint.shutdown().await;
+            }))
+        } else {
+            None
+        };
+
         // Start health server if configured (after sync_status is available)
 
         info!("initializing verifiers");
@@ -513,6 +546,7 @@ impl StartCmd {
                 tx_verifier_setup_rx,
             )
             .await;
+        node_tasks.track(&consensus_task_handles.state_checkpoint_verify_handle);
 
         if let Some(endpoint) = zakura_endpoint.clone() {
             let trace = endpoint.trace();
@@ -617,6 +651,7 @@ impl StartCmd {
             LAST_WARN_ERROR_LOG_SENDER.subscribe(),
             Some(submit_block_channel.sender()),
         );
+        node_tasks.track(&rpc_tx_queue_handle);
         let rpc_impl = rpc_impl.with_end_of_support_height(
             sync::end_of_support::end_of_support_height(&config.network.network),
         );
@@ -628,6 +663,7 @@ impl StartCmd {
         } else {
             tokio::spawn(std::future::pending().in_current_span())
         };
+        node_tasks.track(&rpc_task_handle);
 
         let zcashd_compat_shutdown_timeout =
             Self::zcashd_compat_supervisor_shutdown_timeout(&config);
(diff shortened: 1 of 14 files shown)


evan-forbes:
will get to these tmrw @distractedm1nd thanks for opening

distractedm1nd:
~~marking as not ready yet as i'm still playing around with different ways to embed the node until I find one I like. feel like im pretty close but need more testing~~

found something acceptable that agents couldnt figure out on their own for some reason. tests are vibecoded and a tiny sloppy but I couldn't do better myself.

crates/zakurad/src/commands/start.rs
```diff
@@ -1015,6 +1018,7 @@ impl StartCmd {
         };
 
         info!("exiting Zakura: asking other tasks to stop");
+        zakura_chain::shutdown::set_shutting_down();
```
evan-forbes, line 1050: idk if we have a reset for this we could add later after shutting down, but for testing purposes where we shutdown a node the start it back up again in the same process if we don't reset thing I don't think we can boot back up due to using a global var under the hood.

not sure yet if that's a feature or a bug

evan-forbes (commented):
apologies on the delay!

super dope feature 🔥 

have some very minor things after digging around w/ ai:

we should probably fix these three. I ran it a few times to verify but didn't go through and replicate myself so could be wrong but seems right

```
 2. Meduim: startup cancellation can return success without shutting down the node.

  run_until_shutdown() returns Ok(()) and drops the node future when cancellation arrives before cleanup_ready is cancelled at crates/zakurad/src/components/tokio.rs:114. The code explicitly tests this
  behavior at crates/zakurad/src/components/tokio.rs:175.

  start() creates the Zakura endpoint at crates/zakurad/src/commands/start.rs:462. It does not cancel cleanup_ready until crates/zakurad/src/commands/start.rs:868.

  After setup_tx.send() at line 623, the endpoint-owned inbound service retains state handles. Endpoint background tasks also retain endpoint clones. Those tasks stop only through
  ZakuraEndpoint::shutdown() at crates/zakura-network/src/zakura/handler.rs:816. Dropping the startup future can therefore leave the endpoint, network tasks, and database alive while run() returns
  Ok(()).

  Exactly three post-endpoint ? paths bypass normal endpoint shutdown:

  - crates/zakurad/src/commands/start.rs:625
  - crates/zakurad/src/commands/start.rs:704
  - crates/zakurad/src/commands/start.rs:804

  The same problem occurs if an embedder aborts or drops the run_with_services() future.

  3. Medium: the embedded-node test contacts real mainnet peers and misses the advertised feature.

  The test uses mainnet defaults and clears only initial_mainnet_peers at crates/zakurad/src/node.rs:63. The default cache remains enabled. initial_peers() loads cached peers at crates/zakura-network/
  src/config.rs:417. This checkout has a populated ~/.cache/zakura/network/mainnet.peers, so the test can handshake public peers on port 8233.

  CacheDir::disabled() fixes that test isolation defect.

  The test also selects P2pStack::Legacy and passes no custom services at crates/zakurad/src/node.rs:65. The round-trip custom-service test calls spawn_zakura_endpoint_with_services() directly at crates/
  zakura-network/src/zakura/handler.rs:6457. No test passes a custom service through zakurad::node::run_with_services().

  6. Medium: advertisement does not drive peer selection.

  The discovery dialer requests dial_candidates(&[], …) at crates/zakura-network/src/zakura/discovery/candidate_dialer.rs:199. It never supplies custom service IDs.

  The discovery handle implements service_candidates() at crates/zakura-network/src/zakura/discovery/protocol.rs:2081, but CustomService receives neither that handle nor another selection hook. A custom
  service therefore cannot request peers that advertise its ID.

  The custom service receives peers chosen through general discovery, bootstrap dialing, or legacy upgrades. “Only peers the sync dialer happened to pick” is too narrow, but the design gap is real. The
  changelog overstates service discovery because the PR implements advertisement without service-aware connection selection.

  ## Claim we could remove

  The registry-error path does not leak the bound iroh socket.

  The code binds the raw iroh endpoint at crates/zakura-network/src/zakura/handler.rs:3199. Registry construction fails at line 3317. The code has not yet cloned the endpoint into the handler, Router, or
  background tasks. Returning at line 3330 drops the last endpoint handle. Iroh closes its UDP sockets when the last endpoint clone drops.

  That path does cancel reactor shutdown without awaiting the reactor tasks, so its cleanup is incomplete. It does not retain the bound socket.
```

distractedm1nd:
yeah sounds like it checks out, discovery I haven't used or tested yet. probably should have its own test. Thanks for digging into it a bit, will push some fixes and see if any more adversarial rounds find anything else

distractedm1nd:
Got around to fixing the first two issues. Sorry for slop, but I agree with this take on the third one:

```
Service-aware peer selection — real limitation, but not necessarily this PR’s bug

 The factual observation is correct: the production discovery dialer calls dial_candidates(&[], …), so custom service IDs do not influence connection selection. Custom services also receive no discovery handle.

However, that dialer and the unused service-candidate API predate this PR. The PR promises that custom IDs are advertised through discovery, which it does; it does not explicitly promise automatic service-targeted dialing.
```

I think that diff is probably going to be more hairy than the rest of what I did and would best be reviewed separately so I'll open a PR tomorrow with it that's stacked on top of this one

crates/zakurad/src/commands/start.rs
distractedm1nd: You mean because of the OnceLocks? Yeah, I have some general ideas on how we can handle the lifecycle better. I had to add some weird things in my last commit to fix some issues, but ideally we might be able to give each service a CancellationToken that it can track. That is then a bit more like golang passing ctx :)  OnceLock resolution for things like node id are a bit tougher but sometimes doable. Can look into it but def outside the scope for this one

distractedm1nd:
@evan-forbes ahhh cannot do stacked PR as i'm working from a fork.

The last commit handles the 3rd issue.

In the most recent commit I now distinguish in custom service between `provided` and `seek`ed (sought) ids. As not every client is also a server. They are now also routed through to peer dialing, but this required the "hairy" diff I mentioned because **the default behavior of `dial_peers` is to use the service array as a filter**. If __all services__ are not advertised by a peer, then we won't connect. So I added a pathway to dial for ANY instead of ALL. 

I'm not too satisfied with this, because I'm not sure "ALL" should be the default behavior anyways. It seems like we don't use it as such currently (we only ever pass a single service!), however we have test cases confirming this is desired behavior. [Here](https://github.com/distractedm1nd/zakura/compare/custom-services...distractedm1nd:zakura:service-driven-discovery?expand=1) is a diff that changes the default behavior.

The only reason I can see to argue against ALL not being default is a dos vector where peer slots get filled up by light nodes advertising subsets of the expected behaviors. An ALL behavior could (__sorta__) prevent this by guaranteeing some slots to peers that advertise a full set of services we need, but we don't use it this way currently and honestly the attack could still happen, albeit not as trivally.

evan-forbes (approved):
dope! if we have flups we can open them outside of this PR

thanks again!

evan-forbes:
I'm going to do some smoke testing w/ this change to make sure and see if we can't replace the default behavior as I like this better

evan-forbes:
opened a PR to fix the default to false issue you were discussing @distractedm1nd . imo we should merge that one here then we can merge! https://github.com/distractedm1nd/zakura/pull/1

also got a headersync only and then also a custom reactor only up on mainnet https://github.com/zakura-core/zakura/pull/742

and a dialing issue as well https://github.com/zakura-core/zakura/pull/741

distractedm1nd:
Ah the default behavior I was referencing had to do with `dial_peers` using the passed services as ALL instead of ANY, which we have tests for but never actually use in that way.

I closed https://github.com/distractedm1nd/zakura/pull/1 because custom services handle it themselves very easily and the wrapper makes things more cumbersome. Added an example there in my last comment

The headersync only experiment is dope, tysm!! 🙏🏼 🙏🏼 

Going to rebase this PR on main and then see if I can update any docs for the custom services to make the connection ownership more clear

distractedm1nd:
Docs seem pretty clear in `Service` to me! So leaving alone

PR #597, opened 2026-08-07
feat(consensus): memoize Halo2 proof verification results

evan-forbes:
## Motivation

Every Orchard proof is verified twice: once when the transaction arrives over mempool gossip, and again when it arrives inside the block that mines it. Measured on one box with matched block sizes, a block's halo2 batch weights sum to exactly the block's Orchard action count *plus* the mempool's — the signature of doing the same work twice.

Zebra used to avoid this with a mempool bypass: if a block's transaction had already been verified in the mempool, `tx::Verifier::call()` returned `Ok` immediately. Upstream removed it in [#10494](https://github.com/ZcashFoundation/zebra/pull/10494) and did not replace it, so modern Zebra pays the same cost. Zakura forked from v5.0.0 and never had it.

It was removed for a structural reason, not a fixable bug. It cached a *verdict* — `valid(tx, height, block time, spent outputs)` — under a key that did not determine that proposition:

| | |
|---|---|
| added, #8951 `1974fea88` | `find_verified_unmined_tx` becomes the first statement of `call()` |
| patched, #10425 `2429f97f7` | **CVE-2026-34377** / GHSA-3vmh-33xr-9cqh — the lookup was by txid, which under ZIP 244 commits to no authorizing data, so proof and signatures could be swapped freely. Fixed narrowly by comparing the full `UnminedTxId`; the bypass was kept |
| deleted, #10494 `1f605eca5` | removed entirely, 84 lines, "consensus: remove mempool bypass" |

Everything between the early return and the response was skipped, including `consensus_branch_id`, `is_orchard_temporarily_disabled`, the NU6.2 proof-size rule, `non_coinbase_expiry_height`, `disabled_add_to_sprout_pool`, and `lock_time_has_passed` — all of which move with height, and the last of which also moves with the miner-chosen block time. The deepest one is `nu = request.upgrade(network)`, which selects both the sighash and, via `verifier_for(nu)`, *which Orchard circuit verifying key the bundle is checked against*. Across a network upgrade boundary the bypass was not skipping policy; it was answering a different cryptographic question and returning the stale answer.

Zakura carries both regression tests: `block_with_garbage_orchard_proofs_is_rejected` and `mempool_cached_result_bypasses_expiry_check_for_block_at_next_height`.

## Solution

Memoize the proof verification itself rather than the verdict. `Item::verify_single(self, vk) -> bool` is a pure function, so a hit is bit-identical to the computation it replaces.

|  | #8951 | this PR |
|---|---|---|
| cached proposition | `valid(tx, height, block time, utxos)` | `verify(bundle, sighash, vk) -> bool` |
| pure function of the key? | **no** | **yes, if the key is complete** |
| short-circuits | `tx::Verifier::call()` | one `oneshot` inside `verify_orchard_bundle` |
| consensus checks skipped on a hit | all of them | **none** |

On a hit, `transaction::Verifier::call` still runs end to end at the block's height, with the block's time and the block's spent outputs. Expiry, lock time, the branch id, the soft-fork gates, UTXO resolution, fees and sigops all still run.

**Key completeness is therefore the entire safety argument, and it is where review effort is best spent** — `Item::cache_key` in `primitives/halo2.rs` and the module docs in `primitives/halo2/memo.rs`.

- **`bundle` and `sighash`** are committed to by hashing the bundle's own consensus encoding (`write_v5_bundle`), which is injective and covers everything `BatchValidator::add_bundle` reads: per-action value commitments, nullifiers, validating keys, note commitments and spend authorization signatures, plus the flags, value balance, anchor, proof and binding signature. An explicit `BundleVersion` discriminant is hashed alongside it, because the encoding contains the flag byte but not the version.
- **`vk`** is committed to structurally. There are already three global verifiers, one per Orchard circuit era; each now has its own memo, so an entry can only ever be read back under the key it was written against. Eras cannot mix in a memo for the same reason they cannot mix in a batch.

Two forcing functions guard the key against drift. `cache_key` destructures `Item` exhaustively, so adding a field is a compile error until someone decides whether it belongs in the key; and `bundle_version_discriminant` matches exhaustively on `ValuePool` and `ProtocolVersion`, so a new pool or protocol version is a compile error too.

Deriving the key by hand from things that "obviously" determine the bundle is how this goes wrong, and it is easy to get wrong twice. The txid does not commit to authorizing data at all (CVE-2026-34377). `(auth digest, sighash)` looks complete and is not: a v6 transaction verifies its Orchard bundle and its Ironwood bundle against the same sighash under the same network upgrade, so those two share both components. Hashing the encoding means only byte-identical bundles can share a key, and byte-identical bundles verify identically.

**Only `Ok` is recorded.** A batch error is not per-item evidence, because `Fallback` resolves batch failures by re-verifying each item singly; and an error out of the service need not be a verdict at all — it can report that the batch worker shut down. Recording either as "this proof is invalid" would make the node reject a valid block, which is a chain split in the other direction.

The memo is a small tower middleware wrapping each era's existing stack, built inside `batch_verifier`, so no call site changes and `verifier_for` returns the same shape it did before. Storage is a FIFO-bounded `HashSet` of 32-byte keys, 20,000 entries per era, no new dependency; the mutex is held only across a set operation and never across an await. Metrics: `zakura.consensus.halo2.memo.{hit,miss,insert,evict}` and a size gauge.

There is deliberately **no config knob**. A runtime lever that changes verification behaviour produces irreproducible bug reports for no operator benefit. If a kill switch is ever wanted for incident response, a startup-only, documented-as-debug env var would be the right shape — a `Config` field would make it supported surface.

## Testing

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p zakura-consensus`.

The upstream regression tests are the headline gate, and both pass unchanged:

- **`block_with_garbage_orchard_proofs_is_rejected`** is the direct test of key completeness. The mempool populates a key from the valid bundle; the block carries the auth-substituted twin with an identical txid; the twin encodes differently, so it misses, is fully verified, and is rejected. If the key is ever weakened to something txid-derived, this test fails.
- **`mempool_cached_result_bypasses_expiry_check_for_block_at_next_height`** — the memo does not touch expiry, so this must pass untouched.
- `dont_skip_verification_of_block_transactions_in_mempool` also passes.

New unit tests in `primitives/halo2/tests.rs`:

- the key is deterministic, and changes with the sighash;
- the key changes with the proof, the binding signature, and the spend authorization signatures — asserting in each case that the txid did *not* change, so the test is exercising the CVE's shape;
- the key changes with the bundle version even when the encoded flag byte is identical, which is what the explicit discriminant is for;
- distinct bundles sharing one sighash get distinct keys (the Orchard/Ironwood collision above);
- every `BundleVersion` gets its own discriminant;
- the memo skips the inner service on a hit, does not reuse a result across different keys, never remembers a failure, and re-verifies an evicted entry rather than mis-answering it.

`transaction::tests::v4_and_v5_with_sapling_spends` fails under a parallel full-crate run on this branch. It also fails identically on `origin/main` with these changes stashed, and passes in isolation on both — pre-existing and unrelated.

Not yet measured: the field speedup. The 93.5% figure that motivated this is *crypto* occupancy, which also contains script verification and redpallas batches that this change does not address; the halo2 term has not been sized separately. See follow-up.

## Specifications & References

- Zebra [#8951](https://github.com/ZcashFoundation/zebra/pull/8951), [#10425](https://github.com/ZcashFoundation/zebra/pull/10425) (CVE-2026-34377 / GHSA-3vmh-33xr-9cqh), [#10494](https://github.com/ZcashFoundation/zebra/pull/10494)
- ZIP 244 (txid does not commit to authorizing data), ZIP 229 (Orchard cross-address restriction, which is why the era split exists)
- GHSA-jfw5-j458-pfv6 (the NU6.2 circuit fix behind the per-era verifying keys)

## Follow-up Work

- Split the block transaction phase into halo2 / sapling / redpallas / script to size the actual ceiling for this change, then measure warm-vs-cold and the field hit rate. Note that `worst_case_tx_verification` cannot show this: it passes a closed mempool setup channel, so the population path never runs.
- Worth taking to the Zebra maintainers. #10494 removed the optimization with no replacement, so every Zebra node on the network pays this cost too; if the pure-function framing survives their review, we are not carrying a consensus-critical divergence alone.
- `TransactionWithDepsByMinedId` is still served in Zakura but consumed only by a test — inherited API surface from a bypass we never had. This PR does not need it, so it can be removed separately.
- Sapling groth16 and redpallas admit the same treatment; neither is measured yet, and groth16 is unbatched here, so they should follow rather than ride along.

### DIFF
--- a/crates/zakura-consensus/src/primitives/halo2.rs
+++ b/crates/zakura-consensus/src/primitives/halo2.rs
@@ -12,11 +12,13 @@ use std::{
 use futures::{future::BoxFuture, FutureExt};
 use once_cell::sync::Lazy;
 use orchard::{
-    bundle::{BatchError, BatchValidator},
+    bundle::{BatchError, BatchValidator, BundleVersion},
     circuit::{OrchardCircuitVersion, VerifyingKey},
+    ProtocolVersion, ValuePool,
 };
 use rand::thread_rng;
 use zakura_chain::{parameters::NetworkUpgrade, transaction::SigHash};
+use zcash_primitives::transaction::components::orchard::write_v5_bundle;
 use zcash_protocol::value::ZatBalance;
 
 use crate::{error::TransactionError, BoxError};
@@ -28,8 +30,14 @@ use tower_fallback::Fallback;
 
 use super::spawn_fifo;
 
+mod memo;
+
 #[cfg(test)]
 mod tests;
+
+pub use memo::Memoized;
+
+use memo::{CacheKey, MEMO_CAPACITY};
 
 /// Adjusted batch size for halo2 batches.
 ///
@@ -50,6 +58,12 @@ type Sender = watch::Sender<Option<VerifyResult>>;
 /// The type of a prepared verifying key.
 /// This is the key used to verify individual items.
 pub type ItemVerifyingKey = VerifyingKey;
+
+/// The BLAKE2b personalization for [`Item`] memo cache keys.
+///
+/// Domain-separates these keys from every other BLAKE2b-256 hash in the protocol, so that a
+/// memo key can never be confused with a txid, an auth digest, or a sighash.
+const HALO2_MEMO_PERSONALIZATION: &[u8; 15] = b"ZakuraHalo2Memo";
 
 // The Orchard Action circuit, and therefore its verifying key, changes across protocol eras.
 // A proof produced under one circuit version does not verify under the other keys. So we keep
@@ -150,6 +164,76 @@ impl Item {
         }
         batch.validate(thread_rng())
     }
+
+    /// Returns a key that determines every input to this item's proof verification.
+    ///
+    /// [`Memoized`] reuses a previous `Ok` result whenever this key matches, so the key must
+    /// commit to everything [`BatchValidator::add_bundle`] reads. It hashes the bundle's own
+    /// consensus encoding, which is injective and covers all of it: the per-action value
+    /// commitments, nullifiers, validating keys, note commitments and spend authorization
+    /// signatures, the bundle flags, value balance, anchor, proof, and binding signature.
+    ///
+    /// The verifying key is *not* in the key. It does not need to be: each Orchard circuit era
+    /// has its own verifier and therefore its own memo, so an entry can only ever be read back
+    /// under the key it was written against. See [`verifier_for`].
+    ///
+    /// Deriving this by hand from things that "obviously" determine the bundle is how this goes
+    /// wrong. The txid does not commit to authorizing data at all under ZIP 244 — that was
+    /// CVE-2026-34377 — and even `(auth digest, sighash)` is incomplete, because the Orchard and
+    /// Ironwood bundles of one v6 transaction share both. Hashing the encoding avoids having to
+    /// get that reasoning right.
+    fn cache_key(&self) -> CacheKey {
+        // Destructured exhaustively on purpose: adding a field to `Item` is a compile error here
+        // until someone decides whether it belongs in the key.
+        let Item { bundle, sighash } = self;
+
+        let mut hasher = blake2b_simd::Params::new()
+            .hash_length(32)
+            .personal(HALO2_MEMO_PERSONALIZATION)
+            .to_state();
+
+        hasher.update(&[bundle_version_discriminant(bundle.bundle_version())]);
+        hasher.update(&sighash.0);
+
+        // `write_v5_bundle` is the total encoder: it is `write_v6_bundle` without the precondition
+        // that rejects pre-NU6.3 bundle versions, and the two produce identical bytes otherwise.
+        // The v5/v6 distinction it therefore does not encode is supplied by the discriminant above.
+        let mut encoded = Vec::new();
+        write_v5_bundle(Some(bundle.as_ref()), &mut encoded)
+            .expect("encoding a bundle into a Vec cannot fail: the encoder only reports IO errors");
+        hasher.update(&encoded);
+
+        hasher
+            .finalize()
+            .as_bytes()
+            .try_into()
+            .expect("hash_length(32) produces exactly 32 bytes")
+    }
+}
+
+/// Returns a stable one-byte discriminant for `version`.
+///
+/// A bundle's consensus encoding contains its flag byte but not its [`BundleVersion`], and two
+/// bundles in different value pools can share a flag byte, so [`Item::cache_key`] commits to the
+/// value pool and protocol version separately. Today the pool affects verification only through
+/// the flags, but the bundle version already selects other version-dependent behaviour in the
+/// `orchard` crate, and one byte removes the need to re-check that on every dependency bump.
+///
+/// Both matches are exhaustive: a new value pool or protocol version is a compile error here
+/// until it is given a discriminant on purpose.
+fn bundle_version_discriminant(version: BundleVersion) -> u8 {
+    let pool = match version.value_pool() {
+        ValuePool::Orchard => 0x00,
+        ValuePool::Ironwood => 0x10,
+    };
+
+    let protocol = match version.protocol_version() {
+        ProtocolVersion::InsecureV1 => 0x00,
+        ProtocolVersion::V2 => 0x01,
+        ProtocolVersion::V3 => 0x02,
+    };
+
+    pool | protocol
 }
 
 trait QueueBatchVerify {
@@ -217,12 +301,15 @@ impl Service<Item> for OrchardFallback {
     }
 }
 
+/// The batching-and-fallback stack for one Orchard circuit era, before memoization.
+type BatchFallbackService = Fallback<Batch<Verifier, Item>, OrchardFallback>;
+
 /// The concrete type of a global Halo2 verification service.
 ///
 /// Each Orchard circuit era gets its own instance — see [`VERIFIER_PRE_NU6_2`], [`VERIFIER_NU6_2`],
-/// or [`VERIFIER_NU6_3_ONWARD`] — so that batches, fallbacks, and verifying keys are fully
+/// or [`VERIFIER_NU6_3_ONWARD`] — so that batches, fallbacks, verifying keys, and memos are fully
 /// separated per era.
-type VerifierService = Fallback<Batch<Verifier, Item>, OrchardFallback>;
+type VerifierService = Memoized<BatchFallbackService>;
 
 /// Builds a global Halo2 verifier that validates every item against `vk`.
 ///
@@ -231,7 +318,17 @@ type VerifierService = Fallback<Batch<Verifier, Item>, OrchardFallback>;
 /// passed here, so an item built by this verifier is always checked against exactly one era's key.
 /// Callers select the correct era's key by which `VERIFYING_KEY_*` they pass; there is no runtime
 /// key resolution.
+///
+/// The stack is wrapped in a [`Memoized`] so that a proof gossiped into the mempool does not have
+/// to be verified again when the block that mines it arrives. Because each era builds its own
+/// verifier here, each era also gets its own memo, which is what binds a remembered result to the
+/// `vk` it was produced under.
 fn batch_verifier(vk: &'static ItemVerifyingKey) -> VerifierService {
+    Memoized::new(batch_fallback_verifier(vk), MEMO_CAPACITY)
+}
+
+/// Builds the un-memoized batching-and-fallback stack for `vk`.
+fn batch_fallback_verifier(vk: &'static ItemVerifyingKey) -> BatchFallbackService {
     Fallback::new(
         Batch::new(
             Verifier::new(vk),
--- a/crates/zakura-consensus/src/primitives/halo2.rs
+++ b/crates/zakura-consensus/src/primitives/halo2.rs
@@ -35,9 +35,7 @@ mod cache;
 #[cfg(test)]
 mod tests;
 
-pub use cache::Cached;
-
-use cache::{CacheKey, CACHE_CAPACITY};
+use cache::Cached;
 
 /// Adjusted batch size for halo2 batches.
 ///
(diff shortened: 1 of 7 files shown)


evan-forbes:
Confirmed and fixed in ea7836e. The analysis is correct in every particular, including the two things it rules out — I re-derived both rather than take them on trust.

### What I verified

`Batch::poll_ready` stashes a permit in `self.permit` (`tower-batch-control/src/service.rs:334-343`) that only `Batch::call` consumes (`:349-354`), and returns an error when the worker has exited, panicked, or closed its channel (`:289-318`, `:341-343`). Since callers poll readiness *before* `call`, a memo **hit** never reached `Memoized::call` at all when the worker was down — it was reported as a verification error instead of the `Ok` the memo was holding.

Your framing of that as the more serious of the two consequences is right, and it's the one I'd have wanted flagged: it contradicts this module's own stated reasoning at `memo.rs:33-37` that an error "need not be a verdict" and that rejecting a valid block "is a chain split in the other direction". The memo was written specifically to avoid that failure mode and then reintroduced it one layer up.

Your two exclusions also hold: the permit is released rather than leaked, because the production call site is `verifier_for(nu).clone().oneshot(..)` and `oneshot` drops the clone; and `Batch::call`'s `.expect("poll_ready must be called before a batch request")` cannot fire, because a hit leaves the permit `Some`. Fail-closed, not a soundness bug — agreed on severity.

### The fix

`Memoized::poll_ready` now reports ready without consulting the inner service, and the **miss** path acquires inner readiness inside `call`. The root cause is that `poll_ready` has to decide before the item is known, so it cannot know whether the inner service will be used at all — the readiness acquisition belongs where that is known, which is `call`.

The semaphore still bounds concurrent batch requests: a miss clones the inner service and acquires its permit before calling, exactly as before. Only the *timing* of the wait moves, and for the `oneshot` call sites this service has, that is the same instant.

### Tests

Both were confirmed to fail against the previous `poll_ready` before being committed — I reverted the fix, watched them go red, and restored it:

- `memo_hit_survives_an_inner_service_that_never_becomes_ready` — warms the memo through a healthy inner service, swaps in one whose `poll_ready` always errors, and asserts the hit is still answered. It also asserts the inner service is not polled for readiness **even once**, which is what pins the capacity half of your finding.
- `memo_miss_propagates_an_inner_readiness_failure` — the converse, so the fix cannot degrade into swallowing readiness errors: a miss surfaces the failure, and the failure is not recorded as an `Ok`.

Sapling analogues of both are in the stacked #600, which shares this `Memoized` — so a future change cannot restore the delegating `poll_ready` for one pool while the other's tests still pass.

ValarDragon:
This improvement is huge! Thank you for catching it, and fixing thoroughly, phenomenal work.

This is going to be a slower review, as I will be going through it in more depth by hand + AI due to consensus criticality

crates/zakura-consensus/src/primitives/halo2.rs
```diff
@@ -150,6 +164,76 @@ impl Item {
...
+            .to_state();
+
+        hasher.update(&[bundle_version_discriminant(bundle.bundle_version())]);
+        hasher.update(&sighash.0);
+
+        // `write_v5_bundle` is the total encoder: it is `write_v6_bundle` without the precondition
+        // that rejects pre-NU6.3 bundle versions, and the two produce identical bytes otherwise.
+        // The v5/v6 distinction it therefore does not encode is supplied by the discriminant above.
+        let mut encoded = Vec::new();
+        write_v5_bundle(Some(bundle.as_ref()), &mut encoded)
```
ValarDragon, line 207: The fact we need this comment suggests this should just be renamed to `write_bundle` perhaps?

crates/zakura-consensus/src/primitives/halo2/memo.rs
```diff
@@ -0,0 +1,246 @@
+//! A bounded memo of Halo2 Orchard Action proofs that have already verified.
```
ValarDragon: Ahh, can we rename this file from memo.rs to memoization perhaps?

Really had me confused with tx memo

ValarDragon: or even better, cache.rs ?

crates/zakura-consensus/src/primitives/halo2.rs
evan-forbes: yeah good point https://github.com/zakura-core/zakura/pull/597/commits/2275151451be29ab2c4776a7139d71d366f95b65

crates/zakura-consensus/src/primitives/halo2/memo.rs
evan-forbes: yeah good point renamed to cache

crates/zakura-consensus/src/primitives/halo2/cache.rs
```diff
@@ -0,0 +1,246 @@
...
+//! # Why this is not the mempool bypass
+//!
+//! Zebra used to skip *whole-transaction* verification for block transactions already accepted
+//! into the mempool. That was removed upstream as a security fix (PR #10494), after two rounds of
+//! patching, because the cached proposition was `valid(tx, height, block time, spent outputs)` —
+//! not a function of the key it was stored under. A transaction valid at one height is not
+//! necessarily valid at another: expiry, lock time, the consensus branch id, the Orchard soft-fork
+//! gates and the proof-size rule all move with height, and the network upgrade even selects which
+//! Orchard circuit verifying key applies. Zakura keeps the regression tests for both failures
+//! (`block_with_garbage_orchard_proofs_is_rejected` and
```
p0mvn: Nit: can we shorten and make this AI comment mor human readable

crates/zakura-consensus/src/primitives/halo2.rs
```diff
@@ -150,6 +164,78 @@ impl Item {
         }
         batch.validate(thread_rng())
     }
+
+    /// Returns a key that determines every input to this item's proof verification.
+    ///
```
p0mvn, line 179: nit: same comment on verbosity

p0mvn (approved):
LGTM.

Reviewed the historical reasons of how we got here. Also, execution flows across mempool and consensus + cache semantics.

Had only readability nits

ValarDragon:
We should add more tests to this PR. The first 3 were explicit asks I had, reworded by AI. The last two were AI suggestions.

Expiry cache:

  1. A real V5/V6 Orchard transaction accepted at H+1, warming the actual Halo2 cache.
  2. The same transaction submitted in a block at H+2.
  3. Rejection with ExpiredTransaction.
  
 Real mempool → block cache hit

- Use a valid V5/V6 Orchard transaction. Verify it through Request::Mempool, then the identical transaction through Request::Block. Assert the inner Halo2 verifier runs exactly once.
- This proves the production transaction path and service clones actually share the cache.

Same-txid authorization substitution misses

Warm the cache with a valid transaction, then submit a same-txid twin with a canonical-length bit mutation. Table-drive at least:
      - Halo2 proof
      - Binding signature
      - Spend authorization signature
      - Expiry height

     Assert the mutated item reaches the inner verifier and is rejected. The proof must remain canonical length; the current four-byte garbage proof is rejected before the cache path.

Cache is shared across Cached clones

Warm through clone A, request through clone B, and assert one total inner call. Production uses verifier_for(...).clone(), while the current hit test repeatedly calls the same wrapper instance.


Cancellation does not populate the cache

Start an inner verification, cancel/drop the future before it returns Ok, then retry. Assert the retry reaches the inner verifier.

crates/zakura-consensus/src/primitives/halo2.rs
ValarDragon: dumb Q where did that commit go? https://github.com/zakura-core/zakura/commit/2275151451be29ab2c4776a7139d71d366f95b65

crates/zakura-consensus/src/primitives/halo2.rs
```diff
@@ -150,6 +164,71 @@ impl Item {
...
+    };
+
+    let protocol = match version.protocol_version() {
+        ProtocolVersion::InsecureV1 => 0x00,
+        ProtocolVersion::V2 => 0x01,
+        ProtocolVersion::V3 => 0x02,
+    };
+
+    pool | protocol
 }
```
ValarDragon, line 242: Why do we need this for the pool? Isn't it implied via sighash in the cachekey?

Protoocl makes sense

crates/zakura-consensus/src/primitives/halo2.rs
ValarDragon: ah it got rebased out, but I think its change was removed

ValarDragon:
We should clear the cache when we clear the mempool right?

EDIT: Its not technically needed because this is stateless, and would require coupling the two. NVM lets not do that

ValarDragon:
I like the cache just being for halo2 for now, and we refactor over time to make it easily include the other needed fields.

crates/zakura-consensus/src/primitives/halo2.rs
ValarDragon: Sighash technically implies via consensus branch ID

ValarDragon:
We should do a pass to remove some of the extra stuff as pub. AI slop reply:

  - CacheKey and CACHE_CAPACITY → pub(super).
  - Remove pub use cache::Cached.
  - Keep Cached internal.
  - Hide inner() behind a narrow internal flush helper.
  - Avoid returning the concrete cached service publicly; expose a crate-internal verification function instead.

ValarDragon:
Confirmed all tests added!

crates/zakura-consensus/src/primitives/halo2.rs
evan-forbes: The rebase rewrote `2275151` as `8635354`; its memo-to-cache change was retained. I linked the wrong commit, though: that commit did not change this encoder.

The explicit `write_v5_bundle` call is deliberate. The generic `write_bundle` in `zcash_primitives` is private; `write_v5_bundle` is the public total encoder and delegates to it without the v6 version precondition. Aliasing it locally to `write_bundle` would hide that dependency assumption, so I think keeping the public function's real name and the adjacent explanation is clearer.

evan-forbes:
The cache-specific visibility cleanup is in c195113537c206586462ce3fb38946e9e82637f8:

- `CacheKey` and `CACHE_CAPACITY` are private.
- The `Cached` re-export is removed; the cache module remains private.
- `inner()` is restricted to the parent module and used through the narrow flush helper.
- Cache test instrumentation is exposed only through a test-only helper.

The remaining public verifier globals and `verifier_for()` predate this PR. I opened #609 to move them behind a crate-internal verification API and completely remove the concrete cached service from the public boundary.

ValarDragon (approved):
Great work!

On CacheKey: We're still adding a custom ironwood / orchard field to the cache-key, so I'd rather do the cache key as tx_id || wtx_id || discriminant but that can be a follow-on

PR #599, opened 2026-08-07
perf(chain): defer the Sapling rk small-order check

evan-forbes:
## Motivation

Deserializing a Sapling transaction with 3 spends and 2 outputs takes **322.5 µs** — about thirty times worse per byte than an Orchard transaction. Benchmarked on an otherwise idle 32-vCPU host, effectively all of it is `rk` decoding, and half of that is redundant:

| operation | cost |
|---|---|
| `redjubjub::VerificationKey::<SpendAuth>::try_from([u8; 32])` | 54.0 µs |
| `ValidatingKey::try_from([u8; 32])` (the above, then decompresses again) | 107.8 µs |

3 spends × 107.8 µs = 323 µs, i.e. the entire parse.

For context, the same benchmark over the in-tree mainnet vectors:

| sample | size | deserialize |
|---|---|---|
| V5 coinbase | 201 B | 863 ns |
| Orchard, 2 actions | 9,165 B | 32.5 µs |
| Sapling, 3 spends + 2 outputs | 3,141 B | **322.5 µs** |

**Root cause.** `redjubjub::VerificationKey::try_from` already decompresses the point and keeps it — `reddsa`'s `verification_key.rs:77` stores `point: T::Point` — but the field is `pub(crate)`, so `ValidatingKey::try_from` decompressed the same 32 bytes a second time in order to run the not-small-order check. Point decompression needs a square root in the base field, which is why the second one costs as much as the first.

The check itself is *not* redundant. `reddsa` deliberately stopped rejecting small-order keys and says so at that call site: Sapling has to do it separately. Only the decompression is waste.

## Solution

Finish the job the tree already started for `cv` and `epk`. `crates/zakura-consensus/src/transaction/check.rs` documented the deferral and ended with "Spend rk is still validated at deserialization" — `rk` was the one field left behind, and it is the expensive one.

- `ValidatingKey` now stores the raw 32-byte encoding, exactly like `ValueCommitment` and `EphemeralPublicKey`, with an `is_valid_not_small_order()` that decompresses once.
- `ShieldedData::point_encodings_are_valid` runs that check alongside the ones it already made for the sibling fields, so it is reached in production from `check::sapling_point_encodings_are_valid`.
- `ValidatingKey` becomes `Copy`, dropping the clones at the call sites.

Per Spend description:

| path | before | after |
|---|---|---|
| deserialization | 107.8 µs | ~0 |
| semantic / mempool verification | 54 µs (cv) | 108 µs (cv + rk) |
| checkpoint sync | 107.8 µs | ~0 |

**This moves where a consensus rule is enforced, and reviewers should see it named.** Today a transaction with a small-order `rk` fails to *deserialize*, so it is rejected at the network boundary and under checkpoint sync. Afterwards it deserializes and is rejected by `check::sapling_point_encodings_are_valid` on the semantic and mempool paths — and under checkpoint sync it is not rejected at all. That is the identical tradeoff already accepted for `cv` and `epk`, and it is sound for the same reason: checkpoint sync's authority is the checkpoint hash chain, not per-field validation.

librustzcash is the backstop on the semantic path, and it splits the rule in two: `read_rk` rejects off-curve/non-canonical encodings at read, while `redjubjub` admits small-order keys and the Sapling verifier's `check_spend` (`sapling-crypto-0.7.0/src/verifier.rs`) rejects them at verify.

**Consumer audit.** Every consumer of `Spend::rk` converts straight back to `[u8; 32]`, so nothing anywhere consumes the decompressed group element:

| site | use |
|---|---|
| `sapling/spend.rs` ×3 | moves between the per-anchor and v5-prefix representations |
| `sapling/spend.rs` ×2 | `ZcashSerialize`, via `<[u8; 32]>::from` |
| `transaction/sighash/v4.rs` | v4 sighash digest |
| `transaction/zip244.rs` | ZIP-244 digest |
| `zakura-rpc/.../types/transaction.rs` | RPC output |

Sapling bundle verification runs on the `sapling_crypto` bundle from `to_librustzcash`, not on our `Spend`, so `ValidatingKey`'s decompression was pure overhead once the check moved.

## Testing

- **`sapling_small_order_rk_deferred_but_caught_by_librustzcash`** (new) — pins both halves of the librustzcash backstop: `read_rk` accepts a small-order `rk` and rejects an off-curve one, and `check_spend`'s `is_small_order` is what flags the former. Asserts a small-order `rk` now parses, that the deferred check rejects it, that `to_librustzcash` accepts it and rejects the off-curve one, and that the lazy `rk` round-trips byte-for-byte — the txid and merkle root hash these bytes.
- **`sapling_point_checks_match_librustzcash_predicates`** — gains an `rk` arm. This is the test that makes the new byte-level check trustworthy: it pins `is_valid_not_small_order` against the *conjunction* of `read_rk` and `check_spend` over the existing corpus (valid non-small-order, small-order, off-curve, a byte-pattern sweep, and 64 prime-order points), with the same guard against a vacuous all-accept/all-reject comparison.
- **`sapling_point_encodings_check_rejects_bad_v4_spend_cv`** — now varies `rk` as well as `cv`, covering the distinct V4 `PerSpendAnchor` route for both invalid classes.
- **`sapling_spends_with_invalid_rk_are_rejected_after_roundtrip`** (new) — drives the full transaction verifier over real mainnet V4 and V5 transactions with Sapling spends, for both off-curve and small-order keys, asserting the raw bytes survive the round-trip and the request fails with `TransactionError::SmallOrder`.
- **`sapling_lazy_cv_epk_edge_cases`** — asserted `rk` was *not* lazy; now asserts the deferred check's verdicts instead.

```
cargo fmt --all -- --check
cargo clippy -p zakura-chain -p zakura-consensus -p zakura-rpc --all-targets -- -D warnings
cargo nextest run -p zakura-chain -p zakura-consensus -p zakura-rpc -p zakura-state
cargo doc --no-deps --workspace --all-features --document-private-items
```

All green. Note `transaction::tests::v4_and_v5_with_sapling_spends` fails under plain `cargo test -p zakura-consensus --lib` on `main` as well as on this branch — a pre-existing interaction between parallel tests and the process-global Sapling batch verifier. It passes under nextest, which is what CI runs.

- [Spend description](https://zips.z.cash/protocol/protocol.pdf#spenddesc) — "Check that a Spend description's cv and rk are not of small order"
- `sapling-crypto-0.7.0/src/verifier.rs:48-51` — `check_spend`'s small-order `rk` rejection
- `reddsa-0.5.2/src/verification_key.rs:77` — the retained `pub(crate) point`, and the comment at `:107-110` deferring the small-order check to Sapling

### Follow-up Work

Optional, not a blocker: ask `redjubjub` to expose either the decompressed point or an `is_small_order()` on `VerificationKey`. `reddsa`'s own comment says Sapling has to do this check separately, so it is an obvious gap. After this PR Zakura is no longer a caller that would benefit, so it is a courtesy rather than a dependency.

---

### Update — V12 findings handled, and a citation correction

`librustzcash_check_spend_rejects_a_small_order_rk` (d7f2907) addresses **F-188690**: the `rk` arm of the equivalence test partly modelled itself, so half of it was tautological. It now pins `check_spend`'s behaviour by driving `sapling_crypto::BatchValidator::check_bundle` over a real mainnet bundle. See the [full evaluation](#issuecomment-5223173293).

`ValidatingKey` is no longer `Copy` (a8d28a5) — `cargo-semver-checks` correctly flagged `copy_impl_added` as needing a major bump on a published crate. Consumers take the bytes through `From<&ValidatingKey>`, which avoids both the clone and the move.

**Version citations above are corrected.** I originally cited `reddsa-0.5.1` and `sapling-crypto-0.5.0`, read out of the local registry cache rather than `Cargo.lock`. The workspace locks **reddsa 0.5.2, sapling-crypto 0.7.0, jubjub 0.10.0, redjubjub 0.8.0**. Every upstream claim was re-verified against those and all hold.

**Caveat a reviewer should not take from me on trust:** the semver job is green, but the substantive API changes here — `From<ValidatingKey> for redjubjub::VerificationKey<SpendAuth>` becoming a `TryFrom`, and the inverse changing its associated `Error` from `&'static str` to `Infallible` — look breaking and did *not* fire any lint. 222/223 checks passed with only `copy_impl_added`. `cargo-semver-checks` appears not to lint altered trait impls involving foreign types, so please judge those two by eye rather than by the green check.

### DIFF
--- a/crates/zakura-chain/src/sapling/arbitrary.rs
+++ b/crates/zakura-chain/src/sapling/arbitrary.rs
@@ -114,9 +114,7 @@ fn spendauth_verification_key_bytes() -> impl Strategy<Value = ValidatingKey> {
     prop::array::uniform32(any::<u8>()).prop_map(|bytes| {
         let rng = ChaChaRng::from_seed(bytes);
         let sk = redjubjub::SigningKey::<redjubjub::SpendAuth>::new(rng);
-        redjubjub::VerificationKey::<redjubjub::SpendAuth>::from(&sk)
-            .try_into()
-            .unwrap()
+        redjubjub::VerificationKey::<redjubjub::SpendAuth>::from(&sk).into()
     })
 }
 
--- a/crates/zakura-chain/src/sapling/keys.rs
+++ b/crates/zakura-chain/src/sapling/keys.rs
@@ -368,7 +368,11 @@ impl ZcashDeserialize for EphemeralPublicKey {
 /// [2]: https://zips.z.cash/protocol/protocol.pdf#concretereddsa
 /// [`ValueCommitment`]: crate::sapling::ValueCommitment
 /// [`Transaction::sapling_point_encodings_are_valid`]: crate::transaction::Transaction::sapling_point_encodings_are_valid
-#[derive(Clone, Copy, Deserialize, PartialEq, Eq, Serialize)]
+//
+// Deliberately not `Copy`, unlike its sibling deferred-point types: adding `Copy` to an already
+// published type is a major semver break, and it would buy nothing here — every consumer takes
+// the bytes through `From<&ValidatingKey>`.
+#[derive(Clone, Deserialize, PartialEq, Eq, Serialize)]
 pub struct ValidatingKey(pub(crate) [u8; 32]);
 
 impl ValidatingKey {
--- a/crates/zakura-chain/src/sapling/keys.rs
+++ b/crates/zakura-chain/src/sapling/keys.rs
@@ -345,56 +345,114 @@ impl ZcashDeserialize for EphemeralPublicKey {
 /// represent [SpendAuthSig^{Sapling}.Public][2], which allows any points, including
 /// of small order).
 ///
+/// # Lazy point decompression
+///
+/// Stored as the raw 32-byte compressed encoding, exactly as it appears on the
+/// wire. Decompressing a Jubjub point requires a square root in the base field,
+/// which dominates the cost of parsing a Sapling transaction — so, like
+/// [`ValueCommitment`] and [`EphemeralPublicKey`], the point and its
+/// not-small-order check are deferred off the checkpoint-sync hot path.
+///
+/// # Consensus
+///
+/// Deserialization only checks the byte length; it does not prove the bytes are
+/// a canonical, non-small-order point. The not-small-order check is deferred to
+/// the semantic verifier and mempool, which call
+/// [`ValidatingKey::is_valid_not_small_order`] (via
+/// [`Transaction::sapling_point_encodings_are_valid`]) and also verify the
+/// Sapling bundle through librustzcash, whose `check_spend` rejects a
+/// small-order `rk`. The checkpoint verifier trusts block hashes and skips the
+/// check. Covered by `sapling_small_order_rk_deferred_but_caught_by_librustzcash`.
+///
 /// [1]: https://zips.z.cash/protocol/protocol.pdf#spenddesc
 /// [2]: https://zips.z.cash/protocol/protocol.pdf#concretereddsa
-#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
-pub struct ValidatingKey(redjubjub::VerificationKey<SpendAuth>);
+/// [`ValueCommitment`]: crate::sapling::ValueCommitment
+/// [`Transaction::sapling_point_encodings_are_valid`]: crate::transaction::Transaction::sapling_point_encodings_are_valid
+#[derive(Clone, Copy, Deserialize, PartialEq, Eq, Serialize)]
+pub struct ValidatingKey(pub(crate) [u8; 32]);
 
-impl From<ValidatingKey> for redjubjub::VerificationKey<SpendAuth> {
-    fn from(rk: ValidatingKey) -> Self {
-        rk.0
-    }
-}
-
-impl TryFrom<redjubjub::VerificationKey<SpendAuth>> for ValidatingKey {
-    type Error = &'static str;
-
-    /// Convert an array into a ValidatingKey.
+impl ValidatingKey {
+    /// Returns true if the stored encoding is a canonical, non-small-order
+    /// Jubjub point, i.e. a valid `rk` per the consensus rules.
     ///
-    /// Returns an error if the key is malformed or [is of small order][1].
+    /// This is the not-small-order check deferred from deserialization, run by
+    /// the semantic verifier (not the checkpoint verifier) on untrusted
+    /// transactions.
     ///
     /// # Consensus
     ///
     /// > Check that a Spend description's cv and rk are not of small order,
     /// > i.e. \[h_J\]cv MUST NOT be 𝒪_J and \[h_J\]rk MUST NOT be 𝒪_J.
     ///
-    /// [1]: https://zips.z.cash/protocol/protocol.pdf#spenddesc
-    fn try_from(key: redjubjub::VerificationKey<SpendAuth>) -> Result<Self, Self::Error> {
-        if bool::from(
-            jubjub::AffinePoint::from_bytes(key.into())
-                .unwrap()
-                .is_small_order(),
-        ) {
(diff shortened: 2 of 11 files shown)


evan-forbes:
Evaluated both findings. **F-188690 is correct and is fixed in d7f2907**; F-188693 is accurate but I'm declining the part of it that isn't this PR's. The audit also surfaced a citation error of mine, corrected below.

### F-188690 — equivalence test partially models itself: **valid, fixed**

Correct, and the diagnosis is exact. The `rk` model expressed the `check_spend` half as `jubjub::AffinePoint::from_bytes(bytes).is_small_order()`, which is character for character what `ValidatingKey::is_valid_not_small_order` runs — so that half was tautologically true for every input and could never have detected a divergence.

There is no better *model* available: `sapling-crypto` exposes no byte-level entry point for that half, because `check_spend` itself is those two calls. So the fix is different evidence rather than a better model — `librustzcash_check_spend_rejects_a_small_order_rk` drives the library's own `BatchValidator::check_bundle` over a **real mainnet Sapling bundle** and observes that it rejects once the first spend's `rk` is replaced with the Jubjub identity.

A real bundle turned out to be load-bearing. My first attempt reused the crafted fixture from `sapling_small_order_rk_deferred_but_caught_by_librustzcash`, and its vacuity guard caught the problem immediately: that fixture carries an all-zero proof, which `check_bundle` rejects whatever its `rk` is — the test would have passed for the wrong reason. The accept leg is retained for exactly that.

Replacing `rk` invalidates the spend authorization signature, but `check_bundle` only *queues* signature and proof batch items — validity is decided at `validate()` — so the synchronous verdict isolates the small-order check.

Both sites now document what the equivalence test does and does not pin. Your closing observation is the one I kept: it still covers the non-trivial property that `reddsa`'s decoder and `jubjub`'s decoder agree on acceptance.

### F-188693 — inconsistent decoder provenance: **accurate; `rk` justified, `epk` out of scope**

- **`rk`**: as the finding concedes, structural. librustzcash splits the rule across `read_rk` and `check_spend`, and `redjubjub` exposes no `is_small_order` on `VerificationKey` — deliberately, per the comment at `reddsa-0.5.2/src/verification_key.rs:107-110` saying Sapling must check separately. There is no single library function to call. The residual risk the finding is really pointing at — that nothing pinned the library's actual behaviour — is what F-188690's fix now covers.
- **`epk`**: accurate, but pre-existing. This PR does not touch `EphemeralPublicKey`; that predicate and its doc landed with the earlier `cv`/`epk` deferral. Changing it here would be an unrelated refactor to a consensus predicate in a PR that already moves an enforcement point, which is the wrong shape for review. Filed as follow-up.

I agree with the severity on both, including your note that no byte-level divergence is demonstrable today — `Cargo.lock` resolves one `jubjub` (0.10.0) shared by `reddsa` and `sapling-crypto`, so all three paths bottom out in the same curve code.

### Citation error the audit caught, which I've corrected

Your finding cites `reddsa 0.5.2` and `sapling-crypto 0.7.0`. I had written `reddsa-0.5.1` and `sapling-crypto-0.5.0` in this PR's description — I read stale directories out of `~/.cargo/registry/src/`, which holds every version anything on the machine has pulled, instead of reading `Cargo.lock`. Thanks for the correction.

I re-verified every upstream claim against the actually-locked versions before trusting the argument. All hold:

| claim | locked source | status |
|---|---|---|
| `check_spend` rejects small-order `rk` | `sapling-crypto-0.7.0/src/verifier.rs:48-51` | holds |
| `VerificationKey` retains the decompressed point, skips the small-order check | `reddsa-0.5.2/src/verification_key.rs:77`, `:107-110` | holds |

The PR description is updated. (The same audit prompted me to re-check #600's central premise, which depends on `zcash_primitives` — `0.30.0`, not the `0.27.0` I cited. Sapling's `write_v5_bundle` is still `pub(crate)` at `sapling.rs:452` and orchard's still `pub` at `orchard.rs:355`, so that design stands too.)

No version numbers had leaked into the committed code of either PR — only into prose.

PR #600, opened 2026-08-07
feat(consensus): cache Sapling bundle verification

evan-forbes:
## Motivation

#597 cached Orchard and Ironwood bundle verification, and #613 rekeyed that cache
by witnessed transaction ID. Sapling is the one shielded pool left verifying twice:
once when a transaction is gossiped into the mempool, and again when the block that
mines it arrives.

Per review feedback (#600 (comment)), this covers Sapling **under the same
`CacheKey`** as Halo2 rather than a Sapling-specific key construction. The earlier
revision of this PR hashed a digest of the transaction bytes the bundle was parsed
from; all of that machinery is gone, along with its `zakura-chain` changes.

**What it is worth.** Measured over 582 recent mainnet blocks, Sapling is about 12%
of shielded verification work (0.48 Sapling units per block against 3.55
Orchard/Ironwood actions, ~0.97 ms per unit either way), so a hit saves roughly
0.5 ms per block next to the ~3.5 ms #597 already covers. That share is falling as
Ironwood ramps. The case for landing it now is that it costs one shared key
construction instead of a second one: the cache, its key, and its safety argument
are the ones already reviewed in #597 and #613.

### One cache, one key

`primitives::halo2::cache` moves up to `primitives::cache` and becomes generic over
a `CachedItem` trait, so the Halo2 and Sapling verifiers share one service, one
eviction policy and one set of safety notes. The key is the one #613 introduced,
with two changes:

- `wtx_id: WtxId` becomes `tx_id: UnminedTxId`, because Sapling bundles also appear
  in v4 transactions, which have no witnessed ID; and
- `pool: orchard::ValuePool` becomes a `ShieldedPool` tag with a `Sapling` variant.

```rust
struct CacheKey {
    tx_id: UnminedTxId,
    sighash: [u8; 32],
    pool: ShieldedPool,
}
```

**Both ID forms commit to the bundle**, which is the whole safety argument:

- `UnminedTxId::Witnessed` carries a `WtxId`, whose txid commits to effecting data
  and whose ZIP 244 authorizing-data digest commits to the proofs and signatures.
  The txid alone would not — that is CVE-2026-34377.
- `UnminedTxId::Legacy` is a v1-v4 transaction ID: the hash of the whole serialized
  transaction, which contains the Sapling proofs, `spendAuthSig`s and `bindingSig`
  directly. This is why v4 Sapling bundles can be cached without a witnessed ID.

The sighash stays in the key because it is not a function of the transaction alone:
the amounts and `scriptPubKey`s of spent transparent outputs enter it and come from
the verification context. The verifying keys stay out: each Orchard circuit era has
its own cache instance, and Sapling has one spend and one output verifying key for
all of history, so one cache covers it with no eras to keep apart.

The transaction verifier passes the ID it has already computed (`Request::tx_id()`),
so the lookup path adds no hashing and no allocation. `sapling::Item::new` now takes
that ID; the type is not part of the crate's public API (`primitives` is private and
only `sapling_prover` is re-exported), so this is not a public-API change.

### Benchmark accuracy

`worst_case_tx_verification` replays one workload of repeated mainnet transactions, so with a
process-wide cache every iteration after the first measured cache hits instead of verification —
`full_sapling_limit` reported **38.4 ms** where verifying those 300 Sapling spends actually takes
**283.1 ms**. Since that benchmark is what sizes the ZIP-1271 action limits and `max_block_bytes`
against worst-case block cost, and the worst case is a block of transactions the node has never
seen, it now clears the caches between iterations through a hidden
`clear_shielded_verification_caches`. This also restores the Orchard half of the measurement,
blind since #597.

### Metrics

One cache implementation now reports one set of series. The
`zakura.consensus.halo2.cache.{hit,miss,insert,evict,size}` metrics released in
v1.2.0 become `zakura.consensus.cache.{...}` with a `verifier` label, whose values
are `halo2_pre_nu6_2`, `halo2_nu6_2`, `halo2_nu6_3_onward` and `groth16_sapling`.
Those are the names the same verifiers already use in their batch metrics and flush
logs, so cache and batch series can be joined, and each Orchard circuit era now
reports its own hit rate instead of the three sharing one series.

## Testing

`crates/zakura-consensus/src/primitives/cache/tests.rs` — the shared key: the pool
tag separates the three bundles of one v6 transaction, the authorizing-data digest
separates two witnessed IDs sharing a txid, the sighash separates two verifications
of one bundle, and a legacy ID never collides with a witnessed one.

`crates/zakura-consensus/src/primitives/sapling/tests.rs` — the Sapling key over
real mainnet bundles:

- `v4_cache_key_changes_with_every_piece_of_authorizing_data` — a spend proof, a
  `spendAuthSig`, the binding signature and `value_balance`. These are exactly what
  `check_bundle` batches, and each one changes the v4 transaction ID.
- `v5_cache_key_changes_with_authorizing_data_its_txid_ignores` — the same
  mutations, carried into a v5 transaction, **asserting the v5 txid is unchanged**
  and the key still differs. That is the CVE-2026-34377 shape.
- deterministic keys, distinct transactions, sighash, and v4-vs-v5 carrying the same
  bundle.
- cache behaviour: skips the inner service on a repeat, never reuses a result across
  bundles, never remembers a failure, answers a hit when the inner service can no
  longer become ready, and propagates an inner readiness failure on a miss.
- `cached_verifier_still_rejects_a_corrupted_proof_after_verifying_a_valid_one` —
  the real batch-and-fallback stack behind a fresh cache. A valid mainnet bundle
  verifies and is cached, then the same transaction with a G1-swapped spend proof —
  well-formed, so it passes `check_bundle`'s synchronous checks and can only be
  caught by the batch validation a bad key would have skipped — must still be
  rejected.

- `a_bundle_verified_under_one_upgrade_is_not_reused_under_another` — real verification
  underneath: the same transaction, the same ID, another branch id, and the remembered `Ok` must
  not answer it. This is the end-to-end evidence for keying on the sighash.
- `clearing_the_cache_forces_reverification`, and in `halo2/tests.rs`
  `a_result_cached_under_one_era_is_not_visible_to_another`, which pins what binds an entry to the
  verifying key it was produced under.

`the_sapling_cache_is_reused_only_for_the_transaction_that_earned_it` and
`the_sapling_cache_distinguishes_a_v5_transaction_from_its_corrupted_twin` in
`transaction/tests.rs` — the production path, mirroring the Halo2 test #613 added. The second
drives a v5 transaction and its authorizing-data twin, whose txids are identical.
A real mainnet Sapling transaction verifies from the mempool, the block that mines
it is answered from the cache, the block one height past its expiry is still
rejected, and each authorizing-data mutation gets a fresh verification and fails.
Its fixture is a v4 transaction, so it is also the end-to-end evidence that the
legacy transaction ID covers what a witnessed ID's digest would have.

```
cargo test -p zakura-consensus --lib          # 203 passed
cargo test -p zakura-consensus --lib -- --test-threads=1
cargo clippy -p zakura-consensus --all-targets -- -D warnings
cargo check --workspace --all-targets
cargo fmt --all -- --check
./scripts/changelog.py check
```

## Review

Four independent adversarial reviews covered the key's completeness against everything
`check_bundle`/`add_bundle` reads, the identity plumbing at every call site, the cache service
under load and failure, and the tests. **None found a way for a cache hit to stand in for an
unverified bundle.** Notable confirmations:

- For v4, the legacy txid hashes `zcash_serialize(tx)` and the bundle is parsed by librustzcash
  from the output of that same function, so key and bundle derive from identical bytes.
- Dropping the earlier revision's explicit branch-id commitment is safe twice over: the branch id
  is inside the v4 sighash personalization, and `read_v4` never passes it to the Sapling parser,
  so it cannot change a parsed v4 bundle.
- A mismatched bundle/ID pair is not constructible: `Request::Block::tx_id()` recomputes from the
  carried transaction and never uses the caller-supplied `transaction_hash`, and `UnminedTx`'s
  `id` is private with no constructor that does not derive it.
- Era routing matches `BranchId::orchard_protocol_revision` and `BundleVersion::circuit_version`
  one for one, and Sapling's verification takes no era parameter at all.
- Mutation testing: dropping `tx_id`, the authorizing-data digest, the sighash or the pool from
  the key each fails the suite.

Their findings are fixed in this PR: the benchmark above; metrics reported after the cache lock is
released rather than allocating a label set inside it; the eviction queue evicting before pushing
so it stops doubling its allocation; corrected claims about the v4 sighash, the metric labels and
the cache's memory footprint; and the extra tests listed above.

A fifth review then re-audited the finished branch end to end and found nothing above LOW. It read
the `sapling_crypto` batch verifier to enumerate every field the verification actually consumes,
and traced each one to the component of the key that commits to it — per-spend `cv`, `anchor`,
`nullifier`, `rk`, `zkproof`, `spend_auth_sig`, per-output `cv`, `cmu`, `ephemeral_key`, `zkproof`,
plus `value_balance`, `binding_sig` and the sighash. It also confirmed what a hit does **not**
skip: duplicate-nullifier checks, ZIP 216 point-encoding canonicity, expiry and lock-time, scripts,
and the state-side anchor and nullifier checks all still run. Its two latent findings are addressed
here — `Request::tx_id` now documents the invariant its `TODO` must preserve, since implementing it
with a mined ID would reintroduce the CVE-2026-34377 shape, and `sapling::Item::new` is
`pub(crate)` like its Halo2 twin.

A `core-rust-readability` pass supplied the rest of the final commit: the `CachedItem` contract,
the `VerifiedBundles` invariant and its direct unit test, `VERIFIER_NAME` replacing a repeated
label literal, and `verifier_name` for the metrics-label parameter.

- [ZIP 239: Transaction Identifier Digests and Signature Validation](https://zips.z.cash/zip-0239)
- [ZIP 244: Transaction Identifier Non-Malleability](https://zips.z.cash/zip-0244)

### DIFF
--- a/crates/zakura-consensus/src/primitives/sapling.rs
+++ b/crates/zakura-consensus/src/primitives/sapling.rs
@@ -23,19 +23,15 @@ use zcash_protocol::value::ZatBalance;
 
 use crate::{error::TransactionError, BoxError};
 
-use super::cache::{CacheKey, CacheMetrics, Cached, CachedItem, ShieldedPool, CACHE_CAPACITY};
+use super::cache::{CacheKey, Cached, CachedItem, ShieldedPool, CACHE_CAPACITY};
 
 #[cfg(test)]
 mod tests;
 
-/// The metric names the Sapling cache reports under.
-const CACHE_METRICS: CacheMetrics = CacheMetrics {
-    hit: "zakura.consensus.sapling.cache.hit",
-    miss: "zakura.consensus.sapling.cache.miss",
-    insert: "zakura.consensus.sapling.cache.insert",
-    evict: "zakura.consensus.sapling.cache.evict",
-    size: "zakura.consensus.sapling.cache.size",
-};
+/// The `verifier` label the Sapling cache reports its metrics under.
+///
+/// It matches the label this verifier's batch duration metric already uses.
+const CACHE_VERIFIER_LABEL: &str = "groth16_sapling";
 
 /// Sapling prover containing spend and output params for the Sapling circuit.
 ///
@@ -291,8 +287,13 @@ type VerifierService = Cached<BatchFallbackService>;
 /// gossiped into the mempool does not have to be verified again when the block that mines it
 /// arrives. One cache covers all of Sapling: its spend and output verifying keys have never
 /// changed, so unlike Orchard there are no circuit eras to keep apart.
-pub static VERIFIER: Lazy<VerifierService> =
-    Lazy::new(|| Cached::new(batch_fallback_verifier(), CACHE_CAPACITY, CACHE_METRICS));
+pub static VERIFIER: Lazy<VerifierService> = Lazy::new(|| {
+    Cached::new(
+        batch_fallback_verifier(),
+        CACHE_CAPACITY,
+        CACHE_VERIFIER_LABEL,
+    )
+});
 
 /// Builds the uncached batching-and-fallback stack.
 fn batch_fallback_verifier() -> BatchFallbackService {
--- a/crates/zakura-consensus/src/primitives/sapling.rs
+++ b/crates/zakura-consensus/src/primitives/sapling.rs
@@ -68,11 +68,9 @@ impl Item {
     /// Creates a new [`Item`] from a Sapling bundle, its sighash, and its transaction's ID.
     ///
     /// `tx_id` must identify the transaction containing `bundle`, because the cache treats it as
-    /// determining the bundle — see [`Item::cache_key`]. The transaction verifier passes the
-    /// precomputed ID from its [`Request`](crate::transaction::Request), whose caller must
-    /// preserve this invariant.
-    ///
-    /// [`Item::cache_key`]: Item#method.cache_key
+    /// determining the bundle — see this type's [`CachedItem`] implementation. The transaction
+    /// verifier passes the precomputed ID from its [`Request`](crate::transaction::Request),
+    /// whose caller must preserve this invariant.
     pub fn new(
         bundle: Bundle<Authorized, ZatBalance>,
         sighash: SigHash,
@@ -96,8 +94,12 @@ impl CachedItem for Item {
     /// authorizing data. So unlike Orchard, which only exists in v5 and v6 transactions, Sapling
     /// caches v4 bundles too.
     ///
-    /// The sighash additionally commits to the amounts and scripts of spent transparent outputs,
-    /// which are supplied by the verification context and are not part of the transaction ID.
+    /// The sighash is keyed separately because it is not always a function of the transaction
+    /// alone. A v5 or v6 sighash also commits to the amounts and scripts of the spent transparent
+    /// outputs, which the verification context supplies. A v4 shielded sighash does not — it is
+    /// computed with no input index, so ZIP 143 and ZIP 243 leave the spent output out — but it
+    /// does commit to the consensus branch id of the block, which the transaction ID of a v4
+    /// transaction does not carry.
     ///
     /// The verifying keys are absent on purpose: Sapling has one spend and one output verifying
     /// key for all of history, so unlike Orchard it has no circuit eras to keep apart, and every
--- a/crates/zakura-consensus/src/primitives/sapling.rs
+++ b/crates/zakura-consensus/src/primitives/sapling.rs
@@ -17,11 +17,25 @@ use tower_batch_control::{Batch, BatchControl, RequestWeight};
 use tower_fallback::Fallback;
 
 use sapling_crypto::{bundle::Authorized, BatchValidator, Bundle};
-use zakura_chain::transaction::SigHash;
+use zakura_chain::transaction::{SigHash, UnminedTxId};
 use zcash_proofs::prover::LocalTxProver;
 use zcash_protocol::value::ZatBalance;
 
 use crate::{error::TransactionError, BoxError};
+
+use super::cache::{CacheKey, CacheMetrics, Cached, CachedItem, ShieldedPool, CACHE_CAPACITY};
+
+#[cfg(test)]
+mod tests;
+
+/// The metric names the Sapling cache reports under.
+const CACHE_METRICS: CacheMetrics = CacheMetrics {
+    hit: "zakura.consensus.sapling.cache.hit",
+    miss: "zakura.consensus.sapling.cache.miss",
+    insert: "zakura.consensus.sapling.cache.insert",
+    evict: "zakura.consensus.sapling.cache.evict",
+    size: "zakura.consensus.sapling.cache.size",
+};
 
 /// Sapling prover containing spend and output params for the Sapling circuit.
 ///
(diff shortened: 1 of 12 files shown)


crates/zakura-consensus/src/primitives/sapling.rs
```diff
@@ -46,12 +57,90 @@ pub struct Item {
...
-    /// Creates a new [`Item`] from a Sapling bundle and sighash.
-    pub fn new(bundle: Bundle<Authorized, ZatBalance>, sighash: SigHash) -> Self {
-        Self { bundle, sighash }
+    /// Creates a new [`Item`] from a Sapling bundle, its parse digest, and the sighash.
+    ///
+    /// `bundle` and `parse_digest` must come from the *same* parse — take them together from
+    /// [`SigHasher::sapling_bundle_and_parse_digest`], which is what makes that structural.
+    ///
+    /// [`SigHasher::sapling_bundle_and_parse_digest`]: zakura_chain::transaction::SigHasher::sapling_bundle_and_parse_digest
+    pub fn new(
```
p0mvn: `Item::new` accepts the bundle and `ParseDigest` independently, so the documented same-parse invariant is not structurally enforced. For V5, a valid bundle paired with a proof-corrupted transaction’s digest can warm that invalid transaction’s cache key because the proof change does not change the shielded sighash. Could this constructor instead accept an opaque bundle-plus-digest value produced by `SigHasher`, making mismatched pairing unrepresentable?

crates/zakura-consensus/src/primitives/sapling.rs
```diff
@@ -46,12 +57,90 @@ pub struct Item {
...
+            sighash,
+            parse_digest,
+        }
+    }
+}
+
+impl CachedItem for Item {
+    /// Returns a key that determines every input to this item's bundle verification.
+    ///
+    /// [`Cached`] reuses a previous `Ok` result whenever this key matches, so the key must
```
p0mvn: nit: please consider reducing comment verbosity for human readability

p0mvn (approved):
LGTM

But I was also wondering if the complexity is worth it if, according to PR description, Sapling's contribution is 8.8%.

The implementation differs from Orchard so it's not a direct clone / there is a cost.

Non-blocking.

crates/zakura-consensus/src/primitives/sapling.rs
evan-forbes: sure thing, added a note to my agent and claude files to help avoid letting these slip through in the future, and will go apply to my other PRs as well

evan-forbes:
> But I was also wondering if the complexity is worth it if, according to PR description, Sapling's contribution is 8.8%.

yeah I agree here. I'm very ok w/ just not touching this since it only addresses an increasingly small percent yet incurs the same level of risk of #597 

cc @ValarDragon

crates/zakura-consensus/src/primitives/sapling.rs
evan-forbes: https://github.com/zakura-core/zakura/pull/600/commits/34d58c0269ba2a0edcca0c7d717a69ca1f4475a8

ValarDragon:
I'd like to do both, but under the same CacheKey of tx-ID/wtx-ID matching

p0mvn (approved):

crates/zakura-consensus/src/transaction.rs
```diff
@@ -277,9 +277,17 @@ impl Request {
...
+    /// be answered from the cached verification of a differently-signed twin. Both arms below
+    /// recompute the ID, and the `Request::Block::transaction_hash` field the caller supplies is
+    /// deliberately not used.
     pub fn tx_id(&self) -> UnminedTxId {
         match self {
-            // TODO: get the precalculated ID from the block verifier
+            // TODO: get the precalculated ID from the block verifier. It must be a full unmined
+            // ID recomputed from the transaction, for the reason given above.
             Request::Block { transaction, .. } => transaction.unmined_id(),
             Request::Mempool { transaction, .. } => transaction.id(),
```
ValarDragon, line 292: Naming is odd here, should get renamed

ValarDragon:
This PR should have blocked on two approvals - its consensus critical

PR #602, opened 2026-08-08
feat(chain): public NoteCommitmentTree::from_frontier for Sapling and Orchard

### LINKED ISSUE #601: feature: public NoteCommitmentTree::from_frontier for Sapling and Orchard
**Motivation**

`zindex` (an indexer built on `zakura-chain`) derives note commitment trees in parallel batches. Each batch's closing frontier comes out of the machinery this crate already exposes — `parallel::batch_frontier::append_batch_with_subtree` is public and returns a bare `Frontier` (`src/parallel/batch_frontier.rs:309`) — but there is no public way to wrap a `Frontier` back into a `NoteCommitmentTree`. The supported constructors are `Default` (empty) and `From<Vec<ExtractedNoteCommitment>>`, which re-appends every leaf (`src/sapling/tree.rs:559-574`).

So we advance the authoritative tree by re-hashing every batch's leaves a second time through `append_batch`. On a mainnet genesis backfill that second walk is ~651 s of duplicated hashing (stage timer `derive_finalize_ms=651,079` of `elapsed_ms=3,130,380`). The walk runs concurrently beside the main append pass, so the direct wall saving is smaller than 651 s and unmeasured — the win is deleting the duplicated CPU work and the speculative-placement machinery that exists only to hide it.

The tree is structurally a frontier plus a root cache: `inner: Frontier<sapling_crypto::Node, MERKLE_DEPTH>` (`src/sapling/tree.rs:175`) and `cached_root` (`:192`); Orchard mirrors it (`src/orchard/tree.rs:373`). Your own tests already build trees exactly this way — `Self { inner: Frontier::from_parts(...), cached_root: Default::default() }` (`src/sapling/tree.rs:614-621`, `725-730`, `774-779`, `802-807`). The ask is to export that as an API.

The one external route that exists today is hand-authoring `LegacyNoteCommitmentTree`'s serde payload through `#[serde(from = ...)]` (`src/sapling/tree.rs:155-157`; `src/sapling/tree/legacy.rs:53-60`, `100-127`). We deliberately refuse it: it authors consensus-critical state through a panic-validated legacy DB format (`legacy.rs:88-93`) with no stability contract. This constructor would replace that shape-coupled workaround with a typed API.

**Specifications**

```rust
/// Builds a tree from an already-validated frontier. The root cache
/// starts empty and is recomputed on first use.
pub fn from_frontier(frontier: Frontier<sapling_crypto::Node, MERKLE_DEPTH>) -> Self {
    Self { inner: frontier, cached_root: Default::default() }
}
```

~6 lines each in `src/sapling/tree.rs` and `src/orchard/tree.rs` (parameter type `Frontier<Node, MERKLE_DEPTH>` there); Ironwood inherits via the `ironwood::tree` re-export (`src/ironwood.rs:8-11`). Infallible: frontier well-formedness is enforced by `incrementalmerkletree::frontier::Frontier::from_parts` returning `Result` before the value can exist, and capacity by the `DEPTH = 32` const generic. No hashing, no consensus logic. One ergonomics note: `MERKLE_DEPTH` is `pub(super)` (`src/sapling/tree.rs:40`) — external callers can name `Frontier<Node, 32>`, or the const could be made `pub` (+1 line). An optional symmetric companion is a public `frontier()` accessor, making the constructor the inverse of an existing getter rather than of the `to_rpc_bytes` round-trip (`src/sapling/tree.rs:509-520`), which currently serves frontiers out with no way back in.

Trust surface: identical to your own test literals — validity lives in the already-validated `Frontier` parameter, and verify-then-install stays the caller's job, matching the pattern zakura itself uses when installing VCT final frontiers root-verified (`zakura-state/src/service/finalized_state/commitment_aux.rs:27-43`). On our side every batch frontier is cross-checked against the serially-advanced tree before use (73,952 comparisons, 0 disagreements over the full chain) and against consensus roots (pre-Heartwood `hashFinalSaplingRoot`).

**Testing**

Round-trip property per pool: extract a non-trivial tree's frontier, `from_frontier` it, assert root and position equality with the original — the same sequential-equivalence pattern `append_batch`'s tests already use. The four in-crate test literals can also switch to the public constructor.

**Related Work**

#552 (historical trees by verified replay) and #593 (subtree-root serving) touch adjacent machinery; neither provides a constructor, and this does not propose replacing #552's replay — that replay *is* the verification, this is a constructor for a frontier the caller has already verified. Additive API on a published crate → minor semver bump. Happy to contribute the PR with the changelog fragment.

---

maxdesalle:
## Motivation

Closes #601. `parallel::batch_frontier::append_batch_with_subtree` is public and
returns a bare `Frontier`, but there is no public way to wrap a `Frontier` back
into a `NoteCommitmentTree` — the supported constructors are `Default` (empty)
and `From<Vec<ExtractedNoteCommitment>>`, which re-appends every leaf. A
downstream consumer that derives frontiers in parallel batches must therefore
re-hash every batch's leaves a second time purely to obtain the authoritative
tree type (~11 CPU-minutes of duplicated hashing per mainnet genesis walk in
zindex's case, per the issue).

## Solution

Adds `NoteCommitmentTree::from_frontier` to the Sapling and Orchard trees — the
same two-field construction the in-crate tests already use (`Self { inner,
cached_root: Default::default() }`). Ironwood inherits via the `ironwood::tree`
re-export. Infallible by construction: a malformed frontier cannot exist
(`Frontier::from_parts` validates before a value can be constructed), capacity
is enforced by the `MERKLE_DEPTH` const generic, and the root cache starts
empty. Verifying that a frontier belongs to the chain remains the caller's
responsibility, as with every other way of obtaining one — matching the
verify-then-install pattern the finalized state already uses for VCT final
frontiers.

## Testing

`cargo test -p zakura-chain --release tree`: 47 passed, 0 failed, including the
two new round-trip tests (`from_frontier_round_trips_root_position_and_appends`
per pool): build a tree through the ordinary append path, rebuild it from its
own frontier alone, assert root/count/position equality, then append one more
commitment to both and assert the roots still agree.

## Changelog

`docs/changelog/unreleased/<PR>.md` added after opening this draft.

---
🤖 Drafted with [Claude Code](https://claude.com/claude-code) at @maxdesalle's direction, per the discussion on #601.

### DIFF
--- a/crates/zakura-chain/src/orchard/tree.rs
+++ b/crates/zakura-chain/src/orchard/tree.rs
@@ -391,6 +391,28 @@ pub struct NoteCommitmentTree {
 }
 
 impl NoteCommitmentTree {
+    /// Builds a tree whose state is exactly the given frontier.
+    ///
+    /// A frontier is the rightmost path of an append-only Merkle tree, and is
+    /// all the state this tree keeps: this constructor cannot produce a tree
+    /// that the append path could not have produced, because a malformed
+    /// frontier cannot exist
+    /// ([`Frontier::from_parts`](incrementalmerkletree::frontier::Frontier::from_parts)
+    /// validates before a value can be constructed) and capacity is enforced
+    /// by the `MERKLE_DEPTH` const generic. Verifying that the frontier
+    /// belongs to the chain remains the caller's responsibility, as it is for
+    /// every other way of obtaining one.
+    ///
+    /// The root cache starts empty and is recomputed on first use.
+    pub fn from_frontier(
+        frontier: incrementalmerkletree::frontier::Frontier<Node, MERKLE_DEPTH>,
+    ) -> Self {
+        Self {
+            inner: frontier,
+            cached_root: Default::default(),
+        }
+    }
+
     /// Adds a note commitment x-coordinate to the tree.
     ///
     /// The leaves of the tree are actually a base field element, the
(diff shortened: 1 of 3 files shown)


ValarDragon:
@v12sec pls review

ValarDragon (approved):
Sweet, thank you!

maxdesalle:
Thank you!

PR #608, opened 2026-08-10
fix(release): clarify hotfix mode and harden capability probe

p0mvn:
## Summary

- make the GitHub CLI capability probe reliable under `set -o pipefail`
- require operators to choose and record the hotfix release mode before opening the public PR
- show mode-specific T-0 commands and manual fallbacks in the hotfix checklist

## Motivation

The v1.1.1 emergency release exposed two process hazards: `grep -q` could falsely reject a supported `gh` version after SIGPIPE, and a branch rebased onto `main` was initially treated as a release-tag-base hotfix. Both failures occurred before or around T-0 and made the intended release path ambiguous.

This change keeps mode selection explicit and ensures the checklist remains actionable for both main-base and previous-release-base hotfixes.

## Test plan

- [x] `bash -n scripts/release-t0.sh`
- [x] `shellcheck scripts/release-t0.sh`
- [x] capability probe succeeds under `bash -o pipefail`
- [x] Markdown lint passes for both updated process documents
- [x] checklist documents the main-base and previous-release-base publish paths

## Changelog

Internal release tooling and process documentation only; no user-visible Zakura behavior change.

## AI disclosure

AI assistance was used to implement and validate these release-process safeguards. The maintainer remains responsible for review and merge.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
@@ -96,13 +96,19 @@ allows `hotfix/v*`).
       completed steps):
 
   ```sh
+  # Main base: squash-merge the public PR, then release from main.
+  ./scripts/release-t0.sh publish --hotfix --tag v<version> \
+      --mode main --pr <N> --head-sha <final-commit>
+
+  # Previous-release base: push and release from the hotfix branch.
   ./scripts/release-t0.sh publish --hotfix --tag v<version> \
       --mode branch --head-sha <final-commit>
   ```
 
-  Manual fallback: push `hotfix/v<version>`, dispatch `Create release`
-  **from the branch** with the exact tag, and verify each step landed before
-  the next.
+  Manual fallback: for a main-base hotfix, squash-merge the public PR and
+  dispatch `Create release` from `main`; for a previous-release-base hotfix,
+  push `hotfix/v<version>` and dispatch from that branch. Use the exact tag
+  and verify each step landed before the next.
 - [ ] Approve the `release` environment deployment when the script announces
       it (right commit? right tag?). The workflow publishes a complete
       pre-release and creates the protected tag; the tag push starts Docker
--- a/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
@@ -35,10 +35,14 @@ canary soak, announcement) apply on top of this checklist.
       branch cut from the previous release. Do not rebase the branch onto a
       different base after choosing the mode.
 - [ ] Make the required changes, minimal and with tests.
-- [ ] For a public (non-embargoed) hotfix: open this PR from the branch with
+- [ ] For a public (non-embargoed) previous-release-base hotfix
+      (`--mode branch`): open this PR from the branch with
       `&template=hotfix-release-checklist.md` in the compare URL, and add the
-      `do-not-merge` label — the release is dispatched from the branch itself,
-      and the PR is merged (as a merge commit) only after the release.
+      `do-not-merge` label. Release from the branch, then forward-merge this PR
+      after the release.
+- [ ] For a main-base hotfix (`--mode main`): do not add `do-not-merge`.
+      The T-0 orchestrator squash-merges this PR before dispatching the release
+      from `main`.
 - [ ] Add the `A-release` and `C-exclude-from-changelog` labels **when
       opening the PR**: the changelog fragment check only accepts a
       fragment-consuming release branch in release-PR mode, and relabeling
--- a/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
@@ -30,6 +30,10 @@ canary soak, announcement) apply on top of this checklist.
       those names belong to the regular release process, and disjoint
       namespaces are what prevent an embargo-blind collision with it (see
       the process doc's branch namespace rule).
+- [ ] Choose and review the release mode before opening the public PR: use
+      `--mode main` for a branch cut from `main`, or `--mode branch` for a
+      branch cut from the previous release. Do not rebase the branch onto a
+      different base after choosing the mode.
 - [ ] Make the required changes, minimal and with tests.
 - [ ] For a public (non-embargoed) hotfix: open this PR from the branch with
       `&template=hotfix-release-checklist.md` in the compare URL, and add the
@@ -91,14 +95,14 @@ allows `hotfix/v*`).
       and verifies each step's target state (resumable — re-runs skip
       completed steps):
 
-      ```sh
-      ./scripts/release-t0.sh publish --hotfix --tag v<version> \
-          --mode branch --head-sha <final-commit>
-      ```
+  ```sh
+  ./scripts/release-t0.sh publish --hotfix --tag v<version> \
+      --mode branch --head-sha <final-commit>
+  ```
 
-      Manual fallback: push `hotfix/v<version>`, dispatch `Create release`
-      **from the branch** with the exact tag, and verify each step landed
-      before the next.
+  Manual fallback: push `hotfix/v<version>`, dispatch `Create release`
+  **from the branch** with the exact tag, and verify each step landed before
+  the next.
 - [ ] Approve the `release` environment deployment when the script announces
       it (right commit? right tag?). The workflow publishes a complete
       pre-release and creates the protected tag; the tag push starts Docker
--- a/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
@@ -8,32 +8,35 @@ assignees: ""
 
 A hotfix release should only be created when a bug or critical issue is
 discovered in an existing release, and waiting for the next scheduled release
-is impractical or unacceptable. It ships the fix on top of the release
-operators are already running, instead of everything unreleased on `main`.
+is impractical or unacceptable. Prefer shipping the fix on top of the release
+operators are already running; use `main` only when its unreleased changes are
+safe to include.
 
 For an **embargoed security fix**, follow
 [`docs/security-hotfix-release.md`](https://github.com/zakura-core/zakura/blob/main/docs/security-hotfix-release.md):
-all preparation below happens in the private staging repo, this PR is opened
-at forward-merge time, and the extra steps there (advisory, dress rehearsal,
-canary soak, announcement) apply on top of this checklist.
+all preparation below happens in the private staging repo. Open this PR at T-0
+for a main-base hotfix, or at forward-merge time for a previous-release-base
+hotfix. The extra steps there (advisory, dress rehearsal, canary soak,
+announcement) apply on top of this checklist.
 
 ## Create the Hotfix Branch
 
-- [ ] Cut `hotfix/v<version>` from the tag of the release being fixed (not
-      from `main`). The branch **must** be named for the exact tag it will
-      release — the `Create release` workflow refuses to tag `v<version>`
-      from any other branch. The `hotfix/v*` ruleset blocks deletion and
-      force-pushes.
-- [ ] Embargoed main-base hotfix: the T-0 public PR branch is **also** named
-      `hotfix/v<version>` (based on `main`, pushed at T-0). Whatever the
-      base, the hotfix process never pushes `release/v*` or `bump-v*` —
+- [ ] Choose and review the release mode before creating the hotfix branch:
+      use `--mode main` when the unreleased changes on `main` are safe to
+      include, or `--mode branch` to release only the fix on top of the
+      previous release. Record the choice and do not rebase the branch onto a
+      different base.
+- [ ] Create `hotfix/v<version>` from `main` for `--mode main`, or from the
+      previous release tag for `--mode branch`. The branch **must** be named
+      for the exact tag it will release — the `Create release` workflow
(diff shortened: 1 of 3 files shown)


scripts/release-t0.sh
```diff
@@ -196,7 +197,7 @@ check_globals() {
   gh auth status >/dev/null 2>&1 \
     || die "tools" "gh auth status" "not authenticated" "authenticated" \
         "run: gh auth login"
-  gh pr merge --help 2>/dev/null | grep -q -- '--match-head-commit' \
+  gh pr merge --help 2>/dev/null | grep -- '--match-head-commit' >/dev/null \
```
p0mvn, line 199: Note to reviewer: AI would trip over this and need to search for workarounds

.github/PULL_REQUEST_TEMPLATE/hotfix-release-checklist.md
```diff
@@ -30,6 +30,11 @@ canary soak, announcement) apply on top of this checklist.
       those names belong to the regular release process, and disjoint
       namespaces are what prevent an embargo-blind collision with it (see
       the process doc's branch namespace rule).
+- [ ] Choose the release mode before opening the public PR and verify it with
+      `release-t0.sh preflight`: use `--mode main` when the hotfix contains
+      current `main`, or `--mode branch` only when it was cut from the previous
+      release. If a main-base branch falls behind `main`, update and review it
+      before T-0.
```
p0mvn: Note to reviewer: my AI was strongly encouraging me to release from the branch during the release but not before.

Thus, this added time overhead during the release.

By making an explicit step before the release, the release manager can recall the process and make the appropriate selection in advance

p0mvn:
@cursor review

p0mvn:
@cursor review

scripts/release-t0.sh
```diff
@@ -428,12 +429,66 @@ preflight_expected_head() {
...
+  local main_sha relation compare_status
+  main_sha=$(gh api "repos/${REPO}/branches/main" --jq '.commit.sha')
+
+  if git -C "$repo_root" cat-file -e "${main_sha}^{commit}" 2>/dev/null \
+      && git -C "$repo_root" cat-file -e "${HEAD_SHA}^{commit}" 2>/dev/null; then
+    if git -C "$repo_root" merge-base --is-ancestor "$main_sha" "$HEAD_SHA"; then
+      relation="includes-main"
+    else
+      relation="excludes-main"
+    fi
```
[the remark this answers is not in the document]
p0mvn: [21f7fb8](https://github.com/zakura-core/zakura/pull/608/commits/21f7fb89b9348fe7aa904829d7f2fea5a48afe3f)

p0mvn:
@cursor review

scripts/release-t0.sh
```diff
@@ -428,12 +441,116 @@ preflight_expected_head() {
...
+    esac
+    merge_base_sha=$(git -C "$repo_root" merge-base "$main_sha" "$HEAD_SHA" \
+      2>/dev/null || true)
+  fi
+
+  if [ -z "$relation" ] \
+      || { [ "$MODE" = "branch" ] && [ -z "$merge_base_sha" ]; }; then
+    compare_json=$(gh api "repos/${REPO}/compare/main...${HEAD_SHA}" \
+      2>/dev/null || printf '')
+    compare_status=$(jq -r '.status // "unavailable"' <<<"$compare_json")
```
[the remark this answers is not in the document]
p0mvn: [91a3d57](https://github.com/zakura-core/zakura/pull/608/commits/91a3d57d3af3f03bbebe6b44d3bf310362a7d4ad)

p0mvn:
@cursor review

scripts/release-t0.sh
```diff
@@ -428,12 +441,117 @@ preflight_expected_head() {
   fi
 }
 
+preflight_hotfix_mode() {
```
evan-forbes: given the parsing and many checks in this function I pointed claude at it for correctness and robustness specifically, then rechecked each, and had a bit that still seemed plausible. not blocking, but might be worth passing to yours.

```
1. Branch mode aborts after the forward-merge — scripts/release-t0.sh:525

Once hotfix/vX.Y.Z lands on main with a true merge commit, HEAD_SHA becomes an ancestor of main, so merge-base(main, HEAD_SHA) collapses to HEAD_SHA itself and never equals base_sha. Fixture:

--- T-0, before the forward-merge --------  relation=excludes-main  merge_base==base_sha  PASS
--- after a true-merge forward-merge -----  relation=excludes-main  merge_base==HEAD_SHA  DIE @525
--- after an --allow-squash forward-merge-  relation=excludes-main  merge_base==base_sha  PASS

Re-running preflight or publish for that tag then dies with "the branch contains main commits after the declared release base", which describes the opposite of what happened. Note the asymmetry: the squash path survives, so the check fails only on the forward-merge method cmd_forward_merge actively recommends (scripts/release-t0.sh:965-972).

2. Main mode aborts once main advances past the merge — scripts/release-t0.sh:460

[ "$main_sha" = "$merge_sha" ] || die compares against current main by equality. The pre-existing merged-PR path at scripts/release-t0.sh:623 deliberately uses ancestor_of_main instead. Fixture: after the squash merge and one later commit on main, :460 dies while ancestor_of_main passes.

To be clear, the equality is correct while the tag does not exist yet — dispatch builds the current head of main, which is exactly the hazard scripts/release-t0.sh:671-677 guards. The problem is that it also fires once the tag exists, when step_merge_or_push, step_dispatch, and step_watch all skip and nothing is dispatched. In that state the abort is a false positive that blocks step_verify_publish.

Findings 1 and 2 share a fix: resolve TAG_COMMIT before preflight_hotfix_mode, then skip the ancestry assertions (or downgrade them to warn) when the tag already exists. Once the tag exists the mode decision is settled and there is nothing left to protect.

One constraint on that fix: the HEAD_SHA rewrite at scripts/release-t0.sh:463 was added by bb63d8ca precisely to make resume work, and scripts/release-t0.sh:334 depends on it. So this means hoisting the TAG_COMMIT lookup, not moving preflight_tag_and_release ahead of preflight_hotfix_mode.

3. An orphaned base tag produces a misleading abort — scripts/release-t0.sh:525

If the previous release was branch-base and forward-merged with --allow-squash, its tagged commit is not an ancestor of main. The next hotfix, cut correctly from that tag, then fails the merge-base equality. Fixture:

v1.1.1 is NOT an ancestor of main (orphaned by the squash)
merge_base=eb61ce048a1f  base_sha=fb9aa04f3970  =>  DIE @525

The operator did nothing wrong, and the message blames them for main commits that are not there. An explicit ancestor_of_main "$base_sha" probe with its own message would name the real cause.

Related: scripts/release-t0.sh:984 warns that an orphaned tag breaks make pre-release BASE_TAG=. This PR adds a second consumer of that ancestry, so that warning is now incomplete regardless of what happens to :525.

4. --head-sha has no shape validation

This is a pre-existing gap, not something the PR introduces, but the PR widens it. --tag and --base-tag both get regex checks at scripts/release-t0.sh:214 and scripts/release-t0.sh:218; --head-sha gets none at scripts/release-t0.sh:1045. The value now feeds three incompatible consumers:

- string equality against 40-character remote SHAs (:334, :430, :533, :645),
- git cat-file and git merge-base, which accept abbreviations and refnames (:437 pre-existing, :473, :475, :481 new),
- a URL path segment in compare/main...${HEAD_SHA}, which accepts refnames (:487, the first such use).

An abbreviated SHA or a branch name passes every ancestry check and fails every equality check. A ^[0-9a-f]{40}$ check in check_globals, or normalization via gh api repos/${REPO}/commits/${HEAD_SHA} --jq .sha, closes it.

5. The pin rewrite is silent, and only hotfixes get it — scripts/release-t0.sh:463

HEAD_SHA="$merge_sha" is a good ergonomic: it repairs the tag comparison at :334 and run adoption at :370 on resume. It also fires on every merged-PR resume, since the squash commit is never the PR head the operator pinned. The ok message at :464 does not mention the substitution. A warn when the value changes would make it visible.

The more interesting half is the asymmetry. scripts/release-t0.sh:445 returns early for non-hotfix runs, so a regular release resuming after its tag exists still dies at :334 with "tags are immutable and never reused" — the exact false positive this commit fixes for hotfixes. Worth lifting the rewrite out of the hotfix guard.

6. A failed compare reads as "unavailable" — scripts/release-t0.sh:487

2>/dev/null || printf '{}' maps every failure to the same empty object: HTTP 404, a 5xx, an expired token, and a dropped connection all become compare_status=unavailable and die at :497 telling the operator to fetch refs they already have. gh_api_optional exists specifically so that "network errors never read as 'not found'" (scripts/release-t0.sh:154-155) and would preserve the distinction here.

7. A question about the grep -q change — scripts/release-t0.sh:203

I could not reproduce the failure this hardens against. At gh 2.97.0 I ran both forms 300 times under set -euo pipefail with zero failures. gh pr merge --help emits 1916 bytes, well under the 64 KiB pipe buffer, so gh writes its entire output and exits before grep -q can close the pipe on it; SIGPIPE cannot occur at that size.

The change is harmless either way, so this is not a blocker. But if you hit a real failure here, the reproducer is worth a comment, because the same pattern remains at scripts/release-t0.sh:793 and someone will eventually ask why only one was changed. That instance emits roughly 12 bytes, so it is even further from the threshold.

8. No override on the main-mode ancestry rule — scripts/release-t0.sh:538

Requiring the PR head to contain current main is stricter than GitHub, which squashes a behind-by-N branch onto the tip without complaint. It is a defensible integrity rule: with a squash merge, a branch that does not contain main produces a commit whose tree nobody built or reviewed. Worth confirming it is intentional, since under embargo one unrelated merge to main forces a branch update and a full CI re-run at T-0, and this abort has no --allow-* escape hatch.

```

scripts/release-t0.sh
```diff
@@ -428,12 +441,117 @@ preflight_expected_head() {
...
+    pr_state=$(jq -r '.state' <<<"$pr_json")
+    if [ "$pr_state" = "MERGED" ]; then
+      merge_sha=$(jq -r '.mergeCommit.oid // empty' <<<"$pr_json")
+      [ -n "$merge_sha" ] \
+        || die "preflight: hotfix mode" "PR #${PR_NUM} merge commit" \
+            "missing" "the squash-merge commit" "inspect the merged PR"
+      [ "$main_sha" = "$merge_sha" ] \
+        || die "preflight: hotfix mode" "current main" "$main_sha" "$merge_sha" \
+            "main advanced after PR #${PR_NUM} merged; review the new commits"
+      HEAD_SHA="$merge_sha"
```
czarcas7ic: Could we keep the original `HEAD_SHA` pinned until we've checked that this is actually the hotfix PR? This lookup only asks for `state` and `mergeCommit`, and the merged path in `step_merge_or_push` returns before the base and head checks below.

https://github.com/zakura-core/zakura/blob/bb63d8caa304d7c9b27bda5be54b9cb67079fd13/scripts/release-t0.sh#L451-L465

https://github.com/zakura-core/zakura/blob/bb63d8caa304d7c9b27bda5be54b9cb67079fd13/scripts/release-t0.sh#L609-L648

A typo in `--pr` can therefore be accepted if that PR happens to be the current `main` commit, and the script will release the wrong commit. I was able to reproduce that against the current repo state.

I asked AI to break down the failure mode more, which makes sense to me:

```
1. Your real hotfix head was H.
2. The hotfix was merged.
3. Another unrelated PR was then merged as current main.
4. While resuming, you accidentally pass that latest PR’s number but still pass --head-sha H.

The script should say:
> That PR does not contain the head SHA you supplied.

Instead, it sees that the wrong PR is merged at current main, replaces H with that PR’s merge commit, and continues.

The later code does load the PR’s base, branch name, and head SHA, but for a merged PR it returns before checking any of them. That is why the later checks do not save us.
```

Also, accepting a `--head-sha` equal to either the PR's `headRefOid` or its `mergeCommit.oid` keeps both resume paths working while rejecting a mistyped PR number.

scripts/release-t0.sh
```diff
@@ -428,12 +441,117 @@ preflight_expected_head() {
   fi
 }
 
+preflight_hotfix_mode() {
+  [ "$HOTFIX" = 1 ] || return 0
+
+  local main_sha relation="" compare_status merge_base_status
+  local shallow_repository merge_base_sha="" compare_json="" base_sha=""
+  main_sha=$(gh api "repos/${REPO}/branches/main" --jq '.commit.sha')
```
czarcas7ic: Can we carry this `main_sha` into the merge step? `--match-head-commit` only protects the PR head. If `main` advances after this preflight, GitHub can squash the same head onto the new tip; the post-merge check only confirms that the squash commit is now `main`.

https://github.com/zakura-core/zakura/blob/bb63d8caa304d7c9b27bda5be54b9cb67079fd13/scripts/release-t0.sh#L447-L449

https://github.com/zakura-core/zakura/blob/bb63d8caa304d7c9b27bda5be54b9cb67079fd13/scripts/release-t0.sh#L652-L678

Because this repo doesn't require PRs to be up to date, that can include an intervening commit that wasn't in the reviewed hotfix head. Checking the squash commit's first parent against this SHA before dispatch would close the gap.

I see that the script does ask "Is the merge queue frozen?" before merging, but that is a prompt rather than a check (`--yes` skips it), and a frozen queue does not stop a manual merge, so verifying the squash commit's parent would cover both cases.

p0mvn:
Thanks for taking a look. Upon review, I decided to cut the scope to documentation improvements and fixing the `grep -q` failure, removing the edge-case validation for the hot-fix path.

We can see how problematic it is in the next release and re-add it only if the need is confirmed. Would like to avoid adding clear complexity when the value isn't confirmed to be major.

p0mvn:
@cursor review

p0mvn:
Main-mode freshness is deliberately not enforced for simplicity.

czarcas7ic (approved):

PR #596, opened 2026-08-07
fix(rpc): require mTLS for remote indexer RPC

ValarDragon:
## Motivation

The indexer gRPC server accepted unauthenticated plaintext connections on any configured interface, while `TrustedChainSync` treated received blocks as semantically verified. A publicly reachable or intercepted connection could therefore expose node state, consume streaming resources, or violate the syncer trust assumption.

## Solution

- Require mutual TLS for non-loopback indexer listeners.
- Keep plaintext compatibility only for loopback server and client addresses.
- Configure tonic to authenticate server and client certificates through dedicated trust roots.
- Add production-path handshake tests covering missing and trusted client certificates.
- Bump `zakura-rpc` to 7.0.0 for the breaking configuration and server initialization API changes.

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy -p zakura-rpc --all-targets -- -D warnings`
- `cargo test -p zakura-rpc --lib` (103 passed, 1 ignored)
- `cargo check -p zakura --tests`
- `cargo semver-checks --package zakura-rpc --default-features`
- `./scripts/changelog.py check`
- `npx --yes markdownlint-cli docs/changelog/unreleased/596.md --config .github/.markdownlint.yaml`
- `git diff --check`

The mTLS integration test verifies that a client without a certificate cannot call the service and that a client signed by the configured CA reaches RPC request validation. Separate tests reject non-loopback plaintext server and client configurations.

## Changelog

Added `docs/changelog/unreleased/596.md` under `Security`.

### Specifications & References

- [tonic transport TLS configuration](https://docs.rs/tonic/latest/tonic/transport/index.html)

---

> [!NOTE]
> **High Risk**
> Changes authentication and transport for indexer RPC and the trusted chain sync path (security-critical); operators must configure mTLS for non-loopback deployments and adopt the breaking indexer init/client config API.
> 
> **Overview**
> **Indexer RPC transport and trust**
> 
> Non-loopback indexer listeners must configure **`rpc.indexer_tls`** (server cert/key plus client CA); startup rejects plaintext on public interfaces. Loopback plaintext remains for compatibility, with docs warning that loopback does not authenticate peers. The indexer server enables tonic **`tls-aws-lc`**, loads mTLS PEMs, enforces a **10s** TLS handshake timeout, and limits to **64** concurrent TCP connections and **64** HTTP/2 streams per connection via a semaphore-wrapped accept path.
> 
> **Sync client API**
> 
[...]

PR #612, opened 2026-08-11
feat(rpc): streaming GetBlockRange for finalized raw blocks on the indexer gRPC service

### LINKED ISSUE #611: feature: streaming GetBlockRange for finalized raw blocks on the indexer gRPC service
## Motivation

zindex backfills mainnet genesis-to-tip (~3.44M blocks) through
`zebra.indexer.rpc.Indexer`, one unary `GetBlock` per block, pipelined at
concurrency 64. Serving each request, zakura reads the block's raw consensus
bytes from RocksDB, fully deserializes every transaction (`FromDisk` — the
parse the in-tree TODO at `disk_format/block.rs:296` already flags as heavy),
then re-serializes back to the exact bytes it started from, once per round
trip.

That per-block cost is what bounds supply. Measured against a local archive
zakurad (ratios transfer, absolutes won't): a deliberately cheap consumer —
starved 95–98% of the time, so the number describes the server — floors at
**1,267 blocks/s** in the dense band (height ~1.87M) vs 15–16k in the
fastest bands, and aggregate supply saturates at ~2.1–2.7× one process by
~4 concurrent readers. A direct in-process read + re-serialize of the
same heights runs ~3× the served fetch rate. For a full backfill this makes
the node the eventual floor.

The seam is small: zakura-state already serves block *ranges* internally
(`ReadRequest::BlocksByHeightRange`, used by the P2P sync driver), and a
private `ZakuraDb::raw_block()` already returns the raw header and raw tx
bytes — the wire payload is just header ‖ CompactSize(count) ‖ txs. It is
also a step toward standardizing how data leaves a node: raw finalized
history as one stream, beside the existing change streams that carry
corrections.

## Proposal

One additive method on the existing service (three of its four methods
already server-stream; `GetBlock` is the only unary one):

```proto
message BlockRangeRequest {
  uint32 start_height = 1;
  uint32 end_height = 2;   // inclusive; INVALID_ARGUMENT if < start
};

// height + the Zcash consensus serialization (same encoding as BlockAndHash.data)
message BlockAndHeight {
  uint32 height = 1;
  bytes data = 2;
};

service Indexer {
  rpc GetBlockRange(BlockRangeRequest) returns (stream BlockAndHeight);
}
```

Semantics: finalized-only (clamped at the finalized tip — reorg-stable by
construction); ascending; stops before the first pruned/missing body
(`BlocksByHeightRange`'s own documented convention); backpressure via the
service's existing bounded-buffer streaming pattern; served from the stored
raw bytes, skipping the parse → re-serialize round trip; no auth changes
(same opt-in `rpc.indexer_listen_addr` posture).

Out of scope: tip-following and non-finalized blocks (existing streams cover
them), changes to `GetBlock`, compression, performance guarantees.

## Testing

Same pattern as the service's existing `vectors.rs` tests: real tonic server
+ client; assert ascending order, clamp at tip, `INVALID_ARGUMENT` on
end < start, stop-before-missing on a gap, and byte-for-byte equality of
`data` with `GetBlock`'s answer at the same heights.

maxdesalle:
Closes #611.

## Motivation

zindex backfills Zcash mainnet genesis-to-tip (~3.44M blocks, ~100 GiB)
through the `zebra.indexer.rpc.Indexer` service, one unary `GetBlock` per
block, pipelined at concurrency 64. In the regimes where per-block serving
is the wall (all numbers measured against a local archive zakurad with the
chain in page cache; ratios should transfer, absolutes won't):

- **Dense-band supply trough.** With a deliberately cheap fetch-only
  consumer, our finest supply curve floors at 1,267 blocks/s near height
  ~1.87M vs ~15.5-16.2k blocks/s in the fastest bands — a ~12.8x trough. The
  consumer is starved 95-98% of the time at the floor, so the trough
  describes the node's per-block serving cost.
- **Parallel backfill saturates early.** Aggregate supply saturates around
  2.1-2.7x a single process by ~4 concurrent range readers, and the ceiling
  is a per-band *block rate*, not byte bandwidth — consistent with per-block
  request cost.
- **Per-request overhead bound.** gRPC fetch-alone vs a direct in-process
  read + re-serialize of the same heights: 14,403 vs 42,090 blocks/s at
  height 400k (2.92x), 10,105 vs 30,772 at 2.4M (3.05x). An upper bound on
  the per-request path's overhead, not a deliverable speedup — but overhead
  a range API can amortize.

The per-block cost is visible in the serving path itself: `GetBlock` reads
the block's raw consensus bytes from RocksDB, fully deserializes every
transaction (`FromDisk`; the in-tree TODO at
`crates/zakura-state/src/service/finalized_state/disk_format/block.rs` flags
the crypto-parsing cost), then re-serializes the whole block back to the
exact bytes it started from — once per unary round trip.

This is not only zindex's problem: Zaino's JSON-RPC fetch backend ingests
history one block per `getblock` round trip and has an open slow-sync issue
(zingolabs/zaino#1193, 109k blocks in 20+ hours), and lightwalletd's
ingestor is likewise serial. A range method on the indexer service serves
the block-fetch half of those loops too.

## Solution

One additive method on the existing service (three of its four methods
already server-stream; `GetBlock` is the only finalized-history fetch):

```proto
rpc GetBlockRange(BlockRangeRequest) returns (stream BlockAndHeight);
```

- `BlockRangeRequest { start_height, end_height }`, end inclusive;
  `INVALID_ARGUMENT` if the end is below the start or either height is not a
  valid block height. `BlockAndHeight { height, data }`, where `data` uses
  the same encoding as `BlockAndHash.data`. Ascending height order.
- **Finalized-only.** The stream serves only blocks committed to the
  finalized state and ends cleanly at the finalized tip. This matches the
  existing contract: `GetBlock` is documented as the finalized-gap bridge,
  and the reorg-mutable non-finalized blocks belong to
  `NonFinalizedStateChange`.
- **Stop-before-first-missing.** For pruned or missing bodies the stream
  ends cleanly before the first unavailable height — the documented
  convention of `ReadRequest::BlocksByHeightRange` and
  `HeadersByHeightRange`, adopted unchanged.
- **Raw read path.** A new `ReadRequest::RawBlocksByHeightRange` returns
  each block as its stored raw bytes, assembled as
  header ‖ CompactSize(tx count) ‖ raw transactions by a new
  `ZakuraDb::raw_block_bytes` that reuses the existing private
  `raw_block()` helper — the transactions are never deserialized, deleting
  the parse → re-serialize round trip the `ReadRequest::Block` path performs.
- **Backpressure.** The service's existing streaming pattern:
  `RESPONSE_BUFFER_SIZE` (64) bounded mpsc, `SEND_TIMEOUT` (60 s), drop the
  task on a hung consumer. The range is read from the state in
  buffer-sized chunks (`BLOCK_RANGE_CHUNK_SIZE` = 64 blocks per read), so a
  stream never buffers more than ~two buffers' worth of blocks — ~256 MB
  worst case per stream at pathological block sizes, ~4 MB realistic
  (`MAX_BLOCK_BYTES` = 2,000,000 < tonic's 4 MiB default message limit, so
  no limit changes).
- **No auth changes.** Same posture as the rest of the indexer endpoint:
  opt-in via `rpc.indexer_listen_addr`, off by default.

Out of scope (keeping the scope tight): tip-following and non-finalized
blocks, changes to `GetBlock`, serving pruned ranges, compression, auth.

## Feature gating and semver

The new zakura-state surface is gated behind the existing `indexer` feature,
following the `ReadRequest::SpendingTransactionId` /
`ReadResponse::TransactionId` precedent exactly:
`ReadRequest::RawBlocksByHeightRange`, `ReadResponse::RawBlocks`,
`ZakuraDb::raw_block_bytes`, the service handler arm, and the new tests all
carry `#[cfg(feature = "indexer")]`. **zakura-state therefore has no semver
change**: its default-features public API is untouched (verified by
compiling a probe referencing the variant without the feature — E0599 — and
with it — clean).

One deliberate wiring choice: zakura-rpc does **not** enable
`zakura-state/indexer` unconditionally on its dependency. The state's
`indexer` feature is not a pure API gate — it changes the database format
(`state_database_format_version_in_code()` gains `+indexer` build metadata,
and the format-upgrade set swaps `drop_tx_locs_by_spends` for
`track_tx_locs_by_spends`), so enabling it from zakura-rpc would flip every
default zakurad build into the indexer database format. Instead, zakura-rpc
gains a passthrough feature `indexer = ["zakura-state/indexer"]` — the same
shape zakurad already uses — and zakurad's `indexer` feature now also
forwards to `zakura-rpc/indexer`. Built without the feature, `GetBlockRange`
returns `UNIMPLEMENTED` (the proto method itself cannot be compiled out of
the generated trait); built with `--features indexer` it serves the raw
range path.

**zakura-rpc takes a major bump (6.0.0 → 7.0.0), included in this PR** per
`docs/changelog/guidelines.md` (version bumps for published-API changes land
in the PR that makes the change): the generated `pub trait Indexer` has no
default method stubs (build.rs does not enable `generate_default_stubs`), so
the new required method + its `GetBlockRangeStream` associated type are a
breaking change for trait implementers. We deliberately did **not** enable
default stubs — that would silently change the blast radius of every future
proto addition. The bump was applied as `cargo release version` would:
`crates/zakura-rpc/Cargo.toml` to 7.0.0 and zakurad's workspace-internal
dependency requirement updated to match.

Housekeeping note: `docs/changelog/unreleased/get-block-range.md` is a
placeholder name — once this PR has a number it will be renamed to
`<PR-number>.md`, and the `#0000` placeholder links in the fragment and in
the new `docs/changelog/params.md` row replaced (the numeric-name check does
not catch a stale link body; `0000` is the lychee-excluded placeholder, so
the link checker stays green until then).

## Testing

**CI visibility, stated plainly:** no upstream workflow currently enables
`--features indexer`, so the feature-gated tests below do not run in
upstream CI's current matrix (which is also why the two pre-existing
with-feature failures on `main` noted at the bottom of this section have
gone unnoticed). The one CI-visible assertion on the new method is a
default-features test in `vectors.rs` asserting `GetBlockRange` returns
`UNIMPLEMENTED` when built without the feature, so default CI executes the
new code path on every run. We're happy to contribute a feature-matrix CI
job (e.g. a `zakura-state`/`zakura-rpc` test job with `--features indexer`)
as a follow-up if you want the gated suites enforced.

- `crates/zakura-rpc/src/indexer/tests/vectors.rs` — real tonic server +
  `IndexerClient` against the existing `MockService` pattern. Default
  features: `GetBlockRange` returns `UNIMPLEMENTED`. With
  `--features indexer`: `INVALID_ARGUMENT` on end < start and on
  out-of-range heights (both without touching the state); ascending order;
  byte-for-byte equality of `data` with `GetBlock`'s answer at the same
  heights plus decode + hash round trip; clean end-of-stream when the state
  stops early (clamp at tip / gap); chunked reads for a >64-block range
  (64, 64, remainder) with all 130 messages received in order; and a short
  mid-range read ends the stream without a further state request
  (`expect_no_requests`), making the early-stop load-bearing.
- `crates/zakura-state/src/service/tests.rs` — through the real state
  service: the raw range read returns the exact consensus serialization
  (byte-for-byte vs `zcash_serialize_to_vec`), stops before the first height
  without a committed body (header-only heights), and returns empty for
  `count: 0` and for a start above the body tip — mirroring the existing
  `BlocksByHeightRange` assertions.
- `crates/zakura-state/src/service/finalized_state/zakura_db/block/tests/vectors.rs`
  — every block in the db round-trip suite (including the generated
  multi-transaction and large single-transaction blocks and all
  mainnet/testnet vectors) also asserts `raw_block_bytes` equals the block's
  consensus serialization, pinning the CompactSize count and transaction
  order.
- `crates/zakura-state/src/service/finalized_state/zakura_db/block/tests/prune.rs`
  — `raw_block_bytes` returns the block's exact consensus serialization for
  a retained height and `None` at the online pruning boundary; and a
  read-state-service test where a pruned height sits between retained bodies
  asserts the range read stops before the gap and does not resume above it
  (pinning stop-before-first-missing rather than skip-over-gaps).

Commands run locally (Apple Silicon, rustc 1.91.0), each in both feature
states (default and `--features indexer`):

- `cargo test -p zakura-rpc --lib` — 99 passed (both states)
- `cargo test -p zakura-state --lib` — 458 passed (default); with
  `--features indexer`, 456 passed and the only 2 failures
  (`vct_format_changes_include_root_auth_metadata_updates`,
  `intra_block_self_spend_chain_in_finalized_state`) reproduce identically
  on clean `main` with the feature — pre-existing, unrelated to this change
- `cargo clippy -p zakura-state -p zakura-rpc --all-targets` — clean (both
  states)
- `cargo fmt --all -- --check` — clean
- `cargo check --workspace --all-targets` — clean
- `cargo check --features indexer` (run from `crates/zakurad`; the package
  is named `zakura`) — clean

## Benchmarks

Measured on this branch against the real mainnet archive database (259 GB,
v28), opened as a RocksDB read-only secondary beside the running node.
`GetBlockRange` (one stream) vs `GetBlock` pipelined at concurrency 64,
10,000 blocks per window, A/B alternated across rounds, median within-round
ratio:

| window | mean block | blocks/s | server CPU per block |
|---|---|---|---|
| 400,000 | 15 KB | **2.24x** | 0.028 ms vs 0.340 ms (**12x**) |
| 3,400,000 | 12 KB | **2.69x** | 0.022 ms vs 0.338 ms (**15x**) |
| 1,870,000 | 1.04 MB | **1.44x** | 0.672 ms vs 4.94 ms (**7.3x**) |

The two metrics diverge on the dense band and the reason matters: at
1.04 MB/block both methods saturate the loopback link (2.9 vs 2.0 GB/s), so
the saved CPU cannot turn into wall clock there. Read server CPU per block
for a capacity claim and blocks/s for a latency claim — either alone
misrepresents this.

Parallel range streams reverse direction by block size: four streams roughly
double throughput again on small blocks (134k blocks/s at the tip) and *hurt*
on the dense band (eight streams is 18% slower than one).

A correctness result fell out of the same run: the byte totals from the two
methods are identical across all 30,000 blocks
(10,416,125,249 / 123,709,549 / 149,314,158 B). One path assembles from
stored raw bytes and the other deserializes and re-serializes, so this is an
independent cross-check of the new encoding.

Caveats: the machine was not idle (foreign load 13-19), mitigated by warmup
passes and order alternation — ratios were order-independent in every window,
and the dense window took 8 rounds because its `GetBlock` side is CPU-bound
and scatters under load (834-2035 blocks/s wall clock against a steady
48.8-55.5 s of CPU, which is itself the evidence for the transport-bound
reading). A read-only secondary has an empty non-finalized state, so
`GetBlock` skips a check it would pay on a primary; if anything these numbers
understate the range path.

## Related

- #601 / #602 (`NoteCommitmentTree::from_frontier`): same shape — a small
  seam so a downstream consumer stops re-doing work the node already did.
- #596 (mTLS for remote indexer RPC): a remote deployment of this stream
  would inherit that requirement once it lands.
- zingolabs/zaino#1193: ecosystem demand for the same seam.

🤖 Drafted with [Claude Code](https://claude.com/claude-code) at @maxdesalle's direction


ValarDragon (approved):
Thank you! Probed with AI a bit, great job on this. Will make a follow-on PR for some CI tweaks to ensure this gets properly tested.

I'm not worried about RAM load, tbh I think we can go higher

PR #613, opened 2026-08-11
perf(consensus): key Halo2 cache by witnessed transaction ID

ValarDragon:
## Motivation

The Halo2 cache currently re-encodes the complete Orchard or Ironwood bundle into a new `Vec` and hashes those bytes for every lookup. The transaction verifier has already computed the transaction witness identity before it reaches the Halo2 call site, so the extra allocation and full-bundle scan are avoidable.

## Solution

Thread the existing v5/v6 `WtxId` into production Halo2 items and key each circuit-era cache by:

- the `WtxId`, whose txid component commits to effecting data and whose ZIP-244 authorizing-data digest commits to proofs and signatures; and
- the Orchard/Ironwood value-pool tag, which distinguishes the two bundle slots of one v6 transaction.

`WtxId` already contains the txid, so the key does not store a redundant second copy. The verifying key remains bound by the existing per-circuit-era cache instances.

`Item::new` keeps its existing public signature; items constructed without a transaction identity still verify normally but bypass the cache, while the transaction verifier uses a new crate-private `Item::new_with_wtx_id` constructor with the already-computed ID. There are no downstream public-API changes.

This removes bundle serialization, temporary allocation, and BLAKE2b hashing from the production cache lookup path.

## Testing

- `cargo test -p zakura-consensus --lib` — 178 passed
- `cargo clippy -p zakura-consensus --all-targets -- -D warnings`
- `cargo test -p zakura-consensus --doc`
- `./scripts/changelog.py check-pr --base origin/main --head HEAD --pr 613`

The cache-key tests pin the txid, authorizing-data digest, and pool components, including the CVE-2026-34377 authorizing-data mutations. Cache behavior tests use explicit witnessed transaction identities, and an additional test confirms that items without one never inherit cached results. The production transaction-verifier test still confirms mempool-to-block reuse, height-dependent expiry rejection, and misses for mutated verification inputs.

## Changelog

`docs/changelog/unreleased/613.md` marks this as an internal refinement of the still-unreleased cache feature.

- [ZIP 239: Transaction Identifier Digests and Signature Validation](https://zips.z.cash/zip-0239)
- [ZIP 244: Transaction Identifier Non-Malleability](https://zips.z.cash/zip-0244)

---

> [!NOTE]
> **Medium Risk**
> Changes consensus-adjacent Halo2 cache identity logic; incorrect keys could skip proof verification, though the new key is designed to cover ZIP-244 authorizing data and v6 pool slots and is heavily tested.
> 
> **Overview**
> Replaces the Halo2 Orchard cache lookup key that **re-encoded each bundle into a `Vec` and BLAKE2b-hashed it** with a structured **`CacheKey`** built from the transaction's **`WtxId`** (txid + ZIP-244 authorizing-data digest), **sighash**, and **value pool** (Orchard vs Ironwood on v6).
> 
> Production verification now constructs cacheable items via **`Item::new_with_wtx_id`**, with **`wtx_id` passed from the transaction verifier** into v5/v6 **`verify_orchard_bundle`**. **`Item::new`** still verifies but **does not use the cache** when no witnessed ID is supplied.
> 
> The **`Cached`** service only hits, records, and tests inner calls when a cache key is present. Docs and tests are updated to pin **txid**, **auth digest**, **sighash**, and **pool** separation (including **CVE-2026-34377** authorizing-data mutations). Changelog marks this as an **internal** change to unreleased cache behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6415bbd3ae750c7c95a837e39fd59edc35a175c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


evan-forbes (approved):
benchmarked as we do save 23 us

```
─────────────────────────────────┬───────────────────────────────────┬────────────────────────┐
│                                 │          Old (pre-#613)           │       New (#613)       │
├─────────────────────────────────┼───────────────────────────────────┼────────────────────────┤
│ Cache-key derivation            │ 23.3 µs + a ~9 KB heap allocation │ 20.7 ns, no allocation │
├─────────────────────────────────┼───────────────────────────────────┼────────────────────────┤
│ Full hit path (key + set probe) │ 23.1 µs                           │ 26.8 ns                │
├─────────────────────────────────┼───────────────────────────────────┼────────────────────────┤
│ What a cache hit actually saves │ 5.37 ms (full Halo2 verify)       │ 5.37 ms — unchanged
```

the not needing sighash isn't intuitive to me and we should document that imo

then adding a a test that derives the wtx_id from req.transaction() so the invarient is enforced

edit for more info on the missing enforcement test
```
Existing tests always build UnminedTx with tx.into(), so its stored id matches its transaction. However,
  UnminedTx.id is public. A caller can provide transaction A with transaction B’s WtxId. The verifier currently takes the bundle from req.transaction()
  but the cache identity from req.tx_id() at crates/zakura-consensus/src/transaction.rs:431.

  A normal equality test would not enforce this invariant. The test must deliberately construct a mismatched UnminedTx.

  The minimal correctness change is:

  let tx = req.transaction();
  let tx_id = tx.unmined_id();

  instead of:

  let tx = req.transaction();
  let tx_id = req.tx_id();

  at crates/zakura-consensus/src/transaction.rs:431. This makes the cache identity come from the same transaction as the bundle. It also keeps the
  response identity consistent. The trade-off is that mempool verification recomputes the WtxId. Preserving the precomputation safely would require a
  broader change that makes the UnminedTx invariant type-enforced.
  ```

ValarDragon:
Sighash did wind up being useful, so added it to the tx

ValarDragon:
the security risk you cited is API only - we don't guarantee that, but lets just go make it type safe :)

ValarDragon:
The type safety is merged into main, re-running v12

ValarDragon:
@v12sec pls review

PR #614, opened 2026-08-12
docs: add core Rust readability rule

p0mvn:
## Motivation

Codify recurring human-readability feedback from [#586](https://github.com/zakura-core/zakura/pull/586) so AI-generated changes remain understandable without reconstructing their intent.

## Solution

- Add a focused rule for semantic naming, black-box contracts, cohesive abstractions, and production/test boundaries.
- Scope the guidance to the core Rust library crates.
- Keep the canonical rule under `.agents/rules/` and symlink it into `.cursor/rules/` for Cursor discovery.

## Testing

- `markdownlint .agents/rules/core-rust-readability.mdc .cursor/rules/core-rust-readability.mdc --config .github/.markdownlint.yaml`
- IDE diagnostics: no errors
- `git diff --check`
- Rust compilation and tests were not run because this PR only adds agent guidance.

## Changelog

Not required because this PR does not change Rust source files or `Cargo.toml`.

- [Readability review feedback in #586](https://github.com/zakura-core/zakura/pull/586)


evan-forbes (approved):
sg, fwiw I don't think codex or claude etc use cursor's format here for .mdc (claude does use .md tho)

afaiu as well cursor will load this each time the relevant files are read from and not just written to. perhaps there is a way to trigger only when writing? 

I've been playing around w/ a few skills that require a specific  call to keep context minimum.

the reasoning for being efficient is that we could expand this significantly, even if that means having to remember doing a specific call. could maybe make it a githook?

PR #615, opened 2026-08-12
fix(ci): make test success the required gate

TalDerei:
## Summary

- Keep pull-request path filters for lint, crate-build, and semver workflows.
- Run the unit-test workflow on every PR, while skipping heavy test jobs for irrelevant paths, so `test success` can be the only required check.
- Remove the duplicate status-check workaround and an unsupported Dependabot cooldown option.

## Rollout

After merge, require only `test success`; keep the other aggregate checks advisory. Retrigger existing PRs that do not yet have this check.

## Validation

Latest CI: 28 checks passed and 15 path-filtered checks were skipped. Workflow YAML, zizmor, semver-selector tests, markdownlint, codespell, and `git diff --check` also passed.


ValarDragon:
I actually really like the paths filter - we start taking unnecessary runners if we don't have them on no?

We do actually start hitting concurrency limits in this repository, leading to runner start time delays

ValarDragon (approved):
Lgtm, thank you!

PR #617, opened 2026-08-12
feat(state)!: authenticate VCT roots at header time

evan-forbes:
## Motivation

Fast sync accepts per-block Sapling, Orchard, and Ironwood roots from peers instead of
rebuilding the note commitment trees. Header `H + 1` commits to the roots for `H`, so a
supplied root can be proven as soon as its successor header arrives. Until now the node
waited for `H`'s block body to reach the committer before checking it, which is thousands
of blocks later during fast sync: an invalid root sat unchallenged, and its replacement was
requested only after the committer stalled on it.

`main` already had a lane for this — the height-keyed `AuthenticateHeaderRoots` request and
its durable ascending frontier. That lane cannot run on a fork-aware header chain. It keys
authentication by height, so a fabricated root written at height `H` poisons an index every
competing branch shares, which is why promotion had to refuse to authenticate above the last
completed checkpoint. It also folded and rewrote its frontier on every block commit while
nothing read it.

## Solution

Authenticate at header time, keyed by hash, on the DAG the header chain already maintains.

`crates/zakura-state/src/service/write/vct_sweep.rs` walks the selected path, and for each
block whose successor header has arrived it verifies the supplied roots against that
successor's commitment, marks the delivery `Authenticated` on its node, and evicts and
re-requests the delivery it can attribute a failure to.

The cryptographic kernel is unchanged. The sweep calls
`zakura_chain::parallel::commitment_aux_verify::verify_supplied_roots_from_parts` — the
same function the removed lane called, byte for byte the same file — and supplies it the
same input shape: roots on both sides of the one-header-later boundary, so the boundary
header's pre-activation pins are checked. The only difference is batch size, two items per
step instead of the whole fetched range, which is equivalent because the kernel folds
sequentially and the sweep carries the history tree forward across steps.

Hash keying confines a bad branch's marks to that branch's nodes, so the checkpoint gate
stops being a correctness requirement and authentication runs one header behind the tip
instead of up to `MAX_CHECKPOINT_HEIGHT_GAP` blocks behind the last completed checkpoint.
The DoS surface hash keying opens is bounded by the existing
`MAX_AUX_DELIVERIES_PER_HEADER_V1` (16) and `MAX_AUX_DELIVERIES_TOTAL_V1` (65,536) limits,
and deliveries are pruned with their nodes.

The second commit removes what the sweep replaces: the `AuthenticateHeaderRoots` request,
the `header_root_auth_frontier` column family, the frontier rollback and rebase paths, and
the per-commit fold in `prepare_trees_batch`. That is 2,297 lines out of
`commitment_roots_db.rs`, none of it cryptography — every removed function is storage or
state-machine code. `COMMITMENT_ROOTS_BY_HEIGHT` survives as the body-written serving index.
A database written by an earlier version keeps the column family but has its row cleared on
upgrade; new databases never create it. `DATABASE_FORMAT_PATCH_VERSION` goes 3 to 4.

## Testing

- `crates/zakura-state/src/service/write/vct_sweep/tests.rs` covers admission, attribution,
  eviction, re-anchoring after a branch shift, and the decline-rather-than-panic path.
- A proptest cross-checks the sweep against the committer over corrupted-delivery fixtures:
  every height the sweep marks `Authenticated` must also be accepted by
  `verify_commitment_roots`, which reaches the same ZIP-221 leaf from a block body rather
  than from schema-1 parts. Two independent paths, asserted to agree.
- Disk upgrade tests cover both the frontier drop and the unauthenticated-roots path.

## Changelog

Added after opening this PR.

The kernel `verify_supplied_roots_from_parts` was written by Roman Akhtariev for `main`'s
ahead-of-body header-root authentication lane in #346, #351, #352, and #455. It is reused
here unchanged; this PR only supplies the fork-aware inputs that lane did not have.

`docs/design/verified-commitment-trees.md` describes the header-time check.
`docs/design/header-sync-vct-root-authentication.md` records the removed lane and why it is
superseded.

### Follow-up Work

This PR is based on #532 and stacks on it. The base does not currently build: the
`HeaderNode.body` to `body_validation_state` rename and the removal of
`EligibilityReason::ConsensusBodyInvalid` in `zakura-header-chain` are not carried into
`zakura-state`. `cargo check -p zakura-state --all-targets` reports the identical 34 errors
on this branch and on its base, so none of them come from this PR, but this PR cannot be
green until the base is fixed.


evan-forbes:
haven't done ai reviews or smoke tested properly so not yet ready for review

there's almost certainly some simplifications we can do here

edit: g2g

evan-forbes:
need a migration fix per the smoke test, so waiting on mering

evan-forbes:
merging after conflict fixes

PR #619, opened 2026-08-12
chore: supply chain audit 

TalDerei:
Fixes two advisories (OpenTelemetry baggage, the vulnerable `git2` build backend) and audits the exempted dependency set.

Against `main`, `cargo vet` goes from 154 fully audited / 589 exempted to **733 fully audited / 45 exempted**. `cargo vet check --store-path ./qa/supply-chain/` passes.

## The 45 remaining exemptions

43 are documented holds: each carries an `Audit hold:` note naming the defect, a reproduction, and the condition that clears it. The other two, `console 0.15.11` and `indicatif 0.17.11`, returned with the `number-prefix` revert.

None can be cleared by further auditing, only by dependency work:

- 14 are not in any shipped runtime build (build-only, dev-only, or targets we do not ship)
- 5 are inert behind features we do not enable
- 2 carry live risk: `hickory-proto` and `lru`, both advisory-bearing and both behind the Iroh 0.92 pin
- the rest await upstream fixes

Follow-ups: https://github.com/zakura-core/zakura/issues/661, led by the Iroh 1.x migration, which clears the live-risk pair.

## Scope note

The remaining findings are not simple lockfile updates. Hickory, LRU, `instant`, `paste`, and `atomic-polyfill` are pinned through Iroh 0.92; removing them requires an Iroh 1.x API migration across the native P2P implementation. The direct `bincode` warning backs persisted state data, so replacing it requires byte-compatibility or an explicit database migration. These should not be represented as fixed by merely changing `deny.toml` ignores.


ValarDragon:
@v12sec review this

deny.toml
```diff
@@ -181,6 +181,9 @@ skip-tree = [
     { name = "hmac", version = "=0.12.1" },
     { name = "netdev", version = "=0.36.0" },
     { name = "netwatch", version = "=0.9.0" },
+    # OpenTelemetry 0.32 uses reqwest 0.13; iroh 0.92 remains on 0.12.
+    # Remove this once the iroh 1.x upgrade unifies Zakura on reqwest 0.13.
+    { name = "reqwest", version = "=0.12.28" },
```
ValarDragon, line 186: Cool, updating Iroh is one of the blockers before default v2 p2p stack

ValarDragon (approved):
Approve, but we likely switch away from gitcl soon as well!

PR #631, opened 2026-08-12
docs(header-chain): map the production code in the fork-aware change

evan-forbes:
## Motivation

The fork-aware header chain landed in #586, a change that spans five crates. The
design lives in a minority of those files, and nothing in the tree tells a reader
which files carry a rule and which only update a call site.

## Solution

Adds `docs/design/headerchain/production_code_header-chain.md`, which explains how
the production code implements the design.

- Opens with the single-planner rule the split exists to enforce, then the state
  the engine keeps.
- Follows a header through admission and validation, then states the commit order
  that keeps the staged, committed, and durable copies in agreement, and the write
  paths that reach it.
- Covers fork switching, VCT evidence and authentication, and the branch and
  generation checks that reject stale work.
- Closes with the boundary each crate enforces, one section per crate, linking the
  entry point rather than listing files.

Each `LC-*` citation links to the rule it names, so a reader can check the claim
without searching the specification. The specification gains seven anchors so those
links resolve.

## Verification

Every link target and every backticked identifier resolves against `main` at
`05caf7f6a`. The document states no file counts, line counts, or constants, so
nothing in it goes stale as the tree moves: the specification defines the rules and
the code defines the values.

`markdownlint` and `codespell` pass locally under the repo configs.

## Review notes

Rebased onto `main` now that #586 has merged. This PR previously targeted
`hc/01-engine-model`, and two things in the document depended on that base.

Six source links pinned file contents to the tip of `hc/01-engine-model`, a commit
that is not on main and is not reachable from it. They now name paths in the tree,
which is what the other fourteen source links already did.

`verify_plan` survives on main only as a `#[cfg(test)]` helper that re-runs
verification for tests and fuzzing, so the guide named a test function for the
production invariant gate. It now names `verify_candidate`, the gate a candidate
passes before an `EngineTransition` can exist, and the commit-order diagram returns
`EngineTransition` rather than the old `TransitionPlan`.

Documentation only, no Rust or `Cargo.toml` change, so no changelog fragment per
`docs/changelog/guidelines.md`.

### DIFF
--- a/docs/design/headerchain/production_code_header-chain.md
+++ b/docs/design/headerchain/production_code_header-chain.md
@@ -1,0 +1,572 @@
+# Production Code in the Fork-Aware Header-Chain Change
+
+Parent: `docs/specs/fork-aware-header-chain-engine.md` (rules cited as `LC-*`).
+Source: PR #586, branch `hc/01-engine-model`, merge base `ba9261a73`, measured at
+`664607d4f`.
+
+## The problem all of this answers
+
+Four independent sources want the tip to move. Peers deliver headers, full state
+verifies bodies, an operator invalidates a branch, and finality advances. If each source
+computes the move, the node runs as many fork-choice implementations as it has sources,
+and they disagree. Timing breaks it even when they agree: a source that computed its
+answer against an old view lands late and overwrites a newer one. That is the production
+incident this design replaced.
+
+The answer is one rule. A source states what it observed, and one planner turns that
+evidence into every consequence. The five layers below are that rule under five
+different constraints: keep the planner pure, keep the network crate unable to reach the
+database, make the durable order survive a crash, keep async work from outliving its
+authority, and keep two block-apply paths from running at once.
+
+## The abstractions to know first
+
+The file list below assumes this vocabulary.
+
+### The graph, the overlay, and the delta
+
+`MemHeaderStore` in `graph.rs` is the graph. It holds every retained header as a
+`HeaderNode` keyed by consensus hash, rooted at the finalized `Frontier`. A `Frontier`
+is a height and a hash together, and it is the only way this design names a position on
+a chain. A node carries the header, its parent hash, a height from a checked parent
+increment, the block work derived from the validated compact target, a
+`WorkCoordinate`, a header validation state, an eligibility state, a body validation
+state, and its auxiliary delivery ids. Beside the node map the graph keeps three
+indexes it can rebuild from the nodes alone: children by parent, hashes by height, and
+the set of eligible tips.
+
+Selection reads that graph. `select_best_header_chain` takes the maximum score over the
+eligible tips, so permuting arrivals cannot change the answer (LC-SELECT-04). A node is
+eligible when its header is valid, it carries no exclusion reason, and no ancestor is
+ineligible, which makes ineligibility a set of durable reasons rather than a flag. A
+flag is last-writer-wins, and last-writer-wins is arrival order in disguise.
+
+The engine holds the graph and two projections beside it. The selected projection lists
+the frontier of every prefix of the best header chain from finality to the tip, and the
+verified projection lists the same for the chain full state has verified. A projection
+is a list, so a reader answers a height query without walking the DAG.
+
+`GraphOverlay` in `graph/overlay.rs` is a mutable view over a borrowed graph. Reads see
+the base plus whatever the overlay staged, writes land in the overlay's own maps, and
+the base never changes. `GraphDelta` is what the overlay extracts: a new finalized
+frontier when finality moved, the nodes to put, and the nodes to delete. A node the
+overlay inserted and then deleted appears in neither list, and a node it wrote back
+unchanged appears in neither, so the delta holds the difference and nothing else.
+
+Both types implement `HeaderGraphView` and `HeaderGraphEdit`, so selection, eligibility
+propagation, and the invariant checks run one implementation against committed state
+and against a staged one. The planner therefore derives a complete transition without
+copying the graph and without touching the live one.
+
+`TransitionPlan` is what it derives: the change set the store will write, the private
+`GraphDelta` the engine will install, and `before()`, the snapshot it was derived from.
+
+### Work owners, generations, and branches
+
+The reactor issues requests asynchronously, so an answer can arrive after the question
+stopped meaning anything. A `BranchId` names the work an answer belongs to by the
+anchor it was rooted at and the tip it pursued. It carries no height, because a reset at
+the same height passes every height comparison. Three counters move underneath it.
+`state_version` increments on every committed transition that changed a durable fact.
+`header_generation` increments when selection, header topology, header validation,
+eligibility, or finality moved. `verified_generation` increments when the verified
+projection moved or finality advanced.
+
+`Gate` in `ownership.rs` judges a response against the current snapshot and returns a
+`CompletionDecision`: `Current`, or `Stale` carrying a typed `StaleReason` of
+`HeaderGeneration`, `VerifiedGeneration`, `BranchAnchor`, `MissingOwner`, or
+`OwnerMismatch`. It reads the generations and the branch anchor, and it ignores
+`state_version`, because that counter also moves for transitions on branches the request
+never went near, so binding header work to it would cancel in-flight requests faster
+than they complete for no correctness gain. The planner repeats the same check as
+`TransitionFailure::Stale`, and the scheduler keys its work by `(generation, branch)`.
+
+### Keeping three copies in agreement
+
+The same state exists three times: staged in an overlay, committed in the engine's
+graph, and durable in the store. Six rules keep them from diverging.
+
+The overlay never writes to its base, and `GraphDelta`'s fields are private to the
+crate, so no caller assembles a partial one.
+
+The delta rechecks itself against whichever graph it lands on. `derive_index_changes`
+recomputes each put node's hash, parent link, and work from the compact target, refuses
+a delete of a node the base does not hold, refuses a hash that is both put and deleted,
+and refuses a finalized frontier that is missing or ineligible. `apply_delta` runs that
+check before it mutates, and `validate_delta` runs it without mutating.
+
+`verify_plan` checks the plan against a graph rebuilt from the delta with
+`GraphOverlay::from_delta`, not against the overlay the planner mutated, so the verifier
+approves the transition the engine will install rather than the one it staged.
+
+`install_committed_transition` compares the live snapshot with the plan's `before()` and
+refuses with `StaleSource` when they differ. `plan_transition` takes `&self`, so two
+planners can run against one engine; only one of them can install.
+
+Disk leads memory leads observers, and that order is forced. The next restart reads
+disk, so a crash in this order leaves disk ahead of memory, which the startup audit
+repairs by rehydrating, and a crash in any other order leaves an observer holding a
+frontier that is not on disk, which nothing repairs. One transition makes one atomic
+`db.write`, so a crash leaves the whole batch or none of it. A failure between the write
+and the install fails closed: the runtime returns the store error, publishes nothing,
+and the next open rehydrates the engine from disk.
+
+The durable indexes are caches. The authoritative rows are the per-node rows in
+`header_node_by_hash` and the singleton in `header_engine_meta`. Children, heights,
+eligibility roots, deferrals, and both projections are rebuilt by the startup audit from
+the node rows, which is why a crash that leaves disk ahead of memory is repairable and
+the reverse order is not.
+
+### When the store is written
+
+Once per committed transition, inside the writer lock, and nowhere else in the steady
+state. There is no accumulation and no background flush: the runtime builds one
+`DiskWriteBatch` from the change set and calls `db.write` once (LC-TXN-01). What else
+lands in that one write depends on which call the state writer made.
+
+| Call | What lands in the one write |
+| --- | --- |
+| `apply` | the header-chain rows alone |
+| `apply_combined` | those rows appended to the block batch the state writer already filled |
+| `apply_aux_then_checkpoint_combined` | auxiliary authentication and the checkpoint advance that depends on it, planned into the same batch |
+| a no-change plan | the caller's batch alone, with nothing installed and nothing published |
+| a `ResourceStalled` plan | an alarm-only batch, discarding the caller's rows, so the state writer stops with its own rows unwritten |
+
+Two more writes happen outside that path. When the startup audit finds repairs, the
+runtime commits one repair batch before the engine exists. On first run against a store
+that predates the DAG, `migration.rs` builds the initial nodes and the migrated finality
+pin in its own write.
+
+One transition starts without any component observing anything. The write loop keeps a
+timer for `earliest_deferred()` and submits `ReevaluateDeferred` when it elapses, which
+is how a header held for a local future time becomes eligible. It flushes like every
+other transition.
+
+### What a header is checked against, and when
+
+Validation runs four times before disk, each pass against more context than the last,
+and no pass trusts the one before it.
+
+Off the writer lock, `prepare_header_target` in the driver reads a validation lease for
+the common ancestor through the read service, derives `HeaderRules` from that lease, and
+runs `prepare_context_free_headers` on a blocking thread. Those are the rules that read
+no ancestor: canonical version and hash, checked height inference, commitment structure
+for that height, compact-target domain and network limit, hash at or below target, and
+Equihash. A header whose time runs more than two hours ahead of the local clock comes
+back `DeferredUntil` rather than rejected (LC-VAL-08). The result is a
+`PreparedHeaderBatch` sealed with a receipt naming the parent frontier, the network, and
+the trust-anchor digest.
+
+Under the writer lock and before planning, the runtime reads the predecessor facts
+itself rather than accepting any the caller supplies. It seals the parent's validation
+context, the anchor's when finality moved, and the finality rebase path, then wraps the
+caller's authority in `StateIssuedAuthority` over exactly the leases it just issued, so
+a lease from anywhere else is refused.
+
+In the planner, against the graph, the receipt's parent, network, and trust-anchor
+digest must still match the live config, or the transition fails with
+`StalePreparation` (LC-VAL-11). The planner then re-derives, per header, the parent
+link, the hash, the height increment, and the work from the compact target, and runs
+`validate_contextual_difficulty_and_time` against ancestry it reads from the graph and
+the lease. A peer's framing height is never evidence.
+
+After planning and before disk, `verify_plan` re-derives the projected result: hashes
+and linkage, index round-trips, work coordinates, inherited eligibility, both
+projections' contiguity, that `header_best` really is the maximum eligible score, trust
+pins, protected nodes, frozen limits, generation increments, and auxiliary provenance.
+A disagreement between the planner and the verifier is `InvariantViolation`, and that
+batch never reaches disk.
+
+Startup adds a pass that answers to no caller: `audit_store` re-derives what the DAG
+determines before the engine exists.
+
+## How this list was cut
+
+The change touches 303 files, adding 88,706 lines and removing 46,587. Most of that is
+not the design. Tests and harnesses account for 26,382 added lines, fuzzing for 10,388,
+docs and manifests for 1,180, CI, docker, and policy for 1,150, and a lockfile and two
+data files for 34. What remains is 152 Rust files of production code, 49,572 lines added
+and 23,994 removed.
+
+Twelve of those files are deletions, listed at the end. Sixty-five of the surviving 140
+carry the design. The other 75 are integration: a call site moved, a type re-exported, a
+service registered. This document lists the 65 by layer, then names the integration
+edits and the code the change deletes.
+
+Exclusions applied, in order: `fuzz/`, every `fuzz.rs`, `/tests/` and `tests.rs`,
+`testkit/`, the `header_chain/coherence/` model harness, `bench.rs`, `arbitrary.rs`,
+`crates/xtask/`, `docs/`, `.github/`, `docker/`, `deploy/`, `qa/`, and lockfiles. Two
+excluded files still matter to a reader: `crates/zakura-header-chain/conformance.toml`
+(926 lines) is the machine-checked rule manifest, and
+`crates/xtask/src/header_conformance.rs` (1,285 lines) is the checker that enforces it.
+
+One caveat on every count below. Rust colocates unit tests, so a file's added-line
+figure includes its `#[cfg(test)]` module. In the new crate, 17,176 added lines are
+12,360 lines of code and 4,816 lines of colocated tests.
+
+## Five layers
+
+```mermaid
+flowchart TD
+  ZD["zakurad<br/>drivers, coordinator, lifecycle"]
+  NW["zakura-network<br/>header-sync reactor, wire, scheduler"]
+  ST["zakura-state<br/>runtime, store, schema"]
+  NS["zakura-node-services<br/>typed port traits"]
+  HC["zakura-header-chain<br/>pure, synchronous, decides which chain is best"]
+  ZD --> NW
+  ZD --> ST
+  NW --> NS
+  ST --> HC
+  NS --> HC
+```
+
+Thirteen commits landed the change. Four introduce the production code, and each of the
+first three is one layer. The fourth covers both the reactor and the drivers, because
+the port implementation and the facility that calls it have to arrive together. Five
+later commits revise that code after review, and the remaining four add tests, fuzzing,
+docs, and release plumbing.
+
+| Commit | Adds |
+| --- | --- |
+| `04ed5f077 feat(header-chain)!: add fork-aware graph model` | the pure crate |
+| `dd5b79dd1 refactor(node-services)!: add typed header-sync contracts` | the port |
+| `4fefdd266 feat(state)!: persist and recover fork-aware header chains` | runtime, store, schema |
+| `ebb60ab2d feat(network)!: add bounded fork-aware header sync` | reactor, wire, drivers |
+
+## Layer 1: `zakura-header-chain`, 34 files, +17,176
+
+A new crate, synchronous and pure. It holds every rule about which chain of headers is
+best and performs no effect: no disk, no socket, no clock of its own, no task. Purity is
+what makes fork choice testable without a node, replayable from an event list, and
+usable by a headers-only client that links no database.
+
+A crate stays pure only if its dependency list does, so the boundary sits where a test
+can fail the build. Two architecture tests in `lib.rs` hold it: one fails if the
+manifest gains `tokio`, `tower`, `zakura-state`, `zakura-network`, or
+`zakura-consensus`, and the other scans the crate's own sources for wallet, FlyClient,
+and block-sync surfaces (LC-SCOPE-06).
+
+### Vocabulary: nine files that refuse things
+
+These files name values, but each one exists to make a mistake unrepresentable rather
+than to hold data. A `WorkCoordinate` carries the origin hash it accumulated from, so
+two coordinates from different origins refuse to compare instead of returning a
+plausible number that would silently decide fork choice. Generation counters use
+`checked_next` and fail closed at `u64::MAX` rather than wrap. Every error carries an
+explicit `Attribution`, so no generic conversion can charge a local disk failure to a
+peer.
+
+| File | Added | Role |
+| --- | --- | --- |
+| `src/lib.rs` | 196 | module list, public re-export surface, the two architecture tests |
+| `src/ids.rs` | 443 | identities and generation counters: `HeaderId`, `EvidenceId`, `SourceId`, `BranchId`, `StateVersion`, `FinalityEpoch`, and the work owners and authorities |
+| `src/frontier.rs` | 220 | `Frontier`, `ChainScore`, `WorkCoordinate`, `SuffixWork`, and exact chain-work ordering |
+| `src/header_node.rs` | 351 | `HeaderNode`, the admitted header record, plus header, body, and eligibility state |
+| `src/error.rs` | 412 | `HeaderChainError`, `ErrorCategory`, `RuleId`, and `Attribution`, the peer-blame boundary |
+| `src/config.rs` | 669 | `EngineConfig`, `EngineMode`, `TrustedAnchor`, `CheckpointSet`, settled-upgrade pins, and the versioned `_V1` limits |
+| `src/locator.rs` | 213 | `HeaderLocator` built from a committed selected path, capped at 13 hashes |
+| `src/ownership.rs` | 441 | `PendingOwners` and `Gate`, which decide whether a late response still owns its branch |
+| `src/retention.rs` | 582 | `enforce_retention`, the deterministic eviction planner |
+
+Two of the nine hold algorithms rather than types. `ownership.rs` holds the completion
+gate described above. `retention.rs` holds the eviction planner, and eviction is the one
+place a resource limit could pass for a verdict about a chain. It refuses that twice.
+Eviction writes no reason and no body state, so re-inserting the identical header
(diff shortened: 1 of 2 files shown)


docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
+# Production Code in the Fork-Aware Header-Chain Change
+
+Parent: `docs/specs/fork-aware-header-chain-engine.md` (rules cited as `LC-*`).
+Source: PR #586, branch `hc/01-engine-model`, merge base `ba9261a73`, measured at
+`664607d4f`.
+
+## The problem all of this answers
+
+Four independent sources want the tip to move. Peers deliver headers, full state
+verifies bodies, an operator invalidates a branch, and finality advances. If each source
```
p0mvn: "operator invalidates a branch": this functionality has been confusing me even from reading the code.

Would be great to define what this means

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+is a height and a hash together, and it is the only way this design names a position on
+a chain. A node carries the header, its parent hash, a height from a checked parent
+increment, the block work derived from the validated compact target, a
+`WorkCoordinate`, a header validation state, an eligibility state, a body validation
+state, and its auxiliary delivery ids. Beside the node map the graph keeps three
+indexes it can rebuild from the nodes alone: children by parent, hashes by height, and
+the set of eligible tips.
+
+Selection reads that graph. `select_best_header_chain` takes the maximum score over the
+eligible tips, so permuting arrivals cannot change the answer (LC-SELECT-04). A node is
```
p0mvn: Suggest keeping this document non-normative, focused on humans since we already have a detailed specificaiton

```suggestion
Selection reads that graph. `select_best_header_chain` takes the maximum score over the
eligible tips, so permuting arrivals cannot change the answer. A node is
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+the base plus whatever the overlay staged, writes land in the overlay's own maps, and
+the base never changes. `GraphDelta` is what the overlay extracts: a new finalized
+frontier when finality moved, the nodes to put, and the nodes to delete. A node the
+overlay inserted and then deleted appears in neither list, and a node it wrote back
+unchanged appears in neither, so the delta holds the difference and nothing else.
+
+Both types implement `HeaderGraphView` and `HeaderGraphEdit`, so selection, eligibility
+propagation, and the invariant checks run one implementation against committed state
+and against a staged one. The planner therefore derives a complete transition without
+copying the graph and without touching the live one.
```
p0mvn: It would be helpful to attach a UML diagram below to visualize these relationships for first-time readers.

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+`state_version` increments on every committed transition that changed a durable fact.
+`header_generation` increments when selection, header topology, header validation,
+eligibility, or finality moved. `verified_generation` increments when the verified
+projection moved or finality advanced.
+
+`Gate` in `ownership.rs` judges a response against the current snapshot and returns a
+`CompletionDecision`: `Current`, or `Stale` carrying a typed `StaleReason` of
+`HeaderGeneration`, `VerifiedGeneration`, `BranchAnchor`, `MissingOwner`, or
+`OwnerMismatch`. It reads the generations and the branch anchor, and it ignores
+`state_version`, because that counter also moves for transitions on branches the request
```
p0mvn: This section is particularly dense. I wonder if it can be simplified to key details / fewer types

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+
+`install_committed_transition` compares the live snapshot with the plan's `before()` and
+refuses with `StaleSource` when they differ. `plan_transition` takes `&self`, so two
+planners can run against one engine; only one of them can install.
+
+Disk leads memory leads observers, and that order is forced. The next restart reads
+disk, so a crash in this order leaves disk ahead of memory, which the startup audit
+repairs by rehydrating, and a crash in any other order leaves an observer holding a
+frontier that is not on disk, which nothing repairs. One transition makes one atomic
+`db.write`, so a crash leaves the whole batch or none of it. A failure between the write
```
p0mvn: I think this is an important section that is buried under AI-wording, forcing the reader to skim through key details.

I'd say something along the lines:

There are 3 types of state:
- In-memory base graph
- Copy-on-write overlay on top
- Durable store

The durable store is authoritative. We update the in-memory components only after a store is committed for recovery.

The overlay never writes to the base. Instead, it produces a set of changes after a durable commit, which the state machine validates.

See `HeaderChainEngine` `plan_transition` and `install_committed_transition` and its caller `HeaderRuntime` for details.

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+and refuses a finalized frontier that is missing or ineligible. `apply_delta` runs that
+check before it mutates, and `validate_delta` runs it without mutating.
+
+`verify_plan` checks the plan against a graph rebuilt from the delta with
+`GraphOverlay::from_delta`, not against the overlay the planner mutated, so the verifier
+approves the transition the engine will install rather than the one it staged.
+
+`install_committed_transition` compares the live snapshot with the plan's `before()` and
+refuses with `StaleSource` when they differ. `plan_transition` takes `&self`, so two
+planners can run against one engine; only one of them can install.
```
p0mvn: Not sure if this detail is key tbh. I could as well read the code
```suggestion
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+Once per committed transition, inside the writer lock, and nowhere else in the steady
+state. There is no accumulation and no background flush: the runtime builds one
+`DiskWriteBatch` from the change set and calls `db.write` once (LC-TXN-01). What else
+lands in that one write depends on which call the state writer made.
+
+| Call | What lands in the one write |
+| --- | --- |
+| `apply` | the header-chain rows alone |
+| `apply_combined` | those rows appended to the block batch the state writer already filled |
+| `apply_aux_then_checkpoint_combined` | auxiliary authentication and the checkpoint advance that depends on it, planned into the same batch |
```
p0mvn: The distinction is not immediately clear to me.

Are these operations meant to be committed atomically as part of one operation, or are they driven by a separate execution flow?

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+state-service requests and assembles the reactor's startup inputs, and it is the only
+file where header-sync policy and the state service know each other's names.
+
+`coordinator.rs` owns apply-phase transitions for the process. It holds the phase, a
+watch channel per observer, the in-flight count, a record per block-apply operation, and
+a drain notification, so the native and legacy paths hand off rather than overlap. It is
+the only new file in this layer with no deletions against it, because nothing like it
+existed.
+
+## Integration edits, 75 files, +7,483 / -5,471
```
p0mvn: Everything starting from here is an easy delete. Would only increase the document's value

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+to 1,000 headers by default and 4,000 at the cap, and the chunk budget divides that same
+4,000 across receiving, preparation, and application on every staged target, so one
+target cannot starve the others and the total in flight cannot exceed one transition's
+worth.
+
+`pipe.rs` handles the awkward case that follows from cancelling work. A retired request
+still gets an honest answer from the peer, so the pipe remembers up to 64 cancelled ids
+for 30 seconds and drops those responses instead of scoring them as protocol violations.
+
+## Layer 5: `zakurad`, 4 files, +3,797 / -2,274
```
p0mvn: I think this section could be one sentence or deleted

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+| `header_sync/scheduler/repair.rs` | 503 | 0 | generation- and branch-owned auxiliary VCT repair work |
+| `header_sync/events.rs` | 394 | 551 | `HeaderSyncStartup`, the event enum, the handle |
+| `header_sync/service.rs` | 393 | 574 | the stream declaration and per-peer session spawn |
+| `header_sync/pipe.rs` | 346 | 1,121 | `run_peer`, the per-peer decode loop and its cancelled-response grace window |
+| `header_sync/scheduler/completed_targets.rs` | 147 | 0 | generation- and branch-keyed completed targets |
+| `header_sync/scheduler/status.rs` | 140 | 0 | `StatusPublisher`, coalescing advertisements behind a one-per-second floor |
+| `header_sync/config.rs` | 38 | 130 | `ZakuraHeaderSyncConfig` |
+| `header_sync/mod.rs` | 30 | 55 | module surface |
+| `header_sync/scheduler/mod.rs` | 5 | 0 | scheduler module list |
+| `header_sync/error.rs` | 0 | 194 | reduced to the startup error; the rest became `HeaderChainError` and `PortError` |
```
p0mvn:
```suggestion
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+| `src/service/finalized_state/disk_format/header_chain_values.rs` | 1,825 | 0 | version-one value codecs for the new column families |
+| `src/service/finalized_state/disk_format/header_chain.rs` | 492 | 0 | ordered key encodings for children, heights, eligibility roots, deliveries, deferrals, finality |
+| `src/service/finalized_state/header_chain/migration.rs` | 424 | 0 | first-run construction of the DAG from authenticated full state |
+| `src/service/write.rs` | 1,945 | 727 | the write loop: the only production caller of `apply`, combined batches, the deferred timer |
+| `src/service.rs` | 588 | 265 | runtime construction and the read and write service split |
+| `src/error.rs` | 376 | 274 | state error taxonomy for the new failures |
+| `src/service/finalized_state.rs` | 326 | 117 | opening the store, running the startup audit, wiring the runtime |
+| `src/request.rs` | 293 | 146 | new state requests |
+| `src/header_chain.rs` | 82 | 0 | the public retained-path contract: `RetainedPathLease`, `RetainedPathPage`, outcomes |
+| `src/response.rs` | 59 | 12 | new state responses |
```
p0mvn:
```suggestion
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+This layer exists to make a dependency impossible rather than discouraged.
+`zakura-network` needs header-chain state, and `zakura-state` sits above it, so a direct
+call would invert the stack and put the database within the reactor's reach. The port is
+a trait in a crate below both. The reactor calls seven named operations and can reach
+nothing else.
+
+| File | Added | Role |
+| --- | --- | --- |
+| `src/header_chain.rs` | 693 | the `Port` trait, the `AdapterKey` seal, `PortError`, and the `UnavailablePort` and `InertHeaderChainPort` stand-ins |
+| `src/sync_lifecycle.rs` | 694 | `ApplyPhase` and its declared edges, `LifecycleEpoch`, `HeaderRuntimeStatus`, `SyncServiceDemand` |
```
p0mvn: These file sections aren't helpful. I can navigate to `zakura-node-services` and see the files myself

```suggestion
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+entry point traced above. `prepare_headers` is the other one: it runs parent linkage and
+contextual difficulty and time during preparation rather than leaving them to the
+planner, so it takes a `ValidationLease` that `is_coherent` re-derives. The tests and
+the fuzzer drive it; production does not.
+
+| File | Added | Role |
+| --- | --- | --- |
+| `src/validation/mod.rs` | 466 | per-header rules: link, encoding, compact target, hash filter, future time, commitment structure |
+| `src/validation/prepare.rs` | 691 | the two entry points and the sealed `PreparedHeaderBatch` |
+| `src/validation/contextual.rs` | 963 | `AdjustedDifficulty`, ThresholdBits, the ZIP 205 and 208 testnet rule, and median-time-past |
```
p0mvn:
```suggestion
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+and selection over committed state, and holds the two properties that keep arrival order
+out of fork choice: selection is `max` over a set, and height comes from a checked
+parent increment rather than the height a peer framed. `graph/overlay.rs` holds the
+staging types, which is what lets planning be pure. A plan the caller drops leaves
+nothing behind, so the durable caller chooses the order of its effects.
+
+| File | Added | Role |
+| --- | --- | --- |
+| `src/graph.rs` | 2,538 | in-memory DAG, ancestry and height queries, eligibility propagation, selection, tombstones |
+| `src/graph/overlay.rs` | 1,443 | `GraphOverlay`, `GraphDelta`, and the copy-on-write staging of one transition |
```
p0mvn: For this one, I'd probaly say "graph.rs and overlay.rs are the relevant files"

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+constrain what an event may say. Admission decides who may say it. The event effects
+derive each event's consequences against a staged projection. Settlement and the write
+set turn that projection into a plan. The verifier re-derives that result before disk
+can see it. Recovery is the one path that skips all of them, so it is separate.
+
+| File | Added | Role |
+| --- | --- | --- |
+| `src/transition/mod.rs` | 27 | the module's re-export surface |
+| `src/transition/types.rs` | 1,724 | every typed input and output: the twelve events, `TransitionRequest`, `EngineSnapshot`, change sets, `AuxDelivery`, `ValidationLease` |
+| `src/transition/planner.rs` | 335 | the phase sequence, `PlanCandidate`, and `TransitionPlan` |
```
p0mvn: Suggest simply linking the `transition` package and removing the overload

docs/design/headerchain/production_code_header-chain.md
p0mvn: From reading the code, I think this is similarly AI-overloaded

I'd say:

There are 3 kinds of "apply" functions `HeaderChainWriter` executes against `HeaderChainRuntime` to apply a transition against state:
- `apply`: Used when header sync progresses independently of block execution—such as admitting downloaded headers, reevaluating deferred headers, or recording body evidence. There is no block-state batch to couple with it.
- `apply_combined`: one header transition plus one block-state commit
Used when committing a block also advances the header chain’s verified/finalized view. Both must reach RocksDB atomically so block state and header-chain metadata cannot disagree. apply_combined_expected is the guarded variant used by the regular verified-block path.
- `apply_aux_then_checkpoint_combined`: two dependent header transitions plus one block-state commit. This is the VCT checkpoint fast path. The committed block authenticates auxiliary header data; that authentication can then enable the checkpoint advance. They are planned in order but written atomically with the block.


Side note: the fact it is non-trivial to infer the difference shows that we should probably choose better names and improve abstractions.

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,572 @@
...
+
+Exclusions applied, in order: `fuzz/`, every `fuzz.rs`, `/tests/` and `tests.rs`,
+`testkit/`, the `header_chain/coherence/` model harness, `bench.rs`, `arbitrary.rs`,
+`crates/xtask/`, `docs/`, `.github/`, `docker/`, `deploy/`, `qa/`, and lockfiles. Two
+excluded files still matter to a reader: `crates/zakura-header-chain/conformance.toml`
+(926 lines) is the machine-checked rule manifest, and
+`crates/xtask/src/header_conformance.rs` (1,285 lines) is the checker that enforces it.
+
+One caveat on every count below. Rust colocates unit tests, so a file's added-line
+figure includes its `#[cfg(test)]` module. In the new crate, 17,176 added lines are
```
p0mvn:
```suggestion
```

p0mvn (approved):
LGTM

This is very helpful. Propose to start and keep such human-tailored documents for any future implementations. I would even go one step further and suggest that we write out key details by hand, only using AI for amplifying writing velocity rather than content.

Main recommendations for this PR:
- De-AI (remove unnecessary sections, reduce overload with more key detail)
- Add some visualizations and diagrams to key parts of the architecture
- Consider adding more about VCT handling, how it is tracked in the DAG and authenticated
- What's the process / sequence for switching from one fork to another

I also think we could commit this directly against main after the comments are addressed.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Replaced "operator invalidation" with the behavior it represents. The document now explains that a node administrator uses `invalidateblock` to exclude a block and its descendants from selection, while `reconsiderblock` removes that reversible exclusion. The `Operator*` names now appear only where the text refers to code.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Kept the specification references, but replaced the bare rule ids with linked property names and one-sentence summaries. The guide explains each behavior in plain language while leaving the specification authoritative.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Added a sequence diagram that follows one transition from planning through publication. It shows when the overlay reads committed state, when the delta is created, and why the engine does not update memory before the durable write succeeds.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Rewrote the section around the late-result problem. It now introduces `BranchId` and `Gate` only when they solve that problem, and it explains the deliberate omission of height and `state_version` directly.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Rewrote this around the commit order shown in the sequence diagram: stage in the overlay, write disk, install the delta in memory, then publish. The text now states only the recovery consequence and points to the implementation for the detailed checks.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Deleted. The rewritten section points at both methods, which is enough.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Replaced the table with purpose-first descriptions of the three write paths.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Deleted the counting methodology. The conformance links now sit beside the `zakura-header-chain` discussion.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Deleted the table and linked `graph.rs` and `graph/overlay.rs` from the explanation of committed and staged graph state.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Replaced the transition inventory with one paragraph that follows a request through planning and verification. Detailed file links now appear beside the concepts they implement.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Replaced the file table and move history with the three validation stages. The specification links carry the exact rules.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Removed the file and method inventories. The section now explains the dependency boundary and the sealed preparation handoff.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: alrighty won't include file inventories in the future. The section now explains which rows are authoritative, which rows startup can rebuild, and why the column-family names have a version suffix. The schema link carries the complete list.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: sg, replaced the file table and numeric inventory with the three network behaviors a reader needs: compatibility, branch-owned scheduling, and handling late responses.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Reduced Layer 5 to one sentence that names its two responsibilities.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Deleted the three sections. Relevant code links now appear beside the concepts they implement, so the document no longer ends with another inventory.

evan-forbes:
>I would even go one step further and suggest that we write out key details by hand, only using AI for amplifying writing velocity rather than content.

I'm down but might not have a 100% complete idea on what it means to increase velocity but not content

p0mvn:
> > I would even go one step further and suggest that we write out key details by hand, only using AI for amplifying writing velocity rather than content.
> 
> I'm down but might not have a 100% complete idea on what it means to increase velocity but not content

When reading some of our spec drafts, I often feel that most of the content and structure were developed by AI from a prompt.

I can get the idea of what the AI is trying to convey, but the content is not cohesive: some details are unimportant (probably would not include them if writing manually) while key areas are buried under layers of less critical context.

As a human trying to model the proposal a document is producing, it takes more effort to grok what actually matters. It makes you worse as a reviewer. For an engineer reading, it also increases the chances of mismodeling the system and being ineffective at future extension or bug diagnosis.

That's why, for these specs, we must put deliberate thought and effort into making them focused on key details. The higher the quality of the initial document draft and the code that mirrors it, the greater the velocity and the impact we will make as a team.

**TLDR**

For these specs, the **cases where the use of AI is appropriate are editorial**: syntax corrections, paraphrasing a manually written thought to make it flow better, adding diagrams, etc. I see these as increasing the velocity of producing a high-quality document.

But **the content should be human-driven**.

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+header in a directed acyclic graph (DAG). Each `HeaderNode` is keyed by its consensus
+hash and names its parent. The finalized `Frontier` is the graph root. A `Frontier`
+contains both height and hash, so it identifies one position on one branch.
+
+The engine keeps two paths through the DAG. The selected path ends at the best eligible
+header tip. The verified path ends at the tip whose blocks full state has accepted.
+Each path is stored as an ordered list of frontiers, so readers can answer height queries
+without walking the graph.
+
+Integrated mode advances the verified path only when full state accepts blocks. In
```
p0mvn: We mention "Integrated mode" but don't define what modes are or why we need different modes.

Should first define and explain it before referring to it

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+[`MemHeaderStore`](../../../crates/zakura-header-chain/src/graph/mod.rs) holds every retained
+header in a directed acyclic graph (DAG). Each `HeaderNode` is keyed by its consensus
+hash and names its parent. The finalized `Frontier` is the graph root. A `Frontier`
+contains both height and hash, so it identifies one position on one branch.
+
+The engine keeps two paths through the DAG. The selected path ends at the best eligible
+header tip. The verified path ends at the tip whose blocks full state has accepted.
+Each path is stored as an ordered list of frontiers, so readers can answer height queries
+without walking the graph.
+
```
p0mvn, line 55: Consider adding
```
finalized ── A ── B ── C          verified path (bodies already applied)
              ╲
               D ── E ── F        selected path (more work, headers only)
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+Integrated mode advances the verified path only when full state accepts blocks. In
+headers-only mode, the verified path contains only the finalized frontier.
+
+A node is eligible when its header is valid, it has no direct exclusion reason, and its
+ancestors are eligible. The graph stores exclusion reasons instead of one boolean flag.
```
p0mvn: What are some examples of the "exclusion" reasons?

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+
+Integrated mode advances finality only from authenticated full-state evidence.
+Headers-only mode advances finality with a 1,000-block local depth rule. This local rule
+does not verify block bodies. When finality advances, the engine removes old ancestors
+and competing sibling subtrees. It rebases retained work onto the new root.
+
+## Header admission and validation
+
+The header-sync driver prepares a downloaded batch before it takes the state writer
+lock. It obtains a validation lease for the common ancestor and checks properties that
```
p0mvn: What is a "validation lease"? Would it make sense to define it?

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+checks that the prepared batch still belongs to the current network, trust anchor, and
+parent context.
+
+The planner then checks parent linkage, height, work, difficulty, and time against the
+staged graph. It admits a header only after every required check passes. This implements
+[validation before admission (`LC-VAL-11`)](../../specs/fork-aware-header-chain-engine.md#lc-val-11).
+The engine calculates cumulative work with full 256-bit values and rejects cumulative
+overflow. Its difficulty calculation requires the complete predecessor window. It caps
+target scaling before multiplication can overflow. It reads Testnet's maximum-time
+activation height from network parameters.
```
p0mvn: This does not seem to be a useful detail for a high-level architecture-centric document

```suggestion
overflow. Its difficulty calculation requires the complete predecessor window. It caps
target scaling before multiplication can overflow.
```

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+
+`MemHeaderStore` contains the committed graph.
+[`GraphOverlay`](../../../crates/zakura-header-chain/src/graph/overlay.rs) reads that graph
+and records staged changes without mutating it.
+[`HeaderChainEngine`](../../../crates/zakura-header-chain/src/transition/engine/mod.rs)
+extracts those changes as a `GraphDelta`. The runtime applies the delta to
+`MemHeaderStore` only after the durable write succeeds.
+
+```mermaid
+sequenceDiagram
```
p0mvn, line 168: This is great.

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+nodes before the runtime writes anything.
+
+## Transition commit order
+
+`MemHeaderStore` contains the committed graph.
+[`GraphOverlay`](../../../crates/zakura-header-chain/src/graph/overlay.rs) reads that graph
+and records staged changes without mutating it.
+[`HeaderChainEngine`](../../../crates/zakura-header-chain/src/transition/engine/mod.rs)
+extracts those changes as a `GraphDelta`. The runtime applies the delta to
+`MemHeaderStore` only after the durable write succeeds.
```
p0mvn, line 151: I think it would be valuable to expand on why we actually need the overlay to explain why a single in-memory store was insufficient.

I spent a while talking to AI about this

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+  participant P as Peer
+  participant N as Header-sync reactor
+  participant D as Header-sync driver
+  participant S as State service
+  participant E as HeaderChainEngine
+  participant O as Subscribers
+
+  P-->>N: headers on an alternate branch
+  N->>D: prepare_header_target
+  D->>S: request validation lease
```
p0mvn, line 268: Still think that we under-defined validation lease in this spec

docs/design/headerchain/production_code_header-chain.md
```diff
@@ -0,0 +1,357 @@
...
+fork switch can replace a branch without changing its height.
+
+When a result returns, `Gate` in
+[`completion.rs`](../../../crates/zakura-header-chain/src/work/completion.rs) compares its branch
+and generation with the current snapshot. It accepts current work and rejects stale
+work. It ignores `state_version` because unrelated transitions increment that counter
+and would cancel valid requests. The scheduler uses the same branch and generation to
+retire stale work.
+
+The engine advances `header_generation` when a change makes header work stale. It
```
p0mvn: I wonder if any other temporal abstractions would make sense to document.

For example, I recall "work_epoch" in one of the recent PRs.

p0mvn (approved):
Great work.

Really happy with how this document is turning out.

Have some recommendations still to clarify or expand on some parts.

It would be great to start the development from constructing and revieiwng a spec like this in the future.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Added an "Engine modes" section ahead of the first reference. It defines integrated and headers-only, says which one `zakurad` runs, states that the mode changes only what advances the verified path and what advances finality, and links the headers-only finality disclosure. The two later paragraphs that restated the difference are gone.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Added, with a line saying that header sync follows the selected path while full state moves the verified path behind it.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Listed the five `EligibilityReason` variants: `SettledUpgradeConflict`, `CheckpointConflict`, `FinalityConflict`, `ConsensusBodyInvalid`, and `OperatorInvalid`. Only the last is reversible, which is the reason the graph stores a set rather than a flag.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: :+1: added a definition. 

A validation lease is the durable header context, sealed: the parent frontier, up to 28 headers in reverse height order from the parent, the network parameters, and a trust-anchor digest, all bound by one context digest.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Applied.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: sg added a definition

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Added two paragraphs. The short version: the engine cannot know whether the durable write will succeed, so a single mutable graph would have to change before that write and would then need a second, inverse mutation path to undo a partial change. Staging means `MemHeaderStore` is untouched until the write returns success, so a failed write leaves memory identical to what disk still holds and recovery keeps what it has. Staging also produces a complete result to check.

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Documented the full set: `state_version`, `header_generation`, `verified_generation`, and `finality_epoch` as the four monotonic counters (plus `headers_only_migration_epoch`

docs/design/headerchain/production_code_header-chain.md
evan-forbes: Thanks! sequence diagrams ftw

PR #346, opened 2026-07-22
feat(chain): seal verified header commitment roots

p0mvn:
## Motivation

Header-root authentication needs a compile-time boundary that distinguishes raw peer payloads from roots successfully verified against successor header commitments. This foundation should be independently reviewable before state persistence, migration, and network scheduling are added.

## Solution

- seal successful supplied-root verification behind `VerifiedHeaderCommitmentRoots`
- return only the prefix authenticated by successor headers
- preserve direct pre-activation checks for Sapling, Orchard, and Ironwood roots
- expose the confirmed header hashes and advanced history tree needed by later state integration

This PR intentionally makes no database, state-service, migration, serving, or network-scheduling changes.

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p zakura-chain parallel::commitment_aux_verify::tests` (6 passed)
- `./scripts/changelog.py check`
- `git diff --check main...HEAD`
- IDE diagnostics report no errors in the changed file

The focused tests cover real commitment roots, the one-header confirmation lag, corrupted supplied roots, and pre-activation empty-root constraints.

## Changelog

- Added `changelog-unreleased/346.md` with an internal-only exclusion.

### Specifications & References

- Design: #230
- Extracted from draft #323

### Follow-up Work

State-owned durable frontiers, authenticated persistence, migration cutover, and overlapping network scheduling remain in focused follow-up PRs.

PR #632, opened 2026-08-12
docs: move VCT root authentication under docs/design/headerchain/

evan-forbes:
Follow-up to #631, which introduced `docs/design/headerchain/`.

## Why

`docs/design/` spells one subsystem three ways: the new `headerchain/` directory, the `header-sync-` file prefix, and `docs/specs/blocksync/` on the spec side. This moves the root-authentication document into `headerchain/` and drops the prefix the directory now carries.

## What changed

| Before | After |
| --- | --- |
| `docs/design/header-sync-vct-root-authentication.md` | `docs/design/headerchain/vct-root-authentication.md` |

Git records it as a 100% rename, so the file's contents are untouched. The other seven lines are the links in `verified-commitment-trees.md` that pointed at the old path.

## What deliberately did not change

`verified-commitment-trees.md` stays at the top level of `docs/design/`. It describes the fast-sync machinery rather than the header chain, and fifteen Rust files, two scripts, one workflow, and two READMEs cite it by path, so moving it would put Rust source in a docs-only PR.

The moved document keeps its `# Header-sync VCT root authentication` heading. It specifies the lane that shipped in #346, #351, and #352, and the header-chain stack renames that subsystem in code, so this move leaves the prose to that stack.

`docs/specs/blocksync/` also stays. Renaming a spec directory would collide with the header-chain stack.

## Base

`docs/design/` is byte-identical on `main` and `hc/01-engine-model`, so this targets `main` directly and conflicts with nothing in the stack.

## Checks

`codespell` and `markdownlint` both exit 0. A local resolver over every markdown link in the repo reports zero broken links from this change; the three it does report are pre-existing on `main` and `Docs Check` is green there.

Docs-only, so no changelog fragment per `docs/changelog/guidelines.md`.


evan-forbes:
Closing unmerged. #617 retires `header-sync-vct-root-authentication.md` to historical reference only, so relocating it into `docs/design/headerchain/` would file a retired document in the live header-chain design directory. The move would also break the sibling link `[verified-commitment-trees.md](verified-commitment-trees.md)` that #617's new banner adds, since the file would sit one level deeper.

`docs/design/headerchain/` will hold the new production-code document from #631 and nothing else.

PR #630, opened 2026-08-12
refactor(header-chain): split transition planner into phases

p0mvn:
## Motivation

The transition planner combined admission, replay binding, domain evidence, finality, retention, write derivation, and invariant verification in one large function and event match. That made ordering-sensitive behavior and coupled projected state difficult to review safely.

## Solution

- express transition planning as six named phases with one verified-plan boundary
- encapsulate projected graph, verified-path, and auxiliary mutations in typed state
- dispatch event evidence to discoverable domain handlers
- add table-driven coverage for configuration coherence, finality rebase histories, and predecessor-context reconstruction
- make Lychee installation retry transient release-download failures and verify the pinned archive checksum

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p zakura-header-chain --all-targets -- -D warnings`
- [x] `cargo clippy -p zakura-header-chain --features fuzz-impl -- -D warnings`
- [x] `cargo test -p zakura-header-chain`
- [x] `./scripts/changelog.py check`
- [x] Docs Check workflow: link-check, spell-check, and markdown-lint

## Changelog

Internal-only refactor: `docs/changelog/unreleased/630.md` uses the explicit no-changelog marker.

### Specifications & References

- `docs/specs/fork-aware-header-chain-engine.md`

PR #633, opened 2026-08-12
refactor(header-chain): type transition boundary contracts

p0mvn:
## Motivation

The header-chain transition boundary mixed three different concerns into shared shapes:

- authenticated evidence of what happened
- the write path's observed durable version
- durable store facts the planner is allowed to consume

That made illegal combinations representable: events could be paired with unrelated facts via a shared fact bag, evidence failures were stringly, and `TransitionCause` conflated submitted domain with derived effects.

## Solution

Introduce explicit typed contracts at each ownership boundary:

### `TransitionEvent` — authenticated evidence

Records only **what happened** (header admission, body evidence, operator action, and so on). Built by the state write path after authenticating higher-level work; peers never submit it directly. Fingerprinted for replay. Does **not** carry durable store facts or desired consequences.

### `TransitionRequest` — pre-fact write command

`{ expected_version, event }`. The uniform command the write path builds **before** loading store rows. Still version-qualified for API uniformity even when the eventual engine input is owner-qualified.

### `TransitionInput` — engine-boundary package

The adapter binds each event to **exactly** the durable facts that variant may consume, then calls `plan_transition`. The planner does not read the durable store itself. Exhaustiveness is the contract:

- header insertion → validation leases + finality rebase history
- verified path changes → validation leases
- migrated pin refutation → preserved pin
- most events → `expected_version`
- `InsertHeaders` / `AuxEvidence` → omit version; freshness is owner-qualified

### `TransitionDomain` + `TransitionEffect` — identity vs outcomes

Replace `TransitionCause` with:

- **`TransitionDomain`**: submitted input identity (stable disk discriminants unchanged)
- **`TransitionEffect`**: composable planner-derived outcomes (`HeaderWorkEffect`, `FinalityEffect`, `AuxiliaryEffect`, resource stall)

Callers submit evidence; they never submit desired consequences.

### `InvalidTransitionEvidence` — structured failures

Replace `InvalidEvidence(&'static str)` with nested typed violations (`Limit`, `Finality`, `Header`, `Body`, `Auxiliary`, `Operator`, `Planner`).

State adapters, tests, and fuzz support are migrated to the new API.

## Testing

- [x] `cargo clippy -p zakura-header-chain -p zakura-state --all-targets -- -D warnings`
- [ ] `cargo test -p zakura-header-chain --lib --tests` (interrupted; will rerun)
- [ ] `cargo test -p zakura-state --lib -- selection_and_evidence header_chain::tests::`

Builds on the phased planner from #630.

### Follow-up Work

None required for the typed boundary itself; remaining engine-model review items stay on the `hc/01-engine-model` series.

### DIFF
--- a/crates/zakura-header-chain/src/fuzz.rs
+++ b/crates/zakura-header-chain/src/fuzz.rs
@@ -19,17 +19,18 @@ use zakura_chain::{
 use crate::{
     AlarmSet, AuxDelivery, AuxDelta, BodyCommitmentKind, BodyEvidence, BodyPayloadMismatch,
     BodyRuleId, BodyUnavailableSummary, BodyValidationState, BranchId, ChainScore, CheckpointSet,
-    Clock, ConsensusBodyInvalid, DurableTransitionFacts, EngineConfig, EngineMetadata, EngineMode,
-    EngineSnapshot, EvidenceId, FinalityEpoch, FinalityRecord, FinalitySource, Frontier,
-    FrontierSet, FullStateEvidenceAuthority, FullStateFinalized, GraphError, HeaderBatchInput,
+    Clock, ConsensusBodyInvalid, EngineConfig, EngineMetadata, EngineMode, EngineSnapshot,
+    EvidenceId, FinalityEpoch, FinalityRecord, FinalitySource, Frontier, FrontierSet,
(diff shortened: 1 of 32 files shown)


evan-forbes (approved):
cool beans

PR #634, opened 2026-08-12
refactor(header-chain): decompose transition types and recovery modules

p0mvn:
## Motivation

`types.rs` and `recovery.rs` were overloaded the same way `planner.rs` used to be: high-level APIs were hard to see, and phase helpers lived in the same file. This PR is a pure module move that makes the transition surface easier to navigate without changing behavior.

## Solution

- Split `transition/types.rs` into a private `types/` package (error, snapshot, preparation, auxiliary, event/replay, write-set, outcomes), with `StoreError` folded into `types/error.rs`.
- Split `transition/recovery.rs` into a phased `recovery/` package: contracts → load → authoritative audit → reconstruction → repair classification.
- Keep crate-root exports and public signatures stable (`pub use types::*`, unchanged recovery entry points).
- Reorganize recovery tests by authority vs reconstructible-repair behavior; replace the single-file types layout guard with semantic domain/event coverage.

No intentional behavior, encoding, or recovery-outcome changes.

## Testing

Local verification:

- `cargo test -p zakura-header-chain --lib`
- `cargo test -p zakura-header-chain --test production_transition`
- `cargo test -p zakura-state --lib service::finalized_state::header_chain::tests::startup`
- `cargo test -p zakura-state --lib service::finalized_state::header_chain::tests::crash::every_crash_boundary_reopens_to_complete_transition`
- `cargo check` for direct consumers (`zakura-state`, `zakura-network`, `zakura-node-services`, `zakura-consensus`, `zakura`)

Draft CI will provide the full check suite.

## Changelog

[`docs/changelog/unreleased/634.md`](docs/changelog/unreleased/634.md) — explicit no-changelog for an internal-only module reorganization.

### Follow-up Work

- Curate crate-root exports (drop the types wildcard).
- Make `RecoveryPlan` / `TransitionEffect` opaque with read-only accessors.
- Optionally relocate VCT publication and retired-work types to their owning modules.


evan-forbes (approved):
easy approve v nice thanks!

PR #636, opened 2026-08-12
docs(skills): convert core-rust-readability into an explicit skill

p0mvn:
## Motivation

Auto-attached / glob-scoped rules make context control opaque. Prefer a skill over always-on rules so context is not forced on every core-crate read.

## Solution

- Move the core Rust readability guidance (including Module Packages) to [`.agents/skills/core-rust-readability/SKILL.md`](.agents/skills/core-rust-readability/SKILL.md).
- Recommend the skill for key Rust crates, but confirm with the user during planning instead of auto-applying; if implementation is already underway, do not auto-apply or interrupt.
- Remove the `.agents/rules` / `.cursor/rules` auto-attached copies.
- Do not add an `AGENTS.md` pointer.

## Testing

Docs-only. Verified the skill exists under `.agents/skills/` and rule shims are gone.

## Changelog

Not applicable (no Rust or `Cargo.toml` changes).

### DIFF
--- a/.agents/skills/core-rust-readability/SKILL.md
+++ b/.agents/skills/core-rust-readability/SKILL.md
@@ -3,15 +3,19 @@ name: core-rust-readability
 description: >-
   Keep core Rust code semantically named, discoverable, cohesive, and
   understandable as a black box. Prefer facade modules with private children
-  over overloaded files. Use when modifying key Rust crates (zakura-chain,
-  zakura-consensus, zakura-header-chain, zakura-network, zakura-rpc,
-  zakura-script, zakura-state), or when the user asks for readability,
-  naming, module decomposition, or facade-package refactors in those crates.
+  over overloaded files. For key Rust crates (zakura-chain, zakura-consensus,
+  zakura-header-chain, zakura-network, zakura-rpc, zakura-script, zakura-state),
+  confirm with the user during planning before applying; do not auto-apply
+  mid-implementation. Also use when the user asks for readability, naming,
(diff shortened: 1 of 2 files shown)


evan-forbes (commented):
personally I like having control over the context

the rules on when this get applied are opaque, so would prefer an explicit call

recently I modified things like conorbronsdon/avoid-ai-writing to not even be part of the skills list that gets prepended to each session. while I always want to get rid of sloppy ai writing, always activating a bunch of skills implicitly to me seems needless when I go through and generate a first, draft, then have a separate run after compaction purely dedicated to a specific thing.

then skills like avoid-ai-writing can use 20-50k tokens and it be okay since its never happening on accident

especially since there's only so much that can fit in attention, so we likely need to call multiple times anyway

evan-forbes (approved):

p0mvn:
> personally I like having control over the context
> 
> the rules on when this get applied are opaque, so would prefer an explicit call
> 
> recently I modified things like conorbronsdon/avoid-ai-writing to not even be part of the skills list that gets prepended to each session. while I always want to get rid of sloppy ai writing, always activating a bunch of skills implicitly to me seems needless when I go through and generate a first, draft, then have a separate run after compaction purely dedicated to a specific thing.
> 
> then skills like avoid-ai-writing can use 20-50k tokens and it be okay since its never happening on accident
> 
> especially since there's only so much that can fit in attention, so we likely need to call multiple times anyway

Addressed and coverted to skill

PR #618, opened 2026-08-12
feat(header-chain): add reviewed graph core

evan-forbes:
## Motivation

`docs/specs/fork-aware-header-chain-engine.md` (v1.4) is the normative design
oracle for the fork-aware headers-only chain engine that replaces PR #229. This
PR is the primary feature branch for that engine: it lands the graph core every
later slice builds on, and subsequent reviewed components stack on top of it.

## Solution

Adds the in-memory header DAG and its staged overlay.

**`MemHeaderStore` (`graph.rs`)**

- Hash-keyed fork DAG allowing multiple headers per height, with reconstructible
  parent→children, height→hashes, and eligible-tip indexes (LC-DAG-01, LC-DAG-02).
- Eligibility as durable direct reasons plus inherited ancestry, propagated to
  descendants on operator invalidation, body-validation, and header-validation
  changes (LC-BODY-02, LC-OP-01, LC-OP-02).
- Deterministic greatest-work selection over complete suffixes from the work
  anchor, with the raw `block::Hash.0` equal-work tie-breaker (LC-SELECT-01
  through LC-SELECT-04) on exact checked 256-bit work (LC-WORKCALC-01).
- Finality advance that re-roots at an eligible frontier, prunes every
  non-descendant, and clears inherited ineligibility from the new root
  (LC-FINAL-01, LC-FINAL-02).

**`GraphOverlay` / `GraphDelta` (`graph/overlay.rs`)**

- Stages node, edge, and eligible-tip changes over a borrowed base store without
  mutating it, then extracts a validated delta.
- Gives the single atomic transition path required by LC-TXN-01 and LC-TXN-02
  one place to compute and validate a whole transition before commit.

`HeaderGraphView` and `HeaderGraphEdit` let fork-choice, planning, and invariant
code run identically against the committed graph and against a staged overlay.

## Scope

The component is deliberately unwired here: the crate has no `Cargo.toml` and is
not a workspace member, so its supporting domain types (`HeaderNode`, `Frontier`,
`ChainScore`, eligibility and work coordinates), crate wiring, and graph
unit-test execution land in the next slice on this branch.

## Testing

- `cargo fmt --all -- --check`
- `rustfmt --edition 2021 --check crates/zakura-header-chain/src/graph.rs crates/zakura-header-chain/src/graph/overlay.rs`
- `cargo test -p zakura-header-chain graph --lib` on the feedback branch that
  supplies the supporting types, against byte-identical graph files: 21 passed
- `./scripts/changelog.py check`
[...]

PR #637, opened 2026-08-12
refactor(chain)!: widen block work to exact 256-bit values

p0mvn:
## Motivation

`Work` and `PartialCumulativeWork` store the Zcash specification's implied 256-bit
work values in a `u128`. The specification defines block work as
`floor(2^256 / (ToTarget(nBits) + 1))`, so the narrower type cannot represent work
for hard targets, and `TryFrom<ExpandedDifficulty>` returned `Err` whenever the
computed value exceeded `u128::MAX`. On Mainnet total chain work is around 2^58, so
the truncation is invisible today, but it makes valid hard targets unrepresentable
and converts them into a failed conversion rather than a comparable value.

This is extracted as a standalone PR because it is a breaking change to a published
API and it alters consensus-adjacent conversion behaviour. It should be reviewed on
its own merits rather than merging as a silent dependency of a later crate.

## Solution

Store both types as `U256`:

- `Work::as_u128` / `PartialCumulativeWork::as_u128` become `as_u256`. This is the
  breaking part, so `zakura-chain` takes a major version bump to 5.0.0 and the 13
  workspace requirements move with it.
- `TryFrom<ExpandedDifficulty> for Work` replaces the `u128` overflow branch with
  explicit cases: a zero target has work 2^256 and stays `Err`, and `U256::MAX`
  returns `Ok(1)` because its mathematical denominator is 2^256.
- `PartialCumulativeWork` gains `checked_add` / `checked_sub`. The existing `Add`,
  `AddAssign`, `Sub` and `SubAssign` operators are kept with their previous
  panic-on-overflow semantics, so no caller changes its error handling here.
- Display paths (`difficulty_multiplier_for_display`, `difficulty_bits_for_display`,
  `Debug`) go through a `u256_to_f64` helper instead of `as u128 as f64`.
- `CompactDifficulty` gains `to_le_bytes` / `from_le_bytes` for the exact four-byte
  consensus representation.

Separately, `Solution::validate_shape` is extracted from `check_solution` so callers
can validate a solution's encoded shape against the network parameters without
running the Equihash check. `check_solution` calls it first and keeps its previous
behaviour.

Call-site updates are mechanical:

- `primitives/zcash_history.rs` builds the ZIP 221 V1 node work field from
  `as_u256().to_big_endian()` instead of `as_u128().to_be_bytes()`. Both produce the
  same `U256` for every value below 2^128, so the committed history root is
  unchanged for any real chain.
- `zakura-state` reports `LowerWorkConflict` work through a saturating helper. The
  fork-choice comparison above it stays exact 256-bit; only the diagnostic narrows.
- `read::difficulty::solution_rate` keeps its `Option<u128>` return type, so the
  `getnetworksolps` RPC response type is unchanged. The division is done in `U256`
  and saturates rather than panicking.

### Why this is safe

- **Not persisted.** Neither type has an `IntoDisk` / `FromDisk` impl; cumulative
  work is recomputed at load. No database format change and no migration.
- **Fork choice is unchanged for real chains.** Ordering is derived, and `U256`
  ordering agrees with `u128` ordering for every value below 2^128. Mainnet total
  chain work is around 2^58.
- **The reachable behaviour change is a panic removal.** Previously a target low
  enough to produce work above 2^128 made `to_work()` return `None`, which reached
  `.expect("work must be valid during contextual verification")` in the history-tree
  code. That is unreachable under any Mainnet or Testnet `powLimit` but constructible
  on regtest and custom networks. It now computes the exact value instead.

## Testing

Run on this branch:

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo check --workspace --all-targets` — clean
- `cargo test -p zakura-chain` — 382 passed, 0 failed
- `cargo test -p zakura-state` — 458 passed, 0 failed, 3 ignored, plus 12 passing
  across the integration binaries

The `work/difficulty` vectors and proptests are updated to exercise the widened
type directly, including the new zero-target and `U256::MAX` boundaries and a
round-trip above `u128::MAX` that the previous representation could not hold.
`round_trip_work_expanded` in `zakura-state` now round-trips the exact 256-bit value.
`validate_shape_checks_the_solution_against_the_network` covers the extracted shape
check on both Mainnet and Regtest, in both directions.

## Changelog

Fragment added in `docs/changelog/unreleased/637.md`.

- Work definition: [Zcash protocol specification, section 7.7.5](https://zips.z.cash/protocol/protocol.pdf#workdef)
- nBits conversions: [section 7.7.4](https://zips.z.cash/protocol/protocol.pdf#nbits)

### Follow-up Work

This unblocks the `zakura-header-chain` crate, whose graph and transition code
requires exact 256-bit work and `Solution::validate_shape`. Note that #618 already
calls `work.as_u256()` in `crates/zakura-header-chain/src/graph.rs`, but that crate
has no `Cargo.toml` and is not a workspace member on that branch, so nothing compiles
it yet. This PR should land before the PR that adds the crate manifest.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8357,7 +8357,7 @@ dependencies = [
 
 [[package]]
 name = "zakura-chain"
-version = "4.2.0"
+version = "5.0.0"
 dependencies = [
  "bech32",
  "bitflags",
--- a/crates/zakura-chain/Cargo.toml
+++ b/crates/zakura-chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zakura-chain"
-version = "4.2.0"
+version = "5.0.0"
 authors.workspace = true
 description = "Core Zcash data structures for the Zakura node. Internal crate, published to support cargo install zakura"
 license.workspace = true
--- a/crates/zakura-chain/src/primitives/zcash_history.rs
+++ b/crates/zakura-chain/src/primitives/zcash_history.rs
@@ -403,8 +403,7 @@ impl Version for zcash_history::V1 {
             .difficulty_threshold
             .to_work()
             .expect("work must be valid during contextual verification");
-        // There is no direct `std::primitive::u128` to `bigint::U256` conversion
-        let work = primitive_types::U256::from_big_endian(&work.as_u128().to_be_bytes());
+        let work = primitive_types::U256::from_big_endian(&work.as_u256().to_big_endian());
 
         match network_upgrade {
             NetworkUpgrade::Genesis
--- a/crates/zakura-chain/src/work/difficulty.rs
+++ b/crates/zakura-chain/src/work/difficulty.rs
@@ -103,7 +103,7 @@ pub const INVALID_COMPACT_DIFFICULTY: CompactDifficulty = CompactDifficulty(u32:
 #[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct ExpandedDifficulty(U256);
 
-/// A 128-bit unsigned "Work" value.
+/// A 256-bit unsigned "Work" value.
 ///
 /// Used to calculate the total work for each chain of blocks.
 ///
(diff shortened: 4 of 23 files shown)


p0mvn:
@cursor review

evan-forbes (approved):

p0mvn:
[F-219198](https://v12.sh/runs/6468/219198) was closed.

[F-219197](https://v12.sh/runs/6468/219197) describes a pre-existing accumulator panic, not a regression introduced by the U256 widening. Fixing it requires broader state-layer error propagation outside this change. Since exploitability depends on exposing a PoW-disabled deployment to untrusted peers, I’m omitting it from this review’s scope.

[F-219199](https://v12.sh/runs/6468/219199) if hit is a process panic/availability bug if hit

PR #641, opened 2026-08-13
docs(header-chain): trace one request through the fork-aware path

evan-forbes:
## Motivation

The fork-aware header chain in #586 splits fork choice across crates so that no component
can restate the rule in its own terms. That claim lives in the seams between crates rather
than in any one file, so a reader working through the code file by file cannot check it.
#631 maps the specification onto the crates and files that implement it. This document
follows one request through them and explains what each boundary prevents.

## Solution

Adds `docs/design/headerchain/request_lifecycle.md`, which follows one delivery of headers
from the socket to the planner and back.

- Opens with the problem: four sources can move the tip, and a node that lets each source
  compute the move runs one fork-choice implementation per source and still loses to
  ordering when those implementations agree.
- Names the planner, `HeaderChainEngine::plan_transition`, shows its signature, and gives
  each of the four layers one decision and one prohibition. The sequence diagram arrives
  after that model rather than before it.
- Takes the boundaries in order. Each section states the problem the boundary solves, then
  the mechanism, then the interface as a code block, then the types that make the mistake
  unrepresentable.
- Answers the questions the boundaries raise: when a late response is stale and why
  `state_version` is excluded, why the runtime rather than the peer supplies the validation
  context, how a note-commitment root is authenticated when no lease can seal its history,
  how much history each event needs, and what the four validation passes check.

## Verification

Every backticked identifier, every cited path, and every relative link resolves against
`main` at `7eb6c9f52`, and every `LC-*` citation carries the property name the specification
gives that rule. The document states no line counts, variant counts, or constants, so
nothing in it goes stale as the tree moves; the specification defines the rules and the code
defines the values.

`codespell` passes locally under the repo config, and the markdown structure checks pass.

## Review notes

Rebased onto `main`. #631 has merged, so this guide now links that implementation guide as
the map to read first instead of naming it in passing.

Review feedback said readers got lost in the earlier draft, so this revision rewrites the
document around its argument:

- The order is now the problem, one planner, four layers, then the round trip. The
  decision-authority table sits ahead of the sequence diagram.
- Each interface is a code block: the two `Port` operations, the three runtime `apply`
  shapes, `TransitionInput`, the `EngineTransition` caller surface, `Gate::check`, and the
  `Publisher` surface. The prose labels that stood in for those signatures are gone.
- Headings take title case and drop their subtitles.
- The unnamed production failure is gone. The ordering failure it referred to is stated
  concretely where the sole-publisher rule prevents it.
- Wire framing, schema evolution, and per-message limits defer to a message-expectation and
  DoS-resistance specification, which replaces the `LC-WIRE-15` digression in the peer
  section.

One claim also changed against `main`: `EngineTransition`'s caller surface includes
`snapshot_after_commit()`, which the runtime publishes, and the earlier draft omitted it.

Where the contained specs use different names (`Engine`, `TransitionFacts`, `Authorizer`,
`PreparedHeaders`, `CompletionGate`), this document uses the names `main` has
(`HeaderChainEngine`, `HeaderValidationFacts`, `FullStateEvidenceAuthority`,
`PreparedHeaderTarget`, `Gate`).

Documentation only, no Rust or `Cargo.toml` change, so no changelog fragment per
`docs/changelog/guidelines.md`.


evan-forbes:
okay definitely not perfect but meaningfully improved and should be acceptable

not fully human quality yet, but generally filled in what I was going for

p0mvn (approved):
Thanks for adding this; it will be very helpful for historical reference.

I've lost the deep context of the details, as I haven't been working on this, but I could follow the general architecture from memory while reading.

docs/design/headerchain/request_lifecycle.md
```diff
@@ -0,0 +1,542 @@
...
+of them names a tip, and the planner derives every consequence from what they report: which
+headers the node admits, which chain is best, how far finality moves, what the node retains,
+and which rows the store writes. The documentation on `TransitionRequest` states the rule in
+one line: callers never submit desired consequences.
+
+`&self` means the call mutates nothing, so a plan is a value the caller can inspect, commit, or
+drop. The planner runs six phases in order: it authenticates and admits the evidence, binds
+replay and freshness, projects the evidence onto staged state, derives finality and retention,
+assembles the write set, and verifies the result against the invariants. The same evidence
+against the same state produces the same plan.
```
ValarDragon, line 65: Not quite precise, as we check `context.clock.now()`, which is non-obvious here

ValarDragon: We need that for the 2 hours in the future check I believe?

docs/design/headerchain/request_lifecycle.md
```diff
@@ -0,0 +1,542 @@
...
+// crates/zakura-state/src/service/finalized_state/header_chain.rs
+// Header-chain rows alone.
+pub fn apply(
+    &self,
+    request: TransitionRequest,
+    context: &TransitionContext<'_>,
+) -> Result<ApplyResult, HeaderChainStoreError>
+
+// Those rows appended to the block batch the state writer already filled.
+pub(in crate::service) fn apply_combined<M: FnOnce()>(
```
ValarDragon, line 267: AI is complaining that we didn't also include:

```
apply_combined_expected
```

PR #643, opened 2026-08-13
fix(miner): cancel stale Equihash work promptly

evan-forbes:
## Motivation

Not sure if we want to merge this. For our local testnets w/ low Sol/s, this can make a large difference in the orphan rate. For a mainnet miner, it's almost certainly not worth it.

The internal miner can spend a full Equihash pass on a template whose parent is no longer valid. A solution can also race with the final cancellation check and reach `submitblock` after the template changes.

## Solution

- Add a backward-compatible cancellable solver method with `Continue`, `StopAtNonceBoundary`, and `StopNow` actions.
- Use the cancellable Tromp API to stop invalid work between digit rounds.
- Recheck the action after the final digit and before every internal-miner submission.
- Keep the forked solver in the optional `cpu-equihash-solver` dependency behind `internal-miner`. The registry Equihash dependency continues to serve the default verification graph.

The git dependency pins [valargroup/librustzcash PR #75](https://github.com/valargroup/librustzcash/pull/75) at commit `658533c4c604818610657d9699e82102e7379c76`. This draft must switch to a published Equihash release before merge.

## Testing

- `cargo check -p zakura-chain --features internal-miner --locked`
- `cargo test -p zakura-chain --features internal-miner solver_actions_have_distinct_cancellation_boundaries --locked`
- `cargo test -p zakura --features internal-miner stale_check_observes_a_change_seen_by_the_solver --locked`
- `markdownlint docs/changelog/unreleased/643.md --config .github/.markdownlint.yaml`
- Verified with `cargo tree` that the default `zakura-chain` graph uses only registry Equihash. The git solver appears only with `internal-miner`.
- The pinned Equihash commit passes its 11 crate tests and clippy. Its regression test fails against the old one-row reset and passes after the two-row reset.

Functional, semver, MSRV, and test CI checks pass. The fork is published under the distinct `cpu-equihash-solver` package name. Supply-chain policy narrowly allows the pinned repository and temporarily exempts this package; both entries must be removed when switching to a published upstream release.

## Changelog

Added `docs/changelog/unreleased/643.md`.

The change uses the existing `internal-miner` feature boundary. It does not change `getblocktemplate` or `submitold` semantics.

### Follow-up Work

- Replace the temporary git dependency with the published Equihash release.
- Ask upstream librustzcash maintainers to evaluate the fork-local Equihash draft before starting their upstream contribution process.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1783,7 +1783,7 @@ dependencies = [
 [[package]]
 name = "equihash"
 version = "0.3.0"
-source = "git+https://github.com/valargroup/librustzcash?rev=09fd73db1b49e38ac8830c9ab8a4164bf68eaf7d#09fd73db1b49e38ac8830c9ab8a4164bf68eaf7d"
+source = "git+https://github.com/valargroup/librustzcash?rev=f02666ca85119be817a336313cd28d3c3a9bc9c4#f02666ca85119be817a336313cd28d3c3a9bc9c4"
 dependencies = [
  "blake2b_simd",
  "cc",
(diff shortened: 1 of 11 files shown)


ValarDragon:
Lgtm after my mild edits

ValarDragon (approved):

PR #644, opened 2026-08-13
refactor(header-chain): clarify transition engine responsibilities

p0mvn:
## Motivation

Make the header-chain transition model easier to audit by clarifying lifecycle terms, trust boundaries, and module responsibilities.

## Solution

- Rename transition snapshots and operations around the project, persist, install, and publish lifecycle.
- Split planner, engine, invariant, and recovery-audit responsibilities into focused modules while removing the planner/invariants dependency cycle.
- Document transition contracts, distinguish missing durable facts from invalid evidence, and replace source-layout guards with behavioral verification.

## Testing

- `cargo test -p zakura-header-chain` (158 passed, 1 ignored)
- `cargo check -p zakura-network -p zakura-state --all-targets`
- Markdown lint for both changed Markdown files
- IDE diagnostics and `git diff --check origin/hc/01-engine-model...HEAD`
- `cargo fmt --all -- --check` is not clean: the target branch has an unrelated formatting issue in `crates/zakura-chain/src/parameters/network/testnet.rs`, and direct rustfmt checking also found formatting needed in two PR-owned transition files.

## Changelog

`docs/changelog/unreleased/644.md` marks this internal refactor as having no operator-visible changelog entry.


crates/zakura-header-chain/src/transition/engine/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+            .flatten()
+            .find(|delivery| delivery.delivery_id == delivery_id)
+    }
+}
+
+/// A verified durable write set ready for one post-commit in-memory installation.
+#[derive(Clone, Debug)]
+pub struct EngineTransition {
+    plan: TransitionPlan,
+}
```
evan-forbes: Pretty sure this wrapper predates the current refactor, but am I reading the layering correctly that `TransitionPlan::from_verified` is the actual type-state boundary while `EngineTransition` adds a public façade without another invariant? What do we gain from retaining both verified types instead of making `EngineTransition` the verified type directly? Also, should a value documented for one post-commit installation implement `Clone`?

crates/zakura-header-chain/src/transition/engine/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+    ///
+    /// # Returns
+    ///
+    /// - [`Ok`] — graph, projections, metadata, and auxiliary deliveries match
+    ///   the committed write set and this engine is ready to publish.
+    /// - [`Err`] — [`CommittedTransitionError::StaleSource`] if another
+    ///   transition changed this engine after planning, or
+    ///   [`CommittedTransitionError::Graph`] if the verified delta cannot apply.
+    ///   On either error the engine is unchanged; re-plan against the current
+    ///   snapshot rather than forcing install.
```
evan-forbes: Am I understanding correctly that either error occurs after the durable batch commits and can therefore leave disk ahead of memory? If so, should this direct the runtime to fail closed and recover, as the transition README does, rather than re-plan from the in-memory snapshot?

evan-forbes (approved):
Pretty sure these all come from the existing implementation rather than this refactor, and I may be missing a follow-up, but where are the four V12 findings addressed?

crates/zakura-header-chain/src/transition/engine/mod.rs
p0mvn: Many of these changes were made from a prompt to establish proper crate boundaries for clear separation of concerns.

I think you're right; this example is probably overkill since EngineTransition is just a wrapper type. Modifying

crates/zakura-header-chain/src/transition/engine/mod.rs
p0mvn: Addressed by amending the comment

p0mvn:
v12 has been addressed

PR #646, opened 2026-08-13
test(header-chain): restore architecture guards and broaden transition coverage

p0mvn:
## Motivation

Reviewing #644 surfaced coverage that the transition-model refactor removed or
weakened, plus a few small behaviors worth pinning down before the stack merges.

Two architectural guards were deleted rather than retargeted: one asserting the
retained DAG mutates only through a planned transition, and one asserting the
planner does not derive its write set by diffing whole graphs. Both invariants
still hold, but nothing enforced them. Two corruption assertions were also
switched from the exhaustive test verifier to the shipped incremental path,
which inspects an overlay and only the changed boundary nodes.

## Solution

- Reinstate both guards as one test in `lib.rs`, alongside the existing
  architecture guards. The planner half walks every production source under
  `transition/planner/` rather than reading the single pre-split `planner.rs`,
  and asserts the collected source list is non-empty so the check cannot go
  vacuous if the directory moves again.
- Point the auxiliary and finality corruption cases back at `verify_plan`
  (exhaustive), matching every neighbouring assertion in those files.
  `verify_plan_production` keeps its one deliberate caller in the shared
  `apply_transition` test helper, which covers the shipped path for every test
  in the module.
- Restore the settlement rebase-to-effect mapping to its pre-refactor form
  behind `settlement_header_work_effect`, with a `debug_assert_ne!` recording
  that already-applied header work returns during replay binding, and a unit
  test over all six input combinations.
- Drop `EngineTransition::retired_work`, which had no callers in production or
  tests. `RetiredWork::{from_snapshots, with_owners}` stay; they replace
  hand-rolled field comparisons in `ownership.rs` and the network fuzz harness.
- Report the duplicate tombstone hash during pre-audit row loading instead of
  the finalized hash, so the violation names the offending row.
- Sample the clock once per deferred reevaluation instead of once per candidate
  node.
- Add coverage for the event contract matrix and fingerprint binding, prepared
  header batches, deferred freshness boundaries, exhausted durable counters,
  mode and capability gates, and recovery audit read failures.

## Testing

- `cargo test -p zakura-header-chain --all-targets` — 196 passed, 0 failed
  (157 on the base branch).
- `cargo clippy -p zakura-header-chain --all-targets -- -D warnings` — clean.
- `cargo fmt -p zakura-header-chain -- --check` — clean.
- Negative-tested the restored planner guard: appending
  `// let mut graph = engine.graph().clone();` to `planner/replay.rs` fails the
  guard; reverted.
- Confirmed both reverted corruption assertions pass under the exhaustive
  verifier, so the earlier downgrade was not required by the refactor.

## Changelog

Internal-only; a `<!-- changelog: none -->` fragment follows once this draft has
a number.

### Follow-up Work

The deleted event-surface guard also grepped for consequence-oriented field
names (`pub new_header_best`, `pub prune`, `pub publish`) and obsolete facades.
The new contract matrix covers the behavioral half and gets compile-time
exhaustiveness from a wildcard-free match, but those name checks are not carried
forward; `lib.rs`'s `production_sources` scan is the natural home if we want them
back.


crates/zakura-header-chain/src/lib.rs
```diff
@@ -81,6 +81,81 @@ mod tests {
...
+            include_str!("transition/planner.rs").to_owned(),
+        )];
+        planner_sources(
+            std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("src/transition/planner")
+                .as_path(),
+            &mut sources,
+        );
+        assert!(
+            !sources.is_empty(),
```
evan-forbes, line 135: I realize this restores a guard from the existing implementation, but could we improve it further by enforcing the boundary through visibility or a compile-time API test instead of exact source strings? If the text scan remains, could we remove the nonempty assertion since `sources` already contains `planner.rs`?

crates/zakura-header-chain/src/transition/invariants/mod.rs
```diff
@@ -389,3 +401,269 @@ pub(crate) fn verify_plan_production(
...
+                assert_eq!(
+                    verify_plan_with_mode(&fixture.engine, &plan, mode),
+                    Err(InvariantViolation::SnapshotBeforeCommit)
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn production_and_exhaustive_paths_agree_on_valid_and_corrupt_candidates() {
```
evan-forbes, line 652: Pretty sure the production/exhaustive split comes from the existing implementation, but would it make sense to add a property or fuzz test that compares both paths across generated valid and corrupted candidates, given that the current parity test covers one no-change candidate?

crates/zakura-header-chain/src/transition/invariants/checks/aux.rs
```diff
@@ -60,35 +60,63 @@ pub(crate) fn verify_aux<G: HeaderGraphView>(
...
     for node in nodes {
         let mut deliveries = engine_before_commit.aux_deliveries(node.hash).to_vec();
         deliveries.retain(|delivery| !deleted_ids.contains(&delivery.delivery_id));
-        deliveries.extend(
-            puts.values()
-                .filter(|delivery| delivery.header_hash == node.hash)
-                .map(|delivery| **delivery),
-        );
-        for delivery in deliveries {
+        for delivery in puts
```
evan-forbes, line 93: I think this behavior comes from the existing implementation, but can a candidate contain both `Put` and `Delete` for one delivery ID? If it can, should the verifier replay those changes in installation order, and if it cannot, could the verifier reject that shape explicitly?

evan-forbes (approved):
very minor nit, but since it seems like we're changing behavior and not only tests, perhaps `fix(header-chain)` for PR so its title and description expose those changes?

crates/zakura-header-chain/src/transition/README.md
evan-forbes, line 1: we cover this in #641 fwiw and there are other things also covered there and in #631

evan-forbes: created https://github.com/zakura-core/zakura/pull/656 to remove high level things from this readme that we're already covering and imo shouldn't cover here

imo it feels inconsistent to add a readme w/ all of that info albeit helpful to have somewhere.

p0mvn:
I might have accidentally included this branch as part of another PR when merging the stack. Seeing this change got in

p0mvn:
Addressing comments separately via a follow-up

PR #650, opened 2026-08-13
refactor(header-chain): collapse prepare to context-free API

p0mvn:
## Motivation

Header preparation exposed two entry points: a context-free path used in production, and a lease-based compatibility path that still ran parent-link and contextual difficulty/time checks at prepare time. That dual API made it look like prepare owned graph-dependent rules, even though the transition planner already enforces them against retained ancestry.

## Solution

- Collapse prepare to a single `prepare_headers(input, parent, rules, clock)` API (the former context-free path).
- Remove the lease-based prepare wrapper and the `contextual_lease.is_some()` branch (prepare-time parent-link walk and contextual difficulty/median-time).
- Update call sites to pass an explicit parent frontier (`lease.parent()` where a lease is still loaded for rules/facts).
- Update the state coherence harness to treat planner `InvalidEvidence` as a non-mutating rejection (those failures used to be caught at prepare time).

`ValidationLease` remains for transition predecessor facts and rule binding. Contextual difficulty and median time remain enforced by the planner at admission/validation.

### Why this is safe

The removed prepare-time contextual branch was **not run on production paths**. Production preparation already used `prepare_context_free_headers` (driver and planner full-state validation), which passed `None` for the lease and skipped that branch. The lease-based prepare API was only exercised by tests and harnesses. Graph-dependent checks stay authoritative in the planner.

## Testing

- `cargo test -p zakura-header-chain --locked` — **196 passed**, 1 ignored, 0 failed.
- `cargo test -p zakura-state --locked --lib service::finalized_state::header_chain::coherence::` — **10 passed** (includes the three cases that initially failed after the prepare API change).

## Changelog

`docs/changelog/unreleased/650.md` — `<!-- changelog: none -->` (internal API cleanup only).

### DIFF
--- a/crates/zakura-header-chain/src/validation/prepare.rs
+++ b/crates/zakura-header-chain/src/validation/prepare.rs
@@ -9,15 +9,14 @@ use zakura_chain::{block, parameters::Network};
 
 use super::{
     infer_height, validate_commitment_structure, validate_compact_target,
-    validate_contextual_difficulty_and_time, validate_encoding_version_hash, validate_hash_filter,
-    AdjustedDifficulty, PowPolicy, PowPolicyError, POW_ADJUSTMENT_BLOCK_SPAN,
+    validate_encoding_version_hash, validate_hash_filter, PowPolicy, PowPolicyError,
 };
 use crate::{
-    Clock, EngineConfig, HeaderContextFact, HeaderValidationState, PreparedHeader,
-    PreparedHeaderBatch, RuleId, ValidationLease,
+    Clock, EngineConfig, HeaderValidationState, PreparedHeader, PreparedHeaderBatch, RuleId,
+    ValidationLease,
 };
 
-/// Ordered, nonempty canonical headers to validate against one exact parent lease.
+/// Ordered, nonempty canonical headers to validate against one exact parent frontier.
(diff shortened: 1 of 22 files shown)


crates/zakura-header-chain/src/validation/prepare.rs
```diff
@@ -331,44 +289,9 @@ fn prepare_headers_inner(
...
-            )
-            .map_err(|error| {
-                invalid(
-                    offset,
-                    HeaderRule::ContextualDifficultyAndTime,
-                    error.to_string(),
-                )
-            })?;
-            validate_contextual_difficulty_and_time(header.difficulty_threshold, adjustment)
-                .map_err(|error| invalid(offset, HeaderRule::ContextualDifficultyAndTime, error))?;
```
p0mvn, line 358: This is where prod runs:

https://github.com/zakura-core/zakura/blob/6ac198b004ab567d45162deef66ca441f21f68fd/crates/zakura-header-chain/src/transition/planner/event_effects/header_admission.rs#L100-L102

evan-forbes (approved):
nice, iirc at once point in time we were using these but then stopped so makes sense

PR #653, opened 2026-08-13
refactor(header-chain): package crate root by domain

p0mvn:
## Motivation

Keep the header-chain crate root focused on its client-facing facade instead of exposing the physical organization of graph state, identities, asynchronous work, discovery, fuzzing, and planner policy.

## Solution

- move the retained graph implementation intact under a private `graph` package
- group stable identities and asynchronous work contracts into private `identity` and `work` packages
- package locator discovery and fuzz support, and move retention policy under the transition planner
- preserve every existing flat `zakura_header_chain::Type` API path and feature gate

## Testing

- `cargo test -p zakura-header-chain`
- `cargo check -p zakura-state`
- focused `rustfmt --check` over changed Rust files

`cargo fmt --all -- --check` still reports pre-existing formatting drift in `zakura-chain/src/parameters/network/testnet.rs` and `zakura-header-chain/tests/production_transition.rs`; this PR leaves those unrelated files unchanged. Markdownlint was unavailable in the local environment (`markdownlint` and `npx` are not installed).

## Changelog

`docs/changelog/unreleased/653.md` explicitly marks this private module refactor as having no operator- or crate-consumer-visible effect.


p0mvn:
Rebasing

evan-forbes (approved):
I think we still have the 646 changelog after the rebase not sure if that's on purpose or not

I can see the appeal of the split so lgtm

p0mvn:
The feature branch already has many superfluous changelog entries.

I think we can batch clean it pre-merge

Issue #655, opened 2026-08-13
bug: IPv4-only nodes retry an unreachable IPv6 peer forever, burying getinfo.errors


p0mvn:
## What happens

Every IPv4-only node repeatedly tries to send QUIC datagrams to a peer's global IPv6 address it has no route to. Each attempt fails instantly with `ENETUNREACH` and logs a `WARN`:

```
sendmsg error: Os { code: 101, kind: NetworkUnreachable, message: "Network is unreachable" },
Transmit: { destination: [2a01:4f9:c015:f6a6::1]:8235, src_ip: None, ecn: None, len: 124, segment_size: None }
```

Observed on the whole testnet fleet. None of these hosts have a global IPv6 address or a default IPv6 route, and every occurrence targets that same single destination:

| Node | Global IPv6 addresses | `sendmsg error` lines per 5 MB of log |
| --- | --- | --- |
| zakura-testnet-1 | 0 | 694 |
| zakura-testnet-eu | 0 | 942 |
| zakura-testnet-as | 0 | 943 |

Roughly one per minute per node, continuously. The oldest occurrence in the retained log on `zakura-testnet-as` is `2026-08-08T22:12:00Z`, and `getinfo.errorstimestamp` was 10 seconds old when sampled, so this is an active retry loop rather than a stale message. Port 8235 is not one of the configured testnet bootstrap peers (those are IPv4 on 8234), and the address appears in no other log line — only in the `sendmsg` warnings — so it arrives as a QUIC path candidate rather than a zakura-level dial.

Sync is unaffected: the node held 109 connections and was at tip throughout.

## Why it is worth fixing

Beyond the wasted syscalls and log volume, this silently degrades `getinfo`.

`LastWarnErrorLayer` (`crates/zakurad/src/components/tracing/component.rs:545`) captures the message of the **last WARN or ERROR event from anywhere in the process** and publishes it on a watch channel; `getinfo` returns it verbatim as `errors` (`crates/zakura-rpc/src/methods.rs:1047`). A once-a-minute transport warning therefore overwrites that field continuously, so any genuine condition is buried within 60 seconds.

That is not hypothetical. The chain-stall warning operators actually care about is emitted at the same level (`warn!`, `crates/zakurad/src/components/sync/progress.rs:222`):

```
chain updates have stalled, state height has not increased for 13 minutes.
Hint: check your network connection, and your computer clock and time zone
```

[...]

PR #659, opened 2026-08-14
feat(deploy): add per-node status pages to the cluster dashboard

p0mvn:
## Motivation

The cluster status page answers exactly one question: is the fleet agreeing on a tip and advancing. It shows height, header lag, commit, tip hash and restart time. It cannot answer *why one node is unhealthy* — there is no disk, memory, peer, restart-count or sync-pipeline visibility anywhere in production.

Meanwhile the node already emits ~350 Prometheus series and has a purpose-built health endpoint, but both are opt-in and were only configured on some fleets, so the richest observability surface in the codebase was largely dark.

## Solution

Node names in the fleet table link to `/node/<name>`, a detail page that leads with host vitals, then chain position, the sync pipeline, and peers, with sparklines over a rolling window. Both routes serve the same template and branch on `location.pathname`, so the CSS, formatters and copy-to-clipboard helpers stay shared and there is no second template to keep in sync.

**Probe.** The existing SSH probe additionally collects host vitals (free disk on the state dir, memory, process RSS, load, uptime, systemd `NRestarts`, kernel OOM kills, and a redacted log tail), scrapes the node's Prometheus endpoint against a ~60-name allowlist, reads `/healthy` and `/ready`, and reduces `getpeerinfo`/`getmempoolinfo` on the node to counts. Vitals require no metrics endpoint, so they populate on every node including the zcashd probe. Anything unavailable is reported rather than blanked.

**Scrape cost.** Rendering the exposition costs a live node real CPU, and that cost grows with peer-labelled counter cardinality: measured ~0.03s on a freshly restarted node and ~0.6s (11.6 MB, 129k series) after several days. The scrape interval therefore scales with what the last scrape actually cost, capped at 120s, so a cheap endpoint refreshes every poll and an expensive one cannot burn more than about 1/30th of a core. `--metrics-min-interval` pins it if a fixed cadence is wanted.

**Deployer.** New `health_listen_addr` key renders `[health] listen_addr`, mirroring the existing `metrics_endpoint`. Both are unauthenticated and `zakurad` panics if either address is already bound, so the docs call out loopback binding and a port pre-flight. Testnet now sets it.

**Removals.** The Ironwood activation countdown is deleted now that the upgrade has activated, along with its estimator, height history and the `--upgrade-height`/`--target-spacing` flags — the installed unit files pass those flags, so both workflow `ExecStart` lines drop them in the same change or the service would fail to start. The `zakurad vs zcashd` tile is also removed; note the caveat in follow-ups.

**Compatibility.** `/data` keeps its existing shape for the Slack watchdog and only gains a small per-row `vitals` summary, which surfaces low disk and OOM kills as badges in the fleet table. `/ironwood-status.json`, the public API zakura.com consumes, is untouched and computes activation independently of the removed estimator.

## Testing

`python3 deploy/runner/test_zakura_cluster_status.py` (48), `test_zakura_cluster_watchdog.py` (10) and `deploy/deployer/test_deploy.py` (12) all pass; all three already run in `lint.yml`.

New coverage worth calling out, each written against a bug found while validating on the live testnet fleet:

- The remote probe is executed as its own process against stub HTTP endpoints, since that is how it actually runs on the far side of ssh. Covers allowlist filtering, `/healthy` and `/ready` status-and-body capture, unreachable and unconfigured endpoints, and log-tail redaction.
- `journalctl` exits 1 when its grep matches nothing, which was making "no OOM kills" read as "unknown". Stubbed on `PATH` so the assertion is hermetic.
- `zakurad`'s `getpeerinfo` has no `subver` field, so the peer version breakdown comes from the exporter's `user_agent` label instead.
- The scrape interval scaling, its cap, and the explicit override.
- Every `[data-view]` section is checked against the `[hidden]` override. A class that sets its own `display` beats the UA stylesheet's `[hidden]` rule, which left one unrendered section visible on node pages showing placeholder text.
- Route coverage for `/node/<name>` and `/data/node/<name>` including 404s, plus `Cache-Control: no-store` on all four routes — without it browsers heuristically cached the page and kept serving an old build after a deploy.

Manually verified end to end on the testnet fleet: `[health]` enabled on all three nodes with automatic config restore on failure, then the dashboard deployed and every route, panel and degradation path checked against live data.

## Changelog

 — no Rust or `Cargo.toml` changes; this is deploy tooling only.

### Follow-up Work

- **The `zakurad vs zcashd` tile is removed, and that is a real loss on mainnet.** It compared the best zakurad tip against the best zcashd tip — the cross-implementation consensus check that `zcashd-compat` exists to provide. On testnet there is no zcashd node, so `best_tip_row(rows, "zcashd")` always returned `None` and the tile showed a permanent green "aligned", claiming agreement with a client nobody was observing. The right fix is to render it only when a zcashd node is actually present, rather than delete it; happy to restore it that way.
- Mainnet nodes still have no `[metrics]` or `[health]` endpoints. Mainnet deploys are binary-only (`manage_config = false`), so enabling them needs a systemd drop-in setting `ZAKURA_METRICS__ENDPOINT_ADDR` and `ZAKURA_HEALTH__LISTEN_ADDR`, which is not in this PR. Until then the metrics-backed panels degrade to "not enabled" on the mainnet page.
- Peer-labelled counter cardinality grows without bound between restarts, which is what forces the adaptive scrape. See #655.
- `/data` publishes `ssh = root@<ip>` for every node on a public, unauthenticated page. Pre-existing, worth dropping separately.

### DIFF
--- a/.github/workflows/zakura-mainnet-deploy.yml
+++ b/.github/workflows/zakura-mainnet-deploy.yml
@@ -187,6 +187,9 @@ jobs:
           # not written to any node (manage_config = false).
           rpc_listen_addr = "0.0.0.0:8232"
           log_file        = "/root/logs/zakura-tracing.log"
+          # Verified on the fleet: the nine regional nodes keep state here, the
+          # compat host does not. Used only for the dashboard's disk probe.
+          state_cache_dir = "/mnt/data/zakura-cache"
 
           # Public mainnet node ids (crates/zakura-network/src/zakura/handler.rs). Used by
           # the dashboard to label each host; not deployed to the nodes.
@@ -255,6 +258,7 @@ jobs:
           service_name = "zakurad-compat"
           log_file   = ""
           rpc_listen_addr = "127.0.0.1:8232"
+          state_cache_dir = "/mnt/data/import-zebra"
           EOF
 
       - name: Build deployer binary
@@ -355,6 +359,7 @@ jobs:
           rpc_listen_addr = "[::1]:8232"
           rpc_auth = "cookie"
           rpc_config_path = "/mnt/data/runtime/zcashd/.cookie"
+          state_cache_dir = "/mnt/data/runtime/zcashd"
           # P2P sidecar (post -zebra-compat): supervised with -connect=127.0.0.1:8233.
           process_pattern = "zcashd .*-connect=127.0.0.1:8233"
           EOF
--- a/.github/workflows/zakura-mainnet-deploy.yml
+++ b/.github/workflows/zakura-mainnet-deploy.yml
@@ -369,7 +369,7 @@ jobs:
           [Service]
           Type=simple
           WorkingDirectory=/opt/zakura-mainnet-dashboard
-          ExecStart=/usr/bin/python3 /opt/zakura-mainnet-dashboard/zakura-cluster-status.py --config /opt/zakura-mainnet-dashboard/nodes.toml --host 0.0.0.0 --port 8090 --network mainnet --interval 10 --stale-after 300 --upgrade-height 3428143 --target-spacing 75 --state-file /var/lib/zakura-mainnet-dashboard/orphan-pairs.json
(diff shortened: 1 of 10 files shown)


p0mvn:
Demo: https://status.testnet.zakura.valargroup.dev/node/zakura-testnet-1

evan-forbes (approved):
utack

PR #665, opened 2026-08-14
fix(header-chain): bound protected state before commit

evan-forbes:
## Summary

- Protect the verified best chain, the header best chain, active retained paths, and authorized validation-context parents.
- Reject oversized or unauthenticated retention references before graph traversal.
- Enforce the configured retention-reference limit.
- Refuse eviction when the protected union exceeds the retained-node limit.
- Use constant-time count checks on the ordinary path.
- Rebuild eviction candidates only after permanent eviction changes the graph.
- Recover a durable resource stall through one state-writer reevaluation per stalled state version.
- Record exact protected visits, candidate visits, removals, and graph-sized workspaces.
- Preserve authenticated staged full-state paths across the state-writer boundary.

## Structural gates

- Ordinary retention performs zero graph visits and zero allocations.
- The protected union visits each retained node at most once.
- Candidate construction visits each retained node once per graph state. It rebuilds only after permanent eviction changes the graph.
- Eviction counts each removed node once.
- Repeated stalled insertion exits before graph projection.
- Stall recovery runs at most once for each unchanged stalled state version.

## Performance

The checked-in release benchmark covers 25%, 50%, 90%, and 100% of the current retained-node limit. Run `cargo bench -p zakura-header-chain --bench retention --features test-support` to reproduce the measurements.

## Verification

- `cargo test -p zakura-header-chain --all-targets` — 225 passed, 1 ignored; production integration passed
- `cargo test -p zakura-state --lib` — 430 passed, 3 ignored
- `cargo clippy -p zakura-header-chain -p zakura-state --all-targets -- -D warnings`
- `cargo bench -p zakura-header-chain --bench retention --features test-support -- --test`
- `git diff --check`

## Review state

This PR targets the branch for #667 and must merge after #667. The latest update addresses the configured retention-reference bound, headers-only stall recovery, and post-permanent-eviction candidate indexing.


p0mvn (approved):

crates/zakura-header-chain/src/transition/planner/admission.rs
```diff
@@ -64,11 +64,12 @@ pub(super) fn authenticate_and_admit(
     let snapshot_before_commit = engine.snapshot();
     let metadata = engine.metadata().clone();
     validate_snapshot(&snapshot_before_commit, &metadata, context)?;
-    if context.retention_references.len() > context.config.limits.max_retention_references.get() {
+    if context.retention_references.len() > crate::POW_PREDECESSOR_CONTEXT_SPAN {
```
p0mvn: This bound used to read `context.config.limits.max_retention_references` (frozen v1 = 26). It now uses `POW_PREDECESSOR_CONTEXT_SPAN` (27), the PoW predecessor window — semantically unrelated to retention.

After this change, `EngineLimits::max_retention_references` has no production reader. The only remaining use is `config.rs:433`, which still asserts the frozen spec value with *"one atomic transition can retain every active header target and full-state fork tip."*

Functionally safe today: worst case is `MAX_RETAINED_PATH_LEASES` (16, `zakura-state/src/header_chain.rs:19`) plus `MAX_NON_FINALIZED_CHAIN_FORKS` staged tips (10, `non_finalized_state.rs:268`) = 26, under both bounds. But a versioned-limits crate that pins exact v1 tuples shouldn't keep a spec test for a number the planner ignores.

Please either restore `context.config.limits.max_retention_references` here, or delete the field and update the spec test.

crates/zakura-header-chain/src/transition/planner.rs
```diff
@@ -127,6 +130,12 @@ fn derive_plan_candidate(
     // Phase 1: authenticate / admit
     let (snapshot_before_commit, mut metadata, admitted) =
         authenticate_and_admit(engine, &input, context)?;
+    if snapshot_before_commit.alarms.resource_stalled
+        && matches!(&admitted.event, crate::TransitionEvent::InsertHeaders(_))
+    {
+        let domain = admitted.event.domain();
+        return resource_stalled(engine, snapshot_before_commit, domain, context);
+    }
```
p0mvn, line 138: This short-circuit is the right DoS gate for repeated `InsertHeaders`, but it also removes the only self-heal that used to live on that path. `metadata.alarms.resource_stalled = retention.resource_stalled` only runs in `derive_plan`, so while the durable alarm is set, insertion can never re-measure retention or clear it. The test change is direct evidence: `tests/insertion.rs` now needs a `ReevaluateDeferred` before a within-bound insert can recover.

Integrated/`zakurad` still has a clearing path: `FullStateFinalized` and checkpoint `VerifiedChainChanged` are not `InsertHeaders`, so they still reach settlement. Finality advances first, `advance_finalized_frontier` drops everything not in the new subtree, protected walks stop at the new root, and `derive_plan` can turn the alarm off.

Headers-only does not have that event. `HeadersOnlyDepth` is computed inside `InsertHeaders` settlement (`settlement.rs`, the `EngineMode::HeadersOnly` branch). Once this gate fires, that settlement never runs, so the protected set cannot shrink. The remaining any-mode non-insertion event is `ReevaluateDeferred`, and that is not a heartbeat: `receive_until_deferred_deadline` only submits it when `earliest_deferred()` is `Some`. The alarm is durable metadata, so restart does not help.

Worth an explicit answer that headers-only cannot reach a stall, or a guaranteed clearing path that is not `InsertHeaders`.

crates/zakura-header-chain/src/transition/planner/retention.rs
```diff
@@ -16,146 +16,221 @@ pub(super) struct RetentionPlan {
...
     }
 
-    let plan = RetentionPlan::default();
-    evict_permanently_ineligible(store, &protected)?;
-    if store.view_eligible_header_tips().len() <= limits.max_candidate_tips.get()
+    if protected_header_hashes.len().saturating_sub(1) > limits.max_non_finalized_nodes.get() {
+        return Ok(resource_stalled_plan(plan));
+    }
+
+    let mut candidates = RetentionCandidates::build(store)?;
```
p0mvn, line 77: **Merge blocker:** this builds the eviction candidate set *before* `evict_permanently_ineligible` (line 80). The base branch built it after. Nodes that become leaves only once a permanently-ineligible subtree is removed are therefore never candidates, so the leaf loop can drain a stale snapshot and return `resource_stalled` while an unprotected, evictable node is still in the graph.

Repro:

```text
A ─ P ─ R   (R checkpoint-conflicted)
  └ S       (selected)
enforce_retention(selected=S, verified=A, refs=[], limits(10, 1))
```

`header_leaves` snapshots as `{R, S}`. Permanent eviction removes `R`, making `P` an unprotected leaf. The loop pops `R` (gone), pops `S` (protected), empties, and stalls. Rebuilding candidates after permanent eviction on the same input evicts `P` and does not stall.

This is a regression, not a pre-existing limitation. Please restore the base-branch order (build candidates after permanent eviction) and add this graph to `retention/tests.rs`. The “candidate construction visits each retained node exactly once” gate should be amended: ordinary admission never reaches this path, and one extra bounded scan under pressure is the correct work bound.

A false stall is worse than a missed eviction. Settlement discards the overlay, the alarm is durable, and `InsertHeaders` then short-circuits for as long as it stays set. In `HeadersOnly` with an empty deferred index there is no later event that reliably clears it, including across restart.

p0mvn (approved):
Spent some time thinkig through how likely we are to hit the stall, and I don't think very.

That being said, there seems to be no recovery mode in header-only mode that may be worth addressing.

Also, might be worth addressing the point about nodes becoming evictable after their children are as well.

crates/zakura-header-chain/src/transition/planner/admission.rs
evan-forbes: Good catch. I restored max_retention_references as the production limit and added a test with a non-default configured limit.

crates/zakura-header-chain/src/transition/planner.rs
evan-forbes: Agreed. The state writer now reevaluates each new stalled state version, even without a deferred deadline and after restart. An unchanged stall does not repeat graph work.

crates/zakura-header-chain/src/transition/planner/retention.rs
evan-forbes: Good catch. The planner now rebuilds candidates only when permanent eviction changes the graph. The common pressure path keeps one scan and has coverage.

PR #666, opened 2026-08-14
refactor(header-chain): bound recovery and durable indexes

evan-forbes:
## Summary

- Stack this recovery layer on #667 without merging either PR.
- Use typed `StoreCollection` and `RowLimit` values in limit failures.
- Read all singleton and collection rows from one coherent RocksDB snapshot.
- Reject configuration and singleton mismatches before collection scans.
- Bound nodes, edges, projections, deferred rows, direct reasons, auxiliary rows, validation contexts, finality history, and tombstones before decoding extra rows.
- Stop each RocksDB cardinality scan at `limit + 1` without decoding the extra row.
- Keep normal finality-history append and tombstone admission O(1).
- Create an authenticated finality-history checkpoint at the retained bound.
- Prune finalized tombstones and evict only unretained overflow tombstones in deterministic height/hash order.
- Promote untrusted auxiliary rows only after bounded graph recovery validates their foreign keys and outcomes.

## Verification

- `cargo test -p zakura-header-chain --all-targets`
- `cargo test -p zakura-state service::finalized_state::header_chain --lib`
- `cargo check -p zakura-state --all-targets`
- `cargo check -p zakura --all-targets`
- `git diff --check`

The named F-225519 engine and RocksDB regressions prove the first extra row stops recovery before decoding, hashing, or consensus validation. The named F-225513 regression proves the startup alarm cannot exempt an oversized graph.

## Review state

This PR is a draft. Review #667 before this PR. Do not merge either PR during this review pass.

### DIFF
--- a/crates/zakura-header-chain/conformance.toml
+++ b/crates/zakura-header-chain/conformance.toml
@@ -64,7 +64,7 @@ id = "LC-SCOPE-07"
 name = "Auxiliary-evidence boundary"
 status = "implemented"
 owner = "zakura_header_chain::AuxObservationV1"
-tests = ["IN-06::f_225517_auxiliary_outcomes_derive_from_exact_owned_observations", "IN-06::selected_auxiliary_repair_adds_only_one_exact_provenance_record"]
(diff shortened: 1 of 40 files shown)


p0mvn (approved):

evan-forbes:
cleaning up some of the formatting here then merging

ie some of the v 12 ids got included in test names...

PR #667, opened 2026-08-14
refactor(header-chain): seal auxiliary outcomes and bound recovery

evan-forbes:
## Summary

- Replace caller-selected auxiliary verdicts with sealed VCT observations.
- Derive auxiliary outcomes and boundary evidence from retained deliveries.
- Validate untrusted durable outcomes before recovery promotes them.
- Request replacement data when a retained successor lacks a usable witness.
- Bound recovery collections and maintain the delivery and tombstone indexes.
- Prune finalized tombstones after sparse graph updates.
- Require header-chain disk format version 2.

This PR closes the remaining auxiliary-delivery security gaps. Zakura 1.2.0 shipped disk format version 1. Operators must resync version-1 header-chain databases.

## Verification

- `cargo test -p zakura-header-chain --all-targets` — 224 passed, 1 ignored; production integration passed
- `cargo test -p zakura-state --lib` — 428 passed, 3 ignored
- `cargo check -p zakura --all-targets`
- `git diff --check`

## Review state

This PR targets `main`. The branch includes the bounded-recovery work that previously lived in #666. CI passes. The PR has approval and no unresolved review threads.


p0mvn (approved):

p0mvn:
This is a DB-breaking change that fails closed on V1 format. What is the migration plan for the existing databases?

p0mvn (approved):
LGTM

But would it be possible to develop a non-breaking db migration plan to reduce node roll out overhead?

evan-forbes:
that is definitely a good goal and we could. but then we need to keep the extra info in a separate column w/ a lot of the same info. the info also seems very related as well, where someone sent us a header and we store it, but then need to keep track of if if we derived the validity of that header or not somewhere.

therefore imo having doing the break is worth it

we should definitely add a proper migration tho so will do that. happy to continue to discuss as well

p0mvn:
Propose not to block merge on GitHub CI instabilities. We can follow up with fixes as needed

PR #676, opened 2026-08-15
fix(state,rpc): use exact checked 256-bit work for chain selection

p0mvn:
Authored by @evan-forbes as part of the fork-aware header-chain work in #586.
This PR carries the change on its own so the consensus-critical edit to
non-finalized chain selection can be reviewed in isolation, ahead of the engine,
persistence, and reactor that land in #586. Split out and rebased by @p0mvn.

This PR covers what the split plan listed as two steps (checked chain work, and
the exact solution rate). They turned out to be one change: removing the
panicking `AddAssign` forces `solution_rate` to declare an overflow policy, so
splitting them would have left a PR that does not compile.

## Motivation

Cumulative chain work is a 256-bit quantity, but `zakura-chain` exposed it
through `Add`/`AddAssign` impls whose only overflow handling was:

```rust
.expect("cumulative work stays below the 2^256 boundary on any valid chain")
```

That claim holds for Mainnet and Testnet, where `PoWLimit` bounds per-block
work. It does not hold for a custom network that relaxes the difficulty rules —
and contextual validation deliberately relaxes the difficulty-equality check on
`disable_pow` networks. Every caller inherited a panic that was invisible at the
call site.

## Solution

Remove the three panicking impls and make each of the five summation sites state
its own overflow policy:

| site | policy |
| --- | --- |
| `Chain::push` | new `ValidateContextError::CumulativeWorkOverflow` |
| legacy header store most-work gate | new `CommitHeaderRangeError::WorkOverflow` |
| `solution_rate` | `None`, matching every other case it cannot compute |
| two fabricated-header test helpers | assert; their runs are orders of magnitude below the boundary |

Two details worth the closest reading:

- **`Chain::push` now computes the checked sum before mutating
  `height_by_hash`.** Previously the hash was inserted first and the work added
  after, so a rejected block could leave the chain with a hash recorded and no
  matching work.
- **The most-work gate fails closed.** Dropping an overflowing term there would
  understate whichever side it landed on, which is exactly the comparison the
  gate makes to decide whether a conflicting suffix may replace the stored one.

`Chain::cmp` now scores through `ChainScore`/`SuffixWork` from
`zakura-header-chain` rather than comparing `partial_cumulative_work` inline.
**The ordering is unchanged**: work first, then raw tip-hash byte order.

`solution_rate` widens to `U256` and `get_network_sol_ps` saturates at
`u64::MAX` on the way out, so the zcashd-compatible response width is preserved
at the RPC boundary rather than inside the state read.

## Test coverage

- The full 461-test `zakura-state` suite passes, including the 37 header-store
  coherence tests that drive the most-work gate across reorgs, fork switches,
  and the DAA-window edge.
- `rpc_getnetworksolps_saturates_to_response_width` now feeds `U256::MAX` rather
  than `u128::MAX`, so it exercises the real 256-bit saturation instead of a
  value the old, narrower type could already represent.
- `rpc_getnetworksolps` pins the ordinary path.

## Verification

- `cargo build --workspace --locked`
- `cargo test -p zakura-state` (458 passed, 3 ignored)
- `cargo test -p zakura-rpc -- getnetworksolps` (2 passed)
- Runnable gate: fresh Mainnet cache, stock config, real public v2 peers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8703,6 +8703,7 @@ dependencies = [
  "tower 0.4.13",
  "tracing",
  "zakura-chain",
+ "zakura-header-chain",
  "zakura-node-services",
  "zakura-test",
 ]
--- a/crates/zakura-chain/src/work/difficulty.rs
+++ b/crates/zakura-chain/src/work/difficulty.rs
@@ -658,14 +658,6 @@ impl PartialOrd<ExpandedDifficulty> for block::Hash {
     }
 }
 
-impl std::ops::Add for Work {
-    type Output = PartialCumulativeWork;
-
-    fn add(self, rhs: Work) -> PartialCumulativeWork {
-        PartialCumulativeWork::from(self) + rhs
-    }
-}
-
 /// Partial work used to track relative work in non-finalized chains
 ///
 /// # Consensus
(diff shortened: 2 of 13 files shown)


p0mvn:
[F-228731](https://v12.sh/runs/6579/228731) severity over-stated

p0mvn:
I think it is fine to not do all V12s in this PR to avoid merge conflicts with the feature branch

PR #682, opened 2026-08-15
fix(rpc): expose peer metadata in getpeerinfo

ValarDragon:
## Motivation

CipherScan cannot identify Zakura peers through `getpeerinfo`: the RPC response omits the user-agent and negotiated protocol version even though both are advertised during the P2P handshake.

## Solution

- Retain handshake user-agent and negotiated protocol version in `MetaAddr`.
- Store retained user agents as shared `Arc<str>` data, so address-book/index/cache clones do not duplicate the string allocation.
- Preserve the metadata through address-book updates.
- Add `subver` and `version` to `getpeerinfo`, following Zebra’s design.
- Preserve the existing four-argument `PeerInfo::new` constructor for downstream compatibility.
- Update RPC snapshots, serialization fixtures, benchmarks, and ownership-sensitive network tests for the intentional removal of `MetaAddr: Copy`.

## Testing

Passed locally:

- `cargo check -p zakura-network --all-targets`
- `cargo check -p zakura-rpc --all-targets`
- `cargo clippy -p zakura-network -p zakura-rpc --lib -- -D warnings`
- `cargo test -p zakura-network --lib --no-run`
- `cargo test -p zakura-network meta_addr --lib`
- `cargo test -p zakura-rpc rpc_getpeerinfo --lib`
- `cargo test -p zakura-rpc test_rpc_response_data --lib`
- `cargo test -p zakura-rpc test_get_peer_info_with_ping_values_serialization --test serialization_tests`
- `git diff --check`

The full local network unit run executed 1,062 passing tests; seven environment-sensitive tests failed at runtime (loopback source binding and local block-sync fuzz timing), not during compilation.

The all-target clippy command encounters an existing duplicated `#[cfg(test)]` attribute in `zakura-network/src/zakura/discovery/service.rs`; the changed libraries pass targeted clippy.

## Changelog

Added `docs/changelog/unreleased/682.md` for the user-visible RPC change.

- Zebra peer metadata design

### Follow-up Work

- Keep the PR draft while the refreshed CI checks complete.

---

> [!NOTE]
> **Medium Risk**
> Changes peer metadata storage, Zakura upgrade address-book updates, and RPC output across the network stack; behavior shifts for upgraded peers when native connections end.
> 
> **Overview**
> **`getpeerinfo`** now returns handshake **`subver`** and negotiated **`version`** for live peers, sourced from address-book **`MetaAddr`** entries instead of omitting that data.
> 
> **`MetaAddr`** drops **`Copy`** and stores optional user agent (**`Arc<str`**) and negotiated protocol version from **`UpdateConnected`**, with **`clone()`** used across the address book, codecs, and tests. **`PeerInfo::new`** keeps its four-argument shape for compatibility.
> 
> Successful **Zakura P2P v2** upgrades record connected metadata (including inbound peers) via **`record_zakura_upgrade`**; duplicate upgrades do not. The legacy liveness keeper now emits **`UpdateFailed`** on shutdown when the native connection deregisters so upgraded peers no longer linger in **`getpeerinfo`** after the replacement session ends.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb29cc9cbee632382fde0462d46e77cceab3b12a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/zakura-network/benches/address_book.rs
+++ b/crates/zakura-network/benches/address_book.rs
@@ -104,7 +104,7 @@ fn address_book(c: &mut Criterion) {
         assert!(ban_check.bans().contains(existing_addr.ip()));
 
         let mut batch_check = address_book.clone();
-        batch_check.extend(batch_updates.iter().copied());
+        batch_check.extend(batch_updates.iter().cloned());
         assert_eq!(batch_check.len(), size);
 
         group.bench_with_input(
@@ -183,7 +183,7 @@ fn address_book(c: &mut Criterion) {
             b.iter_batched_ref(
                 || address_book.clone(),
                 |address_book| {
-                    address_book.extend(batch_updates.iter().copied());
+                    address_book.extend(batch_updates.iter().cloned());
                     black_box(address_book);
                 },
                 BatchSize::PerIteration,
--- a/crates/zakura-network/src/address_book.rs
+++ b/crates/zakura-network/src/address_book.rs
@@ -105,12 +105,12 @@ impl AddressBookIndex {
     }
 
     fn get(&self, addr: &PeerSocketAddr) -> Option<MetaAddr> {
-        self.by_addr.get(addr).copied()
+        self.by_addr.get(addr).cloned()
     }
 
     fn insert(&mut self, meta_addr: MetaAddr) {
         let addr = meta_addr.addr;
-        let previous = self.by_addr.insert(addr, meta_addr);
+        let previous = self.by_addr.insert(addr, meta_addr.clone());
         if let Some(previous) = previous {
             assert!(
                 self.by_priority.remove(&previous),
@@ -204,7 +204,7 @@ impl AddressBookIndex {
             }
         }
 
-        let peers: Vec<_> = self.ordered_values().copied().collect();
+        let peers: Vec<_> = self.ordered_values().cloned().collect();
         assert!(peers.windows(2).all(|pair| pair[0] <= pair[1]));
     }
 }
@@ -435,14 +435,14 @@ impl AddressBook {
 
         for (socket_addr, meta_addr) in addrs {
             // overwrite any duplicate addresses
-            new_book.peers.insert(meta_addr);
+            new_book.peers.insert(meta_addr.clone());
             // Add the address to `most_recent_by_ip` if it has responded
-            if new_book.should_update_most_recent_by_ip(meta_addr) {
+            if new_book.should_update_most_recent_by_ip(&meta_addr) {
                 new_book
                     .most_recent_by_ip
                     .as_mut()
                     .expect("should be some when should_update_most_recent_by_ip is true")
-                    .insert(socket_addr.ip(), meta_addr);
+                    .insert(socket_addr.ip(), meta_addr.clone());
             }
             // exit as soon as we get enough addresses
             if new_book.peers.len() >= addr_limit {
@@ -577,7 +577,7 @@ impl AddressBook {
     /// - this is the only field checked by `has_connection_recently_responded()`
     ///
     /// See [`AddressBook::is_ready_for_connection_attempt_with_ip`] for more details.
-    fn should_update_most_recent_by_ip(&self, updated: MetaAddr) -> bool {
+    fn should_update_most_recent_by_ip(&self, updated: &MetaAddr) -> bool {
         let Some(most_recent_by_ip) = self.most_recent_by_ip.as_ref() else {
             return false;
         };
@@ -650,7 +650,7 @@ impl AddressBook {
         let instant_now = Instant::now();
         let chrono_now = Utc::now();
 
-        let updated = change.apply_to_meta_addr(previous, instant_now, chrono_now);
+        let updated = change.apply_to_meta_addr(previous.clone(), instant_now, chrono_now);
 
         trace!(
             peer = %addr_label,
@@ -662,7 +662,7 @@ impl AddressBook {
             "calculated updated address book entry",
         );
 
-        if let Some(updated) = updated {
+        if let Some(ref updated) = updated {
             if updated.misbehavior() >= constants::MAX_PEER_MISBEHAVIOR_SCORE {
                 // Ban and skip outbound connections with excessively misbehaving peers.
                 let banned_ip = updated.addr.ip();
@@ -715,7 +715,7 @@ impl AddressBook {
                 return None;
             }
 
-            self.peers.insert(updated);
+            self.peers.insert(updated.clone());
             self.address_metrics_dirty = true;
 
             // Add the address to `most_recent_by_ip` if it sent the most recent
@@ -724,7 +724,7 @@ impl AddressBook {
                 self.most_recent_by_ip
                     .as_mut()
                     .expect("should be some when should_update_most_recent_by_ip is true")
-                    .insert(updated.addr.ip(), updated);
+                    .insert(updated.addr.ip(), updated.clone());
             }
 
             debug!(
--- a/crates/zakura-network/src/address_book/tests.rs
+++ b/crates/zakura-network/src/address_book/tests.rs
@@ -63,7 +63,7 @@ fn address_metrics_are_updated_once_per_batch() {
     ]
     .map(MetaAddr::new_initial_peer);
 
-    address_book.extend(peers);
+    address_book.extend(peers.clone());
 
     assert_eq!(
         address_book.address_metrics_update_count,
--- a/crates/zakura-network/src/address_book/tests/prop.rs
+++ b/crates/zakura-network/src/address_book/tests/prop.rs
@@ -121,7 +121,7 @@ proptest! {
 
         for (_addr, changes) in addr_changes_lists.iter() {
             for change in changes {
-                address_book.update(*change);
+                address_book.update(change.clone());
                 address_book.peers.assert_consistent();
 
                 prop_assert!(
@@ -146,7 +146,7 @@ proptest! {
         for index in 0..MAX_ADDR_CHANGE {
             for (_addr, changes) in addr_changes_lists.iter() {
                 if let Some(change) = changes.get(index) {
-                    address_book.update(*change);
+                    address_book.update(change.clone());
                     address_book.peers.assert_consistent();
 
                     prop_assert!(
--- a/crates/zakura-network/src/address_book/tests/vectors.rs
+++ b/crates/zakura-network/src/address_book/tests/vectors.rs
@@ -170,7 +170,7 @@ fn address_book_peer_order() {
     );
 
     // Regardless of the order of insertion, the most recent address should be chosen first
-    let addrs = vec![meta_addr1, meta_addr2];
+    let addrs = vec![meta_addr1.clone(), meta_addr2.clone()];
     let address_book = AddressBook::new_with_addrs(
         "0.0.0.0:0".parse().unwrap(),
         &Mainnet,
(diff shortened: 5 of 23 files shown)


evan-forbes:
posting a fable audit:

doubt all of these are purely related to these changes, so 

```
crates/zakura-network/src/peer/handshake.rs
  ● 1150 [correctness]    record_zakura_upgrade dropped the old !connected_addr.is_inbound() guard, so upgraded inbound peers get an address-book entry at their ephemeral source port that the crawler dials and the peer cache persists.
  ● 1144 [security]       record_zakura_upgrade unconditionally creates a Responded entry plus a liveness-keeper task per upgrade, and an inbound peer re-advertising an already-registered node id mints a fresh entry+keeper per handshake with no dedup or cap.
crates/zakura-rpc/src/methods/types/peer_info.rs
  ●   57 [security]       From<MetaAddr> copies the remote-controlled user_agent into getpeerinfo subver with no sanitization; the codec bounds length but allows control/ANSI-escape bytes, whereas zcashd runs SanitizeString over subver before returning it.
  ●   58 [rpc-compat]     getpeerinfo version reports the negotiated protocol version (min of local and remote) instead of the peer's advertised nVersion, diverging from the zcashd contract the PR cites and undercutting its own CipherScan goal; connection_info.remote.version is available but unused.
  ●   14 [rpc-compat]     PeerInfo always serializes subver:"" and version:0 for entries lacking metadata (serde default affects only Deserialize; no skip_serializing_if), so consumers cannot distinguish unknown from a real value, and 0 is an impossible protocol version — unlike the sibling pingtime/pingwait Option fields.
  ●   36 [simplification] The hand-written 4-arg PeerInfo::new replacing the derive_new derive has no production caller (only 4 integration-test call sites), silently zeroes the two new fields, and hard-codes the empty-metadata defaults a third time alongside Default and From<MetaAddr>.
crates/zakura-network/src/meta_addr.rs
  ● 1167 [correctness]    apply_to_meta_addr's concurrent-change guard drops an UpdateConnected when the previous entry is Failed within the 5-second window, and only UpdateConnected carries user_agent/negotiated_version, so a fast reconnect silently loses fresh handshake metadata.
  ● 1017 [correctness]    MetaAddrChange::user_agent() returns Some(Arc::from("")) for an UpdateConnected carrying an empty user_agent, and the .or(previous.user_agent) merge lets that Some("") overwrite previously recorded real metadata; the API conflates absent with empty.
  ● 1237 [correctness]    The both-NeverAttempted merge branch hard-codes user_agent/negotiated_version to None instead of preserving previous values, unlike every sibling field in the same branch and unlike the other merge branch's .or(previous.*), while its comment says preserve original values.
  ●  300 [efficiency]     UpdateConnected stores user_agent: String while MetaAddr stores Option<Arc<str>>, so user_agent() runs a fresh Arc::from on every apply; into_new_meta_addr and apply_to_meta_addr take self/previous by value forcing clones, and AddrV1/V2::from take MetaAddr by value forcing a clone per gossiped address.
  ●  232 [design]         Per-connection handshake metadata is threaded through MetaAddr/MetaAddrChange (the gossip/reconnection address book), costing the Copy impl and clone churn across ~20 files, even though the live peer set already holds this data per connected peer in LoadTrackedClient.connection_info.
crates/zakura-network/src/zakura.rs
  ●  464 [correctness]    A stale liveness keeper's new_shutdown (UpdateFailed) can be applied after a fast re-upgrade's UpdateConnected for the same address; within the 5s concurrent window Failed outranks Responded, so a live re-connected peer is marked Failed and hidden from getpeerinfo.
  ●  463 [correctness]    The keeper detects deregistration via a coalescing tokio::watch of the registered peer-id set; a deregister-then-reregister cycle collapsed into one notification (and the Upgraded-branch connection swap, unconditionally) leaves the stale keeper refreshing its captured ephemeral entry forever and never sending new_shutdown.
  ●  444 [correctness]    The keeper refreshes with MetaAddr::new_responded(addr, None), and the merge sets rtt with no fallback to the previous value, so every 45s refresh wipes rtt/ping_sent_at and getpeerinfo reports pingtime null for exactly the native peers this PR newly exposes.
crates/zakura-network/src/peer/handshake/tests.rs
  ●  975 [test-coverage]  The new metadata test hard-codes the inbound side's ephemeral entry as expected (locking in the ephemeral-port bug), uses a test-only connector with endpoint None so the keeper half is a silent no-op, and the keeper test's post-deregister recv can race the 20ms refresh tick.
```

```
 The latest commit fixes the ordinary duplicate-upgrade issue. Several other findings remain valid. Three findings describe pre-existing behavior that
  this PR exposes or expands.

   Location                                  Verdict                           Scope                             Assessment
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   handshake.rs:1150 inbound ephemeral       Valid                             PR regression                     InboundDirect contains the remote
   address                                                                                                       ephemeral port. record_zakura_upgrade
                                                                                                                 now inserts that address. The crawler
                                                                                                                 can later dial it, and
                                                                                                                 AddressBook::cacheable does not exclude
                                                                                                                 inbound entries.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   handshake.rs:1144 duplicate entries       Mostly fixed                      PR regression                     The latest commit makes
   and keepers                                                                                                   record_zakura_upgrade ignore Duplicate.
                                                                                                                 A winning replacement for the same node
                                                                                                                 ID can still return Upgraded and create
                                                                                                                 another keeper. The “unconditionally”
                                                                                                                 and “no dedup or cap” wording is no
                                                                                                                 longer accurate.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   peer_info.rs:57 unsanitized subver        Valid                             PR regression                     The peer controls this field. Zakura
                                                                                                                 limits its length but accepts control
                                                                                                                 bytes. JSON serialization escapes those
                                                                                                                 bytes, so they cannot corrupt the JSON
                                                                                                                 syntax. A client that parses and prints
                                                                                                                 the value can still interpret ANSI
                                                                                                                 controls. zcashd stores and returns a
                                                                                                                 sanitized value.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   peer_info.rs:58 negotiated rather than    Valid                             PR compatibility regression       zcashd returns the peer’s advertised
   advertised version                                                                                            nVersion. The PR stores min(local,
                                                                                                                 remote). connection_info.remote.version
                                                                                                                 is available at both update sites.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   peer_info.rs:14 empty/zero unknown        Valid design concern              PR-specific                       serde(default) only affects
   metadata                                                                                                      deserialization. Serialization always
                                                                                                                 emits subver: "" and version: 0. This
                                                                                                                 conflates unknown with a legitimate
                                                                                                                 empty user agent. Accepted connections
                                                                                                                 cannot have protocol version zero.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   peer_info.rs:36 manual constructor        Factually true, not a defect      PR-specific                       Only four in-repository integration
                                                                                                                 tests call it. The PR explicitly
                                                                                                                 promises downstream compatibility, so
                                                                                                                 removing it could break external
                                                                                                                 callers. The repeated defaults are
                                                                                                                 minor cleanup debt.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   meta_addr.rs:1167 UpdateConnected         Valid                             Pre-existing guard, new impact    The five-second concurrency rule
   rejected after Failed                                                                                         already made Failed dominate Responded.
                                                                                                                 The PR makes this rule drop the only
                                                                                                                 update carrying the new metadata. The
                                                                                                                 new keeper shutdown race also makes
                                                                                                                 this path more reachable.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   meta_addr.rs:1017 empty user agent        Invalid as correctness finding    PR-specific                       An empty user agent is a legitimate
   overwrites prior value                                                                                        advertised value. UpdateConnected
                                                                                                                 always carries the actual handshake
                                                                                                                 value. Some("") correctly means “known
                                                                                                                 and empty,” while None means “not
                                                                                                                 collected.”
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   meta_addr.rs:1237 NeverAttempted merge    Not reachable in production       PR-specific                       Only UpdateConnected supplies metadata,
   drops metadata                                                                                                and it creates Responded. Production
                                                                                                                 NeverAttempted entries therefore have
                                                                                                                 no handshake metadata. The branch also
                                                                                                                 hard-codes several other connection-
                                                                                                                 only fields to None, contrary to the
                                                                                                                 finding’s wording.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   meta_addr.rs:300 allocations and clone    Valid efficiency observation      PR-specific                       Each applied UpdateConnected allocates
   churn                                                                                                         an Arc<str>. Dropping Copy added clones
                                                                                                                 across the address book and codecs.
                                                                                                                 Most later clones only increment the
                                                                                                                 Arc reference count, so this needs
                                                                                                                 measurement before treating it as
                                                                                                                 significant.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   meta_addr.rs:232 metadata belongs in      Design alternative                PR-specific                       The legacy peer set holds
   the live peer set                                                                                             ConnectionInfo, but the current RPC
                                                                                                                 abstraction cannot access it. Upgraded
                                                                                                                 native peers also do not produce a
                                                                                                                 legacy LoadTrackedClient. Moving the
                                                                                                                 metadata requires a shared live-peer
                                                                                                                 registry rather than a small local
                                                                                                                 change.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   zakura.rs:464 stale shutdown marks a      Valid                             PR regression                     This PR added UpdateFailed on keeper
   new connection failed                                                                                         exit. If it arrives after a same-
                                                                                                                 address re-upgrade, the concurrency
                                                                                                                 rule makes Failed beat the new
                                                                                                                 Responded update.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   zakura.rs:463 stale keeper survives       Valid                             Mostly pre-existing               tokio::watch exposes only the latest
   re-registration                                                                                               registered-ID set. A remove/add cycle
                                                                                                                 can appear as “still present.” An
                                                                                                                 Upgraded replacement sends the same ID
                                                                                                                 set, so the old keeper sees the ID and
                                                                                                                 keeps refreshing its captured address.
                                                                                                                 The PR expands this problem to inbound
                                                                                                                 ephemeral addresses.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   zakura.rs:444 refresh clears RTT          Partially valid                   Pre-existing                      UpdateResponded(None) clears rtt and
                                                                                                                 ping_sent_at. The keeper already did
                                                                                                                 this before the PR. The new
                                                                                                                 UpdateConnected also clears RTT
                                                                                                                 immediately, and native peers do not
                                                                                                                 currently produce a replacement RTT, so
                                                                                                                 the claim overstates the 45-second
                                                                                                                 refresh as the sole cause.
  ────────────────────────────────────────  ────────────────────────────────  ────────────────────────────────  ─────────────────────────────────────────
   handshake/tests.rs:975 coverage and       Valid                             PR-specific plus existing gap     The metadata test requires the inbound
   race                                                                                                          ephemeral entry. Its injected connector
                                                                                                                 has no endpoint, so it never tests
                                                                                                                 keeper creation. The shutdown assertion
                                                                                                                 can receive another 20 ms refresh
                                                                                                                 before UpdateFailed; one run passing
                                                                                                                 does not remove that race.
```

evan-forbes:
closing in favor of #730

PR #692, opened 2026-08-16
fix(header-chain): close remaining V12 findings

evan-forbes:
## Summary

This PR fixes the nine original V12 findings that remain outside PRs #665, #666, and #667. It also fixes the follow-up audit findings in transition policy binding, verified-projection recovery, startup deferral settlement, historical finality authentication, V1 migration, and security documentation.

## Findings and fixes

- F-225509 binds the complete network policy to durable metadata through a stable digest. Transition admission and every verifier path reject policy drift.
- F-225510 rejects conflicts between bootstrap checkpoints, release pins, and local checkpoints. Exact duplicate assertions remain valid.
- F-225511 and F-225516 validate every supplied full-state path member against retained state and body authority. The implementation only scans the supplied indexed path.
- F-225514 binds every planned transition to its exact source engine and revision. Installation rejects detached or stale plans in constant time.
- F-225512 keeps recovery semantically pure. Startup settles a due recovered deferral through one normal transition before publication. Startup propagates a planning failure instead of publishing a runtime that would repeat the failure.
- F-225520 authenticates every historical current frontier through linked predecessor context or the independent canonical index. Recovery authenticates settled headers-only witnesses through the canonical index. Recovery authenticates above-finalized witnesses through a bounded retained-row walk to the current finalized frontier.
- F-225521 records checkpoint effects only when finality appends a checkpoint.
- F-225522 checks replay conflicts first. It returns `AlreadyApplied` for finality-consumed work second. It checks exact fingerprint equality third.

## Follow-up audit repairs

- Transition admission compares the durable network-policy digest with the active configuration. Ordinary, checkpoint, exhaustive, and auxiliary verifier paths reject digest drift.
- Recovery and engine hydration require every integrated verified-projection node after finality to remain eligible and body-verified. Recovery preserves verified body state on side branches outside the verified projection.
- Startup rejects a due deferral when settlement planning exhausts a durable counter. Future deferrals still open without settlement.
- V1 metadata cannot authenticate a configurable Testnet or Regtest policy. Startup migrates Mainnet V1 stores and rejects ambiguous Testnet or Regtest V1 stores before writing.
- Recovery proves that every historical finality frontier belongs to the authenticated chain. It rejects missing or forged historical witnesses.
- The security documentation states the replay order and the historical witness authentication paths.

## Performance

- Ordinary admission adds one constant-time digest comparison and no graph-wide scan.
- Full-state validation makes one pass over the supplied path and uses retained indexes.
- Source binding uses an O(1) capability and revision comparison without graph allocation.
- Startup scans only the bounded recovered deferral rows and runs a transition only when one is due.
- Recovery uses the bounded predecessor context before it reads the canonical index for older historical frontiers.
- Recovery uses one canonical-index lookup for each settled witness.
- Recovery uses one bounded parent walk of at most `local_finality_depth` retained lookups for each witness above finality.

## Durable format gate

This PR bumps the provisional header-chain disk schema from version 2 to version 3. No released Zakura version writes schema version 2, so this provisional schema does not require a migration. Startup continues to migrate released Mainnet V1 stores. Testnet and Regtest V1 stores must rebuild because V1 did not record their configurable network policy.

## Replacement verification

- Focused network-policy, verified-projection recovery, hydration, deferred-startup, V1 migration, and historical finality regressions pass.
- `cargo test -p zakura-header-chain --all-targets` — 250 unit tests and 1 integration test passed; 1 benchmark test was ignored.
- `cargo test -p zakura-state --lib -- --test-threads=1` — 466 passed; 3 ignored.
- `cargo clippy -p zakura-header-chain -p zakura-state --all-features --all-targets -- -D warnings`
- `git diff --check`

## Scope boundary

This PR changes the public Rust API. It makes `EngineConfig::network` private and exposes `EngineConfig::network()` so callers cannot invalidate the cached network-policy digest. It also adds typed variants to public exhaustive error enums. This PR changes the provisional header-chain database schema described above. It does not change the wire protocol or public configuration surface.


crates/zakura-header-chain/src/transition/recovery/audit/authoritative_rows.rs
```diff
@@ -139,14 +139,45 @@ pub(super) fn check_authoritative_rows<S: StoreAuditSnapshot>(
...
+                let witness_is_authentic =
+                    if selected_tip.height <= metadata.frontiers.finalized.height {
+                        store.authenticated_canonical_hash(selected_tip.height)?
+                            == Some(selected_tip.hash)
+                    } else {
+                        witness_descends_to(&by_hash, selected_tip, metadata.frontiers.finalized)
+                    };
+                if !witness_is_authentic {
+                    invalid_history = true;
+                }
```
p0mvn, line 178: **Merge blocker:** this re-authenticates every historical `HeadersOnlyDepth` witness against *current* state — the live canonical index, or a retained walk to the current finalized frontier. Commit-time proof is only `view_header_ancestor(selected_tip, record.current.height)` (`invariants/mod.rs`). The witness is not a second pin; the `depth` blocks above `current` stay reorgable.

A later legal reorg that forks above the then-finalized frontier abandons that witness branch while the append-only history still points at it. Reproduced in the `authority.rs` harness (depth 2; branch A witnesses epochs 0–1 at h3/h4; B forks at h3, wins, finality advances onto B):

```text
Err(Source { violations: [Finality] })
```

Swapping only the two witness frontiers to B's equivalents makes the identical store audit clean, so this check is the sole cause. `AuditViolation::Finality` has no repair classification, so a 2-block headers-only reorg plus restart cannot start again without deleting the header-chain DB. `local_finality_depth = 1` is enough whenever the reorg changes the canonical hash at `current.height + 1`.

The new tests only cover forged or missing witnesses on a chain that is still canonical (`recovery_authenticates_historical_witnesses_against_the_current_frontier`, `recovery_rejects_an_absent_historical_witness`). The abandoned-but-honest case is untested.

The witness's only job is `current.height == tip.height - depth`, and `current` is already authenticated independently in this closure. Either run the live-chain descent check only on the latest record, or treat "witness no longer provable from retained rows" as unauthenticated — keep fatal only for a retained/canonical contradiction.

p0mvn (commented):
Merge blocker: recovery rejects a legitimate headers-only reorg. Details on the witness check below.

crates/zakura-header-chain/src/transition/recovery/audit/authoritative_rows.rs
p0mvn: I will address this

p0mvn: https://github.com/zakura-core/zakura/pull/709

p0mvn (approved):

p0mvn:
Pushed test to fix CI

PR #694, opened 2026-08-16
ci(release): publish crates from the release tag via Trusted Publishing

p0mvn:
## Motivation

[#654](https://github.com/zakura-core/zakura/pull/654) proved the plan and the OIDC identity but deliberately had no upload path, so releases still depend on a maintainer running `cargo login` and a hand-ordered publish loop from a laptop. That is the credential and the drift this initiative removes: the rc3 versions were published by hand from a commit that was never tagged.

This PR adds the upload path. It stays manually dispatched; wiring it into `Create release` is the follow-up.

The `oidc-smoke` job of [run 31961981502](https://github.com/zakura-core/zakura/actions/runs/31961981502) has now passed, so the `crates-io` environment and at least one trusted-publisher entry are confirmed working.

## Solution

`scripts/publish-crates.sh` gains a `publish` mode alongside `plan` and `verify`:

- The publish set still comes from `cargo metadata` filtered against the live sparse index, so a crate already published at its workspace version is excluded. That is what makes publishing **resumable**: a run that fails partway through is recovered by dispatching it again, not by yanking.
- **Hard error on an unpublished crate name.** Trusted Publishing cannot create a crate, because the configuration authorizing this workflow lives on a crate that must already exist. Previously a `NOTICE`; discovering this mid-upload would leave a partial, irreversible publish.
- **Advisory when a selected version sits below the newest already published.** This is expected for a hotfix on an older release line, and otherwise means the run is publishing from a stale tag — `v1.0.3-rc1`, `v1.1.0-rc0`, and `v1.2.0-rc0` were tagged but deliberately never published, and re-dispatching one of those would permanently add stale prereleases. It is an advisory, not an error, precisely because the hotfix case is legitimate.
- `cargo publish --workspace --no-verify --locked` with the computed exclusions. `--no-verify` is required rather than a shortcut: `verify` compiled these exact packages minutes earlier in a token-free job, and repeating those builds would outlast the 30-minute token. Cargo orders the multi-package publish and waits for index availability itself, replacing the hand-ordered loop.

`.github/workflows/publish-crates.yml` splits into `plan-verify` → `publish` → `verify-published` → `install-check`, with a `mode` input. `oidc-smoke` is kept under `mode: verify` as the zero-upload pre-flight.

`scripts/verify-published-crates.sh` is new. crates.io records the repository, run ID, and commit behind every Trusted Publishing upload and serves them publicly as a version's `trustpub_data`; asserting those name this run distinguishes "the release pipeline published this" from "a version with the right number exists".

### Why this workflow keeps the publishing identity

crates.io matches its Trusted Publisher configuration against the OIDC **`workflow_ref`** claim, which names the *top-level* workflow a run starts from — see [`handle_github_token_inner`](https://github.com/rust-lang/crates.io/blob/main/src/controllers/trustpub/tokens/exchange/mod.rs) and [`extract_workflow_filename`](https://github.com/rust-lang/crates.io/blob/main/crates/crates_io_trustpub/src/github/workflows.rs). A `workflow_call` from `create-release.yml` would therefore present `create-release.yml`, requiring every crate to be reconfigured. Release automation must **dispatch** this workflow, which the follow-up PR does. (crates.io also rejects the `workflow_run` event outright, so chaining that way is not available either.)

### Token exposure

The POC kept the token in a checkout-free job. Real publishing cannot: the token and the repo checkout must be in the same job. Mitigations are the step-scoped env var, `persist-credentials: false` on both checkouts, tooling pinned from the dispatched `main` commit, the action's auto-revoke post step, and the `crates-io` environment approval on the exact run.

## Testing

Linters: `shellcheck scripts/*.sh scripts/lib/*.sh`, `actionlint`, `markdownlint --config .github/.markdownlint.yaml`, and `zizmor --min-severity informational` (CI's setting) — all clean.

Behaviour, run against live crates.io:

- `publish-crates.sh plan` on this branch: 14 publishable crates, 5 to publish. `zakura-header-chain 1.0.0-rc0` is correctly classified `to_publish` rather than a new crate, since `0.0.0` is reserved.
- `plan` against the `v1.1.0-rc0` tag, one of the deliberately unpublished ones: all 7 selected crates flagged below the newest published version. This is the exact case the advisory exists for.
- `publish` with no `CARGO_REGISTRY_TOKEN`: fails with the token error, after the plan and before any upload.
- `publish` against an index where no crate resolves: hard-fails with the unreserved-name error, exit 1, never reaching `cargo publish`. The same input under `plan` exits 0 with a notice.
- 11-case unit test of the semver comparator. `sort -V` orders `1.0.0` before `1.0.0-rc1`, the reverse of semver, which would have hidden the stale-prerelease case — hence the small Python key rather than `sort -V`.
- `verify-published-crates.sh` on four paths: existence-only pass; attestation correctly **failing** against hand-published `zakura@1.2.0` (`recorded: none`); attestation **passing** against `cargo-semver-checks@0.50.0`, a real Trusted Publishing upload; and the index-timeout path.

Not exercised: an actual upload. The acceptance test for that is `mode=publish` against `v1.2.0`, where every crate is already published — the token is minted first and the script then no-ops, proving the exchange and the real job shape with zero upload risk. Happy to run that before this is marked ready.

## Changelog

No changelog fragment: this changes release automation only and touches no Rust source or `Cargo.toml`.

- [crates.io Trusted Publishing](https://crates.io/docs/trusted-publishing)
- Supersedes the in-`create-release` approach of [#157](https://github.com/zakura-core/zakura/pull/157); builds on [#654](https://github.com/zakura-core/zakura/pull/654).

### Follow-up Work

- Stacked on this: `ci/crates-io-publish-wiring` dispatches this workflow from `Create release`.
- **Before merging:** audit that every one of the 14 publishable crates has a trusted-publisher entry for `publish-crates.yml` / `crates-io`. `oidc-smoke` passing proves at least one matched, not all of them. A crate missing its entry is silently absent from the token's `crate_ids` and fails mid-publish. Ownership is split — 13 crates are owned by `ebfull`, `zakura-header-chain` by `p0mvn` — so this needs both.
- The `crates-io` environment currently has a single reviewer. Consider a second so releases do not stall.

### AI Disclosure

- [x] AI tools were used: Claude Code (Claude Opus 5) wrote the scripts, workflow, docs, and this description; the crates.io claim semantics were verified by reading the registry's source rather than assumed, and every behavioural claim above was executed.

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -1,26 +1,43 @@
-# No-upload proof of concept for crates.io Trusted Publishing.
+# Publish Zakura's crates to crates.io using Trusted Publishing.
 #
-# This workflow performs Cargo dry-run publishing before it requests an OIDC
-# token. The token is never exported to repository code, and there is
-# deliberately no registry upload command.
-name: Dry-run crate publishing
+# Verification and publication are deliberately separate jobs. The packaging
+# and compile checks run before any registry token exists, because they take
+# far longer than the 30-minute Trusted Publishing token lifetime; only the
+# publish job holds a token, and only the single cargo step that needs it sees
+# it. `mode: verify` stops after those checks and proves the OIDC exchange
+# without an upload path, which is the pre-flight to run before a release.
+#
+# This is a dispatched workflow rather than one create-release.yml calls:
+# crates.io matches its Trusted Publisher configuration against the OIDC
+# `workflow_ref` claim, which names the *top-level* workflow file, so a
+# `workflow_call` would present create-release.yml as the publishing identity
+# instead of this file.
+name: Publish crates
 
 on:
   workflow_dispatch:
     inputs:
       release_tag:
-        description: "Published release tag to inspect, for example v1.0.3"
+        description: "Published release tag to publish crates for, for example v1.0.3"
         required: true
         type: string
+      mode:
+        description: "verify: run the checks and prove the OIDC exchange without uploading. publish: upload to crates.io (irreversible)"
+        required: false
+        default: verify
+        type: choice
+        options:
+          - verify
+          - publish
 
 permissions: {}
 
 concurrency:
-  group: publish-crates-poc-${{ inputs.release_tag }}
+  group: publish-crates-${{ inputs.release_tag }}
   cancel-in-progress: false
 
 jobs:
-  dry-run:
+  plan-verify:
     name: Plan and verify crate publication
     runs-on: ubuntu-latest
     timeout-minutes: 180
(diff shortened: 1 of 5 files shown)


p0mvn:
@cursor review

p0mvn:
@cursor push 6c4d30c67b

p0mvn:
@cursor review

p0mvn:
@cursor review

ebfull (approved):

PR #695, opened 2026-08-16
ci(release): hand crates.io publishing off from Create release

p0mvn:
## Motivation

With #694 the upload path exists, but it is still a manual dispatch, so the tag and the registry artifacts can drift apart. That is not hypothetical: the rc3 crates were published by hand from a commit that was never tagged, and those registry versions now correspond to no tag at all.

## Solution

`Create release` dispatches `Publish crates` for the tag it just created, as the last step of the `publish` job — after the tag exists.

**Dispatched, not called.** crates.io matches its Trusted Publisher configuration against the OIDC `workflow_ref` claim, which names the workflow a run starts from, so a `workflow_call` would present `create-release.yml` as the publishing identity and require reconfiguring all 14 crates. (`on: workflow_run` is not an option either — crates.io rejects that event.) This reuses the existing `handoff-canary` pattern in the same file: `gh workflow run` with `github.token` and `actions: write`, which is confirmed working in this repo.

**Dispatching also keeps the approval out of `Create release`.** This was [#157](https://github.com/zakura-core/zakura/pull/157)'s open design question 1: a publish job gated inside `create-release.yml` holds the `create-release` concurrency group open while parked on the approval, and GitHub expires an unapproved deployment after 30 days. In its own run, the `crates-io` approval can wait for test/sign/promote without blocking anything.

**New `publish_crates` input**, defaulting to `auto`:

- `auto` publishes a stable tag and leaves a release candidate to an explicit dispatch. RCs have genuinely gone both ways — `v1.1.0-rc1` and `v1.1.0-rc2` were published, `v1.0.3-rc1`, `v1.1.0-rc0`, and `v1.2.0-rc0` deliberately were not — so `auto` asks for a decision rather than assuming one. This is where this PR departs from #157, which hard-blocked every pre-release.
- `always` publishes whatever the tag looks like; `never` skips.
- `allow_unpublishable_crate_graph` forces the skip: that override already means the publish graph does not resolve, so the release is GitHub-only by definition.

**Always `--ref main`,** including for a hotfix release. `Publish crates` resolves and checks out the tag itself, so hotfix releases publish without the `crates-io` environment having to accept `hotfix/v*` branches — it stays `main`-only, unlike the `release` environment.

`check-crate-publish-graph.sh` now warns when a selected crate has never been published. Trusted Publishing cannot bootstrap a crate name, so this belongs in the release PR — it runs as step 5 of `make pre-release`, which `create-release.yml`'s `validate` job already invokes — rather than as a hard failure at T-0.

### DIFF
--- a/.agents/skills/release-zakura/SKILL.md
+++ b/.agents/skills/release-zakura/SKILL.md
@@ -103,6 +103,10 @@ For crates.io publishing:
    runs the same script as the `crates.io publish graph` job in
    `tests-unit.yml` (part of `test success`), so a missing cascade bump
    cannot merge. It is also step 5 of `make pre-release`.
+7. Reserve any crate name the release adds. The same check warns when a
+   selected crate has never been published; CI cannot bootstrap it, because
+   its crates.io Trusted Publishing entry lives on a crate that must already
+   exist. Publish the name manually and configure it before the release.
 
 Partial version graphs are allowed, but all tooling must handle them. Do not
 assume every publishable crate has the `zakura` package version.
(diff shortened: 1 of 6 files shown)


p0mvn:
@cursor review

evan-forbes (approved):

PR #698, opened 2026-08-16
ci(release): gate PRs on the crates.io publish graph

p0mvn:
## Motivation

`cargo semver-checks` compares rustdoc APIs and cannot see a dependency's major bump underneath unchanged type names. Until now, `scripts/check-crate-publish-graph.sh` — the check that *can* see a skipped dependent whose crates.io index manifest still pins the old major — only ran at release time (`make pre-release`, `A-release` PRs). That is how #696's silent-duplicate-major debt reached `main`.

## Solution

Run the same script as the `crates.io publish graph` job in `tests-unit.yml`, part of the required `test success` check:

- Path-gated on PRs that touch `Cargo.toml` / `Cargo.lock` (or this check).
- Always runs in the merge queue and nightly, so a docs-only merge cannot land on top of a broken graph.

A later PR that major-bumps a crate without cascade-bumping its published dependents fails that job instead of merging. The error tells the author to patch-bump the pinning dependents in the same PR.

Depends on #696, which is merged, so `main` is publishable and this gate can land.

## Testing

- `./scripts/check-crate-publish-graph.sh` after the #696 cascade: exit 0 against the live crates.io index (`Publish graph resolves at workspace versions: zakura-chain@6.0.0 … zakura-utils@2.2.1`).
- `shellcheck scripts/check-crate-publish-graph.sh`.
- Not run locally: the full workspace build. This is CI-and-docs only; left to draft CI.

## Changelog

No Rust or `Cargo.toml` change, so no fragment is required.

- Follow-up to #696.
- The cascade rule and both failure shapes: `scripts/check-crate-publish-graph.sh`.

### Follow-up Work

- `scripts/check-crate-version-bumps.sh` still reports every crate as "new since v1.2.0" because `v1.2.0` predates the `crates/` move. Separate from this gate.

### AI Disclosure

- [x] AI tools were used: Cursor (Composer) added the PR CI gate.


p0mvn:
@cursor review

evan-forbes (approved):

PR #700, opened 2026-08-17
fix(blocksync): preserve work across header DAG extensions

evan-forbes:
## Motivation

At height 1,842,004, ordinary checkpoint finality advanced the finalized and verified frontiers without invalidating the selected path. The header planner preserved the body-work epoch, but block sync treated the same-epoch verified generation change as a destructive chain-tip reset.

The reset discarded compatible pipeline work and restarted its state query. The old classifier also accepted arbitrary old-projection suffixes, so it could preserve incompatible work after a reanchor.

## Solution

The header-chain planner now owns body-work compatibility.

It removes old selected and verified entries strictly below the committed `FinalityRecord.current.height`. It then requires each retained old projection to be an exact prefix of the new projection. A finality prefix may retire the complete old projection without changing the body-work epoch. Any retained entry replacement or retirement changes the epoch.

Block sync no longer sends same-epoch verified generation changes through the destructive reset path. It handles every same-epoch exact-authority change as a refresh. Finalized and verified frontier growth uses the non-destructive frontier-advance path. The path preserves the floor, queue, peer requests, retries, reorder buffer, applying bodies, submitted bodies, and `reset_epoch`.

`BodyWorkEpochChanged` remains the production destructive-reset authority. The explicit authoritative frontier-reset input remains available for tests and external reset inputs. Assertions reject impossible same-epoch verified retreat and same-height hash replacement.

The existing invalid-body evidence path remains unchanged.

## Testing

The new regressions cover authoritative finalized-prefix trimming, complete projection retirement, incompatible full-state reanchors, repeated compatible finality publication, same-epoch verified growth, checkpoint pipeline preservation, real selected replacement, and invalid-body evidence after compatible finality.

- `cargo test -p zakura-header-chain`: passed.
- `cargo test -p zakura-state`: passed.
- `cargo test -p zakura-network`: passed.
- `cargo test -p zakura`: 399 passed and 2 failed. The per-peer cap test passed alone. The pruned-block logging assertion also fails on unmodified `origin/main`.
- `cargo xtask header-conformance`: passed all 114 rules for specification v1.4.
- `cargo check --locked --all-features --all-targets`: passed.
- Strict all-feature, all-target clippy passed for `zakura-header-chain`, `zakura-state`, `zakura-network`, and `zakura`.
- `cargo fmt --all -- --check`, `git diff --check`, and `./scripts/changelog.py check`: passed.

GitHub Actions passed with 39 successful checks and 12 skipped checks. Initial setup attempts received HTTP 429, 502, and 503 responses while runners downloaded pinned actions. The retries passed without a branch change.

## Continuous-sync validation

The prior `c067c38c1fa5bfc3154dccd563365d5b0e6201be` run is failed invariant validation.

A bounded trace around height 1,842,004 records `block_chain_tip_reset` during monotonic checkpoint finality. The trace also records seven `view_reset` peer wakes. State had five submitted blocks at the reset. The node committed height 1,842,004 successfully, so a real DAG invalidation did not cause the reset.

The corrected head is `e24667fae1ba8153d1fde56c62bf959b4933b07a`. Its binary SHA-256 is `a43d9663d866ce0ee6696421e1c71a25e3e83662ad5614a4d6e71160476d3525`.

The exact-head guarded observation passed. The node advanced 161,421 heights in 1,816 seconds. It crossed the checkpoint handoff at height 3,446,171. The complete appended block-sync trace contained zero `block_chain_tip_reset` events and zero `view_reset` records. The bounded checkpoint sample records 3,248 applying bodies and 401 submitted applies across non-destructive frontier advances.

The controller started clean genesis run `20260817T171411Z-e24667fae1ba` with the exact head after its sentinel-guarded state cleanup. The durable watcher records height progress, the trace offset at height 1,842,004, ordinary reset counts, and the controller result. The run recorded one expected reset at height zero during initial authority establishment. It recorded zero ordinary destructive resets at the first sample.

This PR stays draft until the clean run reaches tip and the controller reports successful completion.

## Changelog

Added `docs/changelog/unreleased/700.md` under `Fixed`.

### Follow-up Work

Complete clean run `20260817T171411Z-e24667fae1ba`. Attach its bounded incident-height trace, near-tip samples, completion metadata, binary hashes, and watcher result. Restore the controller inventory to `main` after collection.


crates/zakura-network/src/zakura/block_sync/reactor.rs
```diff
@@ -847,31 +827,29 @@ impl BlockSyncReactor {
...
-                        scope: current_scope,
+                    .send(SequencerControlInput::BodyWorkEpochChanged {
+                        authority,
+                        frontiers,
                     });
+            } else {
+                self.state.work_queue.refresh_authority(authority);
+                let _ = self
+                    .sequencer_control
+                    .send(SequencerControlInput::WorkAuthorityRefreshed { authority });
```
p0mvn, line 852: Same-epoch refresh leaves live GetBlocks on the old `BodyWorkAuthority`, so a valid selected-header extension can duplicate-fetch.

This path rebases queued work (`WorkQueue::refresh_authority` updates `pending` only) and retry keys (`refresh_body_retry_scope`). In-flight `WorkItem` owners and `PeerRegistry` outstanding `OutstandingMeta.owner` keep the pre-refresh authority, including the old `branch.target_tip_hash`.

The producer then asks `has_outstanding_request_in_scope(current_scope, height, hash)`, which requires `meta.owner.authority() == scope`. After an extension, `body_work_epoch` is unchanged but `target_tip_hash` is not, so a still-valid outstanding request looks absent. `total_unreceived_in_scope` undercounts the same way.

`in_flight_contains` still suppresses a second fetch until a later reject/timeout/`reset_above` returns the height to `pending` while the old outstanding entry remains. The producer filter then allows another `GetBlocks` for the same height+hash.

Please either:

1. Rebase outstanding owners (and `in_flight` owners, so they stay in lockstep with `owner_for_height`) in this control path, the same way pending/retries are rebased, or
2. Key the producer filter and low-water count by height+hash (`has_outstanding_request` / `total_unreceived`). `BodyWorkEpochChanged` already clears outstanding via `reset_epoch`, so full-authority equality is not needed to ignore old-fork requests.

(1) matches the epoch-vs-authority split in this PR: same epoch means keep the download and rebind the exact coordinate.

crates/zakura-network/src/zakura/block_sync/peer_registry.rs
```diff
@@ -276,16 +278,39 @@ impl PeerRegistry {
...
+            }));
+        let retries = std::mem::take(&mut *self.body_retry_all_lock());
+        self.body_retry_all_lock()
+            .extend(retries.into_iter().map(|(mut key, deadline)| {
+                key.header_generation = current.header_generation;
+                key.branch = current.branch;
+                key.body_work_epoch = current.body_work_epoch;
+                (key, deadline)
+            }));
+    }
```
p0mvn, line 323: `refresh_body_retry_scope` rewrites both retry maps non-atomically, so a concurrent reader can observe a state where suppression has vanished.

`body_retry_lock()` / `body_retry_all_lock()` return a `MutexGuard`, and in

```rust
let retries = std::mem::take(&mut *self.body_retry_lock());
self.body_retry_lock().extend(retries.into_iter().map(...));
```

the temporary guard is dropped at the end of the first statement. The map is left **empty and unlocked** for the whole duration of the `into_iter().map(...)` rebuild, and `extend` then re-acquires the lock. There are two distinct windows:

1. **Empty-map window.** Between the `take` and the `extend`, `body_retry_avoid` (and separately `body_retry_all`) is empty.
2. **Split-authority window.** `body_retry_avoid` is rebuilt at the new authority on lines 296-303 while `body_retry_all` is still keyed at the *old* authority until line 304. In that gap neither map answers a lookup at either authority.

Both are reachable. `refresh_body_retry_scope` is called from `SequencerTask::handle_control_input` on `WorkAuthorityRefreshed`, while `is_body_retry_avoided` and `next_body_retry_deadline` are called from the per-peer pipe-routine tasks — genuinely concurrent on the multi-thread runtime, and `WorkAuthorityRefreshed` is now the *common* path (every same-epoch header extension), not a rare one.

The observable effect: `is_body_retry_avoided` returns `false` for a `(source, key)` that is still inside its backoff, so the routine re-requests a body from a peer that should be suppressed — including the global `body_retry_all` suppression, which is the one meant to hold across every peer. `next_body_retry_deadline` can likewise return `None` and let a routine sleep past a live deadline. It is a bounded liveness/politeness bug rather than a correctness one, but it silently weakens exactly the backoff this PR is preserving work across.

Suggested fix — rewrite the keys under guards held across both maps:

```rust
pub(super) fn refresh_body_retry_scope(&self, current: zakura_header_chain::BodyWorkAuthority) {
    // Lock order must match `is_body_retry_avoided`: all-peers map, then per-source map.
    let mut all_retries = self.body_retry_all_lock();
    let mut retries = self.body_retry_lock();
    *all_retries = std::mem::take(&mut *all_retries)
        .into_iter()
        .map(|(mut key, deadline)| {
            key.header_generation = current.header_generation;
            key.branch = current.branch;
            key.body_work_epoch = current.body_work_epoch;
            (key, deadline)
        })
        .collect();
    *retries = std::mem::take(&mut *retries)
        .into_iter()
        .map(|((source, mut key), deadline)| {
            key.header_generation = current.header_generation;
            key.branch = current.branch;
            key.body_work_epoch = current.body_work_epoch;
            ((source, key), deadline)
        })
        .collect();
}
```

One trap worth calling out explicitly, since it is the reason the current shape is *not* a deadlock: `is_body_retry_avoided` holds `body_retry_all_lock()` while acquiring `body_retry_lock()`, whereas this method touches `body_retry_avoid` first. Today that is safe only because each guard here is dropped at its statement boundary and the two are never held together. Any fix that holds both — including the one above — has to acquire them in the `all` → `avoid` order to stay consistent with the reader. Acquiring them in source order (`avoid` → `all`) would introduce a real lock-order inversion against `is_body_retry_avoided`.

p0mvn (approved):
The problem description and the proposed solution makes sense to me on the design level.

The two AI comments I left are non-blocking but are valuable hardenings to address pre-merge IMO.

p0mvn:
Going to commit some tests to gate against regressions and merge

p0mvn (approved):

PR #707, opened 2026-08-17
fix(header-chain): harden durable recovery

evan-forbes:
## Summary

- Rebase durable header recovery onto current `main`, including #706 and #710.
- Introduce disk format v4 with authenticated direct migrations from v1, v2, and v3.
- Replace overlapping headers-only witness rewrites with a bounded content-addressed witness DAG.
- Preserve full-state transition provenance and reject records that disagree with canonical state.
- Recover durable auxiliary rows as unauthenticated provenance and require process-local authentication before use or peer attribution.
- Restore both LC-TXN-02 conformance tests from #692.

This repair changes the database schema and public Rust types. It does not change the wire protocol or user configuration.

## Disk migration policy

The migration accepts v1 and v2 only under the fixed Mainnet network policy because those formats do not store a network-policy digest. It accepts v3 only when the stored digest exactly matches the configured policy.

Integrated migration authenticates the active frontier through canonical full state. Headers-only migration authenticates the active frontier through the selected tip and its complete retained depth proof. Any missing, conflicting, malformed, unsupported, or over-limit source store fails before publication and leaves the legacy store unchanged.

The migration clears unverifiable legacy finality history and writes one authenticated migration record at the existing finality epoch. It preserves monotonic counters, advances `state_version` once, clears `last_transition`, rewrites auxiliary rows as unauthenticated provenance, and commits all v4 metadata, counts, and witnesses in one batch. Repeated migration calls are no-ops.

The public API adds `FinalitySource::DiskMigration`, `DiskMigrationAuthentication`, and the shared immutable `FinalityWitnessProof`.

## Witness representation and recovery

Each witness key contains the four-byte big-endian height and 32-byte header hash. Each immutable row stores the canonical header, root-reference count, and child-reference count. The store authenticates and audits the required witness-row count.

The witness limit is 132,072 rows. The writer evicts finality history early when a new branch would exceed that limit, but it always retains at least one record. Ordinary extension inserts one witness node and updates only the affected references. Reorg work scales with the replaced suffix.

Recovery checks canonical and retained evidence before it consults the witness DAG. It uses exact height-and-hash keys and memoizes durable parent walks across the startup audit. It rejects forged keys, headers, parents, references, row counts, limits, and full-state provenance.

## Auxiliary lifecycle

Durable auxiliary rows preserve provenance, not a derived outcome. Recovery marks them unauthenticated. The VCT window returns `PendingAuthentication` until the bounded authentication sweep derives a process-local authenticated, rejected, or disputed outcome. The writer does not attribute peer failure before that derivation.

## Release benchmark

The benchmark artifacts live under `art/inbox/pr707-finality-witness-benchmark/`.

| Case | Median | p95 | p99 | Witness reads | Witness row writes | Batch bytes | Throughput |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Original PR `59d90ed7d` | 3.228 ms | 5.147 ms | 5.459 ms | 1,000 | 1,000 | 232,669 | 310 advances/s |
| Repaired head | 0.696 ms | 0.713 ms | 0.852 ms | 2 | 2 | 2,185 | 1,436 advances/s |
| Current `main` `6c60b6581` | 0.486 ms | 0.493 ms | 0.503 ms | n/a | n/a | 1,664 | 2,059 advances/s |

The repaired batch is 99.06% smaller than the original PR batch. An ordinary advance stays below four witness reads, three witness row writes, and 16 KiB.

A one-block reorg performs four witness reads and four witness row writes. A 32-block reorg performs 66 witness reads and 66 witness row writes. The counts confirm that reorg work scales with the replaced suffix.

The 65,540-advance run retained 65,536 history rows and 66,535 witness rows. Startup audited that store in 335 ms. The first post-eviction advance completed in 0.759 ms.

The repaired median is 43.33% slower than current `main`. The investigation attributes the difference to the authenticated durability work that `main` does not perform: two witness point reads, two witness row writes, and 521 additional batch bytes per finality advance. Current `main` also cannot audit the generated historical finality after restart because it has no durable historical witness representation.

## Validation

- `cargo test -p zakura-header-chain --all-targets` (272 passed, 2 ignored)
- `cargo test -p zakura-state --lib -- --test-threads=1` (475 passed, 3 ignored)
- `cargo xtask header-conformance` (116 rules, 0 unimplemented, 16 unique audit IDs)
- `cargo clippy -p zakura-header-chain -p zakura-state --all-features --all-targets -- -D warnings`
- `git diff --check`

The previous CI failure came from an Ubuntu `apt-get` setup step that hung for one hour before any tests ran.


p0mvn:
The disk format changed without a version bump — a store written by current main will not open.

I think we need to bump this PR since our nodes are already on v2, having deployed #692

p0mvn:
Performance: headers-only finality now performs O(depth) work on every finalized block. With the selected tip about 1,000 blocks ahead, each advance builds and clones the ancestry, hashes it three times, performs roughly 1,000 RocksDB point reads, and stages about 1.5 MiB of full witness-value rewrites just to update reference counts; history eviction adds another traversal. Since finality advances nearly every block, this is steady-state sync overhead. Please add a realistic release-mode, on-disk benchmark before merge, covering advances before and after history eviction and reporting latency, database I/O, batch bytes, and end-to-end throughput before versus after this PR. If the estimate is confirmed, please use a representation that avoids rewriting overlapping witnesses, such as boundary/delta updates.

p0mvn:
Performance: `durable_witness` is evaluated eagerly for every eligible history record, even when `canonical_witness` or `retained_witness` is already true. Since `historical_witness_descends_to` walks up to the local-finality depth, this can turn headers-only startup into O(history records × depth) point reads. Please short-circuit the durable lookup, e.g. run it only when both cheap witnesses are false.

p0mvn (approved):
Ack on design/approach

Pre-merge steps:
- Merge main / resolve conflcits
- Ensure we can safely migrate on prod nodes (I think we are fine to bump version again)
- Address perf commments

evan-forbes:
Addressed the [steady-state performance review](https://github.com/zakura-core/zakura/pull/707#issuecomment-5323027075) in [`aafba31ab`](https://github.com/zakura-core/zakura/commit/aafba31ab), with artifacts in [`d556faa68`](https://github.com/zakura-core/zakura/commit/d556faa68). The content-addressed witness DAG reduces an ordinary depth-1,000 advance from 1,000 reads, 1,000 row writes, and 232,669 batch bytes to 2 reads, 2 row writes, and 2,185 bytes. That is a 99.06% batch reduction. Median latency improves from 3.228 ms to 0.696 ms. A one-block reorg uses 4 reads and 4 writes. A 32-block reorg uses 66 reads and 66 writes. The 65,540-advance run retained 65,536 history rows, used 66,535 witness rows, and restarted in 335 ms. The PR body records the current-main comparison and the investigation of its 43.33% median difference.

evan-forbes:
Addressed the [startup lookup review](https://github.com/zakura-core/zakura/pull/707#issuecomment-5323072167) in [`aafba31ab`](https://github.com/zakura-core/zakura/commit/aafba31ab). Recovery now checks canonical and retained witnesses first. It consults the durable witness DAG only when both cheap checks fail. Durable lookup takes the exact height-and-hash frontier key, and one startup audit memoizes parent walks across finality records.

evan-forbes:
Addressed the [disk-format review](https://github.com/zakura-core/zakura/pull/707#issuecomment-5322941432) in [`aafba31ab`](https://github.com/zakura-core/zakura/commit/aafba31ab). The store now uses format v4 and directly migrates v1-v3. V1 and v2 require the fixed Mainnet policy. V3 requires the exact configured policy digest. Integrated migration authenticates canonical full state. Headers-only migration requires the complete active depth proof. Every failure leaves the legacy rows unchanged, and a successful migration writes one authenticated migration record atomically with v4 metadata, witnesses, and row counts.

evan-forbes:
CI infrastructure update: I retried the canceled Ubuntu setup jobs once, as planned. The retry cleared setup for the test-binary preparation jobs, which then launched the test shards. The RPC semver job and all newly launched shards are again stalled inside `./.github/actions/setup-zakura-build` before any project test command. That composite action runs `apt-get update` and `apt-get install` without an acquire timeout. Local strict clippy, locked all-feature/all-target checks, header-chain tests, state tests, and conformance all pass. I am leaving the single retry running, but I will not start a third attempt.

PR #709, opened 2026-08-17
fix(header-chain): recover after a legal headers-only reorg

p0mvn:
## Motivation

The recovery audit added in #692 re-proves every historical headers-only depth witness against today's chain. A witness is only commit-time evidence that `current` sat `depth` below the selected tip. A later legal reorg above the then-finalized pin abandons that tip while the append-only history still names it.

`AuditViolation::Finality` is a source failure with no repair, so a short headers-only reorg plus restart cannot start again without deleting the header-chain database. `local_finality_depth = 1` is enough when the reorg changes the canonical hash at `current.height + 1`.

## Solution

Keep fail-closed authentication of each record's `current` pin. For the witness, fail only when retained rows contradict that pin: a retained row at the wrong height, a parent walk whose links do not descend exactly one height, or a retained child at `current.height + 1` that names a different parent. An unretained or retention-truncated witness is unauthenticated, not invalid.

Do not consult the live canonical index or today's finalized frontier for historical tips. The commit-time invariant in `transition/invariants` still proves each witness descends to its pin against the live graph before the record is written.

## Testing

- `cargo test -p zakura-header-chain --lib recovery`
- `cargo test -p zakura-state --lib -- header_chain::tests::startup`, covering `rocksdb_recovery_rejects_a_retained_headers_only_witness_at_the_wrong_height` and `rocksdb_recovery_accepts_an_unretained_headers_only_witness`
- `cargo clippy -p zakura-header-chain -p zakura-state --lib --tests -- -D warnings`

New coverage in the recovery authority harness: a legal depth-2 reorg that abandons epochs 0–1, a depth-1 reorg that changes `current.height + 1`, unretained historical/latest witnesses, a retained hash at the wrong height, and a retained child naming a different parent at `current.height` (the last two still fatal).

## Changelog

- `docs/changelog/unreleased/709.md`

- Follows [#692](https://github.com/zakura-core/zakura/pull/692), which introduced the historical-witness check; that PR is merged, so this one targets `main`
- Addresses the merge-blocker review on the historical-witness check in `authoritative_rows.rs`


evan-forbes (commented):
was comparing with what we are working on in #707, and there are some conflicts that I think we should handle. This could just mean dropping a few things here.

> partially ai generated 👇 

This PR treats the witness as commit-time evidence that may disappear. Recovery accepts an absent witness and rejects only retained rows that contradict the historical finalized frontier. This approach restores restart availability, but it cannot distinguish a legitimately pruned witness from a corrupted witness hash. Corruption can therefore pass recovery when no retained row exposes it.

The repaired PR #707 preserves the complete witness path in a hash-keyed durable index. Recovery can authenticate an abandoned fork without treating it as current canonical history. The hash key also preserves multiple witnesses at the same height across reorgs. Reference counts retain each path only while finality history needs it.

therefore, imo we should use #707's approach to get:
  - Legal reorgs survive restart.
  - Recovery continues to reject modified finality history.

PR #709 solves availability by weakening authentication. PR #707 solves availability by retaining the evidence required for authentication. The additional storage and write-set complexity provide a clear security benefit.

  Adapt:

  - Change the “unretained witness succeeds” RocksDB test into “durably stored abandoned witness succeeds.”
  - Add corruption cases that delete or replace the durable witness path and require startup failure.
  - Rewrite the README to explain hash-keyed witness retention and authentication.

  Drop:

  - headers_only_witness_contradicts_current.
  - The rule that missing evidence counts as acceptable.
  - Tests that expect an arbitrary absent witness hash to recover.
  - Documentation that calls an absent witness “unauthenticated, not invalid.”

p0mvn:
Let's not waste more time on this since #707 is stronger

PR #710, opened 2026-08-17
fix(state): recognize v1 header-chain stores during startup

p0mvn:
## Motivation

Mainnet nodes with released version-one header-chain metadata panic during service construction with `UnsupportedDiskFormat(1)`. The startup existence probe decodes metadata using only the current format before the block writer can run the existing v1 migration.

Production validation also showed that the existing migrator advances metadata without migrating all changed v1 rows or creating the required bounded tombstone and finality-history counts.

## Solution

- Treat the specific released v1 format marker as an initialized durable header runtime during service construction, while preserving fail-closed behavior for other decode failures.
- Migrate v1 auxiliary deliveries, full-state body-evidence authorities, and consensus-invalid tombstones to v2 encodings.
- Derive omitted heights from retained header nodes.
- Persist bounded tombstone and finality-history counts atomically with migration.
- Idempotently complete an interrupted migration whose metadata already advanced.

## Testing

- [x] `cargo test -p zakura-state version_one_migration_downgrades_legacy_verdicts_atomically`
- [x] `cargo clippy -p zakura-state --all-targets -- -D warnings`
- [x] `./scripts/changelog.py check`
- [x] `markdownlint docs/changelog/unreleased/710.md --config .github/.markdownlint.yaml`
- [x] Deployed `ba80ff411e5a903a3e3cd450fd13d215fdff1f7c` to `us-west-0`: [workflow run](https://github.com/zakura-core/zakura/actions/runs/32077504257)

The regression test stages all changed v1 row forms without the current count rows, verifies startup classification and atomic migration, then recreates an interrupted current-format state and verifies idempotent recovery.

On `us-west-0`, recovery migrated 1,005 authority rows and reconstructed counts for 0 tombstones and 1,854 finality records. The node remained active with a stable restart count, advanced from height 3,451,425 to 3,451,495, and returned HTTP 200 from both `/healthy` and `/ready`.

## Changelog

Added `docs/changelog/unreleased/710.md` for the operator-visible startup fix.

### DIFF
--- a/crates/zakura-state/src/service/finalized_state/disk_format/header_chain_values.rs
+++ b/crates/zakura-state/src/service/finalized_state/disk_format/header_chain_values.rs
@@ -427,6 +427,31 @@ impl FallibleDiskValue for ConsensusInvalidBodyTombstone {
     }
 }
 
+/// Decode a released version-one tombstone, supplying the height omitted by v1.
+pub(crate) fn decode_v1_consensus_invalid_body_tombstone(
+    bytes: &[u8],
+    header_height: block::Height,
+) -> Result<ConsensusInvalidBodyTombstone, HeaderChainValueError> {
+    let mut decoder = Decoder::new(bytes);
+    if decoder.u8()? != 1 {
+        return Err(HeaderChainValueError::UnknownDiscriminant {
+            field: "consensus_invalid_tombstone_version",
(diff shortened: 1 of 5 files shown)


p0mvn:
@cursor review

evan-forbes (approved):

PR #711, opened 2026-08-17
docs(design): add the historical treestate serving design

p0mvn:
## Motivation

The design for serving historical treestates on VCT fast-synced nodes currently lives inside
[#703](https://github.com/zakura-core/zakura/pull/703), where it is mixed in with the frontier grid
implementation. The document covers more than that one PR: parts of it are already on `main` (frontier
derivation and the root check, the subtree-root artifact and its serving path, the typed
`HistoricalTreeUnavailable` error), one part is #703, and one part is not started. Splitting it out lets
the design be reviewed and land on its own, and keeps #703 to the code it actually changes.

## Solution

Adds `docs/design/historical-treestate-serving.md` against `main`, with no code changes.

The document describes how an archive-mode VCT fast-synced node restores `z_gettreestate`,
`z_getsubtreesbyindex`, and the verbose `getblock` / `getblockheader` tree sizes across the absent band
`[U, H)`, where the fast commit path writes no per-height note commitment trees:

- **The invariant it rests on** (§3.1): a note commitment root is a binding commitment to its frontier,
  and `commitment_roots_by_height` is gap-free from `U` and authenticated against the node's own header
  chain. A frontier claimed at height `h`, from any source, is accepted only if its root equals the
  stored root at `h`. That is what lets the frontier artifact be coarse, small, and distributed outside
  the binary without carrying trust weight.
- **The frontier grid** (§3.2) and **on-demand derivation with a bounded cache** (§3.3), including why
  the grid is cost-weighted rather than uniform.
- **The subtree-root artifact** (§3.4), pinned by ommer folding against the final frontier rather than
  by a direct root comparison.
- **Failure policy** (§3.6): every failure path returns the typed error, never a `null` treestate or an
  empty subtree list.
- A **component map** (§4) tying the above to the modules and config keys that implement it, and a
  **status table** at the top recording what is on `main`, what is in #703, and what is not started.

Measurements that size the grid are in an appendix rather than the body.

## Testing

Docs-only, no code changes, so no build or test run applies.

Markdown lint and codespell could not be run locally: this environment has neither `npm` (for
`markdownlint-cli`) nor `pip` (for `codespell`). I checked the file against the rules enabled in
`.github/.markdownlint.yaml` by hand instead (MD004, MD009, MD010, MD012, MD022, MD023, MD026, MD031,
MD034, MD047, MD049, MD050, fence balance) and found no violations. Relying on the Docs Check workflow
for the authoritative run of markdownlint, codespell, and lychee.

Both relative links (`verified-commitment-trees.md`, `header-sync-vct-root-authentication.md`) and the
one in-document anchor resolve.

## Changelog

No fragment. This PR changes no Rust source file and no `Cargo.toml`, so the changelog policy in
`CLAUDE.md` does not require one.

- [verified-commitment-trees.md](../blob/main/docs/design/verified-commitment-trees.md) — the fast sync
  design this one builds on, including the final frontier check at the last checkpoint (§7) and the
  authenticated roots index (§9).
- [header-sync-vct-root-authentication.md](../blob/main/docs/design/header-sync-vct-root-authentication.md)
  — how the roots index is authenticated against the header chain.
- [#703](https://github.com/zakura-core/zakura/pull/703) — the frontier grid artifact, config gating, and
  `zakurad export-historical-treestates`.

### Follow-up Work

- #703 still contains an earlier copy of this document. Once this lands, that branch should drop the
  file or be rebased so the two do not diverge.
- §6 (removing the per-block `trees` size dependency from the client side) is not started and is not
  covered by any open PR.

### DIFF
--- a/docs/design/historical-treestate-serving.md
+++ b/docs/design/historical-treestate-serving.md
@@ -164,12 +164,12 @@ nearest anchor. It is a serving backstop, not a correctness bound.
 
 A completed subtree root is an interior node rather than a complete frontier, so it cannot
 be checked by comparing it directly with `commitment_roots_by_height`. It is nevertheless
-pinned by the final frontier. Ommers at levels at or above the tracked subtree height are
-pairwise hashes of the completed subtrees they span, so folding the candidate roots in
-index order must reproduce every corresponding ommer, and the frontier's own subtree-level
-root checks the boundary case where it ends exactly at a subtree completion. A wrong,
-missing, or extra root therefore fails verification without replaying any of the subtree's
-65,536 leaves.
+pinned by the final frontier. The frontier already holds the pairwise hashes of completed
+subtrees at every level at or above the tracked subtree height, so folding the candidate
+roots in index order must reproduce those stored hashes, and the frontier's own
+subtree-level root checks the boundary case where it ends exactly at a subtree
+completion. A wrong, missing, or extra root therefore fails verification without
+replaying any of the subtree's 65,536 leaves.
 
 The set is small and static: ~1,895 records at the last checkpoint (3,418,406), roughly 70 KB
 framed, growing append-only by a handful of subtrees per month at current usage. It ships
@@ -276,8 +276,8 @@ median below 400k, 3.6 ms median through 1.6M–2.0M.
 
 **Entry size.** 715 B measured. Earlier sizing used 1,489 B, the size of
 `mainnet-frontier.bin`, as a conservative upper bound; that figure includes Sprout, which
-this artifact omits. Adjacent entries share nearly all their ommers, so delta encoding
-would shrink the artifact further if it ever mattered.
+this artifact omits. Adjacent entries share nearly all of their stored sibling hashes, so
+delta encoding would shrink the artifact further if it ever mattered.
 
 **Grid sizing at 3.42M heights.**
 
--- a/docs/design/historical-treestate-serving.md
+++ b/docs/design/historical-treestate-serving.md
@@ -1,0 +1,293 @@
+# Historical treestate serving on fast-synced nodes
+
+A verified-commitment-trees (VCT) fast-synced node never builds the per-height
+Sapling/Orchard/Ironwood note-commitment trees below the last checkpoint. That is
+deliberate and consensus-safe (see
+[verified-commitment-trees.md](verified-commitment-trees.md)), but it removes the data
+behind `z_gettreestate`, `z_getsubtreesbyindex`, and the `trees` field of `getblock` for
+that band. Wallets that sync against a fast-synced archive snapshot cannot start.
+
+This design restores all three without adding a trust surface. The mechanism: publish an
+artifact of per-pool frontiers at a sparse height grid, derive any other height on demand
+by replaying retained block bodies forward from the nearest entry, and accept nothing that
+fails to reproduce the root the node already authenticated. Completed subtree roots ship
+separately, pinned by the same final frontier the commit path already verifies.
+
+## Status (2026-08-17)
+
+| Component | State |
+| --- | --- |
+| Frontier derivation, root check, bounded cache | On `main` |
+| Subtree-root artifact and `z_getsubtreesbyindex` serving | On `main` |
+| Typed `HistoricalTreeUnavailable` in place of `null` / empty list | On `main` |
+| Frontier grid artifact, config gating, `export-historical-treestates` | PR [#703](https://github.com/zakura-core/zakura/pull/703) |
+| Client-side removal of the per-block `trees` dependency (§6) | Not started |
+
+## 1. Problem: the absent band
+
+The fast commit path writes anchors, the history tree, and the
+`commitment_roots_by_height` serving index, then returns before writing per-height trees
+or subtrees (`zakura-state/.../zakura_db/shielded.rs`, the `fast_write.anchor_roots`
+branch). The result is a half-open band `[U, H)` where those column families are empty:
+
+- `U` is `vct_upgrade_height`, the first height this binary vct-committed. For a snapshot
+  synced from scratch it is effectively genesis.
+- `H` is `vct_synced_below`, the last checkpoint. The currently embedded Mainnet frontier
+  puts it at 3,418,406. Shipped snapshots can carry an older marker, so `H` is read from
+  the database and never assumed.
+
+Heights at or above `H` get trees again from semantic sync, so the gap is exactly the
+checkpointed band. Three datasets are missing in it:
+
+| Missing dataset | Column family | Consumer | Behavior before this design |
+| --- | --- | --- | --- |
+| Per-height pool frontiers | `{pool}_note_commitment_tree` | `z_gettreestate` | JSON `null` |
+| Completed subtree roots | `{pool}_note_commitment_subtree` | `z_getsubtreesbyindex` | empty list |
+| Per-height tree sizes | derived from the trees | `getblock` v1, `getblockheader` | hard RPC error |
+
+The `null` treestate is the one that motivates the work. Clients following the
+lightwalletd contract treat an absent tree as the _empty_ tree, so a wallet derives a
+birthday anchor asserting an empty commitment tree at a height deep in the chain, with no
+error raised at the point of corruption. The hard RPC error is louder but less dangerous,
+and it is what stops a wallet in practice, since lightwalletd-style clients fetch the
+`trees` sizes once per block.
+
+## 2. Goals and non-goals
+
+Goals:
+
+- Restore `z_gettreestate`, `z_getsubtreesbyindex`, and verbose
+  `getblock`/`getblockheader` across the absent band on an archive-mode fast-synced node.
+- Add no **new** trust surface: every served frontier and subtree root is cryptographically
+  pinned by something the node already authenticated.
+- Keep the artifact small enough to distribute through the existing release-state pipeline,
+  without embedding it in the binary.
+- Keep the consensus commit path untouched.
+
+Non-goals:
+
+- **Pruned mode.** Derivation needs retained block bodies. A pruned node still cannot serve
+  historical treestates, which is unchanged from today.
+- **Historical Sprout treestates.** `z_gettreestate` already returns no Sprout data at any
+  height on every node, because Sprout stores only its tip frontier. Excluding Sprout
+  removes the hardest part of artifact generation and regresses nothing.
+- **Eliminating the recompute globally.** The ordered pass still happens once, on the
+  publisher host. This design removes it from every consuming node, not from the world.
+
+## 3. Design
+
+### 3.1 The invariant: roots pin frontiers
+
+This is the property the whole design rests on.
+
+A note-commitment root is a binding commitment to its frontier. The VCT commit path
+already exploits this at the last checkpoint, where it accepts the embedded final frontier
+only if `frontier.root()` equals the verified root for each pool (VCT design §7, step 3).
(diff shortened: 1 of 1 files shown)


docs/design/historical-treestate-serving.md
```diff
@@ -0,0 +1,293 @@
...
+  the database and never assumed.
+
+Heights at or above `H` get trees again from semantic sync, so the gap is exactly the
+checkpointed band. Three datasets are missing in it:
+
+| Missing dataset | Column family | Consumer | Behavior before this design |
+| --- | --- | --- | --- |
+| Per-height pool frontiers | `{pool}_note_commitment_tree` | `z_gettreestate` | JSON `null` |
+| Completed subtree roots | `{pool}_note_commitment_subtree` | `z_getsubtreesbyindex` | empty list |
+| Per-height tree sizes | derived from the trees | `getblock` v1, `getblockheader` | hard RPC error |
```
ValarDragon, line 46: oh thats weird - I thought this would be populated via VCT aux data? Its part of the auth data root no?

docs/design/historical-treestate-serving.md
```diff
@@ -0,0 +1,293 @@
...
+
+The `null` treestate is the one that motivates the work. Clients following the
+lightwalletd contract treat an absent tree as the _empty_ tree, so a wallet derives a
+birthday anchor asserting an empty commitment tree at a height deep in the chain, with no
+error raised at the point of corruption. The hard RPC error is louder but less dangerous,
+and it is what stops a wallet in practice, since lightwalletd-style clients fetch the
+`trees` sizes once per block.
```
ValarDragon, line 53: For this range, we should only support get treestate at subtree root boundaries, or every 1000th block imo

ValarDragon: The rest is LWD's problem

ValarDragon: (which would bubble to a new requirement in goals and non-goals)

docs/design/historical-treestate-serving.md
```diff
@@ -0,0 +1,293 @@
...
+location is an operational choice, not a source of trust: a consuming node verifies the
+embedded artifact once per process against its embedded frontier before exposing it to the
+read service, and RPC requests do not repeat the proof.
+
+### 3.5 Serving the three RPCs
+
+- **`z_gettreestate`** is served directly by §3.3.
+- **`getblock` / `getblockheader` sizes** need only a count, not a frontier. A frontier
+  knows its own position, so the size at `h` is the nearest anchor's `tree_size()` plus the
+  commitments added in between, which requires no hashing at all.
```
ValarDragon, line 187: TBH this sounds worth editing our sync to go populate

docs/design/historical-treestate-serving.md
ValarDragon: Ok I guess the claim is "lets not require refactoring LWD (yet)".

docs/design/historical-treestate-serving.md
```diff
@@ -0,0 +1,293 @@
...
+              │                         (nearest anchor ≤ h,     │
+              │                          cache or grid)          │
+              ▼                                 │
+  commitment_roots_by_height                    │
+  (authenticated, gap-free from U) ─────────────┤
+                                                ▼
+                                 root(frontier@h) == stored root(h)?
+                                     yes ─▶ cache, then serve
+                                     no  ─▶ HistoricalTreeUnavailable
+```
```
ValarDragon, line 138: I'm a bit confused why do this.

I think an easier approach is:
- Compute the full subtree and save it locally.
- Compute the frontier off of log(N) hashes once you have the frontier

Shouldn't full subtree computation be 1.2s single core at Tal's latest benchmarks?

docs/design/historical-treestate-serving.md
```diff
@@ -0,0 +1,293 @@
...
+§3.1 only root-checked frontiers ever enter the cache.
+
+`state.max_historical_tree_replay_blocks` bounds what a single request may cost from its
+nearest anchor. It is a serving backstop, not a correctness bound.
+
+### 3.4 The subtree-root artifact
+
+A completed subtree root is an interior node rather than a complete frontier, so it cannot
+be checked by comparing it directly with `commitment_roots_by_height`. It is nevertheless
+pinned by the final frontier. Ommers at levels at or above the tracked subtree height are
```
ValarDragon: lets avoid ommer, its terrible, Zcash-only terminology

docs/design/historical-treestate-serving.md
```diff
@@ -0,0 +1,293 @@
...
+
+- **`z_gettreestate`** is served directly by §3.3.
+- **`getblock` / `getblockheader` sizes** need only a count, not a frontier. A frontier
+  knows its own position, so the size at `h` is the nearest anchor's `tree_size()` plus the
+  commitments added in between, which requires no hashing at all.
+- **`z_getsubtreesbyindex`** is served from the subtree-root artifact, never by replay.
+  Wallets doing spend-before-sync fetch the full subtree list in one request at startup,
+  from index 0 to the tip, so a replay-backed cold response would approach a full-band
+  replay rather than one batch forward. The sequential-access argument that makes §3.3
+  cheap does not apply here.
```
ValarDragon, line 192: I'd rather make this the primary unit that we re-compute, and get this faster :D

docs/design/historical-treestate-serving.md
p0mvn: `z_getsubtreesbyindex` are the subtree roots at height 16. We embedded them as a 70 KB artifact. This was implemented here:
https://github.com/zakura-core/zakura/pull/593

What VCT aux actually stores:
- note-commitment roots 
- per-block shielded tx counts 
- auth_data_root

What `getblock.trees` needs is a different number: the cumulative note count at that height (`sapling_tree.count(), orchard_tree.count()`). That lives on the per-height frontier objects in `{pool}_note_commitment_tree.` Fast commit skips those column families, so verbose getblock / getblockheader hit "missing Sapling tree" / "missing Orchard tree".

This is the class of errors Harsh was getting.

docs/design/historical-treestate-serving.md
p0mvn: > "lets not require refactoring LWD (yet)".

That's the claim. The goal is to unbreak existing queries.  But the direction you are suggesting is also implemented.

I have a follow-up that exposes `z_gettreestateanchor(hash | height)` that hands a client the material to
reconstruct the treestate itself: the highest frontier at or below the target this node has already verified.

The cliet can then do the computation themselves.

It is built here: https://github.com/zakura-core/zakura/pull/704

Just as a separate PR, we require LWD/wallets to migrate to over time.

docs/design/historical-treestate-serving.md
p0mvn: Updated

docs/design/historical-treestate-serving.md
p0mvn: We could do this, but populating sizes during sync only helps getblock / getblockheader. It does not replace the frontier work.

- `z_gettreestate` needs the actual tree at h (the serialized frontier), not a count
- `z_getsubtreesbyindex`the subtree-root artifact, pinned by the final frontier.

So, we could add this as an optimization to `getblock,`, but it would not replace the need to "online path recompute"

docs/design/historical-treestate-serving.md
p0mvn: A subtree root is not a treestate. We already store those roots and serve `z_getsubtreesbyindex` from them.

z_gettreestate needs a frontier (current leaf, left-siblings at every level, position). Folding the ~1,895 subtree roots only rebuilds level-16+ nodes. It cannot recover the current leaf or levels 0–15, so it cannot produce that RPC.

Fast commit does not write per-height frontiers, so we publish a sparse grid of real frontiers and replay retained bodies from the nearest one.

Using subtree completions as those anchors does not bound that replay: Sapling subtree 0 runs to height 558,822, so a treestate in the middle is hundreds of thousands of block bodies, not 1.2 s of hashing.

That is why the grid is spaced by measured replay cost rather than by subtree index or uniform block height (Appendix A / [thread](https://zcashzcaloors.slack.com/archives/C0ALB378NLX/p1787022916208559?thread_ts=1787010722.312799&cid=C0ALB378NLX)).

docs/design/historical-treestate-serving.md
p0mvn: This is a spacing policy decision but is consistent with the overall design.

That being said, from benchmarks, embedding frontiers at subtree boundaries would be worse in terms of performance than what is currently proposed.

This is because we must replay note commitments from the block bodies, and for relatively empty block ranges, there must be many block bodies to be read to have all of the leaves necessary to construct a subtree.

docs/design/historical-treestate-serving.md
ValarDragon: Ah got it. But while we are VCT syncing, we do have access to `sapling_tree.count(), orchard_tree.count())` as well, right? (last blocks + a simple counter over this blocks)

Isn't it easier to just do "if !archive, write it then" ?

ValarDragon: Thats kinda funny its just shielded tx in the header, whats the point of that aside from 0 non-zero lol. Either should've been actions / num_outputs, or a boolean lol

docs/design/historical-treestate-serving.md
ValarDragon: Awesome! Great job

docs/design/historical-treestate-serving.md
ValarDragon: Oh what, the get block query adds the frontier's?

Thats kind of crazy. I want to ... just break that tbh. GetBlock should be the block data, any extra state should be some extra query like get processed block

docs/design/historical-treestate-serving.md
ValarDragon: > Compute the full subtree and save it locally.

What I mean by this is save every single internal node in the subtree. I naively assume that that is what a tree state is - let me go better understand this! I clearly have some gaps here!

docs/design/historical-treestate-serving.md
p0mvn: > we do have access to sapling_tree.count(), orchard_tree.count())

We do not because we freeze the tree at an old frontier. We only regain access at the last checkpoint from an embed.

> (last blocks + a simple counter over these blocks)

We could introduce a new running counter from the block bodies. A Sapling output is one leaf; an Orchard action is one leaf.

But it would only help getblock.trees.size (at the expense of extra index). `z_gettreestate` still needs some sort of offline spacing + online recompute solution.

p0mvn: > Thats kinda funny its just shielded tx in the header, whats the point of that aside from 0 non-zero lol. Either should've been actions / num_outputs, or a boolean lol

Looked into why this was done, and it seems that for a light client implementation that ended up not shipping

docs/design/historical-treestate-serving.md
p0mvn: It only needs tree counts in special verbosity modes, which the query gets from reading the trees.

We could have a running counter to address this. But given that we still need to fix/update `z_gettreestate`, we can keep reading the counts from the tree, and a running counter would be an optimization.

evan-forbes (approved):
will defer to other reviewers

but lgtm! seems sensible. idk how else to do this

p0mvn:
Thanks for the reviews. Received Dev's approval over a call

PR #712, opened 2026-08-18
fix(release): rewrite prerelease dependency requirements

p0mvn:
## Motivation

The Prepare release PR workflow failed while preparing v1.3.0-rc0 because converting `zakura-node-services` from `3.2.1` to `3.2.1-rc0` left dependents requiring `3.2.0`. Stable Cargo requirements do not select prerelease versions, so the next `cargo metadata` invocation failed.

## Solution

Rewrite every workspace requirement for a crate when converting its stable version to an explicit prerelease target, rather than only replacing requirements equal to the crate's current stable version. Add regression coverage for older compatible requirements and multiline dependency declarations.

## Testing

- `bash scripts/tests/test_release_packages.sh`
- `bash -n scripts/prepare-release.sh scripts/lib/release-packages.sh scripts/tests/test_release_packages.sh`
- `shellcheck scripts/prepare-release.sh scripts/lib/release-packages.sh scripts/tests/test_release_packages.sh`
- `git diff --check`

The regression test models the failing `3.2.0`/`3.2.1` to `3.2.1-rc0` transition and verifies unrelated requirements remain unchanged.

## Changelog

No changelog fragment is required because this changes internal release shell tooling only, without Rust or Cargo manifest changes.

- Failing workflow: https://github.com/zakura-core/zakura/actions/runs/32095398911/job/95585677981

### DIFF
--- a/scripts/lib/release-packages.sh
+++ b/scripts/lib/release-packages.sh
@@ -36,3 +36,14 @@ list_release_packages_at() {
     printf '%s\t%s\n' "$package" "$manifest_rel"
   done < <(git ls-tree -r --name-only "$revision")
 }
+
+# Rewrite every requirement for a workspace dependency to an explicit
+# prerelease target. Stable requirements never select prerelease versions, even
+# when they select the corresponding stable package version.
+rewrite_prerelease_dependency_requirements() {
+  local manifest="$1" crate="$2" target="$3"
+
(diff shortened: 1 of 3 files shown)


p0mvn:
@cursor review

czarcas7ic (approved):

PR #715, opened 2026-08-18
fix(state): migrate version-two header-chain stores on startup

p0mvn:
## Motivation

Mainnet nodes already running `ba80ff411` (`CURRENT = 2`) migrated header-chain metadata **1 → 2**. Deploying `ec81c294` (`CURRENT = 3`) panics during service construction:

`Codec(UnsupportedDiskFormat(2))`

`is_initialized()` only special-cased format 1, so the writer never reached a v2 → v3 migration. `europe-west-0` crash-looped on that SHA and was rolled back.

Root cause: [#692](https://github.com/zakura-core/zakura/pull/692) inserted `network_policy_digest` and bumped the format to 3. [#710](https://github.com/zakura-core/zakura/pull/710) taught startup to open format 1 only. Version 2 uses the same layout as version 1 (network kind, no policy digest) with marker 2.

## Solution

- Treat formats **1 and 2** as an initialized store during service construction.
- Decode v2 metadata with the version-one layout and inject the running node's network policy digest.
- Rewrite only the metadata row to format 3. Auxiliary and authority rows are already current.
- Keep the Mainnet-only policy rule used for v1, because v2 still cannot authenticate Testnet/Regtest configuration.

## Testing

- `version_two_migration_injects_network_policy_and_leaves_current_aux_unchanged` — a Mainnet v2 store classifies as initialized, migrates, preserves current aux rows, and passes startup audit.
- `version_two_migration_rejects_an_ambiguous_network_policy_without_writing` — a non-Mainnet v2 store is left unchanged.
- Existing v1 migration tests still pass.

Local: `cargo test -p zakura-state --lib version_two_migration` and `cargo test -p zakura-state --lib version_one_migration`.

## Changelog

- `docs/changelog/unreleased/715.md`

- Follow-up to [#710](https://github.com/zakura-core/zakura/pull/710) and the format bump in [#692](https://github.com/zakura-core/zakura/pull/692).
- `europe-west-0` rollback after `ec81c294`: https://github.com/zakura-core/zakura/actions/runs/32101534608

### DIFF
--- a/crates/zakura-state/src/service/finalized_state/disk_format/header_chain_values.rs
+++ b/crates/zakura-state/src/service/finalized_state/disk_format/header_chain_values.rs
@@ -1519,7 +1519,7 @@ impl FallibleDiskValue for EngineMetadata {
     }
 
     fn decode(bytes: &[u8]) -> Result<Self, HeaderChainValueError> {
-        decode_engine_metadata(bytes, None)
+        decode_engine_metadata(bytes, HeaderChainDiskVersion::CURRENT, None)
     }
 }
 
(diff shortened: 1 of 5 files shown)


evan-forbes (approved):

evan-forbes:
merging to ensure #707 correctly fixes its versioning issues in combination

PR #730, opened 2026-08-18
fix(rpc): track peer metadata in a live registry

evan-forbes:
## Summary

This change exposes peer software identifiers and advertised protocol versions through `getpeerinfo`. It stores active connection data in a new `PeerRegistry` instead of adding connection fields to `MetaAddr`.

The registry owns the peer address, direction, sanitized user agent, advertised version, heartbeat send time, and latest round-trip time. Legacy connection tasks control legacy entries. The Zakura supervisor controls native entries with exact connection generations.

## Root cause

The original approach stored handshake metadata in the address book. The address book models discovered addresses, reconnection policy, and recent liveness. It does not model one active connection lifetime.

That mismatch allowed inbound ephemeral addresses to enter persistent discovery state. Address-book merge rules could drop fresh metadata or apply stale liveness changes. Adding an allocated user-agent field also removed `MetaAddr`'s `Copy` behavior and spread clone changes through address codecs and tests.

The first registry implementation attached a completed legacy handshake to a native connection by peer ID alone. An existing connection made the registration wait succeed immediately. A stale or spoofed handoff could therefore replace the incumbent connection's address, user agent, version, and reconnect metadata.

## What changed

- Added a shared registry for active legacy and Zakura connection snapshots.
- Carried the exact native connection generation through each successful upgrade.
- Required the same peer ID and connection generation when the registry attaches native metadata.
- Reserved one handoff owner per peer ID so a duplicate cannot claim an existing maintained dial.
- Used generation keys so stale attachment, cleanup, and heartbeat updates cannot affect replacement connections.
- Sanitized `subver` with the zcashd-compatible safe character set.
- Reported the remote advertised protocol version instead of the negotiated version.
- Made unknown fallback metadata optional instead of emitting an empty string and version zero.
- Kept inbound ephemeral addresses out of the address book.
- Omitted proxy connection addresses when the transport exposes only the node's local ephemeral socket.
- Ignored duplicate Zakura upgrades when recording metadata or starting liveness keepers.
- Preserved outbound Zakura metadata and legacy-dial suppression across reconnects owned by the maintained upgrade dial.
- Restored the original `MetaAddr` schema and its `Copy` behavior.
- Added a real two-sided legacy-to-Zakura upgrade test that verifies registry publication and cleanup.

## Approach tradeoffs

| Concern | Live peer registry | Address-book fields |
| --- | --- | --- |
| Ownership | Matches active connection and supervisor lifetimes. | Matches address discovery and reconnection state. |
| Native identity | Exact peer ID and connection generation binding rejects stale or spoofed handoffs. This requires explicit handoff and maintained-dial ownership. | Peer-ID-only attachment is simpler but can target the wrong connection generation. |
| Stale updates | Generation keys reject stale attachment, cleanup, and heartbeat updates. | Timestamp and state-precedence merges can discard or overwrite metadata. |
| Inbound peers | Keeps transient addresses in process for the connection lifetime. | Risks persisting and redialing an ephemeral source port. |
| Proxy addresses | Omits connections when the transport has no trustworthy remote socket. | Reporting the available socket would identify the local node instead of the peer. |
| Schema | Separates connection metadata from gossip and discovery data. | Reuses an existing type but mixes two lifecycles. |
| Cost | Adds one mutex-protected registry and explicit initialization wiring. | Avoids a new component but adds allocations and clone propagation to address processing. |
| Native reconnects | Retains the last outbound legacy handshake snapshot and suppresses redundant legacy dials while its maintained Zakura dial owns reconnects. Inbound snapshots end with the exact native connection. | Retains metadata and reconnect eligibility according to address-book liveness rather than native connection ownership. |
| Compatibility | Production RPC uses the registry. Tests and downstream `AddressBookPeers` implementations retain the old recently-live fallback and omit unknown metadata. | Produces metadata from any address entry that retained it, including entries that no longer represent an active connection. |

The registry adds synchronization and wiring. In return, it gives the data one owner and one lifetime. Native-only connections that never completed a legacy version handshake do not have `subver` or legacy protocol-version metadata to publish.

## Audit coverage

The audit covered the top-level findings on PR #682, the resolved inline review, and PR #730's generation-binding and proxy-address paths. The new design removes the merge, allocation, inbound persistence, stale attachment, stale shutdown, RTT clearing, duplicate metadata, and misleading proxy-address failure modes from the RPC path.

PR #682 has no unresolved review thread. The first PR #730 CI run exposed a one-shot trace flush race in the regtest harness. PR #734 fixes that test race separately. The canceled `zakura-rpc` semver job stalled during Ubuntu package installation and never reached Cargo.

## Validation

- `cargo fmt --all`
- `cargo clippy -p zakura-network -p zakura-rpc --all-targets -- -D warnings`
- `cargo test -p zakura-network --lib`
- `cargo test -p zakura-rpc --lib`
- `cargo test -p zakura-rpc --test serialization_tests test_get_peer_info`
- `cargo test -p zakura-network mutual_p2p_v2_legacy_upgrade_forms_zakura_connection`

The branch merges current `main`, which contains PR #680's block-sync fuzz harness fix. The full network library run passed 1,087 tests, failed none, and ignored 3. The RPC library run passed 113 tests, failed none, and ignored 1. The peer-info serialization run passed all 3 selected tests. The two-sided upgrade test passed.


evan-forbes:
found one thing in the audit:

```
The native metadata attachment uses only peer_id, not the exact connection generation. The existing registration wait immediately succeeds when that peer ID is already connected. 
A stale or spoofed legacy upgrade can therefore overwrite the incumbent connection’s address, user agent, version, and reconnect metadata. See zakura.pr730-audit-main/crates/zakura-network/src/peer_registry.rs:115, zakura.pr730-audit-main/crates/zakura-network/src/peer/handshake.rs:1147, and zakura.pr730-audit-main/crates/zakura-network/src/zakura.rs:291. 
The upgrade outcome should carry the exact native generation. The registry should reject attachment to any other generation.
```

fixing now brb

crates/zakura-network/src/peer/handshake.rs
```diff
@@ -1101,28 +1123,59 @@ where
...
+    };
+
+    // Keep the upgraded peer's legacy address-book entry live for as long as
+    // its Zakura connection is registered. This prevents the outbound crawler
+    // from re-running the upgrade.
+    connector.spawn_legacy_liveness_keeper(
+        peer_id.clone(),
+        book_addr,
+        address_book_updater.clone(),
+    );
```
p0mvn, line 1196: ### Bug: liveness keeper is one-shot; maintained native ownership is not

`attach_native_metadata(..., retain_for_redial = true)` correctly keeps handshake metadata across native reconnects owned by `ensure_upgrade_native_dial`. But `spawn_legacy_liveness_keeper` still exits forever on the **first** supervisor deregister (`run_legacy_liveness_keeper` in `zakura.rs`), and nothing respawns it when the maintained dial reconnects (that path never re-enters `record_zakura_upgrade`).

**Reachable sequence**
1. Outbound upgrade succeeds → keeper starts, metadata retained.
2. Native blips → keeper sees deregister and exits.
3. Maintained dial reconnects → `native_connected` republishes from retained metadata; **no new keeper**.
4. After ~`MIN_PEER_RECONNECTION_DELAY` from the last `Responded` refresh, `reconnection_peers` can dial the book addr again while Zakura still owns the peer.
5. If that TCP handshake stays on legacy (`TemporaryUnavailable` / no upgrade), `register_legacy` runs while a native registry row is still active.

**Consequence:** `getpeerinfo` can list the same logical peer twice (legacy + native), and you can briefly hold both transports. Not consensus-critical, but it undermines the “one owner / one lifetime” goal of this PR.

**Suggested fix:** tie address-book suppression to upgrade-dial ownership (same lifetime as `upgrade_dials` / `forget_native_metadata`), e.g. keep refreshing across re-register, or spawn/abort the keeper with the maintained dial task—not with a single registration interval. Optional belt-and-suspenders: skip legacy crawler dials (or refuse `register_legacy`) while that peer_id still has retained native metadata / an `upgrade_dials` entry.

p0mvn (commented):
Focused note on outbound upgrade ownership vs the one-shot liveness keeper (details on the inline comment).

p0mvn (approved):
LGTM with non-blocking comment

crates/zakura-network/src/peer/handshake.rs
```diff
@@ -136,6 +138,7 @@ where
             minimum_peer_version: self.minimum_peer_version.clone(),
             nonces: self.nonces.clone(),
             zakura_handshake_connector: self.zakura_handshake_connector.clone(),
+            peer_registry: self.peer_registry.clone(),
```
ValarDragon, line 141: Not obvious to me this is intended behavior for a caller, vs acting off of an arc

crates/zakura-network/src/zakura/handler.rs
```diff
@@ -1016,17 +1121,33 @@ pub enum ZakuraOutboundFrame {
...
+
+    pub(crate) fn new_with_peer_registry(
+        max_connections_per_ip: usize,
+        peer_registry: PeerRegistry,
+    ) -> Self {
+        Self::new_inner(max_connections_per_ip, Some(peer_registry))
+    }
+
+    fn new_inner(max_connections_per_ip: usize, peer_registry: Option<PeerRegistry>) -> Self {
+        let (registration_tx, _) = broadcast::channel(ZAKURA_REGISTRATION_EVENT_CAPACITY);
```
ValarDragon, line 1135: Don't care at all for this PR, but seems like some day we may want some API design here

crates/zakura-network/src/peer_registry.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// The sanitized user agent advertised during the handshake.
+    pub user_agent: Arc<str>,
+    /// The protocol version advertised by the peer.
+    pub version: Version,
+    /// Whether the remote peer initiated the connection.
+    pub is_inbound: bool,
+    /// The latest heartbeat round-trip time.
+    pub rtt: Option<Duration>,
+    /// The latest heartbeat send time.
+    pub ping_sent_at: Option<Instant>,
```
ValarDragon, line 30: Ah the core idea of this PR is to get these two objects out of the address book, sounds great!

ValarDragon: any intuition of why this is tracked on heartbeat instead of just general data processing? At least ping_sent_at?

crates/zakura-network/src/peer_registry.rs
```diff
@@ -0,0 +1,372 @@
...
+            rtt: None,
+            ping_sent_at: None,
+        })
+    }
+}
+
+/// Shared registry for active legacy and Zakura peer connections.
+#[derive(Clone, Debug, Default)]
+pub(crate) struct PeerRegistry {
+    inner: Arc<Mutex<RegistryState>>,
```
ValarDragon, line 50: Ah its already Arc'd, perfect.

Is this mutex a problem though? I guess we can see if we get contended then decide?

ValarDragon (approved):
utACK

PR #742, opened 2026-08-19
smoke(network): run only some native Zakura services

evan-forbes:
Draft, for reference. This is the branch that ran the #594 mainnet smoke test, kept so the header-sync-only capability and its evidence are not lost. It is stacked on #594 (merged into `main` here, three conflicts resolved), so the diff against `main` includes that PR.

Not proposed for merge as-is. The reusable piece is `network.zakura.provide_services` / `seek_services`; the rest is smoke-test scaffolding.

## What it adds

**A gate on which native services a node runs.**

```toml
[network.zakura]
provide_services = ["zakura.discovery.v1", "zakura.service_discovery.v1", "zakura.header_sync.v2"]
seek_services    = ["zakura.header_sync.v2"]
```

`provide_services` is `None` by default and then behaves exactly as before — `advertised_service_ids` returns `default_advertised_services()` unchanged. A list restricts the node: `NativeServiceGate` filters the `ServiceRegistry`, the signed discovery record, and the connection owners together, so a service the node does not run drops its capability bits from the control handshake, drops its id from discovery, and cannot hold a connection open. An unknown id fails startup rather than silently disabling a service.

`seek_services` extends the ids the discovery dialer prefers, feeding #594's `dial_candidates_preferring_any_service`.

Tests: `provide_services_gates_native_services_and_advertisements`, `seek_services_config_extends_the_sought_service_ids`, `gated_off_native_services_are_retained_not_dropped`.

**An embedding example.** `deploy/zakura-smoke-service` is a small binary on `zakurad::node::run_with_services` that registers one custom service (`zakura.smoke_echo.v1`, stream kind 64) and pings its peers. It is what proved the embed API works outside the test harness.

## Bug this branch found and fixes

Gating a native service off *dropped* it, and that killed the node. On a Zakura-only node the legacy gossip sink owns the last handle to the inbound service, so dropping the gated-off service closed the inbound service and startup died at `start.rs:632` — "could not send setup data to inbound service" — before the syncer came up. Reproduced with stock `zakurad`, so it was the gate and not the embed path.

`service_registry` now returns the services it did not register, and the endpoint retains them (`retained_services`). Regression test: `gated_off_native_services_are_retained_not_dropped`.

This is latent on `main`: nothing there constructs a native service without registering it, so nothing hits it. It is only a bug once something can gate a service off.

## What the fleet showed

Four mainnet nodes, tag `fleet-svc-disco-594`, two of them snapshot-synced pruned at height 3452420.

| node | region | services | streams it actually opened |
|---|---|---|---|
| `full-0` | nyc3 | all native | block_sync, discovery, gossip, header_sync |
| `headers-0` | fra1 | discovery + header sync | **discovery, header_sync** |
| `echo-full-0` | sfo3 | all native + `smoke_echo` | all of the above + custom(64) |
| `echo-only-0` | ams3 | discovery + `smoke_echo` | **discovery, custom(64)** |

`echo-only-0` runs no native sync service at all. It found `echo-full-0` through discovery, opened only the custom stream, and held one session with continuous ping/pong at 151 ms across ams3 → sfo3. `headers-0` runs header sync and nothing else and does real work: 250-header `header_request_sent` batches and `header_status_received` replies, with its block height correctly pinned at the snapshot height because it has no block-sync service to advance it.

### Memory

RSS four minutes after start. That is the least contaminated reading this run produced: the #741 reconnect storm adds roughly 3.5 KiB of retained heap per connection cycle, so any later sample measures the storm rather than the service set.

| node | runs | RSS @4m |
|---|---|---|
| `full-0` | all services, state at tip | 1054 MiB |
| `echo-full-0` | all services + custom, syncing from genesis | 1085 MiB |
| `headers-0` | header sync only, state at tip | **589 MiB** |
| `echo-only-0` | discovery + custom service, empty state | **130 MiB** |

Even at four minutes `headers-0` had absorbed 15772 handshakes, worth about 50 MiB at the measured rate, so its service-set cost is nearer 540 MiB. A header-sync-only node costs roughly half a full node on the same state, and a node running no native sync service sits near 100 MiB.

By 67 minutes `headers-0` had reached 1300 MiB and `echo-only-0` 1077 MiB, both tracking handshake count linearly, while `full-0` moved 1054 → 1089 MiB. That growth is #741, not the service set.

## Two bugs the same run found, fixed elsewhere

- Bootstrap peers that share no service reconnect ~11×/s — `main`'s dial supervisor, fixed in #741.
- A `CustomService` must claim its own connection or discovery closes it ~1ms after its stream opens — #594's code, fixed in distractedm1nd/zakura#1. This branch carries the *workaround* instead (`9758edd44`, the smoke service overrides `owns_connection_for_peer`), because that is what actually ran on the fleet.

## Follow-up

`headers-0` sent 332 header requests to a peer syncing from genesis, all `busy` or `no_locator_intersection`, and none to the peer at tip. Service preference selects on whether a peer advertises `zakura.header_sync.v2`, not on whether its tip is ahead — and `headers-0` had 9140 `advisory_header_summary` events carrying peer header tips, so the data is already flowing.


evan-forbes:
this was just a for funzies prototype so closing

PR #744, opened 2026-08-19
fix(ci): fail over stalled Ubuntu package mirrors

evan-forbes:
## Motivation

[#733](https://github.com/zakura-core/zakura/pull/733) bounded the shared Ubuntu package phase, but it gave each `apt-get` attempt the entire remaining 1,200-second budget. A stalled first attempt therefore consumed the budget and prevented the loop from running a second attempt.

PR #594 exposed this behavior in seven jobs across five Azure regions. Each job stopped in `apt-get update` after the first attempt exhausted the full budget.

## Solution

- Move the Linux package setup into a testable script.
- Limit each network attempt to 300 seconds within the existing 1,200-second shared budget.
- Remove `azure.archive.ubuntu.com` from the hosted runner's mirror list after the first failure.
- Retry against the remaining official Ubuntu mirrors.
- Install the downloaded packages once with the remaining shared budget.
- Run the retry regression test and both scripts through the fast-lint job.

## Testing

- `bash -n .github/actions/setup-zakura-build/install-ubuntu-packages.sh .github/actions/setup-zakura-build/test-install-ubuntu-packages.sh`
- `bash .github/actions/setup-zakura-build/test-install-ubuntu-packages.sh`
- `shellcheck .github/actions/setup-zakura-build/*.sh`
- `actionlint`
- `git diff --check`

The regression test makes the first package command return the timeout status. It verifies that the helper removes the Azure mirror, applies the per-attempt limit, and runs a successful second attempt.

## Changelog

Not required. This pull request changes CI only.

- #733
- #594

### Follow-up Work

None.


ValarDragon (approved):
Hit us in zakura-core/libraries too

PR #746, opened 2026-08-19
fix(sync): harden checkpoint provenance and recovery

evan-forbes:
## Motivation

Both continuous-sync nodes halted on `8d38b584`. The controller reported no
height progress for 600 seconds, but the first failure occurred earlier in the
checkpoint commit path:

```text
08:10:17.85  state commit fails provenance validation at height 1723542
08:10:18.65  recovery commits checkpoint range through height 1723596
08:20:24     controller halts after no further verified progress
```

The run exposed three related correctness defects.

### Provenance used the wrong state version

`checkpoint_finality_evidence` hashes `state_version`. The state writer created
checkpoint evidence from the version it read before the combined transition.
The combined path installed the auxiliary transition first, advanced the version,
and then validated the checkpoint evidence against that newer version.

Every checkpoint commit bundled with a state-changing VCT authentication therefore
failed provenance validation. A no-change auxiliary transition did not advance the
version, which made the failure appear intermittent.

### Late reset requests rewound recovered verifier progress

The block-sync driver did retry the failed height. The final `no_work` trace only
described its network work queue. At shutdown, node 1 still held 3,591 applying
bodies. The checkpoint verifier owned 401 submitted applies, and 3,190 bodies
waited behind them. Node 2 ended in the same state shape.

One hard state commit error made the sibling checkpoint commit futures fail. Each
failed future queried the state tip and queued a verifier reset. The verifier
consumed one reset per later call. A late reset could therefore rewind progress
after recovery had already committed a newer checkpoint. The verifier then waited
for an earlier range that the driver would not submit again.

### The combined path accepted a stale checkpoint request

The combined path skipped provenance validation when
`checkpoint_request.expected_version` did not match the current version. It then
discarded that expected version and planned the checkpoint against the
post-auxiliary version. The plain path preserves the request version and can return
`ApplyResult::Stale`.

## Solution

- Validate checkpoint provenance against the pre-auxiliary snapshot that produced
  the evidence.
- Return `ApplyResult::Stale` before staging the auxiliary transition when the
  checkpoint request names an old state version.
- Assign each checkpoint commit to a reset generation.
- Drain reset requests for the current generation and reset to their highest
  durable tip.
- Advance the generation after that reset so late sibling failures cannot rewind
  recovered progress.

The existing driver retry remains unchanged. The verifier now preserves the
progress that retry produces.

## Testing

`combined_checkpoint_rejects_stale_version_and_accepts_pre_auxiliary_evidence`
first submits a stale checkpoint request. It asserts that the combined path returns
`Stale`, does not swap memory, and does not publish any state. The test then submits
evidence bound to the pre-auxiliary version. It asserts that the atomic commit
succeeds and advances the durable version twice.

`checkpoint_grow_provenance_is_bound_to_one_state_version` proves that checkpoint
evidence validates only against the version that produced it.

`stale_commit_reset_must_not_rewind_recovered_checkpoint_progress` queues multiple
resets for one failed generation. It asserts that recovery uses the highest tip.
It then delivers a late reset from that generation and asserts that verified
progress does not rewind.

The existing `newer_request_must_not_rewind_verified_checkpoint_progress`
regression continues to pass.

Validation:

- `cargo test -p zakura-consensus checkpoint::tests:: --lib` — 13 passed
- `cargo test -p zakura-state service::finalized_state::header_chain::tests:: --lib`
  — 52 passed
- `cargo clippy -p zakura-consensus -p zakura-state --all-targets -- -D warnings`
- `git diff --check`

## Changelog

`docs/changelog/unreleased/746.md` now describes the provenance fix, stale-request
result, and reset-generation recovery.

### DIFF
--- a/crates/zakura-consensus/src/checkpoint.rs
+++ b/crates/zakura-consensus/src/checkpoint.rs
@@ -62,7 +62,16 @@ struct QueuedBlock {
     /// The block, with additional precalculated data.
     block: CheckpointVerifiedBlock,
     /// The transmitting end of the oneshot channel for this block's result.
-    tx: oneshot::Sender<Result<block::Hash, VerifyCheckpointError>>,
+    tx: oneshot::Sender<QueuedBlockResult>,
+}
+
+/// The verification result and generation assigned when a queued block resolves.
+#[derive(Debug)]
+struct QueuedBlockResult {
+    /// The queued block's verification result.
+    result: Result<block::Hash, VerifyCheckpointError>,
+    /// The verifier generation that resolved the queued block.
+    reset_generation: u64,
 }
 
 /// The unverified block, with a receiver for the [`QueuedBlock`]'s result.
@@ -71,7 +80,7 @@ struct RequestBlock {
     /// The block, with additional precalculated data.
     block: CheckpointVerifiedBlock,
     /// The receiving end of the oneshot channel for this block's result.
-    rx: oneshot::Receiver<Result<block::Hash, VerifyCheckpointError>>,
+    rx: oneshot::Receiver<QueuedBlockResult>,
 }
 
 /// A list of unverified blocks at a particular height.
@@ -705,6 +714,7 @@ where
         let block = self.check_block(block)?;
         let height = block.height;
         let hash = block.hash;
+        let reset_generation = self.reset_generation;
 
         let new_qblock = QueuedBlock {
             block: block.clone(),
@@ -740,7 +750,10 @@ where
                 // even if the block hash is the same.
 
                 let old = std::mem::replace(qb, new_qblock);
-                let _ = old.tx.send(Err(e));
+                let _ = old.tx.send(QueuedBlockResult {
+                    result: Err(e),
+                    reset_generation,
+                });
                 return Ok(req_block);
             }
         }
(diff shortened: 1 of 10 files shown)


p0mvn:
Merge blocker: the version-advancing combined path persists checkpoint provenance with mismatched fields.

The checkpoint evidence is correctly validated against the pre-auxiliary version `N`, but after the auxiliary transition advances the engine to `N+1`, settlement records the original `H(N, tip)` evidence with `snapshot_before_commit.state_version == N+1`:

https://github.com/zakura-core/zakura/blob/e5913b3a970915f239f219226379aed13b7e8160/crates/zakura-header-chain/src/transition/planner/settlement.rs#L131-L138

Startup audit reconstructs checkpoint evidence from the recorded version, so it computes `H(N+1, tip)` and rejects the finality row:

https://github.com/zakura-core/zakura/blob/e5913b3a970915f239f219226379aed13b7e8160/crates/zakura-header-chain/src/transition/recovery/contracts.rs#L144-L147

The planner needs to carry the pre-auxiliary authorization version separately from the post-auxiliary execution version and persist the former in `FullStateFinalityProvenance`. Please also extend the version-advancing combined-path regression to reopen/audit the store after commit; the current test checks only the live snapshot.

p0mvn (approved):
Pre-approving.

While the edge case is unlikely, unless aux data is state-changing, I think we should address it in this PR.

PR #747, opened 2026-08-20
docs(p2p): add peer message regulation design and specification

evan-forbes:
## Motivation

Zakura must use available P2P capacity while bounding peer-controlled CPU, memory, storage, and buffered output. Metadata exchanges need frequency limits. Block serving must support continuous requests within actual resource capacity.

## Solution

Add four first-draft documents for discovery, header sync, and block sync:

- The design explains cadence, capacity admission, QUIC backpressure, reservations, and adoption order.
- The specification defines bounds and validation for all 14 messages. It retains header push with bounded subscription authorization and credit.
- The testing design describes focused codec, cadence, reservation, capacity, and transport checks.
- The GetBlocks testing plan identifies existing infrastructure and ownership, terminal, load, and bidirectional scenarios.

Requests wait for worker and output capacity. The documents remove byte-rate Work buckets, charges, refunds, mandatory Delay verdicts, and mandatory header push deadlines. Discovery requests gain explicit cadence alongside announcement cadence. Handlers retain sequence, expiry, and state-update policy without requiring separate relevance snapshots.

The scope excludes a new declaration-builder framework, universal panic recovery, mandatory per-decision trace files, and exhaustive model exploration. Policy values still require validation. Stream layouts must demonstrate bidirectional and control progress under backpressure.

## Testing

- Markdown lint and codespell passed on all four documents.
- Local links, reference definitions, code fences, and `git diff --check` passed.
- Checked that all 14 message rules remain and obsolete Work, Relevant, and push-deadline requirements are removed.
- Verified the added implementation-source link through the GitHub API.

This update changes documentation only. It does not implement or validate runtime behavior.

## Changelog

No changelog entry. This PR changes documentation only.


docs/design/peer-message-regulation.md
```diff
@@ -0,0 +1,131 @@
+# Zakura peer message regulation: design
```
evan-forbes, line 1: note to self to change files names before we merge

docs/design/peer-message-regulation.md
```diff
@@ -0,0 +1,226 @@
...
+
+Before Zakura handles an inbound message, it applies every filter required by that message's role:
+
+1. **Safe** — Frame checks precede allocation, and bounded decoding precedes expensive verification.
+2. **Authorized** — Each response matches a live reservation created by a request Zakura sent.
+3. **Useful** — The message can affect a receiver decision.
+4. **Budgeted** — The message fits the work bound for its role.
+
+Only bounded announcement metadata may arrive without a reservation, and a strict cadence limits
+it. Requests pay for the work needed to answer them. Responses consume one-shot, range, or
```
czarcas7ic: > and a strict cadence limits it

On call discussed, this is not true today and that is the goal

docs/specs/peer-message-regulation.md
```diff
@@ -0,0 +1,905 @@
...
+  capacity: messages,
+  refill: messages / second,
+  on_empty: Disconnect,
+}
+```
+
+- Capacity MUST absorb the allowed connect-time send, jitter, and message coalescing.
+- The protocol MUST state a matching sender cadence no greater than half the receiver refill rate.
+- Cadence exhaustion MAY return `Disconnect` only when its configuration prevents a conformant
+  sender from exhausting it.
```
czarcas7ic: An honest peer can send Status updates 30 seconds apart, but if our reads are paused, those updates can pile up and look like spam when we resume. Waiting until we’re allowed to process more Status messages would also hold up block replies behind them on the same stream.

An option AI suggested was a reasonable burst, then disconnecting if it’s exceeded, without banning the peer. 

Are we okay with occasionally closing an honest connection after an unusually long delay, or is “a conformant peer must never hit this limit” a hard requirement? A bigger bucket gives us more room, but doesn’t completely remove that tradeoff.

docs/specs/peer-message-regulation.md
```diff
@@ -0,0 +1,905 @@
...
+   Their mappings MUST use exhaustive matches without wildcard arms.
+2. Each filter MUST cap its per-peer state at a receiver-configured capacity. Peer-provided
+   messages, keys, and counts MAY consume that capacity but MUST NOT increase it. The filter MUST
+   define its behavior at capacity. The receiver MUST size each capacity so its aggregate across the
+   maximum peer count fits the corresponding resource budget.
+3. Every enforced inbound rule MUST have a matching outbound obligation. Local scheduling,
+   finality, reorganization, and work reassignment MUST NOT produce a peer violation.
+4. Every non-`Continue` result MUST record the service, message type, filter, direction, peer, key,
+   result, and reason in `regulation.jsonl`. A full-trace mode SHOULD also record `Continue` results.
+   The implementation MUST rotate the file and bound its total size.
```
czarcas7ic: Can these logs be best effort when the queue is full or the disk fails? For example, if we need to disconnect a peer but can’t write the log, I think we should still disconnect immediately and count the missing record. I don’t think we want message handling waiting on a log write.

czarcas7ic (approved):

PR #748, opened 2026-08-20
feat(mining): advertise admitted blocks optimistically

evan-forbes:
## Summary

- prepare and cache mining candidates by work ID and normalized block fingerprint
- recheck solved header fields before state admission
- advertise prepared mined block inventory after expected-work validation and state queue admission while contextual commit continues
- serve pending block bodies only after commit succeeds
- make cache misses and side-chain candidates use commit-first relay
- add bounded caches, fallback behavior, configuration, metrics, and focused tests

## Safety boundary

Early inventory now requires an exact server-prepared candidate that passed proposal validation. A read-only state preflight validates the submitted time, the exact parent-required target, the body and history commitment, and the selected parent. Only consensus can authorize the admission token for optimistic relay.

This restriction prevents an RPC caller from claiming an easier target, passing context-free proof-of-work checks, and triggering false inventory before state rejects the parent-relative target. A direct submission or prepared-cache miss never receives optimistic-relay authorization.

## Audit fixes

- subscribe to pending-block completion before the state lookup to prevent a commit-removal race
- include caller-supplied work IDs in the prepared-candidate cache size limit
- preserve an existing work ID when the same candidate is prepared without one
- validate expected work and median-time-past before optimistic relay
- require the candidate to extend the selected tip before optimistic relay
- fix all-feature acceptance-test constructors and the strict Clippy failure
- add the required changelog fragment

## Validation

- `cargo clippy -p zakura-state -p zakura-consensus -p zakura-rpc --all-targets --locked -- -D warnings`
- `cargo check -p zakura-consensus -p zakura-state -p zakura-rpc --all-targets --locked`
- `python3 scripts/changelog.py check`
- `cargo test -p zakura-state --lib --locked`: 519 passed; 3 ignored
- `cargo test -p zakura-consensus --lib --locked`: 210 passed
- `cargo test -p zakura-rpc --lib --locked`: 120 passed; 1 ignored
- five focused prepared-relay tests cover selected-tip authorization, side-chain fallback, an easier target, median-time-past, and a forged commitment

## Stack

- Base: `main`.
- Parent: none.
- Children: #754 and #781.
- This PR does not contain #773.


ValarDragon:
woah this is very exciting

crates/zakura-consensus/src/block.rs
```diff
@@ -295,6 +300,46 @@ where
...
+
+                    prepared_block.block = block;
+                    prepared_block.hash = hash;
+                    prepared_block.height = height;
+                    return commit_prepared_block(
+                        state_service,
+                        prepared_block,
+                        request.admission(),
+                    )
+                    .await;
```
p0mvn: Optional hardening: this fast path rechecks PoW, time, and lock times, then commits the submitted block with cached `transaction_hashes` / UTXO metadata. `merkle_root` is in `immutable_candidate_bytes` today, so a changed root already misses the cache — the skip is not a bypass.

Still worth calling `check::merkle_root_validity(&network, &block, &prepared_block.transaction_hashes)?` here. The hashes are already computed, so it is cheap, and it fails closed if the fingerprint contract ever treats `merkle_root` as miner-mutable.

crates/zakura-rpc/src/methods/types/submit_block.rs
```diff
@@ -26,7 +32,173 @@ pub struct SubmitBlockParameters {
     ///
     /// <https://en.bitcoin.it/wiki/BIP_0022#Rationale>
     #[serde(rename = "workid")]
-    pub _work_id: Option<String>,
+    pub work_id: Option<String>,
```
p0mvn, line 38: Optional hardening: `workid` is now a cache key, but deserialization still accepts an unbounded `String`. Zakura issues 32 hex chars, and insert already refuses entries above `MAX_BYTES`, but lookup still allocates and compares the caller-supplied value under the cache mutex.

Cap length at deserialize time (32 hex chars, or a small multiple) so an authenticated `submitblock` cannot turn this field into a large allocation. The same bound would help proposal-mode `GetBlockTemplateParameters.work_id`.

crates/zakura-rpc/src/methods.rs
```diff
@@ -2723,13 +2807,106 @@ where
...
+                        if mined_block_sender.try_send(event).is_ok() {
+                            early_result = Some(receiver);
+                        } else {
+                            pending_blocks.resolve(block_hash, Err(()));
+                        }
+                    }
+                    verification.await
+                },
+                result = &mut verification => result,
+            };
```
p0mvn, line 3217: Optional hardening: `BlockAdmission::wait()` parks until `ADMITTED` or `REJECTED`. State already `reject()`s duplicates, queue replacement, and expiry, but a verifier error *before* `commit_prepared_block` drops the latch still `PENDING`.

This `select!` cancels `wait()` when verification completes, so `submitblock` does not hang today. The latch is still lying: verifier failure is a terminal "never advertise" outcome. Calling `admission.reject()` on that path (or on drop of `CommitMined` if neither `admit` nor `reject` ran) makes `wait()` mean what it says. Safe after `admit()` — first terminal state wins, so a later contextual failure will not rewrite an already-sent early `inv`.

crates/zakura-consensus/src/block.rs
```diff
@@ -295,6 +300,53 @@ where
...
+
+                    prepared_block.block = block;
+                    prepared_block.hash = hash;
+                    prepared_block.height = height;
+                    let admission = request.admission();
+                    if let Some(admission) = &admission {
+                        if check_prepared_mined_relay_eligibility(
+                            &mut state_service,
+                            (&prepared_block).into(),
+                        )
```
p0mvn: Cache unit tests cover identity for parent / version / commitment / difficulty / extra tx, and reuse of solved `time` / `nonce` / `solution`. They insert via `SemanticallyVerifiedBlock::from(block)`, so they never drive `SemanticBlockVerifier` `Prepare` / `CommitMined`. Existing invalid-time / invalid-solution transcripts use `Request::Commit`, which does not take this path.

**Pre-merge**
1. `Prepare` then `CommitMined` with a bad Equihash solution, a far-future `time`, and a header time that fails `lock_time_has_passed`. Those fields are zeroed in the fingerprint, so they must cache-hit and then be rejected by the rechecks here.
2. Failed `Prepare` must not populate the cache; a later `CommitMined` of those bytes must miss and fall through to full verification.
3. Flip `header.merkle_root` in `immutable_candidate_changes_do_not_reuse_work_id`.

**Nice-to-have:** 33rd-entry eviction, `ENTRY_TTL` expiry, oversized `workid` refused by `MAX_BYTES`, re-prepare without a work ID keeps the previous one, `Request::Commit` ignores a populated cache, `insert` does not serve a body until `resolve(Ok)`, inbound state-`None` plus subscribe-before-lookup.

crates/zakura-rpc/src/methods.rs
```diff
@@ -998,6 +1043,38 @@ where
...
     }
 
+    fn prepare_template_in_background(&self, template: &BlockTemplateResponse) {
+        #[cfg(test)]
+        {
+            let _ = (template, self.gbt.try_acquire_template_preparation());
+        }
+
+        #[cfg(not(test))]
+        {
```
p0mvn: Per the readability rules, would it be possible not to have conditional test builds on the prod paths?

p0mvn:
Adversarial pass over the pending-block registry, admission lifecycle, and the rewritten inbound `getdata` path. Two findings are correctness races in the duplicate-submission path; the rest are hardening and regression notes. Deduplicated against the existing inline comments (`workid` bound, merkle-root recheck, latch hygiene, cache test matrix) — none of those are repeated here.

### 1. `PendingBlockRegistry::resolve` is not owner-keyed (`crates/zakura-rpc/src/methods.rs:2855`)

Concurrent duplicate `submitblock` of the same block (pool + failover submitter): task A inserts the hash and sends early inventory; task B's `insert` returns `false`, but B still calls `pending_blocks.resolve(block_hash, Err(()))` when its verification fails fast as a duplicate. B's resolve removes **A's** entry and broadcasts `Failed`, so peers answering A's early `inv` get `notfound` for a block that commits moments later; A's later resolve is a no-op.

Return an owner token/guard from the successful `insert` and require it in `resolve`, so only the inserting lifecycle can terminate the entry. A lifecycle whose `insert` returned `false` should never call `resolve`.

### 2. Queued-duplicate replacement swaps `rsp_tx` but not the `BlockAdmission` (`crates/zakura-state/src/service.rs:951`)

When a commit for a hash still in the queue is replaced, only `rsp_tx` is swapped: the **new** request's admission is rejected and the **old** admission stays on the queued entry — inverting the documented replace-older-with-newer contract. The surviving lifecycle sees `REJECTED` and deterministically loses early inventory even though the block commits; meanwhile the stale old admission can be `admit()`ed on dequeue for a lifecycle whose RPC already returned a duplicate error, which can send an early `inv` and then resolve the registry entry as `Failed` (compounding with finding 1). Swap the admission along with `rsp_tx`.

### Other findings

**NOTE**: I have manually reviewed and think they are okay to skip, but I am attaching for the author's awareness.

- **`getdata` lost the early byte-limit break** (`crates/zakurad/src/components/inbound.rs:656`): all 16 block bodies are fetched concurrently into `lookup_results`, then the `GETDATA_SENT_BYTES_LIMIT` check `continue`s past them. 16 × ~2MB per request across `MAX_INBOUND_CONCURRENCY` slots is multi-GB transient memory plus the wasted disk reads the deleted early-exit prevented. Subscribe pending-waits eagerly, but issue state lookups sequentially and stop at the byte budget.
- **`PendingBlockRegistry::wait()` takes a permit at future-creation time** (`crates/zakura-rpc/src/methods/types/submit_block.rs:168`): a `getdata` repeating one pending hash 16 times burns 16 of the 32 permits per request, so two requests exhaust the semaphore and a peer that genuinely needs to wait gets a spurious `notfound` plus a misleading `pending_peer_wait.saturated` metric. Acquire the permit inside the async block, only when the wait actually begins.
- **`MAX_INBOUND_RESPONSE_TIME` 5s → 18s globally** (`crates/zakurad/src/components/inbound.rs:59`): every inbound request type can now hold a buffered slot ~4x longer, and the 18s↔15s coupling with `PENDING_BLOCK_WAIT` is undocumented at either constant. Scope the longer deadline to the pending-block `getdata` branch, or derive both from one documented constant.
- **Proposal mode lets any RPC client churn the prepared-candidate cache** (`crates/zakura-rpc/src/methods.rs:2480`): proposals issue `Request::Prepare` with a caller-supplied `workid`, so `insert`'s `remove_matching` evicts the background-prepared entry and repeated proposals cycle the 32-entry FIFO, silently defeating the `submitblock` fast path. Content safety holds (byte-equality lookup, only fully validated proposals cached) — consider not caching client proposals, or namespacing them away from server-issued work IDs.
- **Final gossip send is silent and unbounded** (`crates/zakura-rpc/src/methods.rs:2898`): `let _ = mined_block_sender.send(event).await` in a detached task replaced the old `try_send` + RPC error, so a crashed/stalled gossip task means `Accepted` with no advertisement, no error, no log, and detached tasks accumulating on the await. At minimum log the failure and bound the send.
- **Template prep runs CPU-heavy parsing inline** (`crates/zakura-rpc/src/methods.rs:1046`): `proposal_block_from_template` deserializes the coinbase plus every mempool tx on the `getblocktemplate` response path before returning to the miner — move block construction inside the spawned task.
- **Prepared cache stores full bodies alongside the fingerprint** (`crates/zakura-consensus/src/block/prepared.rs:43`): entries keep both the ~2MB serialization and a blake2b fingerprint, lookups do MB-scale memcmps under a `std::sync::Mutex`, and `lookup()` serializes the submitted block even when the cache is empty. The 32-byte fingerprint alone is sufficient identity; drop the stored bytes and short-circuit on an empty cache.
- The drop-based `reject()` suggested in the existing latch-hygiene comment would also cover `Request::CommitMined` at/below `max_checkpoint_height` (`crates/zakura-consensus/src/router.rs:226`), where the checkpoint arm routes via `request.block()`, discards the `work_id`, and drops the admission unresolved — today that path silently loses both the early `inv` and any rejection signal.

Also worth a quick sweep: `docs/changelog/params.md` rows for the new/re-tuned tunables and PR links in `docs/changelog/unreleased/748.md` are missing; `CheckProposal` pays an unconditional `SemanticallyVerifiedBlock` deep clone (`crates/zakura-consensus/src/block.rs:490`); and `BlockAdmission` hand-rolls what `tokio::sync::watch` provides while this PR already uses `watch` for the identical pattern.

p0mvn (commented):
Great work.

The design looks good to me.

Was going to approve but the 2 findings from the adversarial pass look worth addressing.

Would like to review the final changes again before approving.

crates/zakura-consensus/src/block.rs
evan-forbes: ahh nice catch, added the Merkle-root check to the prepared fast path. The cache identity test now covers the mutation case.

crates/zakura-rpc/src/methods/types/submit_block.rs
evan-forbes: makes sense, but I left this unchanged for now.

On the allocation, the server caps the request body at `MAX_BLOCK_BYTES * 2 + 1024`, about 4 MB, so one `workid` cannot exceed that. Retention is bounded as well, because `insert` counts `work_id.len()` against the 64 MiB budget.

You are right that the comparison is the sharper edge, and right that it reaches `GetBlockTemplateParameters.work_id`, since proposal mode caches under a caller-supplied ID. A caller can make `lookup` compare multi-MB strings under the shared cache mutex, which stalls other mining calls, not just their own. Staging those entries needs repeated valid proposals, and `str` equality checks length first, so a normal 32-hex submission short-circuits against them.

What stops me is compatibility rather than effort. A strict length bound would reject work IDs minted by another server, and I would rather not change what `submitblock` accepts inside this PR. Happy to take it as a follow-up with a bound chosen against what real miners send.

crates/zakura-rpc/src/methods.rs
evan-forbes: sg, I traced this through both routes.

`submitblock` is the only `wait()` caller in the tree, and its `select!` cancels the wait when verification finishes, so a pending latch cannot hang anything today. I agree that safety is an accident of having one call site, so I documented the requirement on `wait()`: the future may never resolve, and callers must await it under a cancellation path.

On the checkpoint route, the exposure is smaller than it looks. `max_checkpoint_height` is the last hardcoded checkpoint, so on mainnet it sits far below the tip, and on a custom testnet the list is genesis-only. A mined submission at or below that height is already stale, so that arm never drops a live admission.

I left the redesign for a focused follow-up. `reject()` is `pub(crate)` in `zakura-state`, while the drop guard belongs where `CommitMined` is consumed in `zakura-consensus`, so making the latch honest means widening that boundary.

crates/zakura-consensus/src/block.rs
evan-forbes: makes sense, added real `Prepare` → `CommitMined` tests for invalid Equihash, future header time, and transaction lock time. I also covered failed preparation. The cache identity test now covers Merkle-root mutation.

crates/zakura-rpc/src/methods.rs
evan-forbes: sg, removed the conditional test branches. Tests now use the production preparation path. Template decoding also runs inside the bounded background task.

evan-forbes:
nice catches. I fixed pending registry ownership with owner-keyed registrations, so only the inserting lifecycle can terminate an entry, and a dropped registration now resolves its own entry instead of leaking it until the wait timeout. I fixed queued replacement by swapping the admission along with the response sender.

I also handled the straightforward hardening items. Block loading stops at the byte budget. Waiters acquire permits only when polling begins. Only block requests use the longer timeout, and that constant now documents why it must exceed `PENDING_BLOCK_WAIT`. Final event sends have a timeout, a warning log, and a failure counter. Template parsing runs inside the background task. Proposal checks avoid the unnecessary clone. The changelog lists the new parameters.

On the proposal cache, I took the namespacing follow-up. `Request::Prepare` now distinguishes server templates from client proposals. The cache keeps fixed partitions under the existing total limits: server templates receive 24 entries and 48 MiB, while client proposals receive 8 entries and 16 MiB. A client proposal can evict only another client proposal. Raw `workid` conflicts remain globally checked, so lookup stays unambiguous.

I carried the same split into PR #754’s alias index: 3,072 aliases for server templates and 1,024 for client proposals. This preserves proposal-mode custom-candidate caching while preventing proposal churn from making server-issued `submitsolution` work stale.

I left the fingerprint-only cache rewrite for follow-up work, because it changes cache identity from byte equality to a 32-byte digest. I did take the cheap half now: `lookup` short-circuits on an empty cache, so a cold `submitblock` no longer pays a full block serialization.

I left the admission latch redesign for follow-up work. `submitblock` is the only `wait()` caller in the tree, and its `select!` always terminates, so a pending latch cannot hang anything today. That is an accident of having one call site, so I documented the requirement on `wait()` for the next caller. More on the checkpoint route in that thread.

I left deserialize-time `workid` bounds unchanged, with the reasoning in that thread.

The new regression tests cover registry ownership and queued admission replacement. They also exercise prepared-path PoW, header time, transaction lock time, failed preparation, and Merkle-root changes.

ValarDragon:
@v12sec pls review

evan-forbes:
missed the v 12 stuff, but digging through them now

edit: fixed a while ago but editing now

evan-forbes:
fixing conflicts now

crates/zakura-state/src/service.rs
```diff
@@ -3221,6 +3338,70 @@ impl Service<ReadRequest> for ReadStateService {
...
+    };
+    check::block_commitment_is_valid_for_chain_history(
+        commitment.block.clone(),
+        network,
+        &history_tree,
+        commitment.auth_data_root,
+    )?;
+
+    let relevant_chain: Vec<_> =
+        any_ancestor_blocks(non_finalized_state, db, parent_hash).collect();
```
czarcas7ic: This ends up loading ancestor block bodies all the way back to genesis when a solved block hits the template cache. On an archive node, that's a huge amount of disk work and memory just to submit a mined block. The next check only needs 28 blocks, but its limit kicks in after we've already collected everything. See POW_ADJUSTMENT_BLOCK_SPAN here:

https://github.com/zakura-core/zakura/blob/235c5cfb2c8b1f6995caf2aade55d88630aa7d66/crates/zakura-state/src/service/check.rs#L91-L94

Are we able to limit the iterator before collecting?

crates/zakura-consensus/src/block/prepared.rs
```diff
@@ -0,0 +1,844 @@
...
+                    .any(|entry| entry.work_id.as_deref() == Some(work_id))
+            {
+                inner.proposals.remove_work_id(work_id);
+            }
+        }
+
+        let partition = inner.partition(source);
+        partition.remove_matching(
+            (source == PreparedCandidateSource::ServerTemplate)
+                .then_some(work_id.as_deref())
```
p0mvn, line 209: `Request::Prepare` does not consult this cache until after full semantic and contextual verification. Because every `BlockTemplateResponse` gets a random `new_work_id()`, sequential polls for unchanged content reach this removal with a distinct ID; the fingerprint match discards the previous server entry, and the node rechecks all transaction proofs and signatures before replacing it.

Short-circuit server-template preparation before semantic verification when the immutable fingerprint and bytes already exist in the server partition. On a hit, retain the prepared candidate and add the new work-ID as a bounded/accounted alias instead of removing and reinserting the content, so both repeated polling and submissions using earlier IDs reuse the same verified candidate.

p0mvn: I see that #754 fixes the cache structure: identical candidates share one canonical entry with multiple work-ID aliases.

It does not fix the main performance issue: Request::Prepare still performs full semantic verification before finding/reusing that canonical entry. So unchanged polling still re-verifies proofs and signatures.

crates/zakura-state/src/service.rs
```diff
@@ -1028,23 +1046,40 @@ impl StateService {
...
+            if let Some(old_admission) = old_admission {
+                old_admission.reject();
+            }
+            let _ = old_rsp_tx.send(Err(CommitBlockError::new_duplicate(
+                Some(hash.into()),
                 KnownBlock::Queue,
             )
             .into()));
             rsp_rx
+        } else if self.non_finalized_state_queued_blocks.is_full() {
```
p0mvn: This rejects every new distinct block when the orphan queue is full, including a block whose parent is already available. If the queue is filled by descendants waiting on missing block `B`, `B` receives `QueueFull`; the later call drains children of `B`'s parent, while the queued descendants remain indexed under `B`, so retries cannot restore progress.

Please apply the capacity limit only to blocks that must remain orphaned—for example, reject here only when `is_full() && !self.can_fork_chain_at(&parent_hash)`—and allow an immediately processable block to be queued and drained (or evict an orphan to admit it). Add a state-service regression test that fills the queue with descendants waiting on `B`, submits `B`, and verifies that `B` and its descendants are sent for commit.

ValarDragon:
@v12sec pls review

crates/zakura-state/src/service.rs
evan-forbes: interesting oooh sure thing! 5fb2194c3

I'd be impressed if someone mined 1000 blocks that pass the context checks but definitely doable! dropping does sound like the ideal thing here as we should be recovering even in the event that we drop the winning orphan block

crates/zakura-consensus/src/block/prepared.rs
evan-forbes: yeah fixed in #754 actually as I think that's a better fit b6b31f9c5

crates/zakura-state/src/service.rs
```diff
@@ -1107,13 +1155,37 @@ impl StateService {
...
+                            .non_finalized_block_write_sent_hashes
+                            .contains(&candidate_parent)
+                            || self.read_service.db.finalized_tip_hash() == candidate_parent)
+                        && !self
+                            .optimistic_relay_blocked_parents
+                            .contains(&candidate_parent);
+                    if optimistic_relay_still_authorized {
+                        // Only the first server candidate can reserve early relay for this parent.
+                        // Siblings receive the normal committed relay after contextual validation.
+                        self.optimistic_relay_blocked_parents
```
p0mvn: `optimistic_relay_blocked_parents` is inserted into for every successful optimistic relay and every operator invalidation, but it is never removed from or pruned. Normal mining therefore retains one additional hash per optimistically relayed block for the lifetime of the process, and a successfully reconsidered parent remains permanently ineligible for optimistic relay.

Please separate transient sibling-relay reservations from invalidation state, clear reservations when they are no longer needed, and prune finalized entries. Clear an invalidation entry only after the writer confirms successful reconsideration.

crates/zakurad/src/components/sync/gossip.rs
```diff
@@ -211,26 +215,42 @@ where
...
                 }
-            });
-        } else {
-            tokio::spawn(broadcast_fut);
-        }
+                None => broadcast.await,
+            };
+
+            if succeeded && is_block_submission {
+                let _ = mark_tx.send(hash);
```
p0mvn, line 258: A successful `Early` broadcast is marked here exactly like a successful `Committed` broadcast. A peer can follow the early inventory, exhaust `PENDING_BLOCK_WAIT`, and receive `notfound` while contextual commit continues. If the later committed broadcast then times out or has no ready peer, this earlier mark makes `ChainTipChange` treat the committed hash as already processed, so the committed-tip fallback never prompts that peer to retry.

Only send this mark for a successful `MinedBlockEvent::Committed` broadcast. An early broadcast cannot establish that the body was delivered, so it must not suppress the post-commit fallback. Please cover early success → pending-body timeout → commit success → committed-broadcast failure and verify that committed-tip gossip advertises the hash.

crates/zakura-state/src/service.rs
evan-forbes: Fixed in 05131035c.

crates/zakurad/src/components/sync/gossip.rs
evan-forbes: Fixed in 05131035c.

evan-forbes:
went through all the v 12 feedback and fixed most of them. only left out some lows and a mediums as they weren't worth fixing imo

crates/zakura-network/src/peer/connection.rs
```diff
@@ -1437,7 +1437,9 @@ where
 
                     self.handle_inbound_overload(req, now, PeerError::Overloaded)
                         .await;
-                } else if e.is::<tower::timeout::error::Elapsed>() {
+                } else if e.is::<tower::timeout::error::Elapsed>()
+                    || e.is::<tokio::time::error::Elapsed>()
```
ValarDragon, line 1441: wow thats kind of weird as an issue

ValarDragon (approved):
LGTM! 

- this is safe even if we take longer than 15s to verify our own block
- Could not find any non-MTP / testnet related issues for us mining an invalid block

ValarDragon:
Lets rebase, and use RC tags where possible

evan-forbes:
not blocking, but having fable go through and try cleaning up some of the sloppy code and tests

PR #749, opened 2026-08-20
chore(deps): use Zakura cryptography forks

p0mvn:
## Motivation

Zakura still resolved the upstream Zcash cryptography crates, allowing the original packages to coexist with the coordinated Zakura fork stack.

## Solution

- Replace all direct cryptography dependencies with the `1.0.0-rc.2` Zakura crates.io packages.
- Keep the full 17-crate fork closure on the coordinated release.
- Align `group` traits and use Rand 0.10 at APIs exposed by the fork crates while retaining Rand 0.8 where existing dependencies require it.
- Rename development profile package selectors so cryptography optimizations continue to apply.
- Document narrow Rand duplicate and Cargo Vet exemptions introduced by the migration.
- Update `h2` to 0.4.16, which fixes RUSTSEC-2026-0258.

## Testing

- `cargo check --workspace --all-targets --locked`
- `cargo test -p zakura-chain -p zakura-consensus -p zakura-rpc --lib` (715 passed, 1 ignored)
- `cargo deny` default, release-feature, all-feature, and advisory checks
- `cargo vet check --store-path ./qa/supply-chain/`
- Markdown lint for `docs/changelog/unreleased/749.md`

A full `cargo test --workspace` run reached the network-dependent acceptance suite; 43 acceptance tests passed, 40 were ignored, and `activate_mempool_mainnet` timed out waiting for live Mainnet synchronization.

## Changelog

- Added a user-facing `Changed` entry for crate consumers in `docs/changelog/unreleased/749.md`.

- https://github.com/zakura-core/libraries
- https://rustsec.org/advisories/RUSTSEC-2026-0258

### Follow-up Work

Replace the exact-version Cargo Vet exemptions with direct or delta audits as those reviews are completed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,27 +32,27 @@ edition = "2021"
 
 [workspace.dependencies]
 incrementalmerkletree = { version = "0.8.2", features = ["legacy-api"] }
-orchard = "0.15.4"
-sapling-crypto = "0.7"
+orchard = { version = "1.0.0-rc.2", package = "zakura-orchard" }
+sapling-crypto = { version = "1.0.0-rc.2", package = "zakura-sapling-crypto" }
 zcash_address = "0.13.0"
 zcash_encoding = "0.4"
 zcash_history = "0.5.0"
-zcash_keys = "0.16.0"
-zcash_primitives = "0.30.0"
-zcash_proofs = "0.30.0"
+zcash_keys = { version = "1.0.0-rc.2", package = "zakura-keys" }
+zcash_primitives = { version = "1.0.0-rc.2", package = "zakura-primitives" }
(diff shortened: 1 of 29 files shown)


Cargo.toml
```diff
@@ -112,9 +112,12 @@ prost = "0.14"
 rand = "0.8.5"
 rand_chacha = "0.3"
 rand_core = "0.6"
+rand_10 = { package = "rand", version = "0.10" }
```
ebfull, line 115: Okay, but I'd like to later change this to just `rand` and either remove or rename the ancient `rand` wherever we come into contact with it. (Lots of places I'm sure, this is another project of its own.)

ebfull (approved):

PR #750, opened 2026-08-20
feat(state): add historical frontier artifact format

p0mvn:
## Motivation

Historical treestate serving needs a standalone, bounded format for sparse note-commitment frontiers before generation, embedding, caching, or RPC behavior can be layered on top.

## Solution

- Define `FrontierEntry` and `FrontierArtifact` with deterministic encoding and decoding.
- Enforce record-count and payload-size allocation bounds, strictly increasing heights, complete framing, and non-panicking tree decoding.
- Bind the non-digest header fields and payload into the frontier digest, while consumers independently authenticate entry roots before use.
- Add anchor lookup and maximum cold replay-gap calculation, with minimal state-crate re-exports.

This PR intentionally excludes generators, embedded artifacts, configuration, service caches, RPC changes, and release plumbing.

## Testing

- `cargo test -p zakura-state --lib treestate_artifact::tests` — 22 passed
- `cargo clippy -p zakura-state --lib -- -D warnings` — passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed
- `./scripts/changelog.py check` — passed

The focused codec tests cover round trips, deterministic output, duplicate heights, malformed trees with valid frame digests, trailing bytes, payload and header tampering, truncation, allocation bounds, anchor selection, and replay-gap calculation.

Local Markdown lint could not run because this environment has neither `markdownlint` nor Node/npm; the four-line fragment follows the repository's explicit no-changelog format and will be covered by draft CI.

## Changelog

`docs/changelog/unreleased/750.md` explicitly excludes this internal-only format plumbing from operator-facing release notes.

### Specifications & References

The format follows the historical treestate design's independently root-checked frontier model.

### Follow-up Work

Generators, embedded binaries, configuration, service integration, RPC behavior, and release plumbing belong in later PRs.

PR #751, opened 2026-08-20
feat(state): generate historical frontier grids

p0mvn:
## Motivation

Historical treestate serving needs a generator for sparse note-commitment frontier grids before the grid can be published, embedded, or used at runtime. Part 1 (#750) defined the artifact format; this PR produces grids and ships them from the archive release-state publisher without consuming them on the serving path.

## Solution

- Add `GridSpacing`, `export_frontier_grid_to`, adaptive cost-weighted placement, resume-from-previous-grid, and prefix-extension behavior.
- Add only the dormant replay/storage helpers generation needs (`stored_frontiers_at`, `stored_frontier_before_absent_band`, `verify_against_available_roots`).
- Extend `zakura-checkpoints` with frontier-grid output, spacing/budget flags, resume input, and embedded-checkpoint backfill.
- Move generation onto the archive-node publisher (`zakura-release-state.timer`), which emits `mainnet-frontier-grid.bin` in the R2 bundle.
- Teach `fetch-release-state.py` to accept the grid as an **optional** digest-verified file so four-file publisher bundles do not break today's three-file importer.
- Add `zakurad verify-historical-treestates --frontier-grid-input` for offline framing/coupling checks.

This PR intentionally excludes embedding, config, service caches, RPC serving, provenance/import enforcement, and the Mainnet grid binary.

## Testing

- `cargo test -p zakura-state --lib grid_target_tests` — 4 passed (uniform/adaptive placement and tip-prefix compatibility)
- `cargo test -p zakura-state --lib historical_tree::tests::genesis_absent_band` — 1 passed
- `cargo test -p zakura-utils --features zakura-checkpoints-offline --bin zakura-checkpoints` — 13 passed
- `python3 .github/scripts/fetch-release-state.py --self-test` — 17 passed (optional grid accept/digest)
- `./scripts/changelog.py check` — passed
- Additional clippy/fmt/`bash -n` and generation property coverage are running locally against this draft

## Changelog

`docs/changelog/unreleased/751.md` explicitly excludes this internal-only generation plumbing from operator-facing release notes.

- Builds on #750 (`FrontierArtifact` format).
- Follows the historical treestate serving design's independently root-checked frontier model.
- Stack slice 2 of the #703 split.

### Follow-up Work

- Import/provenance/workflows that *require* the fourth file (part 3).
- Commit the Mainnet grid binary in a release-state-only PR (part 4).
- Runtime serving integration (part 5) and RPC coverage (part 6).

## AI Disclosure

Used Cursor to extract generation from the #703 stack onto current `main` after #750.

### DIFF
--- a/.github/scripts/fetch-release-state.py
+++ b/.github/scripts/fetch-release-state.py
@@ -38,6 +38,13 @@ FILE_LIMITS = {
     "mainnet-frontier.bin": 1 * 1024 * 1024,
     "mainnet-treestate-subtrees.bin": 8 * 1024 * 1024,
 }
+# Produced by the archive publisher once generation lands; still optional here so existing
+# three-file bundles keep fetching until the import follow-up requires the grid.
+OPTIONAL_FILE_LIMITS = {
+    # The cost-weighted grid measures ~3.3 MB on Mainnet. The cap leaves room for a
+    # uniform-spacing run, which is larger, without accepting an unbounded download.
+    "mainnet-frontier-grid.bin": 32 * 1024 * 1024,
+}
 LATEST_REQUIRED_KEYS = {
     "schema_version",
     "network",
@@ -234,9 +241,11 @@ def resolve_bundle(
         raise BundleError(f"release-state bundle is older than {max_age_hours} hours")
 
     files = _object(meta["files"], "meta.files")
-    _check_keys(files, set(FILE_LIMITS), set(), "meta.files")
+    _check_keys(files, set(FILE_LIMITS), set(OPTIONAL_FILE_LIMITS), "meta.files")
     validated: dict[str, dict[str, Any]] = {}
-    for name, max_size in FILE_LIMITS.items():
+    for name, max_size in {**FILE_LIMITS, **OPTIONAL_FILE_LIMITS}.items():
+        if name not in files:
+            continue
         entry = _object(files[name], f"meta.files.{name}")
         _check_keys(entry, {"size", "sha256"}, set(), f"meta.files.{name}")
         size = _integer(entry["size"], f"meta.files.{name}.size", maximum=max_size)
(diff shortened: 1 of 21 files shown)


evan-forbes (commented):
makes sense to me

ai was warning about using rockdb twice, which I think its given me before and can be ignored unless I'm missing that we'd actually be doing that for some reason here

```
1. High: the live RocksDB secondary can fail during primary compaction.
     /home/evan/.tmp/claude-1000/-home-evan-src-valar/d21eb58b-4585-4d7a-96d5-6534cb8c4356/scratchpad/wt751/crates/zakura-state/src/service/finalized_state/disk_db.rs:1525 sets a positive max_open_files. RocksDB 10.4.2 explicitly warns that secondary reads can return IOError:
     file not found when the primary deletes SST files. It recommends primary-secondary coordination or max_open_files=-1. The new service removes that coordination while allowing six-hour scans. This is the strongest blocking finding.

  2. Medium: the authoritative design command always fails.
     /home/evan/.tmp/claude-1000/-home-evan-src-valar/d21eb58b-4585-4d7a-96d5-6534cb8c4356/scratchpad/wt751/docs/design/verified-commitment-trees.md:833 omits --mainnet-frontier-grid-output. I ran the command and reproduced the artifact-set validation error. The section also
     retains false pruned-state and three-file-bundle claims.

  3. Medium: adaptive grids discard their placement policy.
     /home/evan/.tmp/claude-1000/-home-evan-src-valar/d21eb58b-4585-4d7a-96d5-6534cb8c4356/scratchpad/wt751/crates/zakura-state/src/service/finalized_state/treestate_export.rs:943 writes spacing: 0. The artifact records neither the budget nor a cost-model version. A later model
     change can create mixed-policy resumed grids or make fresh regeneration fail the append-only check without explaining the difference.

  4. Medium-low: absent-band generation deserializes transactions twice.
     /home/evan/.tmp/claude-1000/-home-evan-src-valar/d21eb58b-4585-4d7a-96d5-6534cb8c4356/scratchpad/wt751/crates/zakura-state/src/service/finalized_state/treestate_export.rs:855 loads every full block for placement. replay_with_subtrees then reads and deserializes the same
     transactions again. This adds material work to the multi-hour path.
```

evan-forbes (approved):

p0mvn:
On RocksDB, I haven't run into it in publishing, so I would rather not tweak.

Updated documentation.

Reviewed other Mediums and concluded they are okay

PR #754, opened 2026-08-21
feat(mining): submit prepared blocks by solved header

evan-forbes:
## Motivation

PR #748 moves prepared mined-block inventory to the state-admission boundary. A cached `submitblock` submission still sends and decodes the complete block. Peers also wait for contextual commit before they receive the admitted block body.

The near-maximum-size model assigns about 116 ms to client encoding, transport, and server decoding. The fleet observation measured a 78-99 ms origin reduction from compact submission. Immediate admitted-body service reduced origin service delay by another 75-84 ms. The 203 ms change in the complete at-risk window included other cross-leg changes, so it is not an isolated compact-submission measurement.

## Solution

This PR adds a `submitsolution` RPC that accepts a `workid` and the complete serialized solved header. Consensus reconstructs the block from the existing prepared-candidate cache. The existing `CommitMined` request still checks proof of work, header time, transaction lock times, duplicate status, the prepared-relay preflight, and contextual validity.

The cache stores one canonical candidate and multiple bounded `workid` aliases. It compares preserved header fields and decoded transactions directly. It accounts for the canonical block and derived state without retaining a second serialized block copy.

The shared mined-block lifecycle registers an authorized admitted block body before it sends inventory. Peer lookup serves that body immediately while contextual commit continues. Legacy-only and dual-P2P configurations keep the existing hash-inventory and pull flow. This PR does not add native P2P-v2 full-block push.

## Miner compatibility

An unchanged miner can continue to use `submitblock`. The node automatically reuses prepared semantic and proposal validation when the submitted block matches a cached template. That path retains full-block serialization, transport, and decoding.

A miner needs a change to avoid the full-block transfer. It must retain `workid`, submit the complete solved header through `submitsolution`, and fall back to `submitblock` for stale work or a candidate mismatch. A local adapter can translate the calls, but it only moves the full-block serialization and transfer to the adapter boundary unless it integrates directly with the miner.

Immediate admitted-body service does not require a mining software change.

## Testing

- `cargo test -p zakura-rpc --lib` (123 passed; 1 ignored)
- `cargo test -p zakura-consensus block::prepared::tests --lib` (8 passed)
- `cargo test -p zakura --lib components::inbound::tests::peer_block_lookup_` (2 passed)
- `cargo +1.98.0 clippy -p zakura-consensus -p zakura-state -p zakura-rpc -p zakura --all-targets -- -D warnings`
- `python3 scripts/changelog.py check`

The RPC lifecycle test holds contextual commit after state admission. It verifies that the early inventory event and pending body become available before commit completes.

## Changelog

Added `docs/changelog/unreleased/754.md` and recorded the changed limits in the parameter ledger.

## Stack

- Base and parent: #748.
- Children: none.
- This PR does not contain #773.

- BIP 22 and BIP 23 mining RPC conventions

### Follow-up Work

- Run a same-binary maximum-size A/B that isolates compact submission from immediate body service.
- Update mining clients in separate repositories if operators want native `submitsolution` support.


crates/zakura-consensus/src/block/prepared.rs
```diff
@@ -29,59 +34,145 @@ impl std::fmt::Debug for PreparedCandidateCache {
...
+            .entries
+            .iter()
+            .find(|entry| entry.id == entry_id)
+            .ok_or(ResolvePreparedCandidateError::StaleWork)?;
+
+        if !preserved_header_fields_match(&entry.prepared.block.header, &solved_header) {
+            return Err(ResolvePreparedCandidateError::CandidateMismatch);
+        }
+
+        Ok(Arc::new(Block {
```
p0mvn, line 136: `resolve()` packages the caller's `solved_header` into a fresh `Block` after only `preserved_header_fields_match` — it never checks difficulty or equihash. That is correct by design: PoW is re-checked downstream in the `CommitMined` fast path (`difficulty_is_valid` + `equihash_solution_is_valid`), which `submitsolution` now reaches with a fully attacker-controlled header.

That re-check has no test that drives the real `SemanticBlockVerifier` — `rpc_submitsolution_uses_prepared_candidate` mocks the router. Add a `block/tests.rs` case: prepare a candidate, then send `Request::CommitMined` with a solved header carrying an invalid equihash solution / wrong `difficulty_threshold`, and assert rejection. This is the forgery case that proves compact submission does not skip proof-of-work.

crates/zakura-rpc/src/methods.rs
```diff
@@ -2802,171 +3001,61 @@ where
...
-            // Checking the download queues and BlockVerifierRouter buffer for duplicates
-            // might require architectural changes to Zebra, so we should only do it
-            // if mining pools really need it.
-            Ok(_verify_chain_error) => SubmitBlockErrorResponse::Rejected,
-
-            // This match arm is currently unreachable, but if future changes add extra error types,
-            // we want to turn them into `Rejected`.
-            Err(_unknown_error_type) => SubmitBlockErrorResponse::Rejected,
+        let reconstruction_start = std::time::Instant::now();
+        let block = match self.gbt.prepared_candidates().resolve(&work_id, header) {
```
p0mvn, line 3421: The reconstructed block carries the miner's chosen `header.time`, and the only header-time-dependent consensus check is `lock_time_has_passed`, which the `CommitMined` fast path re-runs per transaction against that solved time. Candidates are prepared against the template's `cur_time`, so a solved time can differ from what the tx verifier saw at prepare.

No test exercises a time regression through this path. Add a case: cache a candidate containing a time-locked transaction, submit a solved header whose `time` is below that lock, and assert rejection. Without it, the "lock times are re-evaluated against the solved header" invariant the whole compact path rests on is unverified.

crates/zakura-rpc/src/methods.rs
```diff
@@ -2802,171 +3001,61 @@ where
...
+                metrics::counter!("mining.compact_submission.candidate_mismatches").increment(1);
+                return Ok(SubmitSolutionErrorResponse::CandidateMismatch.into());
+            }
         };
-
+        metrics::histogram!("mining.compact_submission.reconstruction.duration_seconds")
+            .record(reconstruction_start.elapsed().as_secs_f64());
+
+        let response = self
+            .run_mined_block_lifecycle(
```
p0mvn, line 3436: The fast path re-runs `time_is_valid_at(&block.header, Utc::now(), ...)` on the solved header before commit, so a stale or far-future solved time is meant to fail closed here rather than in the structural checks the fast path skips.

Add a `CommitMined` case with a solved header whose `time` is far in the future and assert `VerifyBlockError::Time`. Cheap to add alongside the lock-time case, and it pins the second of the two time-dependent rechecks.

crates/zakura-consensus/src/block/prepared.rs
```diff
@@ -29,59 +34,145 @@ impl std::fmt::Debug for PreparedCandidateCache {
...
+        let entry_id = inner
+            .work_ids
+            .get(work_id)
+            .map(|alias| alias.entry_id)
+            .ok_or(ResolvePreparedCandidateError::StaleWork)?;
+        let entry = inner
+            .entries
+            .iter()
+            .find(|entry| entry.id == entry_id)
+            .ok_or(ResolvePreparedCandidateError::StaleWork)?;
```
p0mvn, line 131: On a miss `resolve()` returns `StaleWork` and never reaches the verifier, but there is a TOCTOU: an entry can be evicted or TTL-expire between this `resolve()` and the `lookup()` re-check inside `CommitMined`. On that miss the fast path must fall through to full semantic verification of the reconstructed body.

Add a test that sends a `CommitMined` whose body is not in the cache (or was evicted after resolve) and confirms full verification still runs — e.g. a reconstructed body with a tampered transaction fails `merkle_root_validity`. This proves there is no reduced-check window when the cache misses mid-flight.

crates/zakura-rpc/src/methods/tests/vectors.rs
```diff
@@ -3445,6 +3445,160 @@ async fn rpc_submitblock_errors() {
...
+#[tokio::test(flavor = "multi_thread")]
+async fn rpc_submitsolution_uses_prepared_candidate() {
+    let _init_guard = zakura_test::init();
+
+    let candidate: Arc<Block> = zakura_test::vectors::BLOCK_MAINNET_GENESIS_BYTES
+        .zcash_deserialize_into()
+        .expect("the genesis test vector is valid");
+    let prepared_candidates = PreparedCandidateResolver::default();
+    prepared_candidates.insert_for_test(candidate.clone(), "prepared-work", &Mainnet);
+    let mut block_verifier_router: MockService<_, _, _, BoxError> =
```
p0mvn, line 3721: This test asserts the plumbing with a `MockService` router, so the reduced-check commit path is never compared against full verification. The fast path reuses the cached `transaction_hashes` / `new_outputs` / `deferred_pool_balance_change` and commits the submitted header — correct only while the cached body is byte-identical to what full verification would accept.

Add an equivalence test against the real verifier: commit the same solved block once via `Request::Commit` (full) and once via `Request::CommitMined` (fast), and assert identical committed hash and state effect. It locks the two paths together so a future change to either cannot silently diverge.

p0mvn (approved):
LGTM, pending the comments in the prior PR in the stack being addressed + tests hardened.

Also, left minor readability requests.

crates/zakura-consensus/src/block/prepared.rs
evan-forbes: Added both real-verifier cases. The updated PR 748 base drives `Prepare` → `CommitMined` with invalid Equihash and asserts `VerifyBlockError::Equihash`. This branch adds the wrong-`difficulty_threshold` case and asserts `BlockError::InvalidDifficulty`. These tests exercise `SemanticBlockVerifier` directly rather than the mocked RPC router.

crates/zakura-rpc/src/methods.rs
evan-forbes: Added this real-verifier regression in the updated PR 748 base. The test prepares a candidate whose transaction passes at the template time, then sends `CommitMined` with the solved time moved back to the lock boundary. It asserts `TransactionError::LockedUntilAfterBlockTime`, which pins the solved-header recheck.

crates/zakura-rpc/src/methods.rs
evan-forbes: Added this real-verifier regression in the updated PR 748 base. It prepares the candidate, moves the solved header three hours into the future, sends `CommitMined`, and asserts `VerifyBlockError::Time`.

crates/zakura-consensus/src/block/prepared.rs
evan-forbes: Added the eviction case. The test resolves a prepared candidate, evicts its cache entry with replacement candidates, and confirms that the work ID is stale. It then sends the already-resolved body with a transaction added but no matching Merkle-root update. `CommitMined` rejects it with `BadMerkleRoot`, which proves that the verifier falls back to full semantic verification when its cache lookup misses.

crates/zakura-rpc/src/methods/tests/vectors.rs
evan-forbes: Added the real-verifier equivalence test. It commits the same block through `Request::Commit` and through `Prepare` → `Request::CommitMined`. The mock state services capture the complete `SemanticallyVerifiedBlock`, and the test asserts that both the returned hash and the state input are identical.

crates/zakura-consensus/src/block/prepared.rs
```diff
@@ -66,54 +113,95 @@ pub(super) struct CachedPreparedCandidate {
...
+            .find_work_id(work_id)
+            .ok_or(ResolvePreparedCandidateError::StaleWork)?;
+        if !preserved_header_fields_match(&entry.prepared.block.header, &solved_header) {
+            return Err(ResolvePreparedCandidateError::CandidateMismatch);
+        }
+
+        Ok(Arc::new(Block {
+            header: Arc::new(solved_header),
+            transactions: entry.prepared.block.transactions.clone(),
+        }))
```
czarcas7ic: There may be an edge case here where the block we recreate above, the size isn't checked before it enters consensus verification. 

A cached proposal and its solved header does not need to have the same serialized header length, so the proposal's earlier size check doesn't establish that the final block fits. The subsequent verifier and contextual commit don't restore this check, allowing compact and full block submission to disagree on acceptance. 

I think we need to enforce `MAX_BLOCK_BYTES` on the reconstructed block, then we should add boundary regression coverage.

crates/zakura-rpc/src/methods.rs
```diff
@@ -743,6 +745,16 @@ pub trait Rpc {
         _parameters: Option<SubmitBlockParameters>,
     ) -> Result<SubmitBlockResponse>;
 
+    /// Submits a solved header for a prepared mining candidate.
+    ///
+    /// The miner must retry with `submitblock` when this method returns `stale-work` or
+    /// `candidate-mismatch`.
+    #[method(name = "submitsolution")]
```
czarcas7ic, line 753: With RPC enabled, the node now crashes on startup because `submitsolution` is missing from `RPC_METHOD_ACCESS`. We need to add the new method to the access table and rerun the RPC startup tests.

czarcas7ic: The feature was something I added after this PR was made so you just need to include it in the access list and it should pass.

crates/zakura-rpc/src/methods.rs
```diff
@@ -1211,6 +1268,158 @@ where
...
+                            ?block_hash,
+                            ?height,
+                            "mined block failed contextual commit after early inventory"
+                        );
+                    }
+                    MinedBlockEvent::Failed {
+                        hash: block_hash,
+                        height,
+                        early_advertised,
+                    }
```
czarcas7ic, line 1549: I think Fable found a DoS but after I looked deeper into it, its only a problem with RPC open so I would just classify as medium. Lets say a client keeps resubmitting a block we already have every few seconds. Each rejection sends this event and restarts the gossip loop’s seven second wait, so normal tip announcements keep getting pushed back as long as those requests continue.

PR #756, opened 2026-08-21
build(rpc): pre-generate RPC artifacts

p0mvn:
## Motivation

Normal RPC builds compile protobuf and OpenRPC generators even though their outputs can be checked in and reviewed.

## Solution

Move regeneration to `cargo xtask generate-rpc-artifacts`, include generated OpenRPC source directly, and add a non-mutating CI drift check.

## Testing

- `cargo check -p zakura-rpc -p xtask`
- `git diff --check`
- `markdownlint docs/changelog/unreleased/756.md --config .github/.markdownlint.yaml`

## Changelog

`docs/changelog/unreleased/756.md` marks this internal-only change as excluded.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -99,6 +99,9 @@ jobs:
       - name: Check locked all-feature build
         if: ${{ !cancelled() }}
         run: cargo check --locked --all-features --all-targets
+      - name: Check generated RPC artifacts
+        if: ${{ !cancelled() }}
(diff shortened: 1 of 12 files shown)


p0mvn:
Fixed the deterministic lint failure by correcting the stale Codex action version comment in `.github/workflows/upstream-sync.yml`. Verified with the same pinned auditor version used by CI:\n\n`uvx --from zizmor==1.29.0 zizmor --min-severity informational .github/workflows`

PR #762, opened 2026-08-21
build: remove stale dependencies and narrow features

p0mvn:
## Motivation

Workspace manifests retain unused declarations and enable test, utility, telemetry, and runtime features more broadly than their consumers require.

## Solution

Remove stale dependencies and scope optional features to the binaries, tests, and telemetry paths that use them.

## Testing

- `cargo check --workspace --all-features`
- `git diff --check`
- `markdownlint docs/changelog/unreleased/762.md --config .github/.markdownlint.yaml`

## Changelog

`docs/changelog/unreleased/762.md` marks this internal-only change as excluded.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4243,7 +4243,6 @@ dependencies = [
  "js-sys",
  "pin-project-lite",
  "thiserror 2.0.18",
- "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


p0mvn:
Restored published Cargo feature names and implied feature edges as no-op/optional compatibility aliases. Verified with:\n\n- `cargo semver-checks --package zakura-node-services --default-features`\n- `cargo semver-checks --package zakura-utils --default-features`\n- `cargo check --locked --workspace --all-features`

PR #763, opened 2026-08-21
refactor(state): publish the Mainnet frontier grid instead of committing it

p0mvn:
Stacked on #703. Base is `feat/vct-treestate-seam-rebased`, not `main`.

Rebased onto #703 after that branch was rebased onto `main` at `f892b907` (which now
carries #775). The pinned version moved with it: #768 regenerated the grid at checkpoint
3,453,771, so the crate to publish is `0.3453771.0`, not `0.3449371.0`.

## Motivation

#703 embeds `mainnet-frontier-grid.bin` — 2,167,503 bytes — and the release-state pipeline
regenerates it on every weekly refresh. Each refresh therefore writes a new multi-megabyte blob
into git history, permanently. `scripts/check-history-growth.sh` exists precisely because of this
failure mode; its own comment records that it was "Added after a 68 MB embedded artifact had to be
removed from history with a force push", and `history-growth.yml` already had to raise the 2 MiB
PR budget to 8 MiB to let the grid through.

The repository solved this once before. `zakura-vct-sprout-history` published a 71.7 MB artifact
as exact-pinned crates.io packages, compiled in through a facade crate. That repository's README
now points at the reserved `zakura-assets` crate for future crates.io-hosted binary data, and that
crate is registered with `repository = github.com/zakura-core/zakura` — this repository, rather
than a separate one.

The grid has not merged to `main` yet: it exists only on #703's branch. Landing this before #703
merges keeps those bytes out of `main`'s history entirely. Afterwards, only a history rewrite
would.

## Solution

Publish the grid as an exact-pinned crates.io package instead of committing it.

**Nothing about node behaviour changes.** The same bytes are compiled in, `FrontierArtifact::decode`
still parses them, every entry is still root-checked against the node's own authenticated
commitment root before it anchors anything, and a stale grid still degrades to unavailable rather
than failing closed. The single production change is `treestate_artifact.rs:77`, from
`include_bytes!` to `zakura_assets::MAINNET_FRONTIER_GRID`.

**The payload is in no git repository.** `assets/zakura-assets/` is excluded from the workspace and
commits only scaffolding — manifest, `lib.rs`, README, `.gitignore`. The payload and its generated
identity constants are gitignored and materialised by `scripts/pack-assets-crate.py` from a
verified release-state bundle immediately before `cargo publish`. An explicit `include` list in the
package manifest is what ships them, because cargo otherwise derives its file list from git. This
is strictly better than the sprout-history layout, which did commit its 8 MiB `part.bin` files and
merely moved the weight to a repository nobody clones.

**The version is the coupling.** `0.<last_checkpoint>.<revision>`, derived from the payload rather
than passed in, so the pin a consumer reads states which checkpoint it covers. Six things hold it
against the release state:

- `check-release-state.sh` requires the pin to be exact, to name the manifest's `finalized_height`,
  and to resolve from crates.io with a checksum in `Cargo.lock`. A path or git override, a loose
  requirement, a pin bump without a manifest bump, or a grid committed by mistake all fail the
  lint job — still without cargo.
- `embedded_mainnet_final_frontiers_parse` hashes the bytes the dependency actually supplies
  against the manifest's digest, size, and entry count, and holds the payload's declared checkpoint
  against the embedded checkpoint list. It checks the crate's own constants too, so a crate that
  misdeclares its payload cannot pass by agreeing with itself. This is **stronger than what it
  replaces**: the grid was previously bound only by the lint script.
- `import-release-state.py` proves each new grid is a byte prefix-extension of the one the
  repository currently pins, materialised from the registry rather than the tree.
- `Cargo.lock` carries the registry checksum; every release build already passes `--locked`.
- The crate self-tests its declared identity against its own payload.
- None of the above is what makes a wrong grid safe. Every entry is root-checked on every node,
  forever.

**Publish then pin.** Cargo can only resolve a version that already exists, so
`update-release-state.yml` splits into `resolve` → `publish-assets` → `open-pr`: the grid is
published before the PR that pins it is opened. A candidate a reviewer rejects is a yanked,
unreferenced version. Only the publish job holds an OIDC token, and it runs in a `crates-io-assets`
environment carrying no other secrets, so the job that uploads to the registry never sees the App
key that opens pull requests, and vice versa — the same split `publish-crates.yml` documents. The
publish is idempotent and refuses outright if the version exists with different bytes.

`fetch-release-state.py` gains a digest-pinned `--meta-url` mode resolving one specific immutable
bundle rather than following `latest.json`. The publish job uses it to derive the bytes it uploads
independently of the run that validated them; it is also the entry point a third party needs, where
fetching an old bundle is the point and the pointer path's freshness rule must not apply.

The history-growth exemption drops from 8 MiB to 3 MiB, since a refresh is now an appended
checkpoint list plus small artifacts. That is what makes an accidentally committed grid fail.

## Testing

| Check | Result |
| --- | --- |
| `cargo test -p zakura-rpc --test historical_treestate` | **4 passed** — serving restored across the absent band from crate-supplied bytes; both negative controls still refuse |
| asset crate `cargo test` | 2 passed, against the real 2.1 MB payload |
| `cargo package` | 8 files, 2.1 MiB, **1.2 MiB compressed**; payload digest inside the `.crate` equals the manifest's `frontier_grid_sha256` |
| `cargo package --list` | ships the gitignored payload via the explicit `include`; `--allow-dirty` confirmed necessary, not defensive |
| `scripts/pack-assets-crate.py --self-test` | 14 passed |
| `fetch-release-state.py --self-test` | 22 passed (16 existing + 6 new for the pinned path) |
| `import-release-state.py --self-test` | 16 passed |
| drift guard | 10/10 cases against a synthetic self-consistent release state |
| `cargo clippy -p zakura-state --all-targets -- -D warnings` | clean |
| `cargo fmt` (workspace and asset crate) | clean |
| `markdown-lint` | passes after `91285ee`; `markdownlint-cli` is not installable locally, so the first push tripped MD014 on prompt-prefixed code blocks and CI caught it |

The drift guard fixture caught a real bug in the first draft of this PR: the lockfile regex ran
past a package block's blank line, found a later crate's `source` and `checksum`, and would have
passed on anything — including the path-override lockfile it exists to reject. The regression case
is in the matrix.

**One thing cannot pass yet, for a reason outside this PR.**

`Cargo.lock` is deliberately unchanged, so every cargo job on this PR fails with
`error: no matching package named 'valargroup-zakura-assets' found` until
`valargroup-zakura-assets 0.3453771.0` is published — the registry checksum only exists after
publication, and `scripts/check-release-state.sh` refuses a path or git override precisely so the
pin cannot be satisfied by anything but a real registry entry. Everything above was verified with
`--config 'patch.crates-io.valargroup-zakura-assets.path="assets/zakura-assets"'`, which touches no
committed file.

## Changelog

Added as `docs/changelog/unreleased/763.md`. This PR deliberately does not touch `703.md`: `changelog.py check-pr` enforces one fragment per PR, which
is what `fix(ci): keep #703 to its own changelog fragment` was fixing earlier on that branch.

- Stacked on #703. Follows the pattern of the retired
  [`zakura-vct-sprout-history`](https://github.com/zakura-core/zakura-vct-sprout-history), whose
  README directs future crates.io-hosted binary data to `zakura-assets`.
- `docs/design/historical-treestate-serving.md` §5 (distribution), §5.1 (the checks that hold the
  pin and the checkpoint together), §5.2 (four-level third-party verification, including
  byte-identical regeneration from an archive node).

### Follow-up Work

Three steps this PR cannot take, in order:

1. Publish `valargroup-zakura-assets 0.3453771.0` from a verified bundle, then
   `cargo update -p valargroup-zakura-assets --precise 0.3453771.0` and commit the lockfile.
2. Create the `crates-io-assets` GitHub environment — branch policy `main`, no required reviewers —
   and the crates.io Trusted Publisher entry naming this repository, `update-release-state.yml`,
   and that environment. Note this would be the organisation's **first** OIDC publish: all 69
   published versions across `zakura`, `zakura-state`, `zakura-chain`, and `zakura-rpc` show
   `published_by: ebfull` with `trustpub_data: null`, because `publish-crates.yml` has not been
   used for a release yet.
3. Retarget from the temporary `valargroup-zakura-assets` to the reserved `zakura-assets`, which
   `ebfull` solely owns. The dependency is renamed, so that is a one-line change in the workspace
   manifest plus the lockfile. It needs `cargo owner --add p0mvn zakura-assets` (only an owner can
   yank, which is this design's remedy for a rejected candidate) and a Trusted Publisher entry on
   the real crate. Then yank the temporary versions.

`qa/supply-chain/config.toml` carries an exemption rather than the intended `[[trusted]]`
publisher record: `cargo vet trust` derives the publisher identity from crates.io and cannot run
before the first publish. Replacing it is part of step 1.

## AI Disclosure

Used Claude Code to design and implement this change and to draft this description.


p0mvn:
Consolidated into #703 and closing here.

#703 now carries this PR's content in full: the `crates/zakura-assets/` packaging crate,
`scripts/pack-assets-crate.py`, the exact pin on `valargroup-zakura-assets 0.3453771.0`, the
pinned-bundle fetch mode, the rewritten `check-release-state.sh`, and the publish step in
`update-release-state.yml`.

One thing changed in the consolidation, deliberately. Landing these as a stack meant #703 added
the 2.1 MB grid and this PR deleted it, so the blob would have entered git history permanently
anyway — the exact outcome this PR exists to prevent. Consolidated, the grid is never committed
at any point in the history: measured growth is **159 KiB**, against ~1.14 MiB for the two-PR
stack.

The changelog fragments were merged into `703.md`, since `changelog.py check-pr` enforces one
fragment per PR.

PR #760, opened 2026-08-21
refactor: use standard library dependency replacements

p0mvn:
## Motivation

Several direct dependencies only provide functionality now available in the standard library or the existing tracing crate.

## Solution

Use inherent `div_ceil`, const assertions, and `tracing::Instrument`, then remove the redundant declarations.

## Testing

- `cargo check -p zakura-chain -p zakura-consensus -p zakura-network -p zakura`
- `cargo fmt --all -- --check`
- `git diff --check`
- `./scripts/changelog.py check`
- `markdownlint docs/changelog/unreleased/760.md --config .github/.markdownlint.yaml`

## Changelog

`docs/changelog/unreleased/760.md` marks this internal-only change as excluded.

PR #764, opened 2026-08-21
build: reduce dependency surface

p0mvn:
## Motivation

Zakura's production and workspace builds compile several unused, duplicate, single-purpose, and build-only dependency subgraphs. This increases clean-build time and the amount of third-party code participating in the supply chain.

## Solution

- Pre-generate protobuf and OpenRPC artifacts and verify drift in CI.
- Replace `vergen-gitcl` with focused build metadata helpers.
- Standardize RPC TLS on Ring and reuse Ureq in the watchdog.
- Replace `mset`, `human_bytes`, `num-integer`, `static_assertions`, and direct `tracing-futures` uses where small local or standard APIs suffice.
- Remove stale declarations and narrow test, utility, telemetry, and runtime features while preserving published Cargo feature compatibility.

The same work remains available as independently reviewable drafts: [#755](https://github.com/zakura-core/zakura/pull/755), [#756](https://github.com/zakura-core/zakura/pull/756), [#757](https://github.com/zakura-core/zakura/pull/757), [#758](https://github.com/zakura-core/zakura/pull/758), [#759](https://github.com/zakura-core/zakura/pull/759), [#760](https://github.com/zakura-core/zakura/pull/760), [#761](https://github.com/zakura-core/zakura/pull/761), and [#762](https://github.com/zakura-core/zakura/pull/762).

## Testing

- `cargo check --locked --workspace --all-targets --all-features`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- RPC library tests: 113 passed, 1 ignored
- Watchdog tests: 18 passed
- Focused consensus, counted-set, and byte-format tests
- `cargo test --locked -p zakura-utils --all-features`
- Semver checks for `zakura-node-services` and `zakura-utils`
- `cargo xtask check-rpc-artifacts`
- Zizmor 1.29.0 workflow audit
- Changelog and Markdown validation

## Changelog

`docs/changelog/unreleased/764.md` marks this internal-only change as excluded.

### DIFF
--- a/crates/zakura-consensus/src/block/check.rs
+++ b/crates/zakura-consensus/src/block/check.rs
@@ -4,7 +4,6 @@ use std::{collections::HashSet, sync::Arc};
 
 use chrono::{DateTime, Utc};
 
-use mset::MultiSet;
 use zakura_chain::{
     amount::{
         Amount, DeferredPoolBalanceChange, Error as AmountError, NegativeAllowed, NonNegative,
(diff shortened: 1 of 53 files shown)


evan-forbes (approved):

crates/zakura-consensus/src/block/check.rs
```diff
@@ -25,6 +24,32 @@ use zakura_header_chain::{CompactTargetError, PowPolicy};
...
+    /// Removes one matching output, preserving duplicate output multiplicity.
+    fn remove(&mut self, expected: &Output) -> bool {
+        let Some(index) = self.outputs.iter().position(|output| output == expected) else {
+            return false;
+        };
+
+        self.outputs.remove(index);
+        true
+    }
+}
```
ValarDragon, line 51: If we need the remove fn, we should be doing this via a hashmap or linked list

ValarDragon: Ah remove is only called in tests it seems

ValarDragon: I'm worried about this change, need to investigate more. Just because the multiset must have joined on something?

EDIT: Oh wait no, its a multi-set so we kept duplicity. So this remove is a regression

ValarDragon: Update: its all safe. Multiset delete only deletes 1 copy

ValarDragon: so were safe

PR #773, opened 2026-08-21
feat(consensus): relay expected-work blocks before contextual commit

evan-forbes:
## Summary

- Set `network.block_relay = "optimistic"` as the default and retain strict `"committed"` behavior.
- Require full semantic verification before optimistic relay.
- Require the header to bind the exact body, authorizing data, and authenticated parent history.
- Require sequential height, parent-derived time, the exact expected target, and proof of work at that target.
- Require the parent to remain the selected tip when state authorizes relay.
- Keep syncing nodes, checkpoint blocks, side chains, and PoW-waived networks on the commit-first path.
- Admit the pending body before inventory and serve it while contextual validation, storage, and publication continue.
- Apply the relay boundary to both P2P stacks while native block sync owns Zakura bodies.
- Suppress supplier scoring for every failure discovered after optimistic eligibility, including strict receivers.
- Retain ordinary scoring for malformed, semantically invalid, commitment-invalid, and expected-work-invalid blocks.
- Add a bounded one-hour rejection cache with replay and descendant suppression.
- Preserve absolute pending-body expiry and commit-first fallback when lifecycle bounds fill.
- Keep local transition, storage, shutdown, and supersession failures unscored and outside the rejection cache.
- Add queued mined-block RPC acknowledgement with a strict default.
- Fail-stop the state writer after storage-uncertain header transition failures.

## Tests

- `cargo test --workspace -- --skip inbound_pruned_block_is_not_advertised_and_getdata_logs_error --skip pruned_storage_mode_prunes_during_regtest_sync`
- `cargo test -p zakura --test acceptance nu6_funding_streams_and_coinbase_balance -- --exact --show-output`
- `cargo test -p zakura --test acceptance pruned_storage_mode_prunes_during_regtest_sync -- --exact --show-output`
- Focused consensus, state, RPC, relay lifecycle, inbound, native block-sync, pending registry, scoring, cancellation, fork, and easy-target regression tests.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`

The workspace run excluded the known unrelated `inbound_pruned_block_is_not_advertised_and_getdata_logs_error` failure. The pruning acceptance test also uses shared log matching, so the workspace run excluded it after a concurrent failure. That test passes alone.

## Validation scope

- The existing fleet result validates commit `17db6d8e4` only.
- We did not rerun the fleet for commit `6564136e8`.

Stacked on #754.


evan-forbes:
this idea is not bad! we just need the entire network to do this in an effort to be overly cautious 

There is a possible yet unlikely network partition between zebra and zakura here, where we verify all the important light weight things like PoW proofs/signatures invalid messages or hashes, but purposely pipeline the heavyweight "contextual/stateful" things. So

Double spends
Unknown anchors
Missing notes


This means a malicious miner could waste work and propagate an invalid block. That's nbd as it's very costly, however we still can't disconnect from peers as we can't determine if it was them being malicious or just still executing the rest of the checks.

Zebra still does disconnect however.

So while very limited, as it can only effect peers that gossip the **_block before the full set of checks finish_**, it still could disconnect many zebra nodes from zakura.

Eventually it would heal and have no noticeable immediate effect, but feels like we're introducing new attack surface area. Like, an attacker would combine this with some other attack like in the last LTC attack.

Therefore, we've decided to delay this approach until we can get the entire network to do it

Issue #628, opened 2026-08-12
bug: indexer mTLS certificate lifecycle — resumption, validity dates, and no reload


maxdesalle:
Three certificate-lifecycle behaviours in #596's indexer mTLS, found while testing that PR.
All reproduced against `90284ad8f`. The PR's core is fine — I confirmed mTLS is genuinely
mandatory: no client certificate gets `CertificateRequired`, a certificate from another CA
gets `UnknownCA`, and only the trusted client reaches request validation. These are about
what happens *after* the three PEM files are read.

Reproduced with a binary calling `zakura_rpc::indexer::server::init(..)` directly — the same
call `StartCmd` makes (`crates/zakurad/src/commands/start.rs:694-701`) — with throwaway
certificates; clients are `tonic` and `openssl s_client 3.6.3`.

## 1. A resumed TLS 1.3 session carries the client's certificate forward without re-validating it

Stock rustls behaviour, not a bug in this PR, but it bounds how quickly any revocation can
take effect. On a resumed handshake rustls restores the stored chain verbatim
(`rustls-0.23.41 src/server/tls13.rs:333-337`) and emits `CertificateRequest` only on the
full-handshake path (`tls13.rs:383-384`). The defaults are an in-process
`ServerSessionMemoryCache::new(256)` (`src/server/builder.rs:113`) and a 24 h stateful ticket
lifetime (`tls13.rs:1290`).

With a client certificate valid for two minutes, against one server process:

```
F1  17:06:00  cert valid    full handshake, cert presented  -> accepted, server HTTP/2 SETTINGS
F2  17:08:39  cert EXPIRED  full handshake, same cert       -> alert certificate expired (45)
F3  17:08:44  cert EXPIRED  resumption of F1, NO cert sent  -> Reused, TLSv1.3, server HTTP/2 SETTINGS
```

F2 and F3 are five seconds apart on the same listener. The identity the server rejects on a
fresh handshake is accepted on a resumed one, and the request path is open — the server's
opening bytes on the resumed-without-certificate connection are byte-identical to a fully
authenticated one.

Two bounds, so this is not overstated. It is **not** a remote bypass: resuming needs the
ticket and PSK of a session that was already fully authenticated, and a ticket minted by a
previous process gets `New,` then `certificate required` (verified). And a restart clears the
cache.

The decision is whether an indexer listener should offer resumption at all: for a small
number of long-lived clients it buys close to nothing and costs a ticket-lifetime revocation
[...]

PR #774, opened 2026-08-22
fix(rpc): warn when an indexer TLS certificate is outside its validity window

jonwalch:
## Motivation

Finding 2 in [#628](https://github.com/zakura-core/zakura/issues/628), reported by
@maxdesalle, shows that an indexer mTLS listener accepts an expired or not-yet-valid
server certificate, logs that the endpoint opened, and then serves no clients because
they reject the handshake.

The root cause is that `load_mtls_config` reads the server certificate as opaque PEM and
passes it to tonic. Rustls correctly does not validate a server's own certificate dates,
so the indexer path never emits an operator-visible diagnostic. Merged PR
[#627](https://github.com/zakura-core/zakura/pull/627) added the equivalent diagnostic to
the existing JSON-RPC TLS path; this PR completes the indexer half.

## Solution

Reuse the existing certificate-validity classification and warnings through a
crate-private, best-effort PEM helper, and call it after the indexer TLS files are read.
Expired and not-yet-valid certificates warn with the configured path, chain position,
relevant certificate timestamp, and current time.

The check remains diagnostic: it does not stop node or listener startup, and tonic stays
responsible for rejecting malformed TLS input.

The behavioral test owns a current-thread Tokio runtime and a thread-local subscriber
with a fixed warning filter. It captures warnings in a private buffer, making the test
independent of the caller's `RUST_LOG` without changing the shared test logging setup.

## Testing

The new behavioral test uses stable expired and future-dated server identities. Before
the production fix, both listeners started but the test reported both warnings missing.
After the fix, both listeners still start and each expected warning is captured.

The focused test passes with both `RUST_LOG=off` and `RUST_LOG=info`. The existing
validity tests continue to cover inclusive boundaries, UTC and Generalized Time,
pre-Unix-epoch dates, and malformed DER. The full RPC library suite reports 114 passed
and one intentionally ignored test.

```text
cargo fmt --all -- --check
cargo clippy -p zakura-rpc --all-targets --all-features -- -D warnings
RUST_LOG=off cargo test -p zakura-rpc --lib indexer::tests::vectors::indexer_server_warns_for_server_certificates_outside_their_validity_window --no-fail-fast
RUST_LOG=info cargo test -p zakura-rpc --lib indexer::tests::vectors::indexer_server_warns_for_server_certificates_outside_their_validity_window --no-fail-fast
cargo test -p zakura-rpc --lib --no-fail-fast
./scripts/changelog.py check
npx --yes markdownlint-cli@0.44.0 docs/changelog/unreleased/774.md --config .github/.markdownlint.yaml
```

## Changelog

Added the user-visible `Fixed` fragment in `docs/changelog/unreleased/774.md`.

- Addresses finding 2 in [#628](https://github.com/zakura-core/zakura/issues/628).
- Completes the indexer counterpart to merged PR
  [#627](https://github.com/zakura-core/zakura/pull/627).

### Follow-up Work

Findings 1 (TLS session resumption) and 3 (certificate reload) in #628 remain out of
scope, so this PR does not close the issue.


evan-forbes (changes requested):
Thanks for adding diagnostics for expired and soon-to-expire indexer TLS certificates. I reviewed the production path, and it preserves the existing TLS behavior. It passes the original certificate, private key, and client CA to tonic after it emits the diagnostic.

The new test currently depends on the caller's `RUST_LOG` value. I reproduced the problem with the focused test: it fails with `RUST_LOG=off` because `LogCapture` receives neither warning, and it passes with `RUST_LOG=info`. The process-wide `EnvFilter` filters the events before the capture layer receives them.

Please make this test independent of the external log filter. If the project intends to require a specific filter for this test, please enforce that requirement in the test setup. Please also document the requirement. I do not see a production TLS defect in this change.

jonwalch:
Addressed in 07c30952b.

The root cause was the process-wide `EnvFilter`: with `RUST_LOG=off`, it discarded the
warning events before `LogCaptureLayer` could observe them. The test cannot safely set
`RUST_LOG` itself because `zakura_test::init` is process-wide and first-call-wins.

I scoped the environment filter to formatted output and tracing error context, then gave
`LogCaptureLayer` a fixed `WARN` filter. Its documentation now states that capture always
includes warnings and errors, independently of `RUST_LOG`, and excludes lower levels.
There are no further production TLS changes.

I verified the focused behavioral test with both `RUST_LOG=off` and `RUST_LOG=info`, and
the full `zakura-test` and `zakura-rpc` library suites plus clippy pass.

jonwalch:
Updated in 574614d17, superseding the shared logging change in 07c30952b.

The root cause was the process-wide `EnvFilter`: with `RUST_LOG=off`, it discarded the
warning events before `LogCaptureLayer` could observe them. Setting `RUST_LOG` inside the
existing test would still race with process-wide, first-call-wins initialization.

The behavioral test now owns a current-thread Tokio runtime and runs under a thread-local
subscriber with a fixed `WARN` filter, capturing its output in a private buffer. The
current-thread requirement is documented beside the setup, and the shared
`zakura_test::init` and `LogCapture` implementations are unchanged. There are no further
production TLS changes.

I verified the focused behavioral test with both `RUST_LOG=off` and `RUST_LOG=info`, and
the full `zakura-rpc` library suite plus clippy passes.

evan-forbes:

PR #776, opened 2026-08-22
perf(consensus): accelerate block verification

evan-forbes:
## Motivation

An 80-node field profile identified transaction parser construction and the
Orchard note commitment tree as two local block-verification bottlenecks.
This change reduces both costs without changing the relay boundary.

## Solution

- Update the coordinated Zakura crypto packages from `1.0.0-rc.2` to
  `1.0.0-rc.3`.
- Enable the weighted Orchard Merkle evaluator.
- Preserve Zakura's parallel frontier while routing Orchard and Ironwood node
  hashes through the library implementation.
- Build each independent `CachedFfiTransaction` on the existing Rayon pool.
- Add warm proof-cache transaction and Orchard tree benchmarks.
- Add a scheduled differential fuzzer for the weighted Orchard Merkle hash.
- Add `safe-to-deploy` delta audits for 15 updated crypto packages.
- Retain exact exemptions for the Pasta curves and Orchard updates until their
  secret-bearing variable-time paths have complete timing-safety arguments.

The weighted evaluator relies on the library's documented discrete-logarithm
relation assumption. Fixed and randomized differential tests compare it with
the direct Sinsemilla reference at every tree level. The scheduled fuzzer
extends that comparison across full-width field elements. It cannot test the
discrete-logarithm assumption that makes exceptional additions infeasible.

## Results

| Benchmark | Baseline | Result | Reduction |
| --- | ---: | ---: | ---: |
| Cold Orchard transaction workload | 139.673 ms | 131.743 ms | 5.68% |
| Warm Orchard transaction workload | 10.476 ms | 3.078 ms | 70.62% |
| Orchard 436-action append and root | 2.917 ms | 1.307 ms | 55.19% |

These component measurements overlap. They do not establish an end-to-end
block-verification saving.

The transaction benchmark uses a 1,513,713-byte workload with 165 transactions
and 330 Orchard actions. The warm mode preserves the production proof caches
between iterations. The tree benchmark preloads 10,000 leaves and appends 436
actions. Both benchmarks run with eight Rayon workers.

## Supply-chain review

The audit replaces 15 exact exemptions with reviewed delta audits. It retains
exact exemptions for `zakura-pasta-curves` and `zakura-orchard`. The reviewed
updates add secret-bearing variable-time inversion and batched key-agreement
paths. Functional and arithmetic tests do not establish timing safety for
those paths. Zakurad does not perform trial decryption or hold incoming viewing
keys.

Cargo Vet reports 679 fully audited packages, 31 partially audited packages,
and 51 exempted packages.

## Testing

- `PROPTEST_CASES=100000 cargo test -p zakura-chain library_merkle_crh_matches_reference_random --lib --locked`
- `cargo +nightly-2026-07-15 fuzz run --fuzz-dir . orchard_merkle_equivalence -- -runs=100000 -max_len=65 -rss_limit_mb=4096 -detect_leaks=1`
  - 100,000 differential inputs passed.
  - The campaign reported no mismatch, crash, or leak.
- `cargo +nightly-2026-07-15 fuzz check --fuzz-dir . orchard_merkle_equivalence`
- `cargo bench -p zakura-consensus --bench worst_case_tx_verification -- full_orchard_limit --noplot`
- `ZAKURA_BENCH_WARM_CACHES=1 cargo bench -p zakura-consensus --bench worst_case_tx_verification -- full_orchard_limit --noplot`
- `cargo bench -p zakura-chain --bench orchard_tree_update`
- `cargo test -p zakura-chain -p zakura-consensus --lib --locked`
  - 398 chain tests passed.
  - 208 consensus tests passed.
- `cargo check -p zakura-chain -p zakura-consensus --all-targets --locked`
- `cargo clippy -p zakura-chain -p zakura-consensus -p zakura-state -p zakura --all-targets -- -D warnings`
- `cargo vet check --locked --store-path ./qa/supply-chain/`
- `actionlint .github/workflows/orchard-merkle-fuzz.yml`
- `git diff --check`

## Changelog

This PR adds `docs/changelog/unreleased/776.md`.

## Stack

- Base: `main`.
- Parent: none.
- Children: none.
- This PR is independent of #748, #754, and #773.

PR #749 introduced the coordinated Zakura crypto packages and their temporary
exact-version Cargo Vet exemptions.


p0mvn (commented):
On merge-blocking issue from the adversarial pass. The reachable Orchard/Ironwood hash still matches published vectors; these are about failure classification and supply-chain coverage of the hash change itself.

crates/zakura-consensus/src/transaction.rs
```diff
@@ -582,8 +582,19 @@ where
...
+            let cached_ffi_transaction = Arc::new(
+                primitives::spawn_fifo(move || {
+                    CachedFfiTransaction::new(ffi_tx, Arc::new(spent_outputs), nu)
+                })
+                .await
+                .map_err(|_| {
+                    TransactionError::Other(
+                        "transaction parser worker dropped its response".to_string(),
+                    )
+                })?
```
p0mvn: **`RecvError` is retryable, so a parser-worker panic can stall the block.**

`spawn_fifo` only surfaces `RecvError` when the oneshot sender is dropped without sending. That is the same signal for process shutdown *and* for a panic inside `CachedFfiTransaction::new`.

This maps that signal to `TransactionError::Other`, which `body_verification_class` treats as `Retryable(VerifierUnavailable)` and `mempool_misbehavior_score` scores as 0. Block sync will retry the same body. If the panic is deterministic on one transaction, the retry hits it again and the height never becomes consensus-invalid.

Catch the panic inside the Rayon closure and send an explicit fatal error. Leave `RecvError` for shutdown only. Do not reuse shutdown’s channel-drop as “verifier unavailable.”

crates/zakura-consensus/src/transaction.rs
```diff
@@ -582,8 +582,19 @@ where
             }
 
             let nu = req.upgrade(&network);
-            let cached_ffi_transaction =
-                Arc::new(CachedFfiTransaction::new(tx.clone(), Arc::new(spent_outputs), nu).map_err(|_| TransactionError::UnsupportedByNetworkUpgrade(tx.version(), nu))?);
+            let ffi_tx = tx.clone();
+            let cached_ffi_transaction = Arc::new(
+                primitives::spawn_fifo(move || {
+                    CachedFfiTransaction::new(ffi_tx, Arc::new(spent_outputs), nu)
+                })
```
p0mvn, line 614: **Parser construction now shares the Groth16/Halo2 Rayon FIFO pool.**

`spawn_fifo` uses the process-wide Rayon pool that already runs Groth16, Halo2, RedJubjub, and RedPallas. `CachedFfiTransaction::new` is sighash setup (librustzcash conversion + ZIP-244 precompute), not a proof. FIFO means a wave of per-tx parser jobs from a many-transaction block can occupy workers ahead of batch proofs.

That is a scheduling change on the cold path this PR is trying to speed up. Put this on `tokio::task::spawn_blocking` or a dedicated pool; keep the crypto FIFO queue for proofs and signatures.

crates/zakura-consensus/src/transaction.rs
evan-forbes: Fixed. Parser construction now returns an explicit `TransactionParserPanicked` error when the blocking task panics. The body classifier maps it to the non-retryable `transaction.parser_panic` rule, while `mempool_misbehavior_score` remains zero.

A canceled task keeps the existing `transaction parser worker dropped its response` error. That shutdown path remains retryable and keeps a zero peer score. Constructor errors still map to `UnsupportedByNetworkUpgrade`.

The new tests induce a deterministic panic and pin both classifications and peer scores.

crates/zakura-consensus/src/transaction.rs
evan-forbes: Fixed. `CachedFfiTransaction::new` now runs on Tokio's blocking pool, so parser construction no longer occupies the crypto Rayon FIFO.

I added a scheduling test that inspects the parser worker and confirms that it is not a Rayon worker. I also made the benchmark's Tokio blocking-pool limit explicit and included it in the workload metadata. With eight Rayon, Tokio worker, and Tokio blocking threads, the cold Orchard workload measured 129.97–130.43 ms.

Issue #681, opened 2026-08-15
state: track indexer schema separately from database format version


ValarDragon:
## Context

The indexer feature currently uses the SemVer build metadata of the state database format (`+indexer`) to indicate that indexer-only tables are present. This marker is fragile: numeric format upgrades can overwrite it, and shutdown/maintenance compatibility checks can accidentally compare build metadata as if it were part of the database schema version.

PR #623 applies the immediate safety fixes by preserving the marker during non-indexer upgrades and using precedence-aware comparison where appropriate.

## Follow-up

Design and implement an explicit, independently persisted indexer schema/version state instead of encoding indexer presence in the database format version build metadata.

The design should specify:

- where the indexer schema version is stored;
- how absent, current, and future indexer schemas are represented;
- crash recovery when index creation, migration, or removal is interrupted;
- whether a non-indexer binary retains, ignores, or removes indexer data;
- which combinations of indexer/default binaries are allowed to run rollback and prune;
- how an indexer binary bootstraps indexes for a database created without the indexer feature;
- regression tests for all supported feature combinations and maintenance operations.

## Related

- PR #623
- v12 findings F-214402, F-214403, and F-214404

PR #778, opened 2026-08-22
fix(state): propagate transparent index removal errors

### LINKED ISSUE #777: state: propagate errors when removing transparent index entries
## Summary

While investigating #681, I noticed that the transparent index range-delete error is ignored during index removal:

https://github.com/zakura-core/zakura/blob/f892b9074002a04a678ef2365ec7658795796572/crates/zakura-state/src/service/finalized_state/disk_format/upgrade/drop_tx_locs_by_spends.rs#L26-L33

`write_batch()` returns a `Result<(), rocksdb::Error>`, but `let _ =` discards the result and any RocksDB error:

https://github.com/zakura-core/zakura/blob/f892b9074002a04a678ef2365ec7658795796572/crates/zakura-state/src/service/finalized_state/column_family.rs#L319-L323

## Impact

If the range deletion fails transiently and later writes succeed, `drop_tx_locs_by_spends::run()` can continue replacing shielded nullifier locations with empty values and then return success. Its caller subsequently removes the `+indexer` marker even though transparent index entries remain readable from the database:

https://github.com/zakura-core/zakura/blob/f892b9074002a04a678ef2365ec7658795796572/crates/zakura-state/src/service/finalized_state/disk_format/upgrade.rs#L595-L602

This can also affect a later index rebuild. `track_tx_locs_by_spends` treats the first existing spend mapping it encounters for a height as proof that the entire height was indexed:

https://github.com/zakura-core/zakura/blob/f892b9074002a04a678ef2365ec7658795796572/crates/zakura-state/src/service/finalized_state/disk_format/upgrade/track_tx_locs_by_spends.rs#L48-L64

A leftover transparent mapping can therefore make the rebuild incorrectly assume that the entire block was already indexed, causing it to skip restoring missing shielded index entries.

Persistent storage failures will normally cause subsequent writes to fail as well, so this is a latent correctness and recovery issue rather than evidence of observed database corruption. RocksDB can, however, automatically recover from eligible retryable I/O and WAL errors, so the removal code should not rely on every later write failing after the initial error.

## Expected behavior

Index removal must not be reported as complete until every deletion and rewrite has succeeded. A storage error should preserve a state that causes removal to be retried on the next writable startup.

## Acceptance criteria

- Propagate the initial transparent range-delete error instead of discarding it.
- Do not remove the `+indexer` marker, or persist a future `Absent` state, when any removal write fails.
- Retry interrupted or failed removal on the next writable startup.
- Add regression coverage for a transient failure of the initial range deletion followed by otherwise successful writes.
- Verify that a later index rebuild cannot skip a height containing partially removed index data.

Related: #681

jonwalch:
## Motivation

The non-indexer state migration ignores the RocksDB result from deleting transparent spending indexes. A transient deletion failure can therefore be followed by successful nullifier rewrites and removal of the `+indexer` marker, leaving a partially removed index that a later rebuild can mistake for complete data.

Closes #777.

## Solution

- [x] Propagate the transparent range-delete failure so the format-change coordinator cannot clear `+indexer` after that write fails.
- [x] Inject a real `rocksdb::Error` at the typed batch-write boundary, ensuring the regression exercises the result returned by `write_batch()`.
- [x] Exercise failed removal, successful startup retry, and the real rebuild loop with a mixed transparent/Orchard-spend transaction.
- [x] Keep the rebuild's cross-feature support test-only; production indexer behavior is unchanged.

## Testing

The regression seeds transparent and Orchard spend locations at the same synthetic transaction location. It verifies that an injected range-delete failure preserves the `+indexer` marker and both locations, a retry removes both locations while retaining the consensus Orchard nullifier, and the real rebuild loop restores both locations and resolves them to the spending transaction hash.

Mutation check: temporarily restoring the original `let _ = ...write_batch()` made the regression fail because database startup incorrectly succeeded.

Completed checks:

- `cargo test -p zakura-state range_delete_error_preserves_and_rebuilds_spend_indexes --lib`
- `cargo test -p zakura-state` (541 unit tests passed, 3 ignored; 2 integration tests and 1 doc test passed)
- `cargo test -p zakura-state --features indexer --lib -- --skip intra_block_self_spend_chain_in_finalized_state` (540 passed, 3 ignored)
- `cargo clippy -p zakura-state --all-targets --all-features -- -D warnings`
- `cargo test -p zakura-state --all-features --all-targets --no-run`
- `npx --yes markdownlint-cli@0.39.0 "**/*.md" --ignore node_modules --ignore target --ignore .git --config .github/.markdownlint.yaml`

The full indexer-feature library run has one pre-existing failure in `intra_block_self_spend_chain_in_finalized_state`; the same failure reproduces on untouched `origin/main`. All remaining indexer-feature library tests pass with the command above.

## Changelog

Added `docs/changelog/unreleased/778.md` under `Fixed`.

- #777
- #681

### Follow-up Work

- Broader indexer schema-state handling remains tracked by #681.


evan-forbes (changes requested):
The production change correctly propagates the initial range-delete error. It returns before the later nullifier deletions and before the coordinator clears the `+indexer` marker.

The regression does not yet cover the final acceptance criterion from #777. It commits genesis, inserts one synthetic transparent mapping, and checks the startup retry. It does not create any shielded spend mappings or run the index rebuild. The test therefore does not prove that a later rebuild cannot skip a height with partially removed index data.

Please add the rebuild regression described in #777. The test should inject the removal failure and confirm that the marker remains. It should then retry the operation and run the index rebuild. Finally, it should verify the rebuilt spend indexes. At minimum, please seed transparent and shielded spend mappings. Prove that the injected failure cannot leave one index removed while another remains.

The production fix looks correct. This missing acceptance coverage is my only requested change.

jonwalch:
Addressed in 7ae1a7d9b. The regression now seeds both transparent and Orchard spend locations, verifies a failed removal preserves the marker and both locations, retries removal and verifies both locations are absent while the consensus nullifier remains, then runs the real rebuild loop and verifies both locations and the spending transaction hash are restored. The mutation check still fails when the original ignored write result is restored.

evan-forbes:

PR #779, opened 2026-08-22
perf(state): avoid contextual UTXO copies

evan-forbes:
## Motivation

The contextual path copied large UTXO maps several times before it validated
a block.

The worst-case transparent benchmark models a 1,999,944-byte block with 26,645
spends and 26,645 new outputs. The old contextual constructor copied every
spent UTXO and then duplicated every new output into the same map.

## Solution

- Move prepared blocks and spent UTXOs into contextual validation.
- Keep new outputs separate from spent outputs.
- Read `OrderedUtxo` values directly during transaction and block value-balance
  checks.
- Document and test the invariant that the spent-output map contains every
  transparent input, including intra-block spends.
- Preserve the offending outpoint and map size in invariant failures.
- Extend the worst-case transparent benchmark to measure each affected step.
- Add an equivalence test for ordered and plain UTXO value-balance paths.

The change preserves contextual validation order. It changes ownership and
lookup representation only.

## Results

| Benchmark | Baseline | Result | Reduction |
| --- | ---: | ---: | ---: |
| Contextual preparation | 13.663 ms | 4.840 ms | 64.58% |
| Remaining-value check | 6.495 ms | 3.486 ms | 46.33% |

Moving the contextual inputs also removes a 2.850 ms median clone from the
worst-case transparent path. These measurements isolate local preparation.
They do not predict the full production contextual interval.

## Testing

- `cargo bench -p zakura-state --bench transparent_spend_worst_case --features internal-bench`
- `cargo test -p zakura-chain -p zakura-state --lib --locked` on the #780 stack tip
  - the state suite passed 515 tests and ignored 3 tests
- `cargo check -p zakura-chain -p zakura-state --all-targets --locked`
- `cargo clippy -p zakura-chain -p zakura-state --all-targets -- -D warnings`
- `cargo check --workspace --all-targets --locked`
- `git diff --check`

## Changelog

This PR adds `docs/changelog/unreleased/779.md`.

## Stack

- Base: `main`.
- Parent: none.
- Child: #780.
- This PR is independent of #748, #754, #776, and #773.

### Follow-up Work

- Use the new-fleet A/B to rank the remaining contextual components.
- Measure nullifier, anchor, UTXO lookup, tree, header transition, and writer
  scheduling work separately.
- Evaluate authenticated contextual preparation during semantic verification.

### DIFF
--- a/crates/zakura-chain/src/block.rs
+++ b/crates/zakura-chain/src/block.rs
@@ -296,6 +296,15 @@ impl Block {
     }
 
     /// Returns the overall chain value pool change using borrowed ordered UTXOs.
+    ///
+    /// The given `utxos` must contain the [`transparent::OrderedUtxo`]s of every
+    /// input in this block. This includes UTXOs created by earlier transactions
+    /// in the same block. The map can also contain unrelated UTXOs, which this
+    /// method ignores.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if `utxos` omits a transparent input's UTXO.
     pub fn chain_value_pool_change_from_ordered_utxos(
         &self,
         utxos: &HashMap<transparent::OutPoint, transparent::OrderedUtxo>,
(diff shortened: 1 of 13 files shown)


p0mvn (approved):
LGTM

ValarDragon:
Nice!!

ValarDragon (approved):
Great job! All these were pretty clearly not needed

Codex xhigh fact checked as well

PR #780, opened 2026-08-22
perf(state): share contextual snapshot maps

evan-forbes:
# Summary

PR #779 removes redundant UTXO copies while it constructs a contextual block.
The non-finalized state still cloned the UTXO maps stored in every historical
contextual block whenever it extended a shared chain snapshot.

This stacked change stores each contextual block's immutable new-output and
spent-output maps behind `Arc`. Chain snapshots now share those maps. A
consuming conversion still returns an owned map and clones only when another
snapshot retains the data.

# Safety

- The change does not alter UTXO lookup, value-balance, rollback, finalization,
  or chain selection.
- The maps remain immutable after contextual block construction.
- Regression tests prove that contextual block clones and sibling forks share
  both immutable maps while each fork retains independent mutable indexes.
- The state suite covers contextual insertion, forks, rollback, finalization,
  persistence, and header-chain coherence.

# Measurement

The release benchmark models 1,000 retained blocks with 165 output entries per
block. Five invocations collected 100 samples each.

| Resource | Baseline | Result | Reduction |
| --- | ---: | ---: | ---: |
| Clone latency | 2.440 ms | 0.008 ms | 99.66% |
| Allocated bytes per clone | 22,848,000 B | 8,000 B | 99.965% |
| Allocations per clone | 1,001 | 1 | 99.90% |

Five current invocations collected 100 samples each. The table uses the median
invocation median. Allocation counts and byte totals were identical in every
sample. The benchmark reuses the repository's counting-allocator pattern.

The benchmark also reports the remaining cumulative `created_utxos`,
transaction-location, and Orchard-nullifier clones. This PR does not share
those mutable indexes. The benchmark measures allocation traffic for one
snapshot clone. It does not claim an exact process RSS reduction or predict the
Orchard-heavy fleet result.

# Verification

- `cargo test -p zakura-chain -p zakura-state --lib --locked`
  - the state suite passed 515 tests and ignored 3 tests
- `cargo test -p zakura-state --lib --features proptest-impl forked_equals_pushed_genesis`
- `cargo check -p zakura-chain -p zakura-state --all-targets --locked`
- `cargo bench -p zakura-state --bench chain_clone_worst_case --features internal-bench`

## Stack

- Base and parent: #779.
- Children: none.
- This PR is independent of #748, #754, #776, and #773.


p0mvn (approved):
LGTM

ValarDragon:
[P2] Do not hide the conditional map clone inside From. At request.rs:758
    (https://github.com/zakura-core/zakura/blob/3aca19c7c0239e7245cd4d8008f94f2341859e3c/crates/zakura-state/src/request.rs#L752-L759), Arc::unwrap_or_clone can copy the entire
    new_outputs map whenever another snapshot retains it. This runs during contextual finalization, so finalized-state write latency depends on unrelated snapshot lifetimes. It also
    directly violates the repository rule prohibiting hidden clones in From implementations. Use an explicitly named conversion that advertises the possible clone, or propagate shared
    ownership farther and make any clone explicit at the caller. A shared-map finalization benchmark would cover this tradeoff.

PR #782, opened 2026-08-22
perf(state): measure contextual verification phases

evan-forbes:
## Summary

- measure the full contextual interval, initial checks, parent-chain work, new
  chain construction, and state-writer snapshot cloning;
- separate the cumulative UTXO snapshot from transparent spend checks;
- measure shielded anchors, contextual block construction, and the parallel
  state update;
- separate the parallel chain clone from `Chain::push`;
- measure header-transition preparation and commit on success and error paths;
- emit matching mined-only series for every phase on the mined-block path.

## Motivation

The preserved memo-enabled field run reduced the median transaction phase to
1.795 ms. Its median state commit remained 34.338 ms. PR #781 separates writer
queue delay from writer execution. This PR splits writer execution so the fleet
can identify the next material optimization without inferring phase costs from
aggregate quantiles.

## Safety

The change adds elapsed-time capture and histograms only. It does not move a
check, change state ownership, or alter chain selection.

The outer `parallel_update` duration includes task scheduling and completion.
The `parallel_task.*` durations overlap because block commitment, Sprout anchor
validation, and chain update run concurrently. Only `parallel_update` measures
their critical-path wall time.

## Benchmark scope

This PR adds observability. It does not claim a runtime speedup. A synthetic
microbenchmark would not show whether these histograms partition the production
contextual interval. The state tests exercise every instrumented success and
error path. The next fleet run must compare each mined-only phase count with
the outer contextual count, then rank phase means from matching campaign
deltas. Concurrent `parallel_task.*` means must not be added together.

## Verification

- `cargo test -p zakura-consensus -p zakura-state --lib --locked`
  - 234 consensus tests passed and 1 test remained ignored
  - 567 state tests passed and 4 tests remained ignored
- `cargo clippy -p zakura-state -p zakura-consensus --all-targets -- -D warnings`
- `cargo check -p zakura-state -p zakura-consensus --all-targets --locked`
- `npm exec --yes markdownlint-cli -- docs/changelog/unreleased/782.md --config .github/.markdownlint.yaml`
- `git diff --check`

This PR adds `docs/changelog/unreleased/782.md` with an internal-only marker.

## Stack

- Base: `main`.
- Parents #748 and #781 are merged.
- Children: none.

### DIFF
--- a/crates/zakura-state/src/service/non_finalized_state.rs
+++ b/crates/zakura-state/src/service/non_finalized_state.rs
@@ -39,14 +39,38 @@ mod tests;
 pub(crate) use backup::write_semantically_verified_backup_block;
 pub(crate) use chain::{Chain, SpendingTransactionId};
 
-macro_rules! record_contextual_duration {
-    ($metric_name:literal, $mined_metric_name:literal, $duration:expr, $is_mined:expr $(,)?) => {{
-        let duration = $duration.as_secs_f64();
-        metrics::histogram!($metric_name).record(duration);
-        if $is_mined {
-            metrics::histogram!($mined_metric_name).record(duration);
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum ContextualMetrics {
+    Disabled,
+    AllBlocks,
+    Mined,
+}
+
+impl ContextualMetrics {
+    pub(crate) fn for_commit(is_mined: bool) -> Self {
+        if is_mined {
+            Self::Mined
+        } else {
+            Self::AllBlocks
         }
-    }};
+    }
+
+    pub(crate) fn record_duration(
+        self,
+        metric_name: &'static str,
+        mined_metric_name: &'static str,
+        duration: std::time::Duration,
+    ) {
+        if self == Self::Disabled {
+            return;
+        }
+
+        let duration = duration.as_secs_f64();
+        metrics::histogram!(metric_name).record(duration);
+        if self == Self::Mined {
+            metrics::histogram!(mined_metric_name).record(duration);
+        }
+    }
 }
 
 /// The state of the chains in memory, including queued blocks.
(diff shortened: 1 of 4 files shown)


crates/zakura-state/src/service/non_finalized_state.rs
```diff
@@ -579,32 +666,62 @@ impl NonFinalizedState {
...
             &new_chain.spent_utxos,
             finalized_state,
-        )?;
+        );
+        contextual_metrics.record_duration(
+            "state.contextual.transparent_spend.duration_seconds",
+            "state.contextual.mined.transparent_spend.duration_seconds",
+            transparent_spend_start.elapsed(),
+        );
+        let spent_utxos = spent_utxos?;
```
czarcas7ic, line 689: The copied unspent outputs now stay in memory for the rest of validation, so we use more memory than before. We could free that copy as soon as the spending checks finish, like the old code did.

czarcas7ic (approved):

crates/zakura-state/src/service/non_finalized_state.rs
```diff
@@ -579,28 +666,58 @@ impl NonFinalizedState {
         // Reads from disk
         //
         // TODO: if these disk reads show up in profiles, run them in parallel, using std::thread::spawn()
+        let unspent_utxo_snapshot_start = Instant::now();
```
ValarDragon, line 669: im worried about Instant::now() being called repeatedly being an atomic slowdown for small operations. (IIRC its around 30ns)

Its probably fine here though

evan-forbes: All ~16 timers here run once per block commit; none are in a per-transaction or per-output loop. On Linux, `Instant::now()` reads the clock through the vDSO with no atomic, and I measured about 22 ns per read. That totals under 1 µs per block against a commit that takes milliseconds, so I'll leave it as is.

PR #792, opened 2026-08-24
feat(runner): alert on a stale or incomplete release state

p0mvn:
Closes the last monitoring gap in the release-state pipeline, on the `#zakura-alerts` webhook the
node alerts already use.

## The gap

The Mainnet release-state publisher is the only scheduled job on the fleet with **no monitoring at
all** — no `OnFailure`, no Sentry, no check timer. Compare what the snapshot publishers get:

| Unit | Monitored by |
| --- | --- |
| `zakura-snapshot.service` | `zakura-snapshot-check.timer` → Sentry cron monitor + staleness alert |
| `zakura-snapshot-pruned.service` | same |
| `zakura-release-state.service` | **nothing** |

A failure surfaces only when the weekly import trips its 48-hour staleness check: up to seven days
of latency, as a red Action rather than a page.

This is not hypothetical. The generator host crashlooped on an end-of-support panic for 4.6 days
and nothing noticed — it was found by hand while wiring something else.

## Watch the artifact, not the unit

Deliberate choice. A systemd `OnFailure=` only fires when a unit *fails*; it says nothing about a
timer that silently stopped firing, and nothing at all if the host disappears. Watching the
published pointer catches all three, and it works from anywhere.

Three conditions:

- **`unreachable`** — the pointer or its `meta.json` cannot be fetched
- **`stale`** — `generated_at` older than the window
- **`incomplete`** — `meta.files` is missing a required artifact

**Missing files rank above staleness on purpose.** A bundle can be perfectly fresh and still
useless — that is exactly what a generator exporting from the wrong database produces, and no
freshness check would flag it. It is also the precise regression the retired snapshot-host
publisher would have caused, since that one could only ever emit three files.

## Thresholds

48 hours by default, matching `fetch-release-state.py --max-age-hours`, so the alert fires *before*
the weekly import would refuse the bundle rather than after. Exports are daily, so that also
tolerates one missed run without paging.

Threshold zero on the alert state machine: a bundle already past its window, or missing a file, is
not a transient blip to wait out — the window is the grace period.

## A bug this caught

`fetch_json` sent no `User-Agent`. Harmless against the dashboard on loopback, but the public
release-state origin sits behind an edge that answers urllib's default agent with **403**. Without
the fix this check would have reported `unreachable` forever — a permanent false alarm, which is
worse than no alerting because it trains people to ignore the channel.

Found by smoke-testing against the real endpoint rather than a fixture. Fixed here.

## Testing

`python3 -m unittest` — **21 pass**: 11 new, and the 10 pre-existing stall-recovery tests still
green. `test_zakura_cluster_status.py` also still passes (51).

The new cases cover the daily-export cadence (a 25-hour-old bundle must *not* alert), the
fresh-but-incomplete regression, missing-file precedence over staleness, unparsable and absent
timestamps, and config validation.

End to end against production, in one dry run with a deliberately bogus fleet URL alongside the
real pointer:

```
release_state: { "mainnet": { "condition": "ok", "height": 3457371, "alerting": false } }
fleets:        { "mainnet": { "condition": "unreachable", "alerting": false } }
```

Both paths exercised. And the alert the regression would produce:

```
:rotating_light: *Zakura mainnet release state* incomplete
missing mainnet-frontier-grid.bin
pointer: https://zakura-release.valargroup.dev/release-state/latest.json
height: 3457371
```

## Rollout

Configuration is **additive**: a config with no `[[release_state]]` section behaves exactly as
before, so this can merge and deploy ahead of the config change without touching current
behaviour.

The matching config lives in `valargroup/zakura-snapshots` (`fleet-watchdog.toml`), which is a
separate repo, so enabling it is a second small PR there:

```toml
[[release_state]]
name = "mainnet"
url = "https://zakura-release.valargroup.dev/release-state/latest.json"
```

### DIFF
--- a/deploy/runner/fleet-watchdog.toml
+++ b/deploy/runner/fleet-watchdog.toml
@@ -7,3 +7,13 @@ dashboard_url = "http://167.99.103.111:8090/"
 name = "mainnet"
 url = "http://127.0.0.1:8090/data"
 dashboard_url = "http://159.65.183.89:8090/"
+
+# The Mainnet release-state pipeline. Watching the published pointer rather than the
+# generator's unit is deliberate: it catches a timer that silently stopped firing and
+# a host that vanished, neither of which a systemd OnFailure would report. The
+# generator exports daily, and the 48h default window matches
+# fetch-release-state.py's --max-age-hours, so an alert fires before the weekly
+# import would refuse the bundle rather than after.
+[[release_state]]
+name = "mainnet"
+url = "https://zakura-release.valargroup.dev/release-state/latest.json"
--- a/deploy/runner/test_zakura_cluster_watchdog.py
+++ b/deploy/runner/test_zakura_cluster_watchdog.py
@@ -198,3 +198,113 @@ class StallRecoveryTests(unittest.TestCase):
 
 if __name__ == "__main__":
     unittest.main()
+
+
+def make_release_state(**overrides):
+    defaults = {
+        "name": "mainnet",
+        "url": "https://example.invalid/release-state/latest.json",
+        "stale_after": 172800.0,
+        "required_files": (
+            "main-checkpoints.txt",
+            "mainnet-frontier.bin",
+            "mainnet-treestate-subtrees.bin",
+            "mainnet-frontier-grid.bin",
+        ),
+    }
+    defaults.update(overrides)
+    return watchdog.ReleaseState(**defaults)
+
+
+ALL_FILES = {
+    "main-checkpoints.txt": {},
+    "mainnet-frontier.bin": {},
+    "mainnet-treestate-subtrees.bin": {},
+    "mainnet-frontier-grid.bin": {},
+}
+
+
+class ReleaseStateConditionTests(unittest.TestCase):
+    """The published artifact, not the unit, is what tells us the pipeline is alive."""
+
+    NOW = 1_800_000_000.0
+
+    def pointer(self, age_seconds=0.0, height=3457371):
+        stamp = watchdog.datetime.datetime.fromtimestamp(
+            self.NOW - age_seconds, watchdog.datetime.timezone.utc
+        )
+        return {
+            "height": height,
+            "generated_at": stamp.isoformat().replace("+00:00", "Z"),
+        }
+
+    def condition(self, pointer, files, target=None):
+        return watchdog.release_state_condition(
+            target or make_release_state(), pointer, {"files": files}, self.NOW
+        )
+
+    def test_fresh_complete_bundle_is_ok(self):
+        cond, _ = self.condition(self.pointer(age_seconds=3600.0), ALL_FILES)
+        self.assertEqual(cond, "ok")
+
+    def test_a_bundle_past_its_window_is_stale(self):
+        cond, detail = self.condition(self.pointer(age_seconds=200_000.0), ALL_FILES)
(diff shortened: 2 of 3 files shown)


p0mvn:
## Correction, and the config is now in this PR

I said earlier the matching config lived in `valargroup/zakura-snapshots`. That was wrong — the
file the deploy actually installs is in this repo:

```
sudo install -m 644 deploy/runner/fleet-watchdog.toml "$watchdog_dir/fleets.toml"
```

`zakura-mainnet-deploy.yml` installs the script, this config, the unit, and the `#zakura-alerts`
webhook env together, all `if: always()`. So this is one PR and one deploy, not two — I had
confused the shipped copy with the snapshots repo's own watchdog wiring.

The config is now included:

```toml
[[release_state]]
name = "mainnet"
url = "https://zakura-release.valargroup.dev/release-state/latest.json"
```

Verified through the real loader against the shipped file: two fleets plus one release-state target
with a 48h window and four required files. 21 tests still pass.

This also means the alerting goes live the next time the mainnet deploy workflow runs, rather than
needing a follow-up in another repo.

p0mvn:
## Tested end to end against `#zakura-alerts`

Both directions verified with the **real webhook** on `us-east-0`, then confirmed by reading the
channel.

**Alert.** A synthetic pointer whose `meta_url` is a genuine grid-less bundle (`v1/3456571`, three
files), so the check walks the same path production will against real metadata:

```
:rotating_light: *Zakura alerting-test release state* incomplete
missing mainnet-frontier-grid.bin
pointer: http://127.0.0.1:8791/latest.json
height: 3456571
```

**Recovery.** Same state key, repointed at the live pointer:

```
:white_check_mark: *Zakura alerting-test release state* recovered from incomplete
pointer: https://zakura-release.valargroup.dev/release-state/latest.json
```

Both landed in `#zakura-alerts`, formatted consistently with the existing node alerts.

Isolation: a separate state file and a temp script path, so the running service was never touched
(`zakura-fleet-watchdog.service` stayed `active` throughout). The fleet entry in the test config was
deliberately bogus and correctly did **not** alert — a fresh state puts `bad_since = now`, so it sat
below its threshold. Fixture and state files removed afterwards; port 8791 closed.

## A bug the test found

The first attempt reported `ok` for a three-file bundle. My test config was wrong — it pointed at
`meta.json` instead of a pointer — but the *reason* it passed was a real defect: with no usable
`meta_url`, the missing-files check was skipped and the target looked healthy.

That is the worst failure mode for monitoring, because it also suppresses the alert that would
otherwise fire. Now an absent or non-object `meta.files` is its own `unreadable` condition, and a
pointer without a `meta_url` is an error rather than an empty meta. Two regression tests cover it.

Worth noting this is exactly the misconfiguration an operator could make — pointing the config at a
bundle instead of a pointer — so it would have shipped as a silent blind spot.

**23 tests pass**, plus `test_zakura_cluster_status.py` (51) unaffected.

## Two unrelated things the channel surfaced

While reading the baseline: **`us-west-0` stalled 70h40m** and **`asia-pacific-0` stalled 69h37m**.
Both real, both pre-existing, and they match the dashboard's `healthy=10 total=12`. Not related to
this PR, but nobody appears to have picked them up.

Also visible: `archive-vct-off` alerted `stalled` and then `recovered` during yesterday's sync
repair — the fleet membership from #787 working live, unprompted.

PR #796, opened 2026-08-24
feat(rpc): add the getchaintips method

### LINKED ISSUE #788: bug: zcashd-compat getchaintips holds cs_main for 8–13 s, stalling all RPC and block commitment
### What happened?

**I expected to see this happen:**

A single `getchaintips` call is answered without affecting unrelated RPC requests, and without preventing the node from committing new blocks.

**Instead, this happened:**

On Mainnet at height ~3,450,000, one `getchaintips` call to the `zcashd-compat` sidecar takes **8–13 seconds of pure CPU** and holds zcashd's global `cs_main` lock for that entire time. While it runs:

* every other RPC on the sidecar blocks — including trivial ones such as `getblockcount`, which otherwise answers in 0.3 ms;
* the sidecar cannot run ProcessNewBlock, so the sidecar's chain view stops advancing for the duration, even though zakurad itself keeps syncing normally

The response itself is tiny — 113 tips, ~10 KB — so the cost is entirely in finding the tips, not in producing or serializing the answer.

This also leaves operators without an alternative: `zakurad`'s native RPC does not implement `getchaintips` at all, so the request can only be served by the sidecar.

```console
$ curl -s -u "$COOKIE" -d '{"jsonrpc":"2.0","method":"getchaintips","params":[],"id":1}' http://node:8232
{"jsonrpc":"2.0","id":1,"error":{"code":-32601,"message":"Method not found"}}
```

### What were you doing when the issue happened?

**1. Cost of the call, on an idle node serving no traffic (load average 0.5):**

```console
$ curl -s -o /dev/null -w 'getchaintips  %{time_total}s\n' -u xxx:*** \
    -d '{"jsonrpc":"1.0","id":"t","method":"getchaintips","params":[]}' http://node:8033
getchaintips  7.438195s
$ curl -s -o /dev/null -w 'getblockcount %{time_total}s\n' -u xxx:*** \
    -d '{"jsonrpc":"1.0","id":"t","method":"getblockcount","params":[]}' http://node:8033
getblockcount 0.000256s
```

Repeated runs: 7.4 s, 9.4 s, 10.1 s, 10.9 s, 11.1 s. Two consecutive calls cost the same, so nothing is cached.

**2. Causal test — poll a trivial RPC every 200 ms, fire exactly one `getchaintips` in the middle:**

```
t=0.0 … 5.9 s   getblockcount = 0.000 s   (30 consecutive samples)
      >>> one getchaintips issued
      >>> returned 113 tips in 7.765 s
t=13.8          getblockcount = 7.675 s   <<< blocked for the whole duration
t=14.0 …        getblockcount = 0.000 s   (immediately normal again)
```

**3. What the sidecar is doing during the freeze** — sampling `/proc/<pid>/task/*/{stat,syscall}` at 20 Hz:

```
thread states: {'S': 73, 'R': 1}   total threads: 74

thread             state  syscall    cpu
zc-http-worker       R    running    35.2 s   <-- userspace, no syscall at all
zcashd               S    futex_wait_queue
zc-txnotify          S    futex_wait_queue
zc-scriptcheck  (x12) S   futex_wait_queue
...
```

One HTTP worker burns a full core entirely in userspace — no syscalls, so no disk or network involved — while every other thread waits on a futex. `perf record` on the same process is dominated by `malloc`/`free`; symbols are unavailable because the managed zcashd binary is stripped.

**4. Production impact — passive packet capture on a node under normal request load, 239 s window, no added load:**

```
paired RPC calls                    : 591
getchaintips calls                  : 5, in flight for 52.9 s total = 22% of the window
slow (>1 s) calls other than getchaintips : 94
   ... overlapping a getchaintips window  : 88  (94%)
calls faster than 50 ms outside those windows : 476
```

Per-method latency from the same capture — note the bimodal shape of every other method, and that `getchaintips` is the only one where *every* call is slow:

```
method               n    p50 ms   p90 ms    max ms
getchaintips         3   10569.7  10804.1   10804.1
getblockchaininfo  106       2.4   4903.4   10222.8
getblockcount       56       0.3   7197.5   10558.1
getrawtransaction   40       0.4      2.6       9.3
```

A health check polling every 15 s therefore fails roughly one time in five, which shows up as the node flapping in and out of a healthy state in monitoring.

### Zakura logs

Nothing appears in the `zakurad` logs — this is the main diagnostic difficulty. The freeze happens entirely inside the supervised `zcashd` child, so from `zakurad`'s point of view nothing is wrong: it keeps its own tip, its own RPC on `:8232` stays responsive throughout, and no warning is emitted. Host metrics look healthy too (one busy core out of 96, load average 11.8, I/O pressure 0.02%).

The only externally visible symptom is that RPC on the sidecar port stops answering for 8–13 s at a time.

### Zakura Version

zakurad 1.1.0 and zakurad 1.2.0 (reproduces identically on both), Mainnet, started with --zcashd-compat and zcashd_source = "embedded"

### Which operating systems does the issue happen on?

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] Other OS

### OS details

Ubuntu 24.04 in Docker, x86_64. Hosts: 32 and 96 cores, 124–251 GB RAM, NVMe. Chain height ~3,450,000, `mapBlockIndex` ≈ 3.45M entries, 113 chain tips reported.

### Additional information

**Root cause.** This looks like the pre-2016 Bitcoin Core implementation of `getchaintips`, which was optimised in [[bitcoin/bitcoin#7827](https://github.com/bitcoin/bitcoin/pull/7827)](https://github.com/bitcoin/bitcoin/pull/7827) ("Speed up getchaintips"). That PR's description matches the profile exactly:

> Previous code made two passes through mapBlockIndex, inserted each block into a new set and then erased all but the chain tips - so hundreds of 1000's of inserts/erases.

zcashd forked before that change, and the vendored zcashd still carries the original version. That explains a fully CPU-bound, allocation-heavy workload producing a 10 KB answer, and it explains why the cost grows with chain height — it is O(size of the block index).

**Why this matters beyond RPC latency.** The lock is also held across `ProcessNewBlock`, so a node that is otherwise perfectly healthy stops committing new blocks for up to 13 s at a time. During that window the node's chain view silently falls behind the network, and a request for a height the network already has fails with `-8 Block height out of range` — even though the node's connectivity and sync are fine. Because the freeze leaves no trace in the logs and none in host metrics, it is easy to misdiagnose as a peer or sync problem; we spent a considerable amount of time looking in the wrong place before profiling the sidecar directly.

**Possible fixes**, in rough order of preference:

1. **Implement `getchaintips` natively in `zakurad`.** The tips of the non-finalized state are already tracked in memory, so the answer requires no scan and no global lock. We prototyped this and measured **0.0003 s versus 13.5 s**, with no effect on concurrent requests on the same port. The trade-off is a narrower answer: only branches within the finalization boundary appear (10 tips instead of 113 in our case), and `status` can only be `active` or `valid-fork`, since the node fully validates what it keeps and discards what it rejects. For reorg detection that is sufficient — a reorg deeper than the finalization boundary cannot occur.

2. **Carry the #7827 patch in the managed zcashd build.** Simpler, but the reported speedup is ~20×, which would still leave roughly 0.5–0.7 s under `cs_main` at the current height, and that figure keeps growing with the chain.

evan-forbes:
## Motivation

Closes #788.

A single `getchaintips` call to the zcashd-compat sidecar costs 8–13 s of pure
CPU on Mainnet at height ~3,450,000, and holds zcashd's global `cs_main` lock
for that entire time. While it runs, every other RPC blocks — `getblockcount`
goes from 0.3 ms to over 7 s — and the sidecar cannot run `ProcessNewBlock`, so
its chain view stops advancing. A health check polling every 15 s fails roughly
one time in five, which shows up as the node flapping in monitoring.

The root cause is in the algorithm, not in the response. zcashd builds the tip
set by iterating all of `mapBlockIndex` twice: once to insert every block, once
to erase every block that is some other block's `pprev`
([`src/rpc/blockchain.cpp:1270-1283`](https://github.com/zcash/zcash/blob/master/src/rpc/blockchain.cpp)).
That is O(all blocks ever seen) under the global lock, and it grows with chain
height forever. The answer itself is tiny — 113 tips, ~10 KB — so the cost is
entirely in finding the tips.

Operators had no alternative: `zakurad` did not implement `getchaintips` at
all, so the request could only be served by the sidecar.

## Solution

Implement `getchaintips` natively in `zakurad`, per the direction on the issue.

Zakura does not have zcashd's problem to solve. It keeps only the non-finalized
chains in memory, and that set is bounded by `MAX_NON_FINALIZED_CHAIN_FORKS`
(10) and `MAX_INVALIDATED_BLOCKS` (100). So the new
`ReadRequest::ChainTips` is bounded by the number of tracked forks rather than
by the height of the chain, it never touches the finalized state's block index,
and it holds no global lock.

`read::chain_tips` reports the tips that are still live, covering four sources:

| Source | Status |
| --- | --- |
| The best chain tip | `active` |
| Non-finalized forks | `valid-fork` |
| The `invalidated_blocks` map | `invalid` |
| The selected header chain, when some block bodies are unavailable | `headers-only` |

`branchlen` is measured against the best chain, matching zcashd's
`chainActive.FindFork(block)`. Tips are sorted by descending height, as in
zcashd, with hash as a tiebreak so repeated calls on an unchanged state return
an identical list.

One subtlety is worth calling out for review. Zakura keeps a **shortened**
chain in `chain_set` after a reorg or an `invalidateblock`, and that chain's
tip is frequently still an interior block of a longer chain. Such a block has a
known successor, so it is not a tip and zcashd would not report it. Without a
filter, `getchaintips` would invent a phantom zero-length fork after every
invalidation. `chain_tips` skips those, and
`chain_tips_report_invalidated_branches` pins the behavior.

### Compatibility

The two nodes deliberately report different tips, and integrations should know
this before switching endpoints:

- zcashd never prunes its block index, so it lists **every stale tip it has
  ever seen** — 113 in the issue report.
- Zakura drops a fork once it falls below the finalized tip, so it lists only
  live tips — typically a dozen or fewer.
- Zakura never returns zcashd's `valid-headers` or `unknown` statuses. Every
  block in its non-finalized state is contextually verified, so a tip is either
  fully valid, invalidated, or known only by its header.

Callers using this for liveness or reorg monitoring get the same answer far
faster. Callers using it for historical fork archaeology will see less. Matching
zcashd's historical view would need stale tips persisted in the state database,
which means a state format version bump and a migration — and it would re-import
the unbounded growth that makes zcashd slow in the first place. That is
deliberately not done here.

## Testing

State layer (`crates/zakura-state/src/service/read/tests/vectors.rs`), ten
vectors covering each branch of the tip walk:

- no blocks at all returns an empty list;
- a single chain reports exactly one tip, as `active`;
- a fork is reported below the active tip with the correct `branchlen`,
  measured from the block it shares with the best chain;
- an invalidated branch is reported as `invalid` with its own `branchlen`, and
  the shortened chain left behind is **not** reported — this is the phantom-fork
  regression above;
- repeated invalidations on one branch produce one invalid tip;
- an invalid branch keeps the correct fork point when another branch becomes active;
- a selected header tip above the block tip is reported as `headers-only`;
- a taller selected header fork is measured from its fork with the active chain;
- a shorter higher-work selected header fork is still reported;
- a selected header tip with an available body is not reported again.

RPC layer (`crates/zakura-rpc/src/methods/tests/vectors.rs`): one vector asserts
the zcashd JSON contract end to end — the `height`/`hash`/`branchlen`/`status`
field names and the exact status spellings (`headers-only`, `valid-fork`) — for
a response containing all four statuses; another asserts that an empty state
returns an empty array rather than an error.

Also run and passing: the full `zakura-rpc` suite, `zakura-state --lib` (528
passed, 3 ignored), `cargo fmt --all -- --check`, `cargo clippy --all-targets -D warnings`
on both changed crates, `cargo xtask check-rpc-artifacts` (`rpc_openrpc.rs` is
regenerated in this PR), and `cargo check --workspace --all-targets`.

**Not run:** the `zcashd-compat` regtest interop suite, which is manual-only in
CI and spawns real nodes. End-to-end confirmation against a live zcashd has not
been done.

`docs/zcashd-compat.md` gains a short section telling operators to send
`getchaintips` to Zakura, with the cost and coverage differences.

## Changelog

`docs/changelog/unreleased/796.md`, under `Added`.

- zcashd's implementation: `src/rpc/blockchain.cpp:1235` (`getchaintips`) and
  the `CompareBlocksByHeight` ordering at `:1220`.
- Upstream Zebra does not implement `getchaintips`, so there was nothing to
  port.

### Self-audit

Review passes after the first push found and fixed five defects:

- **Phantom tips after repeated `invalidateblock`.** A second call on the same
  branch leaves a lower branch whose tip is the parent of the upper one. That
  parent has a known successor, so zcashd does not report it. The same applies
  to a fork tip whose child was invalidated. The active tip stays exempt,
  because zcashd always reports it.
- **`branchlen` short by the depth of the lower branch** in that same case. It
  was measured from the parent of the branch root, which is itself invalidated
  after a second call. It now walks back to the deepest invalidated ancestor,
  whose parent is on the best chain, which is what zcashd measures.
- **Wrong or missing selected header forks.** The first implementation required the
  selected header tip to have a greater height and assumed it extended the active
  tip. Header selection uses cumulative work, so a shorter fork can win. The state
  read now uses the selected projection to find the common ancestor and detect body
  availability.
- **Invalid `branchlen` after the active chain changes.** The invalidation point
  can be on a former active branch. The state read now finds that branch in the
  tracked chain set and measures it from the current active fork.
- **A tautological hash assertion.** The RPC test compared serde output with
  `Hash::to_string()`; both use the same encoder, and the test hash repeated one
  byte. It now uses 32 distinct bytes and a literal string, so it fails if the
  hash is not in zcashd's big-endian display order.

The audit also removed `ChainTipStatus::as_str` and its `Display` impl, which
nothing used and which held a second copy of the wire strings, and corrected the
changelog fragment link, which still named a guessed PR number.

### Follow-up Work

- End-to-end verification against a live zcashd sidecar, comparing both nodes'
  output on the same chain.


ValarDragon:
@v12sec pls review

ValarDragon (approved):

PR #800, opened 2026-08-25
fix(sync): break the VCT root-repair livelock

evan-forbes:
## What broke

Two continuous-sync nodes (`temp-zakura-sync-test-1`, `temp-zakura-sync-test-2`) wedged on
`origin/main` @ `5c33befdf`, ~8k blocks short of the network tip, and stayed down:

```
14:09:57 WARN  vct_write_retry: VCT: supplied root not yet verifiable; retrying checkpoint
               commit in place height=3452169 retry_cause=MissingRoot{replacement_required:true}
14:10:27 ERROR vct_write_retry: VCT: checkpoint commit stalled waiting for a verifiable supplied
               root or successor witness; the node will not recompute against the frozen frontier
```

The committer parks a checkpoint block waiting for a VCT root and asks header sync to supply
it. Header sync scheduled the repair, resolved its context, and then found zero eligible
suppliers — 10,872 `header_vct_repair_state` trace rows across 1,238 repair generations, every
one with `supplier_count: 0`, never reaching `assignment`.

This is a policy-level liveness deadlock. Three defects combined; fixing any one alone does not
restore liveness.

### 1. The supplier filter selects for a peer that cannot exist

`try_assign_vct_repair` required:

```rust
status.selected_tip_hash == task.owner.branch.target_tip_hash
status.oldest_retained_height <= predecessor.height
```

`task.owner.branch.target_tip_hash` is **our own** `frontiers.header_best.hash`
(`BodyWorkAuthority::for_snapshot`), so the first predicate demands a peer sitting on our
stalled tip. `oldest_retained_height` is the peer's finalized anchor, so the second demands a
peer whose finalized frontier is *below* our stall height — a peer less synced than we are.

Measured on node 2: our branch target was `0000…129e39…` @ 3455238; every peer advertised
`0000…4554fc…` @ 3460252 with anchor 3459251. Both predicates are unsatisfiable against a
healthy network, and get more unsatisfiable the longer the node is stuck.

The tip-hash predicate is not even guarding the thing being requested: the wire request carries
`target_tip_hash = context.target.hash`, the header **at the repair height**, not our tip.

### 2. The serving path refuses a finalized target

`acquire_retained_path` resolved the target only through the in-memory header graph, which holds
the retained suffix. A repair target is deep-finalized on any healthy peer, so every supplier
would have answered `TargetNotRetained` even with defect 1 fixed.

### 3. The committer's 500 ms poll destroys the reactor's backoff

`missing_vct_successor_retry` returned `MissingRoot { replacement_required: true }` whenever the
successor header existed but its auxiliary record did not — a branch that disputes nothing. That
forced a repair-generation bump on every poll, which retired and re-inserted the reactor's task,
discarding its `SupplierBackoff` and re-querying context. Result: a 2 Hz spin, the 1 s
`VCT_REPAIR_RETRY_INTERVAL` never applied, and a 2.4 GB `header_sync.jsonl` in eight hours.

The backoff is self-healing on its own — `resume_retry_cycle` clears `tried_sources` when
`SupplierBackoff` elapses — so the churn bought nothing and cost the backoff.

## What this changes

- **`reactor.rs`** — drop both unsatisfiable predicates. A peer serves the exact target either
  from its retained graph or from its finalized state, and those bands are exhaustive below its
  selected tip, so `selected_tip_height >= context.target.height` plus the capacity, schema, and
  idle checks is the complete condition. Branch membership is proved by the response: the
  supplier must return the exact target hash after the exact predecessor hash, and the existing
  admission path rejects anything else. The enforcement site is `handle_headers`, which checks
  the response target against the locally derived target, the returned common ancestor against
  the exact predecessor frontier sent, exactly one entry, complete, schema V1, and
  `entries[0].header.hash() == target`. A header hash pins `previous_block_hash`, so an
  off-branch or substitute header cannot pass. A peer that cannot serve the height answers
  `TargetNotRetained` / `HistoryPruned`, is recorded as a failed source, and the cycle is bounded
  by `MAX_SUPPLIERS_PER_CYCLE = 3` and the 1 s backoff.
- **`header_chain.rs`** — resolve an exact canonical finalized target through the finalized
  indexes when the header graph does not hold it, and serve it as a contiguous finalized run from
  the locator's canonical ancestor up to the target. An unknown hash, or one at or above the
  finalized frontier, still returns `TargetNotRetained`.
- **`write.rs`** — a missing successor auxiliary record continues the open repair episode instead
  of declaring a new one. Only a genuinely rejected delivery (`attributed_failure_repair_height`)
  starts a new episode. `missing_vct_successor_retry` now takes the successor height rather than
  the whole window, which is what it always used.
- **Observability** — an unassignable repair reports which requirement excluded the network
  (peers considered, plus counts rejected by height, capacity, schema, busy, already-tried),
  counts `sync.header.vct.repair.no_supplier.total`, escalates to error level after 60 s on one
  generation, and samples its trace row at 10 s. Previously an operator saw only
  `supplier_count: 0`, which counts `tried_sources` — not candidates — and so said nothing about
  why the network was excluded.
- **Audit fixes.** A `ResourceStalled` admission dropped the repair task. With defect 3 fixed the
  committer no longer bumps the repair generation while it polls, so nothing rescheduled a
  dropped task except an unrelated snapshot change, which a stalled node cannot produce. The task
  now backs off and retries another supplier. Separately, the no-supplier escalation clock reset
  on a non-empty candidate set, so a candidate that failed to send restarted the 60 s clock and
  suppressed the report; it now clears only once a request is actually assigned.

No wire-format change and no disk-format change. A new requester asking an old supplier for a
finalized target gets `TargetNotRetained`, which is already a handled outcome.

## Not changed, deliberately

- **No local recomputation inside the frozen VCT band.** The node deliberately lacks the
  intermediate frontiers there; safe recomputation would need a replay from a known frontier and
  may be impossible after pruning. The committer's refusal is correct.
- **No new advertised status field.** `oldest_retained_height` already carries the band split.
- **No reschedule on `VctRepairContextResult::Stale`.** That arm still waits for the next
  committed snapshot. A stale context means the state rejected our owner, so rescheduling from
  the reactor's own snapshot would re-derive the same stale owner and spin.
- **Roots serving needed no work.** `serve_block_roots` already joins tree-derived roots below
  the VCT upgrade height with indexed roots at and above it, and `finalized_tree_aux_for_page`
  already serves any page ending at or below the finalized tip. Both shapes are covered by
  existing tests (`serve_block_roots_serves_at_or_above_upgrade_from_index`,
  `finalized_pages_load_contiguous_tree_aux_from_state`).

## Tests

- `retained_path_serves_an_exact_finalized_target_below_the_header_frontier` — real RocksDB
  fixture: a finalized target below the header frontier acquires a lease, serves one complete
  page ending at the target, and still returns `NoLocatorIntersection` for a locator at or above
  the target and `TargetNotRetained` for an unknown hash.
- `vct_repair_accepts_a_supplier_ahead_of_the_stalled_branch` — the outage shape: the peer
  advertises a different selected tip, ahead of us, with `oldest_retained_height` above the
  repair predecessor. The repair request is now sent.
- `vct_repair_defers_when_no_peer_reaches_the_repair_height` — a peer below the repair height
  produces no wire request. The test captures the trace and asserts the `no_supplier` row's
  rejection tally, rather than only asserting the absent request.
- `vct_aux_selection_prefers_authenticated_complete_nonrejected_provenance` — updated to assert
  a missing successor record polls the open episode.

`cargo test -p zakura-state --lib` (519 passed) and `cargo test -p zakura-network` (1093 passed)
are green; `cargo clippy --all-targets` and `cargo fmt --check` on both crates are clean.

## Not included

A full two-node cluster end-to-end test — supplier ahead on a different branch, target and roots
only in supplier finalized state, parked commit resumes, repeated 500 ms polls leave the repair
generation unchanged — is not in this PR. The testkit has no VCT-stalled node fixture, so it is a
substantially larger piece of work than the fix. The seams it would cover are each covered by the
unit and reactor tests above.

## Validation still owed

The two wedged databases should resume in place on a fixed binary: the parked block and its
header state are intact, and the repair simply becomes assignable. That is not confirmed yet —
both nodes are still down (`Restart=no`). Opening for code review while that runs; please treat
the in-place resume as an open item on the PR rather than a blocker on reviewing the change.

### DIFF
--- a/crates/zakura-network/src/zakura/header_sync/reactor.rs
+++ b/crates/zakura-network/src/zakura/header_sync/reactor.rs
@@ -45,7 +45,7 @@ const VCT_REPAIR_RETRY_INTERVAL: std::time::Duration = std::time::Duration::from
 const VCT_REPAIR_NO_SUPPLIER_TRACE_INTERVAL: std::time::Duration =
     std::time::Duration::from_secs(10);
 /// Time one repair generation may stay unassignable before the reactor reports it at error level.
-const VCT_REPAIR_NO_SUPPLIER_WARN_AFTER: std::time::Duration = std::time::Duration::from_secs(60);
+const VCT_REPAIR_NO_SUPPLIER_REPORT_AFTER: std::time::Duration = std::time::Duration::from_secs(60);
 /// Minimum interval between unchanged header-snapshot refresh trace rows.
 ///
 /// The trace records every frontier advance and reanchor.
@@ -1551,7 +1551,19 @@ impl HeaderSyncReactor {
                     }
                 }
                 HeaderTargetAdmissionResult::ResourceStalled(_) => {
-                    self.vct_repair.remove(repair_owner);
+                    // A local resource alarm stalled the commit, so nothing judged the delivery.
+                    // Back the task off instead of dropping it: the committer no longer bumps the
+                    // repair generation while it polls, so a dropped task would wait for an
+                    // unrelated snapshot change that a stalled node has no way to produce.
+                    let deferred = self.vct_repair.get_mut(repair_owner).is_some_and(|task| {
+                        task.retry(source).is_ok()
+                            && task
+                                .defer_retry_until(Instant::now() + VCT_REPAIR_RETRY_INTERVAL)
+                                .is_ok()
+                    });
+                    if !deferred {
+                        self.vct_repair.remove(repair_owner);
+                    }
                     metrics::counter!("sync.header.vct.repair.resource_stalled.total").increment(1);
                 }
             }
@@ -1764,28 +1776,7 @@ impl HeaderSyncReactor {
         outcome: Option<&'static str>,
     ) {
         self.startup.trace.emit_with(HEADER_SYNC_TABLE, |row| {
-            row.insert(
-                hs_trace::EVENT.into(),
-                hs_trace::HEADER_VCT_REPAIR_STATE.into(),
-            );
-            insert_header_scope(row, task.owner.header);
-            row.insert(hs_trace::SESSION_ID.into(), task.owner.session_id.into());
-            row.insert(
-                hs_trace::REQUEST_ID.into(),
-                task.owner.request_id.get().into(),
-            );
-            row.insert(hs_trace::HEIGHT.into(), u64::from(task.height.0).into());
-            row.insert(
-                hs_trace::REPAIR_GENERATION.into(),
-                task.repair_generation.into(),
-            );
-            row.insert(hs_trace::PHASE.into(), phase.into());
-            row.insert(
-                hs_trace::SUPPLIER_COUNT.into(),
-                u64::try_from(task.tried_sources.len())
-                    .unwrap_or(u64::MAX)
-                    .into(),
-            );
+            insert_vct_repair_identity(row, task, phase);
             row.insert(
                 hs_trace::OUTCOME.into(),
                 outcome.map_or(serde_json::Value::Null, Into::into),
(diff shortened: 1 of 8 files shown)


evan-forbes:

PR #803, opened 2026-08-26
fix(sync): retry transient block failures in place

evan-forbes:
## Summary

- retry one required block after a transient peer or transport failure
- preserve the other download and checkpoint verification tasks in the sync round
- restart the round when the same hash exhausts its bounded retry budget

## Incident evidence

The preserved `temp-zakura-sync-test-5` database reproduced a legacy sync round failure after restart. One block request ended with:

`SharedPeerError(ClientDropped, Other)`

The syncer treated that request failure as a round-wide error. It canceled 1,311 other tasks while checkpoint commits were still advancing. The next round started from checkpoint height 3,428,356 and continued syncing, which showed that the selected chain and checkpoint verifier remained valid.

This failure uses the legacy block sync path. PR #800 changes the Zakura header and VCT repair path, so it does not cover this failure.

Diagnostic run: https://github.com/zakura-core/zakura/actions/runs/32989882566

## Tests

- `cargo test -p zakura transient_download_failure --lib`
- `cargo test -p zakura not_found_download --lib`
- `cargo test -p zakura registry_miss_ --lib`
- `cargo test -p zakura components::sync::tests::vectors --lib`
- `git diff --check`

### DIFF
--- a/crates/zakurad/src/components/sync.rs
+++ b/crates/zakurad/src/components/sync.rs
@@ -94,6 +94,14 @@ fn block_error_peer_label(error: &BlockDownloadVerifyError, expose_peer_addresse
 /// accountability in `zakura-network` disconnects such peers so this bound is rarely reached.
 const MISSING_BLOCK_DOWNLOAD_RETRY_LIMIT: usize = 8;
 
+/// Controls how many times the syncer requeues a required block hash after a transient peer or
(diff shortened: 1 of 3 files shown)


evan-forbes:
Superseded by consolidated PR #819, which targets main and includes the audited fixes and regression tests from this PR.

PR #804, opened 2026-08-26
fix(sync): try every VCT repair supplier

evan-forbes:
## Summary

- retain every failed supplier for the current VCT repair cycle
- try a fresh fourth supplier after three deterministic failures
- clear the complete supplier set only when the cycle backs off

## Incident evidence

The PR 800 preserved-state validation connected two upgraded failed nodes to each other. One upgraded node completed a repair while it was among the first three connected suppliers. After three old suppliers connected, both nodes repeatedly retried only those same suppliers. Each old supplier returned `TargetNotRetained`. The upgraded fourth supplier remained connected and eligible but never received a repair request.

This PR stacks on #800 because it fixes the supplier rotation exposed by the PR 800 validation.

## Tests

- `cargo test -p zakura-network supplier_cycle_tries_a_fresh_peer_after_three_failures --lib`
- `cargo test -p zakura-network supplier_cycle_retains_every_distinct_source_until_backoff --lib`
- `cargo test -p zakura-network zakura::header_sync --lib`
- `git diff --check`

### DIFF
--- a/crates/zakura-network/src/zakura/header_sync/reactor.rs
+++ b/crates/zakura-network/src/zakura/header_sync/reactor.rs
@@ -2679,14 +2679,6 @@ impl HeaderSyncReactor {
         let RepairPolicyState::Ready { context } = &task.state else {
             return;
         };
-        if task.supplier_cycle_exhausted() {
-            let _ = self
-                .vct_repair
-                .get_mut(task.owner)
-                .expect("the ready repair was cloned above")
-                .defer_retry_until(now + VCT_REPAIR_RETRY_INTERVAL);
-            return;
-        }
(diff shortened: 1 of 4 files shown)


evan-forbes:
Preserved-state validation on the original failed test-1 and test-2 databases confirms that this PR fixes supplier starvation. Test-1 admitted 14 repairs and test-2 admitted 2 repairs after an upgraded peer became reachable beyond the original three failed suppliers.

The nodes then exposed a separate state-side priority inversion: an older authentication-sweep repair masked each next committer repair while the nonempty checkpoint queue prevented the sweep from running. PR #806 fixes that follow-on defect. The VCT recovery merge order is #800, #804, then #806.

evan-forbes:

PR #806, opened 2026-08-26
fix(state): prioritize VCT committer repairs

evan-forbes:
## Summary

- prioritize a missing committer root over an older authentication-sweep repair
- resume the sweep repair after the blocked checkpoint commits
- cover the observed higher-committer, lower-sweep priority inversion

## Preserved-state evidence

PR #804 let both preserved nodes accept repairs. Each node then committed one or more blocks and stalled again. The active sweep repair remained below the next committer repair. The writer only runs the sweep after the checkpoint queue becomes empty, so the sweep could not clear while the committer kept that queue nonempty.

## Validation

- cargo test -p zakura-state service::write --lib
- cargo fmt --all -- --check
- git diff --check

Stacked on #804.

### DIFF
--- a/crates/zakura-state/src/service/write/vct_write_retry.rs
+++ b/crates/zakura-state/src/service/write/vct_write_retry.rs
@@ -233,20 +233,17 @@ impl VctWriteRetryManager {
         self.publish_effective_repair_status(None);
     }
 
-    /// Publishes the lowest outstanding repair height.
+    /// Publishes the repair that can unblock the writer.
     ///
-    /// The repair channel stores one latest value. The manager publishes the lower height because
-    /// its replacement unblocks the higher repair.
+    /// The repair channel stores one latest value. A committer repair takes priority because the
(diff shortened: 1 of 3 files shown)


evan-forbes:
Preserved-state validation at head a536286908ea32d319d3a801bf3dd0a91ccccfb9 confirms the priority fix.

- Test 1 advanced finalized state from 3452358 to 3452477 while its authentication-sweep frontier remained at 3452371.
- Test 2 advanced finalized state from 3452168 to 3452477 while its authentication-sweep frontier remained at 3452168.
- Both peer services remained active and recorded no panics.
- Both nodes then requested root 3452479 and reported no supplier. The isolated pair does not contain that root, so this final boundary reflects the validation topology rather than the prior priority deadlock.

The committer repair therefore ran ahead of the older sweep repair and advanced the preserved checkpoint queues by 119 and 309 blocks. I stopped the transient validation services after capturing the final metrics so the updated states remain available for further inspection.

evan-forbes:

PR #807, opened 2026-08-26
fix(block-sync): keep retry-filtered work nonempty

evan-forbes:
## Summary

- Evaluate shared retry avoidance once for each selected work item.
- Keep the first nonempty contiguous allowed run.
- Prevent peer routines from indexing an empty work list after retry filtering.

## Production evidence

Canada, Europe West, and Asia South reported the same peer-routine panic at line 944. The nodes recorded 1, 3, and 6 occurrences respectively before block sync stopped issuing requests.

## Validation

- cargo test -p zakura-network zakura::block_sync
- cargo fmt --all
- git diff --check


evan-forbes:
Superseded by consolidated PR #818, which targets main and includes the audited fixes and regression tests from this PR.

PR #808, opened 2026-08-26
test(block-sync): harden retry filter boundaries

evan-forbes:
## Summary

- Test every retry-filter decision pattern through six items.
- Verify the predicate evaluates each visited item once.
- Exercise the production drain and split retention geometry.
- Guard request construction against an unexpected empty filtered range.
- Name the reactor and sequencer as the concurrent retry-state writers.

## Stack

- Depends on #807.
- Targets `fix/empty-retry-filter` so #807 remains unchanged.

## Validation

- `cargo test -p zakura-network zakura::block_sync` (278 passed)
- `cargo test -p zakura-network` (1,092 passed, 3 ignored; acceptance test passed)
- `git diff --check`


evan-forbes:
Superseded by consolidated PR #818, which targets main and includes the audited fixes and regression tests from this PR.

PR #811, opened 2026-08-26
test(sync): verify transient retry preserves sync round

evan-forbes:
## Motivation

PR #803 changes a transient block failure from a round-wide restart into a per-hash retry. Its focused tests did not prove that unrelated work survives the failure or that a successful replacement clears retry state.

## Solution

Add a two-hash test through the production `sync_round` path. The test exhausts the service-level retry layer for one hash while another hash downloads and commits. It then lets the queue-level replacement succeed and verifies that the original round completes with no transient retry state or downloader task left behind.

## Testing

- `cargo test -p zakura transient_download_failure --lib`
- `cargo test -p zakura components::sync::tests::vectors --lib`
- `cargo test -p zakura components::sync::tests --lib`
- `git diff --check`

The new test sends four `ClientDropped` failures to exhaust the configured inner retry limit. It commits the unrelated block before answering the replacement request. It then commits the replacement and checks round completion and retry-state cleanup.

## Changelog

This test-only change has no user-visible changelog entry.

- Builds on #803.

### Follow-up Work

The combined inner retry, hedge, and queue-level retry layers can still amplify requests for a persistently failing hash. This PR does not change that policy.

### DIFF
--- a/crates/zakurad/src/components/sync/tests/vectors.rs
+++ b/crates/zakurad/src/components/sync/tests/vectors.rs
@@ -2226,6 +2226,107 @@ async fn transient_download_failure_restarts_after_retry_limit() {
     peer_set.expect_no_requests().await;
 }
 
+/// A transient failure for one hash preserves other work in the sync round and retries only the
+/// failed hash.
+#[tokio::test]
+async fn transient_download_failure_preserves_sync_round() -> Result<(), crate::BoxError> {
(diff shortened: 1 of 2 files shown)


evan-forbes:
Superseded by consolidated PR #819, which targets main and includes the audited fixes and regression tests from this PR.

PR #818, opened 2026-08-26
fix(block-sync): keep retry-filter work nonempty

evan-forbes:
## Summary

- evaluate each retry-filter item once
- retain the first nonempty allowed run
- guard later first-item access in release builds
- exhaustively test every Boolean mask through six items

This PR consolidates and supersedes #807 and #808.

## Validation

- cargo test -p zakura-network: 1,092 passed, 3 ignored
- acceptance test: 1 passed
- doc tests: passed
- git diff --check: passed


ValarDragon (approved):
Good fix!

PR #817, opened 2026-08-26
fix(sync): retain VCT local operation ownership

evan-forbes:
## Motivation

PR #815 correctly attributes VCT repair failures, but its diagnostic timeout retired a still-pending local preparation or application. The reactor could then download the repair again and start another local operation while the first future remained live.

## Solution

- Keep the repair task and peer work assigned while a local preparation or application remains pending.
- Rearm a diagnostic deadline and feed pending local work into the existing generation-scoped stall escalation.
- Preserve that ownership across supplier disconnects and session replacement.
- Retire ownership when the repair generation becomes obsolete, then ignore the late completion.
- Record aggregate local-capacity stalls without charging a supplier attempt.

## Tests

- `cargo test -p zakura-network`
- `cargo test -p zakura-state vct_ --lib`
- `cargo clippy -p zakura-network --lib --tests -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

The direct-port tests hold real preparation and application futures pending across several diagnostic intervals. They verify that the reactor keeps exactly one future and one local call until completion or generation retirement.

Stacked on #815.

PR #815, opened 2026-08-26
fix(sync): preserve VCT retry attribution

evan-forbes:
## Motivation

PR #810 bounded VCT supplier cycles, but reactor retry paths still inferred supplier blame from broad terminal outcomes. Admission failures also recreated the repair task and discarded its policy state. Source-ID sorting let newly inserted identities move ahead of established suppliers, and the stall escalation depended on unrelated maintenance deadlines.

## Solution

This PR adds explicit local and supplier retry attribution at each VCT reactor boundary. It preserves repair context, attempt counts, and tried suppliers across admission failures. Receiving timeouts and unavailable ordered sends rotate suppliers, while Preparing, Applying, and encoding failures keep the supplier eligible under local backoff.

The scheduler now uses one bounded admission-order queue for authenticated peers. Supplier failures rotate one admitted peer to the back. Replacements keep their position, disconnects remove their entry, and new identities enter behind established peers.

The repair-generation stall record now owns its trace and escalation deadlines. It records typed local or supplier outcomes in traces and error diagnostics. Resource-stalled admissions use the exact `resource_stalled` terminal label.

## Testing

- `cargo test -p zakura-network`
- `cargo test -p zakura-state vct_ --lib`
- `cargo clippy -p zakura-network --lib --tests -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

The new tests cover adversarial identity insertion and removal, replacement sessions, bounded send-failure cycles, timeout attribution by phase, admission-failure state preservation, ordered-send attribution, the resource-stalled label, and the independent escalation deadline.

## Changelog

Added `docs/changelog/unreleased/815.md`.

### Specifications & References

- Stacked on #813.
- Corrects the remaining scheduler and reactor blockers found while auditing #810.

### Follow-up Work

No known merge-blocking follow-up remains in this scope.

PR #813, opened 2026-08-26
fix(state): prioritize VCT committer repair

evan-forbes:
## Summary

- prioritize a missing committer root over an older authentication-sweep repair
- restore the sweep repair after the blocked checkpoint commits
- cover production missing-root and reactor replacement behavior

## Design

The state writer publishes one effective VCT repair. The committer repair takes priority because the authentication sweep runs only after the checkpoint queue becomes empty. The patch keeps the existing single-slot design and changes only its arbitration rule.

## Validation

- cargo test -p zakura-state a_missing_committer_root_preempts_a_lower_sweep_need --lib
- cargo test -p zakura-state --lib service::write
- cargo test -p zakura-network committer_repair_replaces_and_then_restores_a_sweep_repair --lib
- cargo test -p zakura-network --lib zakura::header_sync
- cargo fmt --all -- --check
- git diff --check

Stacked on #810. This PR replaces #806 without changing it.

PR #810, opened 2026-08-26
fix(sync): bound VCT supplier rotation

evan-forbes:
## Summary

- keep each VCT repair cycle bounded to three distinct authenticated suppliers
- preserve a round-robin SourceId cursor so the next cycle starts after the last failed supplier
- keep generation stall evidence across failed supplier and ordered-send cycles
- back off local failures without rejecting the supplier or clearing the current supplier cycle

This replaces the unbounded supplier-set design in #804. It keeps the identity-churn bound while fixing deterministic starvation of a fourth supplier.

## Tests

- cargo test -p zakura-network bounded_supplier_cycles_rotate_to_the_fourth_supplier --lib
- cargo test -p zakura-network send_failures_preserve_the_stall_clock_across_bounded_cycles --lib
- cargo test -p zakura-network replacement_session_keeps_the_authenticated_supplier_identity --lib
- cargo test -p zakura-network local_retry_preserves_supplier_eligibility_and_cursor --lib
- cargo test -p zakura-network supplier_cycle_bounds_identity_churn_and_preserves_cursor --lib
- cargo test -p zakura-network zakura::header_sync --lib
- cargo fmt --all -- --check
- git diff --check

PR #821, opened 2026-08-26
fix(sync): complete VCT repair liveness

evan-forbes:
## Problem

- VCT auxiliary repair could livelock at one height, starve exact-height serving, or keep the node alive indefinitely while a selected local operation remained pending.

## Solution

- Complete exact-height VCT auxiliary serving and repair while preserving supplier attribution and local backoff.
- Rotate bounded supplier cycles and restart same-height repair after a rejected delivery.
- Prioritize checkpoint committer repairs, then restore authentication-sweep repairs after commit.
- Reserve serving capacity for one exact finalized fallback header so general leases cannot starve VCT repair.
- Terminate the node with an error when a selected VCT repair prepare or apply operation remains pending for 30 minutes.

## Design

The reactor owns one exact repair requirement. The requirement carries its repair generation, body authority, branch, and operation phase. Supplier failures rotate the supplier. Local dispatch and operation failures preserve the supplier and retry the same operation.

The writer publishes a committer repair while a checkpoint remains parked. It retains the authentication-sweep need and republishes it after commit. A typed repair trigger distinguishes an observed missing root from a rejected delivery. Missing-root polls share one episode. Each rejected delivery can advance the generation once and fetch a replacement at the same height. The state adapter reserves one retained-path slot for an exact one-header finalized fallback. General leases cannot consume this slot, and finalized fallback leases cannot renew into long paths.

Selected VCT repair prepare and apply operations have a fixed 30-minute local deadline. The reactor retains the original operation future until completion or the deadline. A biased selection gives completion priority when both become ready together. The existing 30-second diagnostics and one-minute VCT stall error remain unchanged. Normal header operations do not use this deadline.

At the deadline, the reactor emits one error with the phase, owner, repair generation, branch, target, and elapsed time. It increments `sync.header.vct.repair.local_operation_fatal.total` with a `phase` label. It sends one private fatal event to the node root and starts no replacement state operation. The node root returns an error and starts normal shutdown. Shutdown drops the endpoint and the original operation.

Production systemd services use `Restart=always`, so they restart after this fatal exit. Continuous-sync services use `Restart=no`, so their controller records the bounded failure instead of an indefinite height stall.

## Authorship and superseded PRs

Evan Forbes authored the consolidated work across #800, #804, #806, #810, #813, #815, and #817. This PR supersedes those PRs. It does not close or modify them.

## Validation

- `cargo test -p zakura-network --lib` (1116 passed, 2 ignored)
- paused-time VCT deadline tests for prepare, apply, singular ownership, completion priority, and non-VCT exclusion
- rejected-delivery tests for same-height refetch and generation deduplication
- finalized-path tests for reserved capacity, one-header completion, and non-renewal
- `cargo test -p zakura-state --lib` (536 passed, 3 ignored)
- node-root fatal event shutdown and error test
- `cargo clippy -p zakura-network -p zakura-state -p zakura --all-targets -- -D warnings`
- `cargo clippy --workspace --all-targets --features default-release-binaries`
- `cargo xtask check-rpc-artifacts`
- `cargo xtask header-conformance`
- `git diff --check`
- all required GitHub checks


evan-forbes:
@v12-auditor

evan-forbes:
as I was fixing things over today, the testing scope creeped a lot so trying to reduce that without reducing coverage

ValarDragon (approved):

PR #820, opened 2026-08-26
fix(ops): complete fleet stall alert handling

evan-forbes:
## Motivation

Synchronized fleet stalls produced duplicate node alerts and lacked bounded, atomic diagnostics. Later ownership transitions could lose or duplicate alerts when node membership, tip identity, or Slack delivery changed. Whole-probe failures also reported metrics as available, and prefix-only payload truncation could remove the dashboard link.

This PR consolidates and replaces the effective changes from #802, #805, #809, #812, #814, and #816.

## Solution

- identify shared stalls by height and validated block hash
- maintain one alert owner across node, shared, dashboard, and Slack failure transitions
- preserve conservative membership timer boundaries and legacy state compatibility
- attach atomic, bounded sync and VCT diagnostics to fleet rows
- mark whole-probe and timestamp-free metric samples unavailable
- sanitize and bound dynamic alert fields
- truncate payloads on line boundaries while preserving the incident prefix and dashboard link
- keep a shared owner latched when malformed rows cannot prove recovery

## Testing

- `python3 deploy/runner/test_zakura_cluster_watchdog.py` (59 passed)
- `python3 -m unittest discover -s deploy/runner -p "test_zakura_cluster_watchdog.py"` (59 passed)
- `python3 deploy/runner/test_zakura_cluster_status.py` (57 passed)
- `python3 -m unittest discover -s deploy/runner -p "test_zakura_cluster_status.py"` (57 passed)
- `python3 -m py_compile deploy/runner/zakura-cluster-status.py deploy/runner/zakura-cluster-watchdog.py deploy/runner/test_zakura_cluster_status.py deploy/runner/test_zakura_cluster_watchdog.py`
- `systemd-analyze verify deploy/runner/zakura-fleet-watchdog.service`
- `actionlint .github/workflows/zakura-mainnet-deploy.yml`
- `git diff --check origin/main...HEAD`

The tests cover shared-tip identity, threshold skew, membership changes, state migration, failed Slack transitions, dashboard outages, whole-probe failures, malformed rows, bounded diagnostics, and oversized formatted payloads.

## Changelog

No Rust source file or Cargo manifest changed.

### Follow-up Work

A versioned fleet diagnostic contract could replace the duplicated metric-name maps in a later PR.

PR #823, opened 2026-08-27
fix(sync): complete legacy fallback checkpoint handoff

evan-forbes:
## Motivation

The legacy fallback handoff can wait forever on an incomplete checkpoint range. Native block sync submits checkpoint bodies to the shared checkpoint verifier. The verifier holds each response until every body in the checkpoint range arrives. When fallback starts, the coordinator stops new native admission but treats every accepted checkpoint response as too late to cancel. The coordinator then waits for those responses before it authorizes legacy sync. Legacy sync cannot supply the missing body until it receives that authorization.

This circular wait left five Mainnet fleet nodes without block verification progress. The affected nodes retained 18–401 native applies after the VCT repair restored state progress.

## Solution

- transfer incomplete checkpoint-range responses when fallback starts draining native applies
- keep the shared checkpoint verifier responsible for transactional checkpoint commits
- let legacy sync supply the missing bodies through that same verifier
- notify the native block-sync reactor that transferred submissions were abandoned
- record transferred operations with an explicit lifecycle state and metric
- keep full semantic commits behind the existing strict drain barrier

The completion branch has priority when a checkpoint response and the transfer signal become ready together. A completed response therefore retains its committed or rejected result.

## Testing

- `cargo test -p zakura --lib legacy_fallback_completes_transferred_checkpoint_range`
- `cargo test -p zakura --lib fallback_yield_releases_queued_submit_blocks`
- `cargo test -p zakura --lib buffered_native_commit_blocks_legacy_fallback_past_every_diagnostic_interval`
- `cargo test -p zakura --lib commands::start::zakura::coordinator::tests`
- `cargo clippy -p zakura --lib --tests -- -D warnings`
- `git diff --check`

The regression test uses a real checkpoint verifier and an ephemeral state. It submits every block in a checkpoint range except one body. It proves that fallback acquires the handoff while the range remains incomplete. It then submits the missing body through the shared legacy verifier and verifies that the finalized tip reaches the checkpoint.

The full `cargo test -p zakura --lib` run passed 410 tests. `components::inbound::tests::real_peer_set::inbound_pruned_block_is_not_advertised_and_getdata_logs_error` failed its pruning-log assertion. PR #821 reports the same failure at its unmodified base, and this PR does not change inbound pruning or logging.

## Changelog

`docs/changelog/unreleased/823.md` records the operator-visible fix.

This bug is separate from the VCT root-repair livelock fixed by #821. The fleet validation branch `validate/sync-fixes-complete-20260826` combined #818, #819, #820, and #821. That validation cleared the VCT repair backlog and exposed the checkpoint handoff residue.

- #818 keeps block-sync retry-filter work nonempty.
- #819 retries transient legacy block failures in place. It does not break this handoff cycle because legacy sync cannot start.
- #820 reports the resulting fleet stalls. The alerts predate the validation deployment and remain accurate.
- #821 fixes VCT repair liveness. It does not cause or fix this checkpoint handoff cycle.

### Follow-up Work

Validate the fix on the five affected fleet nodes and confirm that `zcash_chain_verified_block_height` resumes.

### DIFF
--- a/crates/zakurad/src/commands/start.rs
+++ b/crates/zakurad/src/commands/start.rs
@@ -4943,18 +4943,17 @@ mod zakura_header_sync_driver_tests {
         reactor_task.abort();
     }
 
-    /// Drives the block-sync apply loop against the *real* checkpoint verifier and a *real*
-    /// ephemeral state, reproducing the checkpoint-range batch-commit that Regtest (genesis
-    /// checkpoint only) cannot exercise.
+    /// Drives the block-sync apply loop against the real checkpoint verifier and an ephemeral
+    /// state. This test reproduces the checkpoint-range batch commit that Regtest cannot exercise.
     ///
     /// A checkpoint is placed at height 10 so an 11-block range covers a full checkpoint gap
     /// without 400 real blocks. The whole range is submitted except one mid-range body, which
     /// the verifier holds the entire range for (it commits nothing until the range is
-    /// contiguous to the next checkpoint). Delivering the withheld body must let the whole
-    /// range commit — i.e. a transiently-missing body recovers instead of wedging the floor,
-    /// which is the production "drop-through" failure mode.
+    /// contiguous to the next checkpoint). The fallback handoff must transfer the held requests
+    /// to the shared verifier. The legacy driver can then deliver the withheld body and commit the
+    /// range.
     #[tokio::test]
-    async fn block_sync_driver_recovers_checkpoint_range_after_withheld_body() {
+    async fn legacy_fallback_completes_transferred_checkpoint_range() {
         const CHECKPOINT_HEIGHT: u32 = 10;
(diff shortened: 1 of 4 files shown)


crates/zakurad/src/commands/start.rs
```diff
@@ -5060,30 +5069,44 @@ mod zakura_header_sync_driver_tests {
...
             "checkpoint range must not commit while a mid-range body is missing",
         );
 
-        // Deliver the withheld body; the verifier can now commit the whole range.
-        let (withheld_height, withheld_block) = chain
+        let fallback = tokio::spawn(async move {
+            handoff
+                .acquire_legacy_fallback(Duration::from_secs(1))
+                .await
+        });
```
ValarDragon, line 5093: wow why does this logic appear in start.rs

PR #824, opened 2026-08-27
feat: bump libraries crates to 1.0.0-rc.4 and adopt their new optimizations

ebfull:
## Motivation

The `zakura-core/libraries` crates shipped `1.0.0-rc.4` with performance work that was not well cataloged in their changelogs: a position-weighted fixed-length Sinsemilla evaluator, and opt-in prepared-MSM state (the `orbits` feature's prepared fixed-base zero-check and commitment tables) exposed through ungated arming entry points. Zakura should track the new release and pick up the optimizations that apply to a validator.

## Solution

Three commits, one per concern:

- **`build(deps)`**: bump the 12 workspace dependencies published from `zakura-core/libraries` from `1.0.0-rc.2` to `1.0.0-rc.4` (plus the 5 transitive crates in `Cargo.lock`), and bump the matching exact-version cargo-vet exemptions. rc.4 drops four transitive dependencies (`home`, `known-folders`, `num_cpus`, `xdg`) via the `zcash_proofs` default-feature change.
- **`perf(chain)`**: switch `MerkleCRH^Orchard` in `zakura-chain` to the new weighted fixed-length Sinsemilla evaluator, packing the fixed 52-word message directly into 10-bit words. This halves the per-node hash cost that dominates Orchard note-commitment-tree updates during sync (`orchard_combine`: 14.76 µs → 7.5 µs, −48–49%, with an unchanged Sapling/Pedersen control), for a one-time 3.75 MiB table and ~28 ms of lazy construction. The evaluator omits Sinsemilla's incomplete-addition exceptional-case checks; an input where that changes the output would exhibit a discrete-log relation between the Sinsemilla bases (see the security argument in `sinsemilla::weighted`).
- **`feat`**: arm halo2's prepared-MSM state on both sides of the process, unconditionally (the entry points are documented no-ops while `orbits` is off, which remains our policy): each of the three Orchard circuit-era verifying-key statics arms `prepare_batch_validation()` at construction, and the miner's shielded-coinbase path arms `prepare_proving()` on `zcash_primitives`' process-wide cached proving key when a shielded reward output is added. If `orbits` is ever enabled, verification and coinbase proving pick up the prepared paths with no further changes.

Everything else perf-relevant in rc.4 (pasta `multicore`, `deferred`, `glv`, `aarch64-asm`, halo2 `batch`) is already active through default features or halo2's own dependency declarations; `orbits` stays deliberately off, and this PR resolves it off everywhere (verified with `cargo tree -e features`).

## Testing

- Weighted Merkle CRH: existing differential tests now pin the weighted path against this crate's independent from-scratch Sinsemilla implementation across all 32 layers, edge-case and random full-width field elements (proptest), plus the fixed consensus vectors for empty roots and incremental anchors. Measured with `benches/note_commitment_hash.rs` (criterion baselines, Pedersen control).
- Prepared-MSM arming: new `prepared_msm_arming_does_not_change_proving_or_verification` test pins parity in both directions — a real mainnet pre-NU6.2 proof is accepted by its own era's key and rejected by the wrong era's key identically under armed and unarmed keys, and proofs produced before and after arming a test `ProvingKey` verify under both. Repeat arming is asserted idempotent. The `#[ignore]`d `shielded_coinbase_paths` test exercises the miner arming hook with the real prover and passes.
- Workspace: `cargo test --workspace` passes (all 43 test binaries, 0 failures) with two groups of environment-only skips on the local macOS dev machine: the live-network acceptance tests that CI's default nextest filter also excludes (`sync_one_checkpoint_mainnet`, `activate_mempool_mainnet`, `restart_stop_at_height`, etc. — they run in the dedicated network-enabled integration jobs), and three `zakura-network` listener tests that bind 127.0.0.x loopback aliases (Linux-only; that crate is untouched by this PR). Without the skips, exactly those tests fail locally for environmental reasons; nothing else.
- `cargo check --workspace --all-targets --locked`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `cargo vet check`, and `./scripts/changelog.py check` all pass. Release `zakurad` binary size is unchanged (+1,744 bytes, +0.0014%).

## Changelog

`docs/changelog/unreleased/824.md` added (Changed: dependency update, Merkle hashing speedup).

### Follow-up Work

- `../libraries`: share one process-wide `Params` across Orchard key builds and arm `cached_orchard_proving_key` at init (prompt prepared for a libraries-side agent). Once that ships in the next rc, the `arm_shielded_reward_proving_key` hook in `zakura-rpc` becomes redundant and should be removed.
- The `sinsemilla::weighted` batch APIs (`hash_words_batch_with_workspace`) could amortize inversions across nodes in the `append_batch`/`batch_frontier` path for a further gain beyond this drop-in swap.

### DIFF
--- a/crates/zakura-chain/src/orchard/tree.rs
+++ b/crates/zakura-chain/src/orchard/tree.rs
@@ -17,7 +17,6 @@ use std::{
     io,
 };
 
-use bitvec::prelude::*;
 use halo2::pasta::{group::ff::PrimeField, pallas};
 use hex::ToHex;
 use incrementalmerkletree::{
@@ -28,7 +27,7 @@ use lazy_static::lazy_static;
 use thiserror::Error;
 use zcash_primitives::merkle_tree::HashSer;
 
-use sinsemilla::HashDomain;
+use sinsemilla::{weighted::UncheckedFixedLengthHashDomain, HashDomain, K};
 
 use crate::{
     serialization::{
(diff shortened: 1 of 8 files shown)


crates/zakura-chain/src/orchard/tree.rs
```diff
@@ -48,15 +47,42 @@ pub type NoteCommitmentUpdate = pallas::Base;
...
 
+/// Bits in one Merkle child encoding: a 255-bit little-endian Pallas base
+/// field element (`l_MerkleOrchard` in the protocol spec).
+const L_ORCHARD_MERKLE: usize = 255;
+/// Bits in one `MerkleCRH^Orchard` message: the 10-bit layer prefix followed
+/// by the left and right child encodings.
+const MERKLE_CRH_BITS: usize = K + 2 * L_ORCHARD_MERKLE;
+/// `MerkleCRH^Orchard` inputs always fill this many Sinsemilla words exactly,
+/// which is what lets the fixed-length weighted evaluator apply.
+const MERKLE_CRH_WORDS: usize = MERKLE_CRH_BITS / K;
```
ValarDragon, line 58: Note to future selves, we should go push these constants into Orchard / Sinsemilla

crates/zakura-chain/src/orchard/tree.rs
```diff
@@ -74,21 +100,58 @@ lazy_static! {
...
+    ORCHARD_MERKLE_CRH_DOMAIN.hash_words(&merkle_crh_words(layer, left, right))
+}
+
+/// Packs a `MerkleCRH^Orchard` input into its 10-bit Sinsemilla words.
+///
+/// The message is `I2LEBSP_10(l) || left || right` with 255-bit little-endian
+/// child encodings, so the words are: the layer prefix `l`, 25 complete words
+/// of `left`, one word spanning `left`'s top 5 bits and `right`'s low 5 bits,
+/// and 25 words covering the remaining bits of `right`.
+fn merkle_crh_words(layer: u8, left: pallas::Base, right: pallas::Base) -> [u16; MERKLE_CRH_WORDS] {
```
ValarDragon, line 112: This brings up a good TODO, need to work-out parallel API's + handling imbalanced trees

ValarDragon: Actually im confused why were doing any of this packing, should have been happening in sinsemilla I thought

ValarDragon: oh its because the function in orchard is private, got it

ValarDragon (approved):
LGTM, we will do some sinsemilla API cleanup + re-working of API's to get the full 20x (mac-OS) gain exposed here

PR #831, opened 2026-08-27
fix(sync): retain full-state header forks

evan-forbes:
problem:
- Native block sync can keep a side fork in full state after the header DAG evicts the matching header branch.
- The next combined commit then fails with `StagedPathMismatch`, queued descendants remain pending, and legacy fallback can wait forever for the apply drain.

solution:
- Retain every full-state fork tip in the header DAG and reserve one additional candidate slot for an independent selected header tip.
- Restore and verify all full-state header branches at startup.
- Complete queued and later-arriving descendants when a local ancestor write fails.
- Terminate the node when native applies prevent legacy fallback from acquiring ownership for 30 minutes.

design:
- The full-state writer replaces the retained fork-tip set in the same serialized commit that swaps the non-finalized state.
- A bounded recent-failure index covers descendants that arrive before or after the writer reports the ancestor failure.
- The writer preserves retryable local-context failures and marks only consensus failures or inherited invalid ancestry as invalid.
- The fallback coordinator returns a private fatal error after the hard deadline. The syncer propagates that error to the node supervisor.

testing:
- Header candidate pressure preserves every full-state tip.
- Startup side-path restoration remains idempotent.
- A failed ancestor completes its queued subtree.
- A descendant submitted after an ancestor failure completes immediately.
- Invalid and retryable write failures retain their distinct classifications.
- A pending apply fails fallback acquisition at 30 minutes.
- Apply completion wins at the deadline.
- The syncer propagates the fatal deadline error to the node supervisor.


ValarDragon (approved):
utACK, I have some spec-level concerns to review around the state tracking, but the incremental change seems like a strong improvement

PR #834, opened 2026-08-27
fix(zcashd-compat): allow non-Unix executable no-op

evan-forbes:
## Motivation

The post-merge Windows workspace check fails because make_executable excludes its Unix-only body and leaves path unused. The workspace denies warnings.

## Solution

Prefix the parameter with an underscore and use that binding in the Unix block. Rust then accepts the intentional non-Unix no-op while Unix behavior stays unchanged.

## Testing

- cargo fmt --all -- --check
- cargo check --locked -p zakura --lib --all-features
- git diff --check

The cross-platform workflow will verify the Windows target on this branch.

## Changelog

- docs/changelog/unreleased/834.md records this internal-only build fix.

- Windows failure: https://github.com/zakura-core/zakura/actions/runs/33107995430/job/98642947646

### DIFF
--- a/crates/zakurad/src/components/zcashd_compat/managed.rs
+++ b/crates/zakurad/src/components/zcashd_compat/managed.rs
@@ -354,14 +354,14 @@ fn sha256_hex_file(path: &Path) -> Result<String, Report> {
 /// Makes `path` executable on Unix targets.
 ///
 /// Non-Unix targets currently no-op because managed release targets are Linux.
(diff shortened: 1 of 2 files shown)


evan-forbes:
The branch-specific cross-platform run passed on both Windows and macOS: https://github.com/zakura-core/zakura/actions/runs/33116935533

ValarDragon (approved):

PR #835, opened 2026-08-27
fix(deploy): suppress watchdog alerts during planned restarts

czarcas7ic:
## Motivation

Planned mainnet deploys refreshed only the central fleet watchdog marker, so the node-local compatibility watchdog emitted Sentry failure and recovery events while `zakurad-compat` restarted cleanly.

## Solution

When a restart includes `zakura-compat`, the mainnet workflow now writes that host's existing 20-minute suppression marker and restarts the active watchdog before deploying. The bounded SSH step is best effort and stays in workflow-owned code so deployments of older refs remain covered.


evan-forbes (approved):
I have less context on the alerting but utack

PR #847, opened 2026-08-31
fix(vct)!: require authoritative repair progress

evan-forbes:
## Summary

- Derive each VCT repair episode from the selected target, authentication
  boundary, and every durable semantic payload and outcome.
- Reuse #845's payload-only fingerprint to detect idempotent admission.
- Bind rejected and disputed input to a separate boundary-aware fingerprint.
- Carry the claimed target and episode through network completion.
- Check the claim inside serialized state admission.
- Reject a payload that state already retains, even under a new delivery ID or
  supplier.
- Exclude input that durable state already rejected or disputed.
- Exclude a supplier that already occupies the target's rooted payload slot.
- Rotate suppliers after supplier-attributed failures.
- Retain a failed supplier identity while that supplier remains connected.
- Release disconnected supplier identities so peer churn cannot exhaust repair
  progress.
- Retry local queue and storage failures with bounded backoff without consuming
  a supplier.
- Wait after an auxiliary-capacity refusal until a newer state context reports
  available capacity.

## Recovery Rule

A deterministic failure cannot repeat unless an authoritative coordinate
changes. The selected target, durable episode, semantic input, supplier, or
relevant committed state must change first. A transient local failure can retry
after bounded backoff.

Header sync tries each connected supplier once during an episode. It waits when
no eligible connected supplier remains. It removes failure-only identity state
when a supplier disconnects. A newly connected supplier can then make progress.
Durable state still excludes retained, rejected, and disputed semantic input.
A new state episode or branch replaces the old task.

## State Ownership

State remains authoritative. The repair-context query reads the exact selected
target, authentication boundary, durable outcome rows, state version, and
current auxiliary admission capacity. The network stores that context in the
repair task. State compares the claimed episode with the current durable episode
under the writer lock before it applies the supplied input.

Rejected input remains excluded after restart. A new peer, session, request, or
delivery ID cannot make the same rejected input count as progress. A disputed
pair requires independent input for the current episode. A later authenticated
outcome can remove the valid member from the exclusion set. The dispute does
not permanently reject both inputs.

State also rejects a semantic payload that it already retains. This check
prevents #845's idempotent semantic deduplication from appearing as a successful
repair admission. State rejects a distinct rooted payload from a supplier that
already owns one rooted slot for the target.

An auxiliary-capacity refusal keeps the task state-blocked. A newer committed
snapshot triggers a new context read. Header sync sends no peer request until
state reports that one distinct rooted payload can be admitted.

Old completions fail the target, episode, semantic-payload, or supplier
authority check. They produce no state effects.

## Relationship to #845

This PR includes #845 through the current `main` branch. That PR raises the
per-header limit to 32 semantic payloads. It also retains one copy of each
semantic payload and one rooted payload per supplier. This PR uses those
retained rows as recovery constraints.

The network can discard a duplicate durable payload row because header sync
still knows which connected peers announced the header. A different peer cannot
replace a distinct retained payload with duplicate bytes.

A VCT repair failure does not affect fork choice.

## Scope

This PR does not add the sparse durable requirement schema from the larger VCT
refactor. State already persists the selected target and bounded auxiliary
outcome rows. The committer and authentication sweep reconstruct the repair
reason after restart.

This PR does not add a state-owned replacement transition for a full auxiliary
set. The existing repair deadline remains the terminal bound when no relevant
state change frees capacity.

## Tests

- Complete `zakura-header-chain` library test suite.
- Complete `zakura-network` library test suite.
- Complete `zakura-state` library test suite.
- Restart reconstruction and semantic redelivery under a new delivery ID and
  supplier.
- Same-supplier replacement rejection under #845's rooted-slot rule.
- Stale episode completion with no effects.
- Supplier rotation to every untried connected peer without timed supplier
  reuse.
- Disconnected supplier-history release after the previous 512-entry limit.
- Established supplier priority during adversarial identity churn.
- Local send backoff without supplier consumption.
- State-gated waiting and immediate context refresh after an auxiliary-capacity
  refusal.
- Retained rooted supplier exclusion.
- Warning-denied clippy for all `zakura-network` targets.
- Header-chain conformance and changelog validation.
- `cargo-semver-checks` for `zakura-network` and `zakura-rpc`.

## SemVer

- Bump `zakura-header-chain` from 1.1.0 to 2.0.0 because the public
  `VctRepairContext` shape and `TargetCompletion::SelectedAuxiliaryRepair`
  variant changed.
- Update every workspace production and development dependency requirement to
  2.0.0.
- Keep the repair episode in private `zakura-network` queue state. This avoids a
  breaking change to the public `HeaderTargetPurpose` enum.
- Patch the broken alloc-only import in `tinyvec 1.13.0` inside the semver jobs.
  This patch applies to current and published-baseline rustdoc builds.


ValarDragon (approved):
We need to fix semver + we have this:

  - P2 — Supplier-history exhaustion creates a persistent remote sync stall. The implementation stores up to 512 failed peer identities, then returns before considering any supplier—
    including newly connected honest peers (limit, terminal branch). Disconnects and ordinary negative outcomes consume identities. Because peer identities are inexpensive to generate
    and the full-history state has no retry deadline, an attacker can induce a repair and cycle identities until the checkpoint commit remains parked indefinitely. Recovery requires a
    new repair generation/branch or process restart. Bounded memory should not imply terminal progress: for example, expire disconnected failure-only entries while retaining durable
    rejected-input constraints, or reserve continued attempts for established outbound peers.

evan-forbes:
Addressed both review findings in 51890415e.

The repair scheduler now retains failure-only supplier identities only while the supplier remains connected. It removes disconnected identities before assignment. Durable retained, rejected, and disputed semantic input stays in the state-owned repair context. The fixed 512-identity terminal state no longer exists. Tests cover a full previous-capacity history, adversarial churn, reconnects, and established supplier priority.

The concurrent SemVer commit correctly bumps `zakura-header-chain` to 2.0.0. I removed the new public field from `zakura-network::HeaderTargetPurpose` and bound the episode in private queue state instead. Local `cargo-semver-checks` now passes all 223 checks for both `zakura-network` and `zakura-rpc`.

The SemVer jobs also hit an independent `tinyvec 1.13.0` alloc-only rustdoc failure on current `main`. The workflow now patches that exact source in an isolated Cargo home, so the patch covers both current and published-baseline builds.

PR #851, opened 2026-09-01
feat(consensus): reduce block target spacing to 25 seconds at NU7

evan-forbes:
> [!NOTE]
> **Split:** the ZIP 218 shielded action limits moved to #1027, so this PR now covers only the 25-second target spacing, averaging window, subsidy and halving, difficulty context, and the remaining timing changes. The two PRs are independent and can merge in either order. The limit items below are kept for history.
>
> **Split:** the end-of-support halt, warning heights, and matching `getdeprecationinfo` estimate moved to #1043. That generic timing fix is independently mergeable and should land before this PR so the NU7 spacing automatically feeds it.
>
> The V12 run 7913 findings are addressed in [this comment](https://github.com/zakura-core/zakura/pull/851#issuecomment-5702485352).

## Motivation

The Zcash coinholder poll that closed on 2026-09-14 adopted [ZIP 218](https://zips.z.cash/zip-0218) for NU7 with 99.9% of the vote. ZIP 218 reduces the block target spacing from 75 seconds to 25 seconds. It also adds per-pool shielded action limits.

NU7 has no activation heights yet. The deployment ZIP lists them as TBD for both networks.

## Solution

ZIP 218 is keyed directly to NU7 activation. Every rule checks whether NU7 is active at the relevant height on the selected network. Mainnet and the default Testnet therefore keep current consensus until a deployment PR sets the NU7 activation heights.

The implementation:

- changes the post-NU7 target spacing to 25 seconds;
- changes the difficulty averaging window from 17 to 102 blocks;
- preserves issuance per unit of time by scaling the subsidy by the spacing ratio;
- enforces limits of 330 Orchard and Ironwood actions, 300 Sapling inputs and outputs, and 25 Sprout JoinSplits;
- enforces the global cost limit of 330, with each JoinSplit counting twice;
- applies the limits to block validation, mempool admission, and block-template selection; and
- reserves shielded capacity used by the coinbase transaction before it selects mempool transactions.

The implementation derives halving heights from the complete spacing schedule. On a configured Testnet whose Blossom height is below the slow-start shift, `halving()` matches ZIP 218's formula and the code on `main` in every build.

The state migration backfills the wider difficulty context before it runs the startup consistency check. Every build retains 112 predecessor headers, the maximum context needed by the post-NU7 difficulty adjustment. Existing databases with the old 27-header context are backfilled once at startup.

Zakura already meets or exceeds the ZIP's related node-policy recommendations. It permits reorgs up to 1,000 blocks and retains at least 10,000 blocks. Its native block sync bounds work by bytes and peer capacity. Zakura does not create ordinary wallet transactions, so it has no default wallet expiry delta to change.

### Changes in this update

- **Height-only activation.** The `ZIP218_ENABLED` build flag and the new `nu7` Cargo feature plumbing have been removed. The configured NU7 activation height alone selects the new consensus rules.
- **ZIP 2003 moved out.** This branch no longer carries the ZIP 2003 rule from #856. #959 implements it off `main`. This branch restores `main`'s v4 handling and drops the `Network::v4_deprecation_height` and `Network::is_v4_deprecated` helpers.
- **Ironwood counts against the Orchard limit.** See "Beyond the ZIPs".
- **Halving on configured Testnets.** The era sum no longer clamps each era at the slow-start shift, which changed default-build results when Blossom activates below it.
- **Non-consensus timing follows the 25-second spacing.** These values now read the target spacing at the relevant height, so they change only after NU7:
  - The Testnet `getblocktemplate` minimum-difficulty heuristic allows 2 target spacings instead of a fixed 150 seconds. It also reads the minimum-difficulty gap at the candidate block's height, as consensus does. Before this change, almost every post-NU7 Testnet template would have used minimum difficulty.
  - `getnetworksolps` uses the averaging window at the requested height (102 blocks after NU7).
  - The legacy syncer's lookahead grows 3 times after NU7, as ZIP 218 recommends.
  - The sync stall warning and checkpoint hint use the spacing at the estimated network tip.
  - The 53-second inventory rotation and the 73-second mempool crawler stay the same. Their doc comments explain why.
- **Difficulty context.** Every build retains the 112 predecessor headers needed by the largest averaging window. Older 27-header contexts are backfilled at startup.
- **Versions.** `zakura-consensus` moves to 9.0.0 because the NU7 limit checks add public error variants. The `zakura-chain` 8.0.0 change and its dependent-crate cascade are now inherited from #1007 rather than changed by this PR.

## Beyond the ZIPs

- **Ironwood actions count against the Orchard limit and the global budget.** ZIP 218 names only Orchard actions. NU6.3 (ZIP 258) moves new Orchard-protocol value to the Ironwood pool, so uncounted Ironwood actions would bypass the limits. Zakura counts Orchard plus Ironwood actions against the 330-action Orchard limit and in the 330 global budget.
- **The mempool rejects a transaction that exceeds a per-block limit on its own.** ZIP 218 does not specify this rejection. No block can include such a transaction, so the verifier gives it misbehavior score 100.
- **The rules stay dormant until NU7 activation.** No NU7 activation height exists. The deployment PR only needs to add the heights.

## Testing

Tests cover both sides of the NU7 activation boundary. They cover the exact subsidy, halving inversion, the 102-block averaging window, every per-pool limit, the global limit, mempool admission, block validation, block-template selection, and shielded coinbase capacity. New tests cover:

- Ironwood actions alone and mixed with Orchard actions, over and exactly at the Orchard limit, in block validation, mempool admission, and the block template, including coinbase Ironwood actions;
- `halving()` against ZIP 218's formula on a configured Testnet whose Blossom height is below the slow-start shift;
- startup backfilling the old 27-header context to the new 112-header maximum; and
- each remaining timing change, before and after NU7.

Post-rebase local runs passed:

- `cargo test --locked -p zakura-chain -p zakura-header-chain -p zakura-consensus -p zakura-rpc -p zakura-state -p zakura --lib`;
- `cargo check --locked -p zakura --features nu7`;
- `cargo clippy --locked --workspace --all-targets --features zakura/nu7 -- -D warnings`;
- `cargo fmt --all -- --check`;
- `./scripts/changelog.py check`; and
- Markdown lint for `docs/changelog/unreleased/851.md`.

The current CI crates.io publish-graph check passed. Local `actionlint .github/workflows/tests-unit.yml` also passed.

## Changelog

Updated `docs/changelog/unreleased/851.md`.

- [NU7 coinholder poll](https://forum.zcashcommunity.com/t/nu7-coinholder-vote/56912)
- [ZIP 258](https://zips.z.cash/zip-0258)
- #959 (ZIP 2003)
- #1007 (signed funding-stream address periods)

### Follow-up Work

- #1043 updates end-of-support timing and should merge before this PR.
- A deployment PR will set the NU7 activation heights.
- A ZIP 218 amendment should count Ironwood actions in the limits.

### Review follow-up

Merged current `main` and retained both the ZIP 218 mempool limit test and ZIP 2003 activation-boundary tests. The NU7 CI job now includes `zakura-state`, as requested. Local Rust 1.97 validation passed 246 consensus tests with `nu7` enabled, formatting, and the changelog check.

The follow-up also fixes a CI race in `dont_skip_verification_of_block_transactions_in_mempool`: each mock UTXO request now receives its response before the test starts the next verification. The fix was already present and tested in #859.

### End-of-support split verification

After moving the end-of-support timing to #1043, focused local runs passed:

- `cargo test --locked -p zakura-rpc rpc_getnetworksolps_uses_averaging_window_at_height`;
- `cargo test --locked -p zakura --test end_of_support`;
- `cargo fmt --all -- --check`;
- `./scripts/changelog.py check`; and
- Markdown lint for `docs/changelog/unreleased/851.md`.


crates/zakura-header-chain/src/validation/contextual/constants.rs
```diff
@@ -9,7 +9,12 @@ pub const POW_MEDIAN_BLOCK_SPAN: usize = 11;
 ///
 /// `PoWAveragingWindow + PoWMedianBlockSpan` in the Zcash specification based on
 /// > ActualTimespan(height : N) := MedianTime(height) − MedianTime(height − PoWAveragingWindow)
-pub const POW_ADJUSTMENT_BLOCK_SPAN: usize = POW_AVERAGING_WINDOW + POW_MEDIAN_BLOCK_SPAN;
+///
+/// ZIP 218 widens `PoWAveragingWindow` at NU7, so this span covers the largest
+/// window the build can use at any height. A `zip218` build therefore carries
+/// this wider context from genesis onwards and ignores the entries beyond the
+/// window in force at the candidate height.
+pub const POW_ADJUSTMENT_BLOCK_SPAN: usize = MAX_POW_AVERAGING_WINDOW + POW_MEDIAN_BLOCK_SPAN;
```
czarcas7ic, line 17: An existing synced node fails startup when upgraded to a zip218 build. Its database retains 27 predecessor headers, but startup now requires 112. Existing databases skip the initialization code updated here, and the migration doesn't backfill the missing headers before the startup consistency check rejects them.

We should backfill that context before the startup consistency check and add a test that upgrades an existing database from a default build to a zip218 build.

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -417,28 +417,38 @@ pub fn halving_divisor(height: Height, network: &Network) -> Option<u64> {
...
+    // Each target spacing era contributes (blocks in the era * era spacing) to a
+    // running total of block seconds, which the pre-Blossom halving interval
+    // measured in seconds then divides. This is the spec's segmented sum of
+    // fractions with the common denominator factored out, so it stays in integer
+    // arithmetic no matter how many spacing eras a network has. ZIP 218 adds a
+    // third era at NU7.
+    let pre_blossom_spacing_seconds = NetworkUpgrade::Genesis.target_spacing().num_seconds();
+    let mut total_block_seconds: HeightDiff = 0;
+
+    let mut eras = NetworkUpgrade::target_spacings(network)
```
czarcas7ic, line 426: height_for_halving() still returns heights from the old schedule. With the network parameters in post_nu7_spacing_halving_and_subsidy, it returns 6,720,001 for halving 4, but halving() returns 3 at that height. The fourth halving actually starts at 10,080,001.

Lets update height_for_halving() to account for NU7 and extend the round trip test to cover that activation.

crates/zakura-header-chain/src/validation/contextual/constants.rs
evan-forbes: Fixed in e76fe4900. The format migration now backfills the complete ZIP 218 validation context before startup validation. New migration tests cover existing 27-row databases, idempotent upgrades, and an interrupted backfill.

crates/zakura-chain/src/parameters/network/subsidy.rs
evan-forbes: Fixed in e76fe4900. height_for_halving now inverts the complete spacing-aware halving function. The boundary test covers the NU7 transition. I also fixed signed funding-stream periods for configured networks that activate NU7 before their first halving.

.github/workflows/tests-unit.yml
```diff
@@ -358,8 +358,10 @@ jobs:
       - uses: ./.github/actions/setup-zakura-build
       - name: Test experimental NU7 builds
         run: |
+          cargo test --locked --features zakura-chain/nu7-experimental \
+            -p zakura-chain -p zakura-header-chain --lib
           cargo test --locked --features nu7-experimental \
-            -p zakura-consensus --lib
+            -p zakura-consensus -p zakura-rpc --lib
```
czarcas7ic: Can we add `zakura-state` to this test command? The new migration test needs `nu7-experimental`, so CI currently skips it.

czarcas7ic (dismissed):

ValarDragon:
Ah we need to make Ironwood count towards the orchard limit, writing a zip update rn

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -279,24 +279,19 @@ impl ParameterSubsidy for Network {
...
     let height_after_first_halving = height - network.height_for_first_halving();
 
-    let address_period = (height_after_first_halving + network.post_blossom_halving_interval())
-        / network.funding_stream_address_change_interval();
-
-    address_period
-        .try_into()
-        .expect("all values are positive and smaller than the input height")
+    (height_after_first_halving + network.post_blossom_halving_interval())
+        .div_euclid(network.funding_stream_address_change_interval())
```
ValarDragon, line 301: Ok this is just for testnets. We should document this more, but also I think pull out of this PR

crates/zakura-chain/src/parameters/network_upgrade.rs
```diff
@@ -251,10 +251,84 @@ const PRE_BLOSSOM_POW_TARGET_SPACING: i64 = 150;
...
+/// JoinSplits count twice because each produces two shielded outputs.
+pub const GLOBAL_SHIELDED_BUDGET: u32 = 330;
+
+/// The largest averaging window this build can use, which bounds the number of
+/// relevant blocks a difficulty adjustment reads.
+pub const MAX_POW_AVERAGING_WINDOW: usize = if ZIP218_ENABLED {
+    POST_NU7_POW_AVERAGING_WINDOW
+} else {
+    PRE_NU7_POW_AVERAGING_WINDOW
+};
```
ValarDragon:
```suggestion
/// The largest averaging window this build can use, which bounds the number of
/// relevant blocks a difficulty adjustment reads.
pub const MAX_POW_AVERAGING_WINDOW: usize = POST_NU7_POW_AVERAGING_WINDOW;
```

crates/zakurad/src/components/sync/end_of_support.rs
```diff
@@ -103,9 +97,40 @@ pub async fn start(
...
+/// The estimate follows the target spacing at each height, so ZIP 218's 25 second
+/// spacing after NU7 fits three times as many blocks into each day.
+fn estimated_height_after_release(network: &Network, days: u32) -> Height {
+    let mut height = i64::from(ESTIMATED_RELEASE_HEIGHT);
+    let mut remaining_seconds = i64::from(days) * 24 * 60 * 60;
+
+    let target_spacings: Vec<_> = NetworkUpgrade::target_spacings(network).collect();
+    for (index, (_, target_spacing)) in target_spacings.iter().enumerate() {
+        let target_spacing = target_spacing.num_seconds();
+        let remaining_blocks = remaining_seconds / target_spacing;
```
ValarDragon: LOL this is funny, didn't see this coming as a need

ValarDragon:
Still need to review DAA + subsidy + GBT, but the block time component looks good to me! If we had a split out of just the 75s -> 25s change, I'd hit merge :)

ValarDragon:
@v12-sec pls review

ValarDragon:
@v12sec please review

evan-forbes:
Responses to V12 run [7913](https://v12.sh/runs/7913):

| Finding | Outcome |
|---|---|
| F-280622 Low: support window ends two blocks early | Fixed in 92bbb595a. The old era now ends at `next_height - 1`, and the tests cover NU7 activating mid-window and right after release. |
| F-280623 High: implicit NU7 activation skips shielded limits | Fixed in 69b49189f. `is_nu7_active` is now `current(network, height) >= Nu7`, the same check as the target spacing and averaging window. This only affects `zcash_unstable="zfuture"` builds, which don't compile on `main` today, so the new test checks that the three activation rules agree rather than the ZFuture case itself. |
| F-280626 High: early NU7 activation overmints slow-start rewards | Won't fix, as this matches the spec. ZIP 218 adds its NU7 subsidy case after the slow-start cases, as the Blossom case does. Protocol spec §7.8 and zcashd's `GetBlockSubsidy` also apply slow start before any spacing ratio. e22a3a986 pins this behavior. |
| F-280627 Medium: headers-only upgrades fail context migration | Won't fix, because it's unreachable in `zakurad`. The node only opens the header-chain store in Integrated mode, and both legacy migration and startup recovery reject a stored mode that doesn't match. HeadersOnly stores only exist in tests, benchmarks and fuzzing. |
| F-280628 Medium: scaled refills exceed the memory lookahead | Already on `main` with the unscaled limit. Tracked in #1026. |
| F-280629 Medium: checkpoint crossing can deadlock synchronization | Already on `main`, where the limit in the reported scenario is lower (21 instead of 61). Tracked in #1026. |
| F-280630 Medium: high body latches sync pause | Already on `main`. Tracked in #1026. |
| F-280631 Low: scaled request window exceeds drop cap | Fixed in e0ccefb78. The drop limit now scales with the same multiplier as the lookahead. |

Our follow-up review also found that the wider context loaded 113 full ancestor blocks on every block commit and block template. 49954486e reads headers instead.

crates/zakura-chain/src/parameters/network/subsidy.rs
czarcas7ic, line 249: AI comment that checks to me

---

On a Regtest with every upgrade at height 1, `height_for_first_halving` still returns 287, but the NU7 aware height_for_halving(1) returns 859 because 25 second blocks need more blocks per halving. The subsidy then halves at 859 while funding_stream_address_period still anchors on 287. ZIP 207 rotates each funding stream's recipient address every PostBlossomHalvingInterval / 48 blocks (6 on Regtest) counted from the first halving, and the verifier rejects a coinbase missing the required funding stream payment, so on a Regtest with configured funding streams whose recipients rotate, this build and one that derives the height can demand different addresses at the same height. We should derive the Regtest value from `height_for_halving(1, self)`, as the configured Testnet branch of `height_for_first_halving()` already does.

crates/zakurad/src/components/sync.rs
```diff
@@ -2417,24 +2442,20 @@ where
...
-            .0
-            .try_into()
-            .expect("fits in usize");
+        let verified_height = self.latest_chain_tip.best_tip_height().unwrap_or(Height(0));
+        let multiplier = lookahead_limit_multiplier(&self.network, verified_height);
+        let verified_height: usize = verified_height.0.try_into().expect("fits in usize");
 
         if verified_height >= max_checkpoint_height {
-            self.full_verify_concurrency_limit
+            self.full_verify_concurrency_limit * multiplier
```
czarcas7ic, line 2450: Just noting that the x3 multiplier also triples the configured in flight download count that the `MIN_CHECKPOINT_CONCURRENCY_LIMIT` doc says must stay small for RAM, so we may want to either scale just the height window or update that doc and accept the larger bound.

The same point applies to the inbound gossip downloader, which still uses the unscaled `full_verify_concurrency_limit` as its window, so after NU7 the two paths read the same config value as 20 blocks and 60 blocks.

https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakurad/src/components/sync.rs#L153-L155

crates/zakurad/src/components/sync.rs
czarcas7ic: The multiplier that was added here is applied on the syncer path, but the inbound gossip downloader still computes its window as `tip + full_verify_concurrency_limit` with no multiplier. After NU7 that window is about 8 minutes instead of 25, so a node more than 20 blocks behind drops gossiped blocks and falls back to the syncer 3x sooner in wall clock terms. The drop is unscored, so there is no ban risk. 

Worth calling `lookahead_limit_multiplier` there too, or noting in the description that only the syncer scales.

https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakurad/src/components/inbound/downloads.rs#L687-L695

czarcas7ic: `AT_OR_NEAR_TIP_THRESHOLD` derives its 20 minute margin from 75 second blocks and was not scaled, so after NU7 stall tracking engages at about 7 minutes behind.

https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-chain/src/chain_tip.rs#L19-L27

crates/zakura-state/src/service/finalized_state/header_chain/migration.rs
```diff
@@ -254,6 +269,97 @@ impl HeaderChainStore {
...
     }
 
+    /// Resize the retained validation context to the consensus maximum.
+    ///
+    /// Older builds retained fewer rows, so the authenticated full-state header
+    /// index supplies the missing rows. This step runs before the startup audit
+    /// because that audit requires the complete context.
+    ///
+    /// Returns the number of rows added or removed.
+    pub(in crate::service) fn resize_validation_context(
```
czarcas7ic, line 279: Rolling back to the previous release after this resize fails with `LimitExceeded` because the old reader caps the column at 27 rows and the disk format version remains unchanged.

crates/zakura-state/src/service/write.rs
```diff
@@ -539,6 +539,7 @@ impl HeaderChainWriter {
         let store = HeaderChainStore::new(finalized_state.db.header_chain_disk_db());
         store.migrate_to_current(&config)?;
         let runtime = if store.is_initialized()? {
+            store.resize_validation_context(&finalized_state.db)?;
```
czarcas7ic, line 542: nit: we should move the resize after the `FinalizedDivergence` check so a rolled back full state keeps its specific error. We should also log the returned row count when it's nonzero.

crates/zakura-chain/src/parameters/network_upgrade.rs
```diff
@@ -402,12 +437,13 @@ impl NetworkUpgrade {
     pub fn target_spacing(&self) -> Duration {
         let spacing_seconds = match self {
             Genesis | BeforeOverwinter | Overwinter | Sapling => PRE_BLOSSOM_POW_TARGET_SPACING,
-            Blossom | Heartwood | Canopy | Nu5 | Nu6 | Nu6_1 | Nu6_2 | Nu6_3 | Nu7 => {
+            Blossom | Heartwood | Canopy | Nu5 | Nu6 | Nu6_1 | Nu6_2 | Nu6_3 => {
                 POST_BLOSSOM_POW_TARGET_SPACING.into()
             }
+            Nu7 => POST_NU7_POW_TARGET_SPACING.into(),
```
czarcas7ic, line 443: Just pointing out that ZIP 218 is height gated only, while ZIP 2003 still requires nu7 feature for example, so trying to understand if we want to feature gate this as well or if we just remove the feature tag from everything else?

crates/zakura-state/src/service/check.rs
```diff
@@ -115,7 +113,7 @@ where
     //
     // TODO: accept a NotReadyToBeCommitted error in those tests instead
     #[cfg(test)]
-    if relevant_chain.len() < POW_ADJUSTMENT_BLOCK_SPAN {
+    if relevant_headers.len() < POW_ADJUSTMENT_BLOCK_SPAN {
```
czarcas7ic, line 116: Full AI comment but this checks to me:

---

`POW_ADJUSTMENT_BLOCK_SPAN` is now `MAX_POW_AVERAGING_WINDOW + POW_MEDIAN_BLOCK_SPAN`, so this test-only bypass widened from 28 to 113 without its own line changing. State unit tests reaching this helper with between 28 and 112 ancestor headers now skip its difficulty check entirely, which includes the code this PR rewrites.

Two header-chain tests lost their edge the same way: `predecessor_context.rs` iterates `[0, 27, 28, 35]` over a 35-block chain, so `.min(POW_ADJUSTMENT_BLOCK_SPAN)` never binds, and `validation.rs` iterates `[1, 11, 16, 17, 27]`, where 27 used to be the span minus one. Deriving those fixtures from the span constant, as `preparation.rs` now does, restores the coverage.

Separately, no test feeds a 113-entry context through `AdjustedDifficulty` at a post-NU7 height to check that 102 thresholds are averaged and the older median comes from offsets 102..113. Could we add a small window-sensitivity test on a configured testnet with NU7 at 1000 that does this?

https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-header-chain/src/validation/contextual/constants.rs#L11-L17

https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-header-chain/src/transition/planner/tests/predecessor_context.rs#L87

https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-header-chain/src/validation/contextual/tests/validation.rs#L64

crates/zakurad/src/components/sync.rs
```diff
@@ -177,6 +180,21 @@ pub const DEFAULT_ZAKURA_BLOCK_APPLY_CONCURRENCY_LIMIT: usize = 32;
 /// If the concurrency limit is 0, Zebra can't download or verify any blocks.
 pub const MIN_CONCURRENCY_LIMIT: usize = 1;
 
+/// Returns the multiplier for the syncer's block-count lookahead limits at `height`.
+///
+/// ZIP 218 scales block-count download limits by `NU7PoWTargetSpacingRatio` where the target
+/// spacing is 25 seconds, so the limits cover the same time as at 75 second spacing. The
+/// multiplier is 1 at every other spacing.
+fn lookahead_limit_multiplier(network: &Network, height: Height) -> usize {
```
czarcas7ic, line 188: `AT_OR_NEAR_TIP_THRESHOLD` derives its 20 minute margin from 75 second blocks and was not scaled, so after NU7 stall tracking engages at about 7 minutes behind.

https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-chain/src/chain_tip.rs#L19-L27

p0mvn:
**AI review comment — blocking recommendation**

**Recommendation: request changes until the post-NU7 difficulty calculation has known-answer and rejection tests.** This is a validation gap, not a confirmed arithmetic defect or exploit. Reviewed commit: `97be29a2c0c659cf6d4e36f5d653d99df87b7426`.

The changes in [adjusted_difficulty.rs:202–245](https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-header-chain/src/validation/contextual/adjusted_difficulty.rs#L202-L245) and [lines 316–333](https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-header-chain/src/validation/contextual/adjusted_difficulty.rs#L316-L333) switch the early-chain cutoff, target averaging, and older-median offset from 17 to 102. These calculations determine which headers are accepted.

The new [activation test](https://github.com/zakura-core/zakura/blob/97be29a2c0c659cf6d4e36f5d653d99df87b7426/crates/zakura-chain/src/parameters/network/tests.rs#L539-L574) checks that the window accessor returns 102. Existing difficulty validation vectors use networks without NU7 activation; they do not establish that post-NU7 `nBits` matches ZIP 218. Retaining a 17-entry mean or older-median offset on the NU7 path would not be caught by the accessor test or existing pre-NU7 vectors. That could reject the canonical chain or accept invalid headers. This is a coverage argument, not an experimentally demonstrated mutation.

Please add:

- Independently calculated expected `nBits` vectors immediately before, at, and after NU7, using nonuniform targets and timestamps that distinguish the 17- and 102-block calculations.
- A negative validation case rejecting a target calculated under the old rules.
- Early-activation boundary coverage at candidate heights 102 and 103.
- Testnet minimum-difficulty coverage at exactly 150 seconds and 151 seconds after the parent timestamp.

Local verification with Rust 1.97: 33 network parameter tests passed; existing header-chain contextual tests had 6 passes and 2 ignored tests. An independent second AI review found no concrete subsidy or halving arithmetic defect. The missing post-NU7 consensus-calculation coverage is the blocking recommendation.

crates/zakurad/src/components/sync.rs
```diff
@@ -2417,24 +2442,20 @@ where
...
         // When the state is empty, we want to verify using checkpoints
-        let verified_height: usize = self
-            .latest_chain_tip
-            .best_tip_height()
-            .unwrap_or(Height(0))
-            .0
-            .try_into()
-            .expect("fits in usize");
+        let verified_height = self.latest_chain_tip.best_tip_height().unwrap_or(Height(0));
+        let multiplier = lookahead_limit_multiplier(&self.network, verified_height);
```
p0mvn, line 2446: Optional note: a few non-consensus block-count policies differ numerically from ZIP 218. For example, Zakura retains a 1,000-block reorg limit rather than 600 and a 10,000-block pruning window rather than 864.

These are not consensus mismatches and both exceed the ZIP's intended wall-clock safety windows, but documenting them as deliberate implementation-specific choices would make the compliance story clearer.

crates/zakura-state/src/service/check.rs
evan-forbes: #1048 covers this. It keeps the active difficulty span at 28 headers and adds separate 113-header retention limits, so the test bypass and the `predecessor_context.rs` and `validation.rs` fixtures keep their original edges.

crates/zakurad/src/components/sync.rs
evan-forbes: We previously raised the max reorg length to 1000 because it is not a consensus-breaking value, so we are keeping it for now. If we want to match the spec exactly, happy to do that in a follow-up.

evan-forbes:
Do not merge until #1048 and #1057 have merged and `main` has been merged into this branch.

crates/zakura-chain/src/parameters/network_upgrade.rs
evan-forbes: We went the other way: every NU7 rule activates by height alone, following d078fd901. #1060 removes the `nu7` feature, so ZIP 2003's V4 rejection applies from NU7 in every build. #1060 also sets the estimated Mainnet and Testnet NU7 heights.

crates/zakura-state/src/service/finalized_state/header_chain/migration.rs
evan-forbes: Fixed in cd9aecec9. The header-chain disk format is now 5, and startup records it before it widens the context. The previous release rejects the database with `UnsupportedDiskFormat(5)` before it reads a context row. The changelog says a downgrade needs a header-chain resync.

crates/zakura-state/src/service/write.rs
evan-forbes: Fixed in a277f5ce3. The resize now runs after the `FinalizedDivergence` check and logs the row count when it is nonzero. `rolled_back_full_state_reports_divergence_before_resizing_the_context` covers it.

crates/zakurad/src/components/sync.rs
evan-forbes: Fixed in 622042602. We kept the ×3, because ZIP 218 scales `BLOCK_DOWNLOAD_WINDOW` and `MAX_BLOCKS_IN_TRANSIT_PER_PEER` by the same factor, and the `MIN_CHECKPOINT_CONCURRENCY_LIMIT` doc now states the larger RAM bound. Inbound gossip now scales its height window by the same multiplier. Its queue capacity stays unscaled, because `MAX_INBOUND_CONCURRENCY` bounds its RAM.

crates/zakurad/src/components/sync.rs
evan-forbes: Fixed in e12f1f087. The threshold now scales with the target spacing at the estimated network tip. It is 16 blocks before NU7 and 48 after, so it stays about 20 minutes.

crates/zakura-chain/src/parameters/network/subsidy.rs
evan-forbes: Fixed in e6e5b0e8b. Regtest now derives its first halving from `height_for_halving(1)`, as configured testnets do. That gives 859 with NU7 at height 1 and still 287 without NU7.

evan-forbes:
Re https://github.com/zakura-core/zakura/pull/851#issuecomment-5737481077: the precondition is met. #1048 and #1057 have merged, and `main` is merged into this branch (51b31065b). The open review threads are fixed in cd9aecec9 through e6e5b0e8b.

p0mvn:
Superseded by the reviewed split stack:

- #1062 — header-chain migration and startup hardening (merged)
- #1063 — candidate-height difficulty-context readers
- #1064 — v1.4.0 compatibility fixture (merged)
- #1065 — spacing-aware node policies (merged)
- #1066 — remaining ZIP 218 consensus activation, stacked on #1063

The replacement commits retain Evan Forbes as co-author and reference this PR for provenance. Closing #851 in favor of #1066.

PR #853, opened 2026-09-01
fix(block-sync): supervise the critical block driver

evan-forbes:
## Motivation

The node tracked the critical block-sync driver only for endpoint teardown. An unexpected return or panic could leave the node and native services running after block application stopped.

Consensus-invalid body persistence intentionally fails closed when it cannot record evidence. The node root must observe that driver exit instead of leaving the process half-alive.

## Solution

Wrap the driver in a drop guard that reports every unexpected return, panic, or cancellation to the node root. Construct the guard before the future receives its first poll so pre-poll cancellation cannot bypass the notification.

Treat the notification as fatal and start normal node shutdown. Suppress the notification during requested shutdown.

## Testing

- `cargo test -p zakura --lib commands::start::tests -- --test-threads=1` (21 passed)
- `git diff --check`

The regressions cover an unexpected return, panic unwinding, pre-poll cancellation, and requested shutdown.

## Changelog

Added `docs/changelog/unreleased/853.md`.

- Useful-item-loss audit finding F022

### Follow-up Work

PR #850 uses this supervision when a required block-sync action receiver closes.

### DIFF
--- a/crates/zakurad/src/commands/start.rs
+++ b/crates/zakurad/src/commands/start.rs
@@ -1720,6 +1720,41 @@ mod tests {
         assert!(shutdown.is_cancelled(), "the node starts normal shutdown");
         assert!(error.to_string().contains("block-sync driver exited"));
     }
+
+    #[tokio::test]
+    async fn block_sync_driver_panic_notifies_the_node_root() {
+        let shutdown = CancellationToken::new();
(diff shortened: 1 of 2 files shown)


ValarDragon (approved):
Oh great job, yes this should definitely be handled in start

PR #854, opened 2026-09-01
perf(rpc): cache getblockchaininfo disk size

evan-forbes:
## Motivation

Lightwalletd calls `getblockchaininfo` for each `GetLatestBlock` request. A CPU profile on a synchronized full-state node showed that the `UsageInfo` state read dominated this RPC. Each request called `DB::ListColumnFamilies`. RocksDB then replayed the manifest and summed SST metadata for every column family.

## Solution

Cache the database disk-size estimate in `DiskDb`. Database startup computes the initial value. The existing RocksDB metrics task normally refreshes it every 30 seconds. The `UsageInfo` state read now performs one atomic load.

Keep the public `DiskDb::size()` and `ZakuraDb::size()` methods as on-demand measurements. This preserves their behavior for direct database users and offline tools. An incomplete RocksDB property scan also retains the last complete cached value.

Add `scripts/rpc-bench.py` to measure JSON-RPC latency and throughput over persistent HTTP connections.

## Freshness rationale

The RPC defines `size_on_disk` as an estimated size. The underlying RocksDB property only counts SST file bytes, so the prior on-demand value was already not an exact filesystem measurement. The 30-second cache interval adds bounded operational staleness without changing consensus or chain-tip data.

The mainnet target block interval is 75 seconds. Lightwalletd does not deserialize or use `size_on_disk`. It uses the block height and best block hash for `GetLatestBlock`. Operators that need exact free-space information must use filesystem monitoring rather than this RPC estimate.

## Benchmark

I reran an isolated A/B benchmark after the audit fixes. The baseline used unpatched `main` at `0f435507338a19fe0b8aff51875c6dac784e39c2`. The candidate used the same commit plus this PR at `d2339c5f575793504807864228edac9df7859c8f`. Both binaries used the same synchronized full-state database, configuration, hardware, and localhost RPC path.

| Concurrency | Metric | Unpatched `main` | Patched `main` | Improvement |
| ---: | --- | ---: | ---: | ---: |
| 1 | p50 | 22.56 ms | 0.84 ms | 27.0x |
| 1 | p95 | 26.81 ms | 1.29 ms | 20.8x |
| 1 | p99 | 36.05 ms | 1.80 ms | 20.0x |
| 1 | requests/s | 43.48 | 1,102.34 | 25.3x |
| 4 | p50 | 41.68 ms | 1.90 ms | 22.0x |
| 4 | p95 | 55.65 ms | 3.48 ms | 16.0x |
| 4 | p99 | 64.22 ms | 4.73 ms | 13.6x |
| 4 | requests/s | 96.69 | 1,932.11 | 20.0x |

Every benchmark request succeeded. The candidate's `size_on_disk` value of 277,008,746,976 bytes exactly matched `zakura_state_rocksdb_total_disk_size_bytes`. I restored the node's exact original binary and verified RPC health afterward.

## Testing

The state test writes and flushes an SST file. It verifies that direct size measurements do not use the cache. It also verifies that the metrics export refreshes the cached nonzero disk size.

The following checks passed:

- `cargo test -p zakura-state --lib` (541 passed, 3 ignored)
- `cargo test -p zakura-state exporting_metrics_refreshes_cached_disk_size --lib`
- `cargo test -p zakura-rpc get_blockchain_info --lib`
- `cargo clippy -p zakura-state --lib --tests -- -D warnings`
- `cargo build --release --bin zakurad`
- `python3 -B scripts/rpc-bench.py --help`

## Changelog

Added `docs/changelog/unreleased/854.md` under `Fixed`.

### DIFF
--- a/crates/zakura-state/src/request.rs
+++ b/crates/zakura-state/src/request.rs
@@ -1440,7 +1440,10 @@ impl Request {
 /// [`ReadStateService`](crate::service::ReadStateService).
 pub enum ReadRequest {
     /// Returns [`ReadResponse::UsageInfo(num_bytes: u64)`](ReadResponse::UsageInfo)
-    /// with the current disk space usage in bytes.
+    /// with a recent estimate of the disk space usage in bytes.
+    ///
+    /// The state service refreshes this estimate with its RocksDB metrics, normally every 30
+    /// seconds.
     UsageInfo,
 
     /// Returns [`ReadResponse::PruningInfo`] with
--- a/crates/zakura-state/src/response.rs
+++ b/crates/zakura-state/src/response.rs
@@ -387,7 +387,7 @@ pub struct BlockSyncBodyMetadata {
 /// A response to a read-only
 /// [`ReadStateService`](crate::service::ReadStateService)'s [`ReadRequest`].
 pub enum ReadResponse {
-    /// Response to [`ReadRequest::UsageInfo`] with the current best chain tip.
+    /// Response to [`ReadRequest::UsageInfo`] with a recent disk space usage estimate.
     UsageInfo(u64),
 
     /// Response to [`ReadRequest::PruningInfo`] with this node's pruning status.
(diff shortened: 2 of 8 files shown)


ValarDragon (approved):
Awesome., looks good to me. Agreed its silly this query doesn't use a cache. Anyone who cares about disk size either can do the dedicated probe, or accept risk of a 60s stale measurement

PR #856, opened 2026-09-01
feat(consensus): disallow version 4 transactions at NU7

evan-forbes:
## Motivation

The candidate NU7 decision deprecates Sprout by disallowing version 4 transactions at NU7. Zakura needs one fixed consensus boundary for that decision. NU7 remains subject to change before deployment.

## Solution

The off-by-default `nu7-experimental` Cargo feature makes Zakura reject version 4 transactions at the exact NU7 activation height. A default build and a network without an explicit NU7 activation continue to accept version 4 transactions.

The implementation does not expose a configurable deprecation policy or height. This removes the former `AtNu7`, `AtHeight`, and `Never` choices from consensus configuration.

The exact NU7 lookup matters for networks that omit NU7 but configure a later upgrade. A general network-upgrade lookup can substitute the later activation height. This branch searches the activation list for NU7 itself.

## Testing

Tests cover the last valid height, the NU7 activation height, and later heights. They cover builds with and without `nu7-experimental`. They also cover an omitted NU7 and a later upgrade without NU7.

The audit passed the chain and consensus unit suites. Formatting and diff checks also passed.

## Changelog

Added `docs/changelog/unreleased/856.md`.

- [Current NU7 coinholder poll](https://forum.zcashcommunity.com/t/nu7-coinholder-vote/56912)
- [Current ZCAP poll](https://forum.zcashcommunity.com/t/zcap-poll-now-open-nu7-august-2026/57223)
- [ZIP 2003](https://zips.z.cash/zip-2003)

### Follow-up Work

A deployment PR will replace the experimental feature with finalized NU7 rules and set the activation parameters.


crates/zakura-chain/src/parameters/network.rs
```diff
@@ -342,6 +369,40 @@ impl Network {
...
+    ///
+    /// [`V4Deprecation::AtNu7`] resolves to the NU7 activation height, so a network
+    /// that does not activate NU7 keeps accepting version 4 transactions.
+    pub fn v4_deprecation_height(&self) -> Option<Height> {
+        match self.v4_deprecation() {
+            V4Deprecation::AtNu7 => self
+                .activation_list()
+                .iter()
+                .find_map(|(height, upgrade)| (*upgrade == NetworkUpgrade::Nu7).then_some(*height)),
+            V4Deprecation::AtHeight(height) => Some(height),
```
czarcas7ic: If `AtHeight` is used and we are between network upgrades, this doesn't trigger mempool re verification, so previously accepted v4 transactions can remain eligible for block templates after they become invalid.

For example, suppose NU7 activates at height 1,000 and v4 is disabled with `AtHeight(2000)`. A v4 transaction accepted for block 1,999 can still be waiting in the mempool, unmined and unexpired, when the tip reaches 1,999. The node can then include it in a template for block 2,000 even though its block validator would reject that block.

We should trigger mempool re verification at the independent cutoff, as already happens for `AtNu7` through the network upgrade reset, and add a regression test covering this transition.

evan-forbes: Resolved by removing the configurable cutoff. Version 4 transactions now become invalid only at the exact NU7 activation height, so the existing network-upgrade mempool reset covers the transition. Boundary tests cover explicit and omitted NU7 activation.

PR #858, opened 2026-09-01
feat(consensus): implement ZIP 235 behind a build feature

evan-forbes:
Stacked on #857 (ZIP 233) → #856 (ZIP 2003) → #851 (ZIP 218). Fourth of the remaining NU7 consensus changes.

Implements [ZIP 235](https://zips.z.cash/zip-0235) "Remove 60% of Transaction Fees From Circulation" behind the off-by-default `zip235` cargo feature, which enables `zip233` because that ZIP's `zip233Amount` field carries the burn, and because ZIP 235 "MUST be deployed at the same time or after ZIP 233".

Ported from the merged upstream [zebra#10357](https://github.com/ZcashFoundation/zebra/pull/10357). ZIP 235 is on the `zcash/zips` NU7 milestone rather than the poll ballot, but it is the third leg of the NSM alongside ZIPs 233 and 234.

## Consensus

- `minimum_fee_burn` is `floor(block_miner_fees * 6 / 10)` — the ZIP's 60%, with the rounding in the miner's favour.
- `miner_fees_are_valid` rejects a coinbase whose `zip233Amount` is below that minimum, from NU7 activation onwards.
- It also adds the coinbase's own `zip233Amount` to its total output value, because that value comes out of the same block subsidy and fees as its outputs. Without this a compliant coinbase would fail the NU6 "total output MUST equal total input" rule.

## Block templates

`TransactionTemplate::new_coinbase` burns exactly the minimum and pays the miner the block subsidy plus the remaining 40% of the fees, so a template this node produces passes its own verifier. Burning more than the minimum is the miner's choice to make, not this node's, so templates do not.

The burn is set through `Builder::set_zip233_amount` after the outputs are added, so the builder's value balance already reflects the reduced miner reward. That setter only exists in a build that compiles `zakura-primitives`' ZIP 233 code, so the call is `#[cfg(feature = "zip235")]`.

## Tests

`zip235_coinbase_must_burn_sixty_percent_of_fees` covers the boundary in both build configurations: burning exactly 600 of 1000 zatoshi in fees is valid, burning more is valid, and burning 599 or nothing is rejected with `InsufficientFeeBurn` under the feature. Without the feature there is no minimum, and the test asserts the burn still balances against the coinbase's total output value.

Both configurations pass all 731 tests across `zakura-chain`, `zakura-consensus`, and `zakura-rpc`; clippy clean. A `zip235` CI job mirrors the `zip233` one.

### DIFF
--- a/.github/workflows/tests-unit.yml
+++ b/.github/workflows/tests-unit.yml
@@ -393,6 +393,36 @@ jobs:
           cargo test --locked --features zip233 \
             -p zakura-consensus -p zakura-rpc --lib
 
+  zip235:
+    name: ZIP 235 feature tests
+    needs: changes
+    # Compiles its own binaries, because the shared nextest archive is built
+    # without this feature, and because the feature needs its own RUSTFLAGS.
+    if: needs.changes.outputs.rust == 'true' || github.event_name != 'pull_request'
+    permissions:
+      contents: read
+      id-token: write
+      statuses: write
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    env:
+      # ZIP 235 enables ZIP 233, which needs this cfg to reach the matching
+      # `zakura-primitives` code.
+      RUSTFLAGS: --cfg zcash_unstable="nu7"
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 #v7.0.1
+        with:
+          persist-credentials: false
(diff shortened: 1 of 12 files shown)


evan-forbes:
The August 2026 NU7 poll explicitly excludes transaction-format changes, so ZIP 235 is outside the current NU7 scope. Closing this superseded PR. PR #859 has been rebuilt without ZIP 233 or ZIP 235 and now stacks directly on #856.

PR #859, opened 2026-09-01
feat(consensus): implement halving-preserving NU7 reissuance

evan-forbes:
## Replacement PR stack

Review the split stack below. This PR is closed as superseded and remains the original source reference.

| Order | PR | Scope |
| --- | --- | --- |
| Prerequisite | #1055 | Run normal CI for PRs targeting split branches |
| 1 | #1049 | Fix deferred funding in slow-start replay and rollback |
| 2 | #1050 | Await the actual parent's committed state across forks |
| 3 | #1051 | Calculate cumulative scheduled issuance |
| 4 | #1052 | Persist the signed deficit and migrate existing databases |
| 5 | #1053 | Enforce parent-dependent reissuance and negative-deficit rejection |
| 6 | #1054 | Integrate reissuance into mining templates and subsidy RPCs |

Each PR targets its predecessor. All seven PRs carry the `nu7` label. State
accounting and consensus rules have separate reviews and keep their own tests.
The split preserves ZIP 218's newer base changes. It also fixes the cumulative
schedule's walk through zero-subsidy halvings at large Regtest heights.

The stack adds activation and arithmetic-limit matrices, migration fault cases,
signed-counter tests, and stronger parent/RPC tests. Extended NU7 suites pass
with 10,000 arithmetic cases and 256 histories per rollback/replay property.
All eight deliberate runtime mutations trigger test failures. Each replacement
PR reports its chunk-level test results. Extended default suites also pass. Both node configurations pass 442 tests
and reproduce one pruning-log assertion failure on the unchanged base. Build,
lint, formatting, packaging, publish-graph, changelog, and workflow checks pass.
These results cover the extracted stack ending at 3dc4b5811; subsequent fixes
require their own validation.
Production activation still requires the policy and real-transaction validation
work described below.

---

## Motivation

Implement halving-preserving reissuance behind the disabled-by-default `nu7` feature. Store its balance so verification, RPCs, forks, and rollback use the same parent state.

## Solution

From the reissuance start height, the subsidy is the halving subsidy plus `ceil(parent_deficit * 4126 / 10^10)`. The default start maps the February 2031 crossing rule through ZIP 218's halving clock. Configured testnets and Regtest can override it. The fraction stays fixed per block across spacing changes.

The signed deficit starts at zero immediately before NU7. From NU7, each block adds scheduled issuance minus its change to the monetary pools. Pool transfers do not create reissuance. The deficit itself is excluded from monetary totals. Both state commit paths reject negative deficits from the reissuance start height.

**Historical funds remain excluded pending guidance.** Should pre-NU7 unclaimed subsidy and fees seed this balance? We need guidance before adding them. The [baseline comment](https://github.com/zakura-core/zakura/blob/95bfd18abc03aae16db60286a43c2ad9b56f7005/crates/zakura-chain/src/block.rs#L353) identifies the decision and the [matching migration](https://github.com/zakura-core/zakura/blob/95bfd18abc03aae16db60286a43c2ad9b56f7005/crates/zakura-state/src/service/finalized_state/disk_format/upgrade/issuance_deficit_pool.rs#L151). This provisional baseline differs from zips#1354's genesis-based deficit. Changing it requires updating runtime accounting, migration, and their tests together.

Database format 29.0.0 moves the existing v28 database to the v29 cache path and backfills the unseeded balance. Migration preserves monetary pools and block sizes. It propagates arithmetic, corruption, and write errors. Cancellation and failure leave the version marker unchanged so startup can retry. Older binaries cannot use the expanded records; downgrade requires a pre-upgrade backup or separate sync.

See [the accounting contract and activation requirements](https://github.com/zakura-core/zakura/blob/95bfd18abc03aae16db60286a43c2ad9b56f7005/docs/design/reissuance-accounting.md).

## Testing

Local chain, state, consensus, RPC, header-chain, and node tests pass with and without NU7. Workspace clippy passes in both configurations. The all-feature/all-target build, formatting, packaging, changelog, and workflow checks pass. Both arithmetic properties pass 10,000 cases. The extended database property passes 1,000 histories covering rollback, reopen, replay, alternate branches, and fresh replay equivalence. Fault injection detects rounding, parent selection, historical seeding, migration, transfer, and rollback errors.

The properties cover integer rounding, schedule sums, historical exclusion, transfers through every monetary pool, reductions in issued value, independent non-finalized forks, finalized rollback, replay, alternate branches, reopen, and fresh replay equivalence. Migration tests use legacy bytes and cover batch boundaries, failed writes, cancellation, corruption, startup retry, and version preservation.

Checkpoint fixtures isolate accounting and can intentionally underclaim. The semantic subsidy test accepts the exact reward and rejects a one-zatoshi overclaim, but mocks transaction verification because NU7 lacks a production branch ID. These tests do not establish ZIP 233 transaction support.

## Changelog

Updated `docs/changelog/unreleased/859.md`.

- [Draft halving-preserving issuance specification](https://github.com/zcash/zips/pull/1354)
- [NU7 coinholder poll](https://forum.zcashcommunity.com/t/nu7-coinholder-vote/56912)

### Follow-up Work

Production activation requires historical-funds guidance, confirmation of the later negative-deficit rejection height and spacing policy, a production NU7 branch ID, and activation heights. It also requires independent accounting review, historical balance verification, representative migration measurements, and a private-network test covering real transactions, bonus claims, forks, restart, migration, and convergence. The feature remains disabled by default.


crates/zakura-consensus/src/block.rs
```diff
@@ -328,8 +328,39 @@ where
...
+                        .await
+                        .map_err(|source| VerifyBlockError::Depth { source, hash })?
+                        .call(zs::Request::BlockInfo(parent_hash.into()))
+                        .await
+                        .map_err(|source| VerifyBlockError::Depth { source, hash })?
+                    else {
+                        unreachable!("wrong response to Request::BlockInfo");
+                    };
+
+                    let parent_info = parent_info.ok_or_else(|| {
```
czarcas7ic: The subsidy used to depend only on block height, so a child could pass this check while its parent was still being verified, then wait in the state service’s commit queue. With ZIP 234, calculating the child’s subsidy requires the money reserve balance after the parent block has been applied. This lookup only sees committed blocks, so if the parent is still being processed, the child fails before reaching that queue and the legacy sync round restarts. I think we need to wait for the parent’s committed state within the verification timeout before calculating the child’s subsidy.

crates/zakura-consensus/src/checkpoint.rs
```diff
@@ -675,7 +675,11 @@ where
 
         // See [ZIP-1015](https://zips.z.cash/zip-1015).
         let expected_deferred_amount =
-            funding_stream_values(height, &self.network, block_subsidy(height, &self.network)?)?
+            // Checkpoint verification only covers blocks below the mandatory checkpoint,
+            // which is far below any NU7 activation, so there is no ZIP 234 money reserve
+            // to supply. `block_subsidy` returns `MissingMoneyReserve` rather than a wrong
+            // subsidy if that ever stops being true.
+            funding_stream_values(height, &self.network, block_subsidy(height, &self.network, None)?)?
```
czarcas7ic: Adding a checkpoint at or after ZIP 234 activation would prevent nodes from syncing through that activation. They would send those historical blocks through the checkpoint verifier, which passes None for the money reserve. That works before ZIP 234 because the subsidy depends only on height, but after activation it returns MissingMoneyReserve and stops sync. Please update checkpoint verification to supply the required reserve and test syncing across activation with a later checkpoint.

evan-forbes: Fixed in 6ad0de7ab. The checkpoint verifier now defers the active ZIP 234 subsidy calculation until the parent commit is available. The checkpoint_sync_crosses_zip234_activation test covers a checkpoint above activation.

crates/zakura-consensus/src/block.rs
evan-forbes: Fixed in 6ad0de7ab. The semantic verifier now waits for the parent through AwaitBlockInfo before reading the reserve. State commit notifications wake the waiter after checkpoint and semantic commits. A forced race test covers the wait.

czarcas7ic (dismissed):
Confirmed both my comments are addressed. Side note, we should probably set a reminder for us to check against Zebra whenever they implement all NU7 features just to ensure consensus agrees.

evan-forbes:
Historical funds remain excluded: the balance starts at zero at NU7. We need guidance on whether pre-NU7 unclaimed subsidy and fees should seed it.

The [source comment and baseline rule](https://github.com/zakura-core/zakura/blob/d50d40ba447d8133149162d36cd99714714addb6/crates/zakura-chain/src/block.rs#L353) mark the decision. The [migration baseline](https://github.com/zakura-core/zakura/blob/d50d40ba447d8133149162d36cd99714714addb6/crates/zakura-state/src/service/finalized_state/disk_format/upgrade/issuance_deficit_pool.rs#L151) excludes the same funds. Any policy change must update both implementations and their accounting tests.

evan-forbes:
@v12sec please review

evan-forbes:
Responses to V12 run [7922](https://v12.sh/runs/7922):

| Finding | Outcome |
|---|---|
| F-280652 High: zero seed disables all reissuance | Confirmed, but it needs a policy decision rather than a code fix. From NU6, a coinbase must claim exactly the subsidy plus fees, and no supported transaction removes value from the monetary pools. So with the deficit seeded at zero, it stays zero and the bonus is never paid. Reissuance only pays once the historical-funds seed is decided (see the open question above) or a value-removal mechanism such as ZIP 233 exists. Until then, this PR only adds the accounting. |
| F-280651 Medium: orphan waits exhaust verification pipelines | Won't fix in this PR. `AwaitBlockInfo` has the same exposure as `AwaitUtxo` on `main`: a block spending an unknown outpoint already waits until the verification timeout. The wait can't return early, because during sync a parent is often still being verified and hasn't reached the state queue yet. |
| F-280655 Medium: slow-start funding breaks legacy upgrade | Fixed in e125dd1c3. The format 27 replay and offline rollback now compute the deferred funding change at every height, as block commits do. The slow-start guard was already on `main`, so rollback is fixed too. |
| F-280650 Low: side fork contaminates subsidy RPC | Fixed in 7219035a7. `ReadRequest::BlockInfo` now resolves heights on the best chain only; hashes still resolve on any chain for parent lookups. |
| F-280656 Low: future subsidy queries no longer work | Documented in 2880f8bab. From the ZIP 234 start, the subsidy depends on the parent's value pools, so `getblocksubsidy` can't answer heights more than one block above the tip. The docs and error now say so, and with the F-280650 fix a side fork can no longer answer. |
| F-280657 Medium: implicit NU7 skips shielded limits | Duplicate of F-280623 on #851. The shielded limits moved to #1027, which includes the fix, and this branch no longer contains `is_nu7_active`. |

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -460,10 +468,451 @@ pub fn halving(height: Height, network: &Network) -> u32 {
         .expect("halving index is non-negative and fits in u32")
 }
 
+/// The numerator of [ZIP 234]'s `BLOCK_SUBSIDY_FRACTION`.
+///
+/// [ZIP 234]: https://zips.z.cash/zip-0234
+pub const BLOCK_SUBSIDY_FRACTION_NUMERATOR: u128 = 4_126;
```
czarcas7ic: The draft ZIP changed on 2026-09-16, after the alignment commit, and now defines the fraction per height: `BLOCK_SUBSIDY_FRACTION(height) = floor(LN2_SCALED / HalvingInterval(height)) / 10^10`, which is 1375 at every NU7 height under ZIP 218, not 4126.

https://github.com/ShieldedLabs/zips/blob/60720df9e971e19d8b6f67f1869426d78c96d250/zips/draft-judah-nsm-halving-preserving-issuance.md#L153-L165

It also explicitly rejects a fixed 4126 for the reason the denominator's doc comment gives (a roughly 1.33-year half-life), so that rustdoc now states the opposite of the spec:

https://github.com/ShieldedLabs/zips/blob/60720df9e971e19d8b6f67f1869426d78c96d250/zips/draft-judah-nsm-halving-preserving-issuance.md#L296-L316

Every reissuance height is at or above NU7 and this branch switches to 25-second spacing there, so in practice the value is never 4126. The crossing-height search already treats 4126 as a per-75-second rate, so it and the bonus disagree with each other today:

https://github.com/zakura-core/zakura/blob/8b67f1080a50de3f80de4deab844c1453afdcc58/crates/zakura-chain/src/parameters/network/subsidy.rs#L593-L594

Could this become a function of the halving interval in force, `floor(6_931_680_000 / halving_interval(height))`, with tests pinning 4126 below NU7 and 1375 above it, and the rustdoc here plus the "starts reissuance at NU7" and `ExpectedIssuedSupply` notes (L497, L774, and `check.rs` L58) updated to the current draft? Dormant while the deficit is zero, but it is a consensus constant.

crates/zakura-consensus/src/block.rs
```diff
@@ -436,8 +436,44 @@ where
...
+            // value pools. Wait for the parent commit if its verification is still running.
+            let issuance_deficit =
+                if zakura_chain::parameters::subsidy::is_zip234_active(&network, height) {
+                    let parent_hash = block.header.previous_block_hash;
+
+                    let zs::Response::BlockInfo(parent_info) = state_service
+                        .ready()
+                        .await
+                        .map_err(|source| VerifyBlockError::Depth { source, hash })?
+                        .call(zs::Request::AwaitBlockInfo(parent_hash))
```
czarcas7ic: This wait also runs for proposal requests, since the `is_proposal()` early return is below it:

https://github.com/zakura-core/zakura/blob/8b67f1080a50de3f80de4deab844c1453afdcc58/crates/zakura-consensus/src/block.rs#L591

The RPC gets the router unwrapped, so unlike sync and inbound there is no `Timeout` around it:

https://github.com/zakura-core/zakura/blob/8b67f1080a50de3f80de4deab844c1453afdcc58/crates/zakurad/src/commands/start.rs#L717-L726

So on a network past the reissuance start, a `getblocktemplate` proposal whose `previousblockhash` this node never received, or reorged out, never returns: `AwaitBlockInfo` has no timeout and no unknown-parent exit, and each retry parks another verifier future that wakes on every commit. Before this change the state's `CheckBlockProposalValidity` rejected it immediately. `getblocktemplate` is `RpcAccess::Unauthenticated` and proposals need no PoW:

https://github.com/zakura-core/zakura/blob/8b67f1080a50de3f80de4deab844c1453afdcc58/crates/zakura-rpc/src/methods.rs#L223-L224

A `submitblock` whose queued parent later fails can pin one of the `MAX_MINED_SUBMISSIONS` slots the same way, since that work is spawned and only detached on cancellation.

Could proposals use a plain non-waiting `ReadRequest::BlockInfo` and fail fast when the parent is missing (it has to be the tip anyway), or the RPC's router get the same `Timeout` layer as sync and inbound?

crates/zakura-chain/src/parameters/network/subsidy.rs
czarcas7ic, line 474: AI comment, checked over

---

The September 16 ZIP draft now makes the reissuance fraction depend on the halving interval. With ZIP 218's 5,040,000-block interval, the numerator is 1375 rather than 4126:

https://github.com/ShieldedLabs/zips/blob/60720df9e971e19d8b6f67f1869426d78c96d250/zips/draft-judah-nsm-halving-preserving-issuance.md#L153-L165

The draft also explains why a fixed 4126 is rejected, which is the opposite of what the `BLOCK_SUBSIDY_FRACTION_DENOMINATOR` doc comment below says the draft does:

https://github.com/ShieldedLabs/zips/blob/60720df9e971e19d8b6f67f1869426d78c96d250/zips/draft-judah-nsm-halving-preserving-issuance.md#L296-L316

For example, a 100 ZEC eligible parent balance would receive 4,126 zatoshis here instead of the draft's 1,375. Could we use the height-dependent fraction for the bonus, and update the tests and that doc comment to match?

crates/zakura-consensus/src/block.rs
czarcas7ic, line 450: AI comment, checked over

---

With reissuance active, a getblocktemplate proposal referencing an unknown parent waits here before reaching the best-tip check. If that parent never commits, validation has no deadline because validate_block_proposal awaits the router directly:

https://github.com/zakura-core/zakura/blob/8b67f1080a50de3f80de4deab844c1453afdcc58/crates/zakura-rpc/src/methods/types/get_block_template.rs#L909-L918

Proposals don't require proof of work and `getblocktemplate` is served unauthenticated, so otherwise well-formed proposals with unknown parents can leave requests pending without mining a block. Could we reject missing proposal parents without waiting and add a regression test for that case?

evan-forbes:
closing in favor of the split

crates/zakura-consensus/src/block.rs
evan-forbes: fixing in the split, thanks!

crates/zakura-chain/src/parameters/network/subsidy.rs
evan-forbes: fixing in the split, thanks!

PR #869, opened 2026-09-02
fix(network): preserve replacement sessions across stale lifecycle events

czarcas7ic:
## Motivation

Delayed block-sync lifecycle events can arrive after a replacement session for the same peer. Either an old disconnect or an old connect could then remove or replace the live reactor session.

## Solution

Give each session an internal generation and apply peer lifecycle events only when that generation still owns the peer. The generation-bearing events stay on a crate-private channel, so the public `BlockSyncEvent` API, wire protocol, and `zakura-network` 7.0.0 version remain unchanged.

The reactor also retains a sender for its public control channel. Without that keepalive, moving the public handle into `BlockSyncService` after the private-channel split closed the old channel and stopped the reactor before peer admission.

## Testing

Focused regressions cover stale teardown; newer-connect, older-connect, older-disconnect ordering; the private lifecycle path bypassing the bounded driver queue; and moving the handle into the service without stopping the reactor. The pre-existing Status and supplier tests that exposed the channel-closure regression also pass locally.

### Reachability evidence

Disposable probes compared affected revision `5f1e12367` with this PR at `28b4aef9e`. Two native nodes communicated over real Iroh transport and framed stream 6. The probes delayed only service-generated lifecycle events immediately before enqueue; they did not synthesize lifecycle events.

- For delayed `Connected`, affected code replaced the collision-winning session with the older losing session in 6 of 6 runs. Fixed code preserved the winner in 6 of 6 runs. A stronger wire replay failed to complete a subsequent real GetBlocks exchange in 3 of 3 affected runs and completed it with the fix.
- Intermediate revision `19214c75e`, which had session-owned disconnect but not generation-gated connect admission, still failed. This isolates the second outcome to the delayed-connect protection in this PR.
- For stale `Disconnected`, resetting the old ordered stream and reconnecting the same authenticated peer produced a real delayed disconnect. Restoring historical unconditional removal made the replacement request `GetBlocksSpam` with no query; fixed session-owned removal preserved the replacement and its query.

Both findings are **peer-triggered, controlled schedule**. The probes bypassed only the dialer's already-connected suppression to model simultaneous open and included a testkit-only endpoint wiring correction to match production construction. The permanent deterministic regressions protect the schedules without wall-clock dependence.

## Changelog

Added the numbered bug-fix fragment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7961,7 +7961,7 @@ dependencies = [
 
 [[package]]
 name = "zakura-network"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "bitflags",
  "blake2b_simd",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7961,7 +7961,7 @@ dependencies = [
 
 [[package]]
 name = "zakura-network"
-version = "8.0.0"
+version = "7.0.0"
 dependencies = [
  "bitflags",
  "blake2b_simd",
--- a/crates/zakura-network/Cargo.toml
+++ b/crates/zakura-network/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zakura-network"
-version = "8.0.0"
+version = "7.0.0"
 authors.workspace = true
 description = "Networking code for the Zakura node. Internal crate, published to support cargo install zakura"
 # # Legal
--- a/crates/zakura-network/Cargo.toml
+++ b/crates/zakura-network/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zakura-network"
-version = "7.0.0"
+version = "8.0.0"
 authors.workspace = true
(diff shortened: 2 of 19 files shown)


evan-forbes:
Are these network-breaking or API-breaking changes? As I understand it, the wire protocol stays unchanged, so could we remove the `network breaking` label and confirm whether the internal crate API change needs a new major version?

czarcas7ic:
These are API breaking! Not network breaking, will fix

evan-forbes:
Would it make sense to keep peer lifecycle events private to `zakura-network` and reserve `BlockSyncEvent` for the public? That split could avoid exposing `session_id` publicly and remove this specific API break.

evan-forbes:
Could we also generation-gate `PeerConnected` before installing the session, since concurrent `add_peer` calls could enqueue a replacement before an older canceled session and let the stale event reinstall the older session? A focused property for this ordering would help confirm the behavior.

PR #870, opened 2026-09-02
fix(network)!: bind GetBlocks responses to live requests

czarcas7ic:
## Motivation

GetBlocks completions were matched by peer and start height, while serving capacity was tracked separately. A delayed, duplicated, or mismatched completion could therefore send blocks for an obsolete request or release capacity owned by another live request. A second handoff gap let a GetBlocks request decoded by a superseded peer routine be accepted against the replacement session.

## Solution

Assign each accepted request a distinct `BlockRangeRequestId` and keep its start height, original wire count, accepted query count, start time, and slot in one serving ledger. The state driver echoes the ID, while response framing and slot release use only metadata owned by the matching live ledger entry.

Each `ServeGetBlocks` handoff also carries the originating routine generation. The reactor drops a handoff unless that generation still owns the peer, before emitting a query, frame, or misbehavior record. This changes public action and event shapes, so `zakura-network` moves to 8.0.0; it does not change the stream-6 wire format or version.

## Testing

Focused serving tests cover stale request handoffs, stale response ownership, duplicate slot release, count and byte clamps, and queue backpressure. The stale-handoff regression queues an older and current routine request in FIFO order and proves only the current range reaches state; removing the generation guard makes it fail. One completion deliberately echoes `u32::MAX` for a one-block request and proves the ledger-owned count controls the real wire response. A zero-result state response after a count of 10 is clamped to 2 proves that `RangeUnavailable` still echoes the original count of 10. The request-ownership properties in the test PR exercise longer generated histories.

### Reachability evidence

A disposable peer process used the real Iroh transport, native handshake, ordered stream 6, and canonical Status and GetBlocks messages against a separate Regtest `BlockSyncService`. The affected source snapshot was `38538d460`; the fixed snapshot containing the response-ownership change was `6d6ab411f`. Each peer was limited to one in-flight request.

The only timing control was a delay immediately before the production driver's real state read. No response or completion event was forged, injected, or altered. Session A requested genesis and disconnected, the same authenticated identity reconnected as B and requested it again, and request C arrived while B remained pending.

- Affected code sent the delayed result for A to B, released B's sole slot from A's stale completion, and admitted C. The peer observed three `Block`/`BlocksDone` responses, no `RangeUnavailable`, and three state queries.
- Fixed code discarded A's result and kept B's slot occupied, so C received `RangeUnavailable` without a state query. The peer observed one `Block`/`BlocksDone` response and one `RangeUnavailable`.
- Two recorded affected/fixed pairs and an independent repeat produced the same split.

Both response findings are **peer-triggered, controlled timing**. The forged-identity property is **internal robustness only** because peers cannot create driver completion events over the wire.

The stale-request regression separately uses two real framed peer routines. It delays only the older routine-to-reactor handoff until the replacement is live. Without the guard, state receives the older range through the replacement; with the guard, only the replacement range reaches state. This finding is **peer-triggered, controlled timing**.

## Changelog

Updated the numbered bug-fix fragment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7961,7 +7961,7 @@ dependencies = [
 
 [[package]]
 name = "zakura-network"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "bitflags",
  "blake2b_simd",
--- a/crates/zakura-network/Cargo.toml
+++ b/crates/zakura-network/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zakura-network"
-version = "7.0.0"
+version = "8.0.0"
 authors.workspace = true
 description = "Networking code for the Zakura node. Internal crate, published to support cargo install zakura"
 # # Legal
--- a/crates/zakura-network/src/zakura/block_sync/events.rs
+++ b/crates/zakura-network/src/zakura/block_sync/events.rs
@@ -432,6 +432,8 @@ pub(super) enum RoutineToReactor {
     ServeGetBlocks {
         /// Peer that requested the range.
         peer: ZakuraPeerId,
+        /// Registry generation of the session routine that decoded the request.
+        session_generation: u64,
         /// First requested height.
         start_height: block::Height,
         /// Requested block count.
--- a/crates/zakura-network/src/zakura/block_sync/events.rs
+++ b/crates/zakura-network/src/zakura/block_sync/events.rs
@@ -1,7 +1,7 @@
 #[cfg(any(test, feature = "proptest-impl"))]
 use super::state::BlockSyncFrontiers;
 use super::{request::*, *};
-use std::num::NonZeroU64;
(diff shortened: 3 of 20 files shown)


evan-forbes:
Are these network-breaking or API-breaking changes? As I understand it, the wire protocol stays unchanged, so could we remove the `network breaking` label and confirm whether the internal crate API change needs a new major version?

czarcas7ic:
API breaking, will fix

evan-forbes:
Would a small `BlockRangeRequestId` newtype be safer than a type alias to `NonZeroU64`? A distinct type would prevent callers from mixing this ID with other `NonZeroU64` identifiers while still allowing the driver to echo it.

evan-forbes:
Could the serving ledger retain the accepted request count and use that value when processing the completion? The current handler matches the request ID and start height but trusts `requested_count` from the event, so the ledger does not yet own the complete request metadata.

PR #874, opened 2026-09-03
DRAFT: feat(network): add reusable byte regulation mechanics

czarcas7ic:
## Motivation

Message-specific P2P regulation needs reusable accounting whose ownership follows work through admission, pending queues, transport writes, cancellation, and failure. Independent counters make leaks and double releases difficult to prevent.

## Solution

Add crate-private primitives for refundable byte-rate charges, non-refilling outstanding-byte reservations, drop-released pending-input permits, and frame leases. Add an internal leased-send path behind the unchanged framed-stream API. Cancelling a partially written leased frame releases its accounting and resets only that ordered stream, leaving sibling streams usable.

This PR adds mechanics only. No message policy attaches these owners yet, so normal production traffic is unchanged.

## Testing

- 15 generic regulation tests, including generated ownership histories
- 6 framed transport compatibility and lease tests
- pending-write cancellation regression
- cargo fmt, cargo check, and clippy with warnings denied

## Changelog

Internal-only fragment in docs/changelog/unreleased/874.md.


crates/zakura-network/src/zakura/transport/io.rs
```diff
@@ -8,60 +8,410 @@
...
+        match sender.try_reserve() {
+            Ok(slot) => {
+                let lease = make_lease();
+                slot.send(QueuedFrame::leased(frame, lease));
+                Ok(())
+            }
+            Err(mpsc::error::TrySendError::Full(())) => Err(LeasedSendError::Full(frame)),
+            Err(mpsc::error::TrySendError::Closed(())) => Err(LeasedSendError::Closed(frame)),
+        }
     }
```
czarcas7ic, line 158: Where GetBlocks will call to reserve a queue slot

crates/zakura-network/src/zakura/handler.rs
```diff
@@ -3980,13 +3987,13 @@ fn spawn_persistent_stream_worker(
...
-    let (from_service_tx, from_service_rx) = mpsc::channel(queue_depth);
+    let (from_service_tx, from_service_rx) = worker_framed_channel(queue_depth);
     let admitted = AdmittedOrderedSession {
         kind: prelude.stream_kind,
         version: prelude.stream_version,
         session_id: context.stream_id,
         recv: FramedRecv::new(to_service_rx),
-        send: FramedSend::new(from_service_tx),
+        send: from_service_tx,
         cancel_token: context.stream_token.clone(),
```
czarcas7ic, line 3997: This allows a frame to carry a lease for its reserved capacity, with the reservation policy defined separately for each message type.

crates/zakura-network/src/zakura/handler.rs
```diff
@@ -4116,21 +4123,41 @@ async fn persistent_stream_worker(
...
                 match outbound {
-                    Some(frame) => {
-                        if let Err(error) = write_ordered_frame(
-                            &mut send,
-                            frame,
-                            context.limits,
-                            context.outbound_frame_cap,
-                        ).await {
+                    Some(queued_frame) => {
+                        let write_result = write_queued_frame_until_cancelled(
```
czarcas7ic, line 4140: We keep the capacity that we charged reserved here until QUIC finishes handling this frame.

crates/zakura-network/src/zakura/handler.rs
```diff
@@ -4116,21 +4123,41 @@ async fn persistent_stream_worker(
...
+                        )
+                        .await;
+                        let write_result = match write_result {
+                            QueuedFrameWriteOutcome::Completed(result) => result,
+                            QueuedFrameWriteOutcome::ConnectionCancelled => break,
+                            QueuedFrameWriteOutcome::StreamCancelled => {
+                                let _ = send.reset(VarInt::from_u32(ZAKURA_CLOSE_STREAM_CANCELLED));
+                                break;
+                            }
+                        };
```
czarcas7ic, line 4161: This lets stream cancellation immediately stop a blocked write, release its lease, and reset only the affected stream instead of the whole connection.

crates/zakura-network/src/zakura/handler.rs
```diff
@@ -4443,14 +4470,27 @@ async fn read_frame(
     let mut payload = vec![0; payload_len];
     timeout(read_timeout, recv.read_exact(&mut payload))
         .await
-        .map_err(|_| ZakuraHandlerError::Timeout("frame payload"))??;
+        .map_err(|_| ZakuraHandlerError::Timeout("frame payload"))?
+        .map_err(classify_ordered_read_error)?;
```
czarcas7ic: This teaches the receiving peer that reset code 6 closes only the cancelled service stream, not the entire peer connection.

PR #876, opened 2026-09-03
fix(rpc): segment unauthenticated and admin methods

czarcas7ic:
## Motivation

Unauthenticated Mainnet and Testnet RPC listeners registered administrative chain-control methods, allowing any network client to call `invalidateblock` or `reconsiderblock`. Normal query, transaction submission, mining, lightwalletd, and fleet-health uses should remain available without credentials by default.

## Solution

Every registered RPC method now has a fail-closed access classification. On Mainnet and Testnet, `rpc.listen_addr` without cookie authentication exposes the normal compatibility surface but omits chain-control and test methods. Authenticated primary listeners and Regtest retain the full surface. Operators that disable authentication but still need administrative methods can configure a different loopback port with `rpc.admin_listen_addr`; that listener always requires the existing cookie.

Existing authenticated deployments need no configuration change. Existing unauthenticated deployments lose `invalidateblock` and `reconsiderblock`, while methods such as `getrawtransaction`, `getinfo`, `getpeerinfo`, transaction submission, and mining remain available. Operators that want credentials on every method can set `rpc.enable_cookie_auth = true`. The unauthenticated surface is defense in depth, not Internet hardening, so its network exposure still needs an appropriate firewall, private network, or access-controlled gateway.

The `zakura-rpc` crate moves to version 9 because the new public configuration field is a breaking change for Rust struct-literal consumers.

## Testing

- `cargo clippy -p zakura-rpc -p zakura --all-targets --locked -- -D warnings`
- `cargo test -p zakura-rpc --lib --locked` (125 passed, 1 ignored)
- `cargo test -p zakura --test config --locked` (23 passed)
- `cargo test -p zakura --test acceptance invalidate_and_reconsider_block --locked -- --nocapture`
- `./scripts/changelog.py check` and repository-wide Markdown lint

The RPC tests exercise the actual HTTP listeners: unauthenticated discovery and batches omit administrative methods, current unauthenticated compatibility methods remain available, unauthenticated admin requests are rejected, authenticated admin calls remain registered, and incomplete classifications fail closed. The acceptance test confirms Regtest can still invalidate and reconsider blocks.

## Changelog

Added `docs/changelog/unreleased/876.md` under Security.

### DIFF
--- a/crates/zakura-rpc/src/server/tests/surface.rs
+++ b/crates/zakura-rpc/src/server/tests/surface.rs
@@ -1,4 +1,4 @@
-//! Tests for public and authenticated RPC method surfaces.
+//! Tests for restricted unauthenticated and authenticated RPC method surfaces.
 
 use std::{
     collections::BTreeSet,
@@ -54,15 +54,15 @@ fn access_policy_matches_the_openrpc_method_set() {
 }
 
 #[test]
-fn public_surface_contains_only_allowlisted_methods() {
+fn restricted_surface_contains_only_unauthenticated_methods() {
     let mut module = classified_module();
-    configure_rpc_methods(&mut module, RpcSurface::Public)
+    configure_rpc_methods(&mut module, RpcSurface::Restricted)
         .expect("the reviewed method classification should be complete");
 
     let actual: BTreeSet<_> = module.method_names().collect();
     let expected: BTreeSet<_> = RPC_METHOD_ACCESS
         .iter()
-        .filter_map(|(name, access)| (*access == RpcAccess::Public).then_some(*name))
+        .filter_map(|(name, access)| (*access == RpcAccess::Unauthenticated).then_some(*name))
         .collect();
 
     assert_eq!(actual, expected);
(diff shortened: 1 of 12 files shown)


crates/zakura-rpc/src/server/tests/surface.rs
```diff
@@ -0,0 +1,329 @@
...
+        module
+            .register_method(method_name, |_params, _context, _extensions| true)
+            .expect("classified method names should be unique and valid");
+    }
+
+    module
+}
+
+#[test]
+fn access_policy_matches_the_openrpc_method_set() {
```
czarcas7ic, line 41: This prevents us from forgetting to add new RPC methods to the access list when they are introduced

evan-forbes (approved):
nice do we need to change any other deployments

czarcas7ic:
> nice do we need to change any other deployments

I am only tracking the Zakura fleet we own to be redeployed after tag, there are no compatibility concerns, if that is what you are asking.

PR #877, opened 2026-09-03
docs(p2p): separate property testing from message regulation

czarcas7ic:
## Motivation

The parent proposal combines property-testing methodology, runtime traffic regulation, and speculative protocol changes. That makes it difficult to tell what is a general rule, what belongs to one exchange, and what has executable evidence.

## Solution

Separate property testing from message regulation and organize normative contracts under `docs/specs/native-p2p/`. Regulation is owned by an exchange, so one request contract covers the response work it causes while each response keeps its own wire and receiving-side contract.

Preserve the earlier Discovery v2, Header Sync v9, and block-sync declarations in exchange-specific draft specifications. The block-range exchange remains the first contract with stable `GB-WF`, `GB-SM`, and `GB-RL` requirements, but this PR does not choose concrete regulation limits, deadlines, or load schedules. Those belong to the implementation evidence.

### Review map

- `docs/design/property-testing.md`: evidence standard and claim strength
- `docs/design/peer-message-regulation.md`: traffic-control architecture
- `docs/specs/native-p2p/regulation.md`: common production requirements
- `docs/specs/native-p2p/README.md`: exchange catalog and implementation status
- `docs/specs/native-p2p/block-range.md`: scoped GetBlocks contract
- the remaining files in `docs/specs/native-p2p/`: preserved draft exchange proposals

## Testing

- Markdown lint
- Codespell

Link checks also run in docs CI.

## Changelog

Not needed; this PR changes documentation only.

This PR targets the open peer-message-regulation documentation branch so the final parent diff can become the shared foundation for the implementation stack.

### Follow-up Work

The implementation stack supplies the concrete GetBlocks limits, workloads, and measurements without expanding this shared contract.


docs/specs/native-p2p/block-range.md
```diff
@@ -0,0 +1,324 @@
+# Block-range exchange (`GetBlocks`)
```
czarcas7ic, line 1: This is the one that is fully specd, the others are just retaining the information that Evan originally added in their respective files that will get expanded similar to this one when the time comes to work that message.

czarcas7ic:
@codex review

czarcas7ic:
@codex review

czarcas7ic:
@codex review

czarcas7ic:
@codex review

czarcas7ic:
@codex review

docs/specs/native-p2p/block-range.md
```diff
@@ -0,0 +1,356 @@
...
+  schema-versioned scenarios, direct replay, and repeat-run comparison.
+- A successor stream version may add a wire request ID and block hashes to make
+  response ownership explicit.
+
+## Implementation evidence
+
+The GetBlocks properties above specialize the shared requirements in
+[the regulation specification](regulation.md). The implementation's
+machine-checked test manifest is the authoritative mapping from requirement IDs
+to tests.
```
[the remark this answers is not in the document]
czarcas7ic: This is intentionally left to the implementation evidence. This PR keeps the layer at Specified; the implementation PR adds the shared-to-exchange mapping and machine-checked test manifest before marking it Implemented. That keeps the evidence synchronized with the implementation rather than duplicating a future mapping here.

docs/specs/native-p2p/block-range.md
```diff
@@ -0,0 +1,356 @@
...
+  behavior.
+- **All:** invariants checked after each settled step.
+
+| ID | Class | Requirement |
+| --- | --- | --- |
+| GB-SM-01 | Peer | A replacement connection cancels the preceding session for the same peer. |
+| GB-SM-02 | Peer | A stale disconnect does not close or mutate the current session. |
+| GB-SM-03 | Peer | A peer without retained valid `Status` cannot start a request; the attempt is recorded as `GetBlocksSpam`. |
+| GB-SM-04 | All | Each peer has an independent committed-request ledger bounded by the configured local in-flight cap. |
+| GB-SM-05 | Peer | A request rejected by the full committed-request ledger emits no state query and receives `RangeUnavailable` echoing its original wire count while output capacity is available. |
```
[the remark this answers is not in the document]
czarcas7ic: No change planned here. Committing the resource permit is not the same as successfully inserting the request into the serving ledger. The session-gated attempt commits and binds its permit first; if ledger insertion rejects it, that same bounded permit funds the RangeUnavailable frame lease and then settles.

docs/specs/native-p2p/block-range.md
```diff
@@ -0,0 +1,356 @@
...
+| GB-RL-07 | Time refills rate tokens but never outstanding-byte capacity. |
+| GB-RL-08 | Handoff, action, driver, and output failures settle each attempt or permit exactly once and never blame the peer for local failure. |
+| GB-RL-09 | Session end settles permits without moving them to a replacement; frame leases survive until their frames leave the application transport. |
+| GB-RL-10a | Generated hostile histories vary peer count and every configured bound without exceeding peer or node accounting. |
+| GB-RL-10b | A named native reading-flood workload preserves useful service for an honest peer within the declared local deadline. |
+| GB-RL-10c | A named native stopped-reader workload stays within the declared application and transport envelopes, releases ownership after failure or timeout, and restores honest service. |
+| GB-RL-11 | Responses to Zakura's downloads continue within the request timeout behind admission-delayed serving requests on the same stream. |
+| GB-RL-12 | Supported configurations use checked arithmetic, fit the largest legal request, one maximum-size block, and one session's pending-input window, and reject insufficient limits or capacities. |
```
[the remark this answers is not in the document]
czarcas7ic: No change planned here. P2P-RG-11 already normatively requires aggregate startup validation, and GB-RL-10c covers the corresponding GetBlocks transport-envelope evidence. GB-RL-12 does not need to repeat that shared transport requirement.

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
-# Zakura peer message regulation: design
+# Native P2P message regulation: design
```
evan-forbes, line 1: meh

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
-> rules. This document gives the reasons and the shape of the implementation. Scope is the three
-> Zakura native reactors: discovery, header sync, block sync.
+> **Status: proposal.** The
+> [message-regulation specification](../specs/native-p2p/regulation.md)
+> defines the required behavior. This document explains why the controls are
+> separate and how they compose. The
+> [property-testing design](property-testing.md) defines how that behavior is
+> tested.
+
+Message regulation is traffic control for Zakura's native P2P services. It
```
evan-forbes, line 10: not a fan of "traffic control" here as that can mean congestion control, which we're not wanting to handle here

why not just use a single term here and avoid explaining: ambiguous term A is ambiguous term B

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
-4. **Budgeted** — The message fits the work bound for its role.
+1. **Safe** — Can it frame, allocate, decode, and validate the message within
+   fixed bounds?
+2. **Expected** — If it is a response, does it match a live request or
+   reservation?
+3. **Actionable** — If it is valid but stale or unable to affect Zakura's state,
+   should it be ignored without penalizing the peer?
+4. **Affordable** — Can the peer and node admit all work and response bytes
+   caused by the exchange? Otherwise, use the contract's declared wait, reject,
+   or drop outcome.
```
evan-forbes, line 33: personal opinion I guess but questions don't belong here, just state what we are trying to say

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
+> [message-regulation specification](../specs/native-p2p/regulation.md)
+> defines the required behavior. This document explains why the controls are
+> separate and how they compose. The
+> [property-testing design](property-testing.md) defines how that behavior is
+> tested.
+
+Message regulation is traffic control for Zakura's native P2P services. It
+decides how much peer-caused work may enter the node, how long that work owns
+capacity, and what happens at a limit. Property testing is one way to verify
+those decisions; it is not the regulation itself.
```
evan-forbes, line 13: we should describe the problem first, why was the previous intro lacking here?

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
-2. **Authorized** — Each response matches a live reservation created by a request Zakura sent.
-3. **Useful** — The message can affect a receiver decision.
-4. **Budgeted** — The message fits the work bound for its role.
+1. **Safe** — Can it frame, allocate, decode, and validate the message within
+   fixed bounds?
+2. **Expected** — If it is a response, does it match a live request or
+   reservation?
+3. **Actionable** — If it is valid but stale or unable to affect Zakura's state,
+   should it be ignored without penalizing the peer?
+4. **Affordable** — Can the peer and node admit all work and response bytes
```
evan-forbes, line 31: why change terms here

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
 
-1. **Safe** — Frame checks precede allocation, and bounded decoding precedes expensive verification.
-2. **Authorized** — Each response matches a live reservation created by a request Zakura sent.
-3. **Useful** — The message can affect a receiver decision.
-4. **Budgeted** — The message fits the work bound for its role.
+1. **Safe** — Can it frame, allocate, decode, and validate the message within
+   fixed bounds?
+2. **Expected** — If it is a response, does it match a live request or
+   reservation?
+3. **Actionable** — If it is valid but stale or unable to affect Zakura's state,
```
evan-forbes, line 29: actionable is worse here than usesful imo

receiving a block part is not always actionable for example. inherently sometimes the thing to do is nothing but wait. its useful. 

right we're wanting to prevent DoS, therefore is some message is useful, inherently its not spam.

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
+- A **request** asks Zakura to perform work and reply, such as `GetBlocks`.
+  Zakura limits the request and all work and response bytes it can cause.
+- A **response** answers a request Zakura previously sent, such as `Block`.
+  Zakura accepts it only when it is valid and matches the request.
 
-Each message declaration builds an admission path from reusable filters and message-specific
-configuration. The declaration selects the message role, the applicable filters, and their bounds
-and keys. The filters implement the shared admission behavior. The declaration contains the policy
-that differs between messages.
+### When a peer requests data from Zakura
```
evan-forbes, line 56: pure opinion but I have a strong distates for ai seo maxing these headings and its usually one of the first things I change or include in a skill

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
+    participant P as Peer requester
+    participant Z as Zakura responder
+    P->>Z: Send GetBlocks
+    Z->>Z: Validate and reserve peer and node capacity
+    Z->>Z: Query state
+    loop For each response frame
+        Z->>Z: Move reserved bytes into a frame lease
+        Z->>P: Send Block or terminal response
+    end
+    Z->>Z: Settle request accounting
```
evan-forbes, line 74: the purpose of this early flow is to be general, therefore I don't think we should be getting into specific messages here

I also don't see what value we're adding with things like:
```
Z->>Z: Query state
```
besides not always being true as an implementation detail, it doesn't add important details.

docs/design/peer-message-regulation.md
```diff
@@ -1,226 +1,294 @@
...
-}
+```text
+frame header and per-message cap
+  → bounded decode and wire validation
+  → peer/session and protocol prerequisites
+  → provisional resource admission
+  → reactor or service handoff
+  → committed handler work
+  → leased response frames
+  → settlement
```
evan-forbes, line 89: word salad imo

I don't find this accurate or useful

evan-forbes (commented):
submitting the first part of the review but tbh I disagree with the direction of the design document, maybe I'm missing some rationale

docs/design/peer-message-regulation.md
evan-forbes: why suddenly start using a different name

docs/design/peer-message-regulation.md
evan-forbes: this doesn't need to attract views on a blog, therefore should instead focus on actually organizing the document in a clear standard way

PR #882, opened 2026-09-04
feat(metrics): report first block receipt source

czarcas7ic:
## Motivation

Operators need to know whether Zakura or legacy TCP delivered each requested full block body first. Commit order is not a reliable proxy for arrival order.

## Solution

Expose `sync.block.first_received.count` with a bounded `source="zakura"` or `source="legacy"` label. Native block sync and the dual-stack legacy service share one per-node first-seen tracker, so the first complete, hash-matching body for a block wins and a later duplicate from the other transport does not increment the metric.

This records receipt before consensus verification or canonical-chain commitment. It measures block propagation, including competing blocks, rather than verifier scheduling.

## Testing

Formatting, clippy, changelog validation, and focused network tests pass. The remaining network suite passes locally apart from three unchanged macOS loopback-alias tests that this host cannot bind. Focused tests cover both source orderings, native receipt, legacy receipt, Zakura fallback, late matching bodies, and rejection of unrequested bodies.

## Changelog

Added the new operator metric to the unreleased changelog.

[Operator discussion](https://zcashzcaloors.slack.com/archives/C0ALB378NLX/p1788467414801909)

### AI Disclosure

OpenAI Codex was used to implement and test this change and prepare the PR description.

### DIFF
--- a/crates/zakura-network/src/zakura/block_sync/peer_routine.rs
+++ b/crates/zakura-network/src/zakura/block_sync/peer_routine.rs
@@ -45,7 +45,9 @@ use super::{
     BlockSyncAction, BlockSyncMessage, BlockSyncMisbehavior, BlockSyncPeerSession, BlockSyncStatus,
     ZakuraBlockSyncConfig, ZakuraPeerId, ZakuraTrace, MSG_BS_BLOCK,
 };
-use crate::zakura::{Admit, FramedRecv, OrderedSendError, SinkReject, ZakuraConnId};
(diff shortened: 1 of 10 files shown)


czarcas7ic:
@codex review

ValarDragon (approved):

ValarDragon:
Trace does make sense as a filename, though previously I thought that meant the other trace feature we had. If this works with no tracing, seems good to me. @evan-forbes can comment later on if we should move that code!

PR #884, opened 2026-09-04
fix(sync): retry timed-out parked-block refreshes

evan-forbes:
## Motivation

Continuous sync test 5 repeatedly downloaded the same partial checkpoint range. The verifier parked those blocks until the range completed. A 45-second tip refresh timeout then ended the sync round and canceled every parked block.

PR #137 added tip refreshes for partial checkpoint ranges. Closed PR #141 explored preserving the checked frontier after empty extensions. Neither change preserved a live sync round when the refresh itself timed out.

## Solution

Keep the sync round alive after a tip refresh timeout. Retry discovery while the existing verifier work remains parked. Preserve the existing verifier timeout as the final bound for a genuinely stalled round. Propagate state-service and other refresh errors so they still restart the sync round.

## Testing

- `cargo test -p zakura --lib components::sync::tests::vectors::` (48 passed)
- `cargo test -p zakura --lib incomplete_checkpoint_range_retries_refresh_timeout_without_verifier_timeout`
- `git diff --check`

The regression test parks a partial checkpoint range, stalls the first refresh for the production 45-second timeout, and verifies that the next refresh completes the range before the verifier timeout.

## Changelog

Added `docs/changelog/unreleased/884.md`.

- Continuous sync test 5 incident on 2026-09-03 at height 3,453,371.
- Prior fixes: #137 and closed draft #141.

### Follow-up Work

- Deploy this branch to continuous sync test 5 after review and confirm that later refresh attempts complete the parked checkpoint range.

### DIFF
--- a/crates/zakurad/src/components/sync.rs
+++ b/crates/zakurad/src/components/sync.rs
@@ -1681,10 +1681,7 @@ where
 
                         let refreshed =
                             timeout(SYNC_RESTART_DELAY, self.obtain_tips(checkpoint_bootstrap))
-                                .await
-                                .map_err(Into::into)
-                                // TODO: replace with flatten() when it stabilises (#70142)
-                                .and_then(convert::identity)?;
+                                .await;
 
                         // A refresh is not progress, even when it returns hashes.
                         //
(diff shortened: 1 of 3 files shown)


crates/zakurad/src/components/sync.rs
```diff
@@ -1681,10 +1681,7 @@ where
 
                         let refreshed =
                             timeout(SYNC_RESTART_DELAY, self.obtain_tips(checkpoint_bootstrap))
-                                .await
-                                .map_err(Into::into)
-                                // TODO: replace with flatten() when it stabilises (#70142)
-                                .and_then(convert::identity)?;
```
ValarDragon, line 1687: ahh this is the change

ValarDragon (approved):
LGTM! Not merging, b/c draft

PR #892, opened 2026-09-04
feat(network)!: regulate GetBlocks with paired streams

czarcas7ic:
## Motivation

This draft is retained as the reference for the replacement stack: #942 (owned storage), #943 (generic paired transport), #944 (outgoing request ownership), and #945 (serving activation). The initial stack preserved this PR at `953b85b18`; #941 was later closed and its backport removed from the replacement drafts in favor of the separate Iroh upgrade in #935. This reference still records the original backport-based implementation. See the [audit map](https://github.com/zakura-core/zakura/blob/d3485291b/docs/design/getblocks-stack.md).

## Solution

Use paired native block-sync streams and sequential serving, with one response slot per authenticated identity and 64 per node. Retain capacity through storage, results, encodes, and writes across cancellation and reconnects. Settle outgoing requests across publication, expiry, reset, and writer startup. Preserve cooldowns for unanswered remote resets and defer session-capacity races without closing healthy connections. The previous native layout is replaced while other services and legacy fallback remain available. Property coverage remains separate in #896.

## Testing

The latest two fixes pass 29 targeted regressions and network Clippy. Before withdrawing the backport, the identical completed stack passed 405 focused tests, 22 cluster integration tests without retries, and workspace all-target Clippy, formatting, Markdown lint, and changelog checks. Those results do not validate the replacement stack with the separate Iroh upgrade; activation needs that integration check.

Earlier long transport results and their limitations remain in the [validation evidence](https://github.com/zakura-core/zakura/blob/3b8f8caf1/docs/design/getblocks-refactor-results.md). Those long gates were not rerun. The earlier full workspace run had three unavailable loopback source addresses and two legacy mainnet-sync timeouts; it was not repeated for this split.

## Changelog

The combined fragment remains here; each replacement draft has its own fragment. Used Codex for implementation, tests, and documentation.


czarcas7ic:
@codex review

crates/zakura-network/src/zakura/block_sync/serving_regulation.rs
```diff
@@ -0,0 +1,1281 @@
...
+        requested_count: u32,
+        mut acquired: Option<AcquiredAdmissionSlot>,
+    ) -> Result<AdmissionAttempt, AdmissionBlocked> {
+        let cost = match serving_cost(&self.regulator.inner.config, requested_count) {
+            Ok(cost) => cost,
+            Err(error) => panic!(
+                "GetBlocks serving arithmetic remains valid after configuration validation: {error}"
+            ),
+        };
+        let peer_rate = reserve_rate(BoundKind::PeerRate, &self.peer_rate.budget, cost.charge)?;
```
czarcas7ic: A small GetBlocks request can cause megabytes of block data to be read and queued. Reserving its maximum cost before starting the read bounds aggregate serving work, including when requests come from many different peers.

crates/zakura-network/src/zakura/block_sync/reactor.rs
```diff
@@ -1449,69 +1550,130 @@ impl BlockSyncReactor {
...
-        let local_inflight_cap = self.startup.config.advertised_max_inflight_requests();
-        if !self.state.peers.contains_key(&peer) {
-            self.report_misbehavior(peer, BlockSyncMisbehavior::GetBlocksSpam)
-                .await;
+        // A routine can be superseded after reserving resources but before the
+        // reactor receives its message. Dropping the stale attempt rolls back
+        // every reservation without running state work or emitting a frame.
+        debug_assert_eq!(attempt.peer(), &peer);
+        let session_id = attempt.session_id();
+        if !self.registry.owns_generation(&peer, session_id)
```
czarcas7ic: A peer can reconnect while a request from its previous connection is still queued. This check discards that old request and returns its reservations, keeping it from being served through the replacement connection.

crates/zakura-network/src/zakura/block_sync/peer_routine.rs
```diff
@@ -586,6 +647,66 @@ impl PeerRoutine {
...
+    }
+
+    /// Queue a request or install a bounded wait without suspending the routine.
+    fn retain_serving_request(&mut self, start_height: block::Height, count: u32) {
+        match self.serving.try_retain_input(start_height, count) {
+            Ok(request) => self.enqueue_serving_request(request),
+            Err(blocked) => {
+                self.record_pending_input_delay(&blocked);
+                let serving = self.serving.clone();
+                self.pending_input = Some(Box::pin(async move {
```
czarcas7ic: When the pending queue fills, the routine still needs to move earlier requests forward to free space. This pauses new stream reads while the main loop continues processing queued requests and enforcing a local timeout.

crates/zakurad/src/commands/start/zakura/block_sync_driver.rs
```diff
@@ -33,6 +33,108 @@ use super::{
...
+            warn!(?peer, "timed out reading Zakura block-sync serving range");
+            trace.trace_block_range_finished(&peer, start, count, 0);
+            let _ = block_sync.send_control(BlockSyncEvent::BlockRangeResponseFinished {
+                request_id,
+                peer,
+                start_height: start,
+                requested_count: count,
+                returned_count: 0,
+            });
+            drop(query.await);
```
czarcas7ic: A query timeout does not stop a blocking storage read that is already running. That read keeps its reservation until it actually finishes, so timed-out work still counts against the node's serving limits.

crates/zakura-network/src/zakura/block_sync/service.rs
```diff
@@ -177,6 +215,83 @@ impl BlockSyncPeerSession {
...
+            )))
+        })?;
+        if !permit.can_transfer_frame(accounted_bytes) {
+            return Err(OrderedSendError::Encode(Box::new(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "encoded GetBlocks response exceeded its admitted byte cap",
+            ))));
+        }
+        self.send
+            .try_send_leased(frame, || permit.transfer_frame(accounted_bytes))
```
czarcas7ic: A completed request can still have response frames buffered in memory. The byte reservation moves to those frames, so they still count against peer and node limits until the application write finishes or the frames are discarded.

crates/zakura-network/src/zakura/block_sync/service.rs
```diff
@@ -177,6 +215,90 @@ impl BlockSyncPeerSession {
...
+
+    /// Transfer response bytes only when a transport queue slot is available.
+    pub(super) fn try_send_regulated_message(
+        &self,
+        msg: BlockSyncMessage,
+        permit: &mut super::serving_regulation::GetBlocksServingPermit,
+    ) -> Result<(), OrderedSendError> {
+        let (frame, accounted_bytes) = Self::encode_regulated_message(msg, permit)?;
+        self.send
+            .try_send_leased(frame, || permit.transfer_frame(accounted_bytes))
```
czarcas7ic: A completed request can still have response frames buffered in memory. The byte reservation moves to those frames, so they still count against peer and node limits until the application write finishes or the frames are discarded.

crates/zakura-network/src/zakura/block_sync/reactor.rs
```diff
@@ -1449,69 +1567,130 @@ impl BlockSyncReactor {
...
         if !self.registry_received_status(&peer) {
-            self.report_misbehavior(peer, BlockSyncMisbehavior::GetBlocksSpam)
-                .await;
             return;
         }
 
-        if count == 0 {
-            self.report_misbehavior(peer, BlockSyncMisbehavior::GetBlocksTooLong)
-                .await;
+        let mut permit = attempt.commit();
```
czarcas7ic: Reservations remain refundable until the reactor accepts the request for the current connection. Acceptance spends the fixed request-processing charge even if no blocks are returned, so empty responses still consume rate allowance.

crates/zakura-network/src/zakura/block_sync/reactor.rs
```diff
@@ -1449,69 +1567,130 @@ impl BlockSyncReactor {
...
         &mut self,
+        request_id: BlockRangeRequestId,
         peer: ZakuraPeerId,
         start_height: block::Height,
-        requested_count: u32,
+        reported_requested_count: u32,
         blocks: Vec<(block::Height, Arc<block::Block>, usize)>,
     ) {
-        let prepare_elapsed = self.serving_blocks_elapsed(&peer, start_height);
+        let Some(mut request) = self.finish_serving_blocks(&peer, request_id, start_height) else {
```
czarcas7ic: A storage result can arrive after its request has ended, while another request for the same height is active. Matching the unique request ID prevents that old result from completing the newer request or sending its blocks through a replacement connection.

crates/zakura-network/src/zakura/block_sync/reactor.rs
```diff
@@ -1998,48 +2187,109 @@ impl BlockSyncReactor {
...
+                    _ => metrics::counter!("sync.block.unavailable.serve_queue_full").increment(1),
+                }
+                self.trace_message_sent(peer, &message, "waiting", started.elapsed());
+                // A terminal message is required even after a partial response.
+                // Keep the permit until queue admission, cancellation, or the
+                // local deadline; never resolve this wait through a newer session.
+                let peer = peer.clone();
+                let cancelled = session.cancel_token();
+                let deadline = time::Instant::now() + self.startup.config.request_timeout;
+                self.pending_serving_terminals.push(Box::pin(async move {
```
czarcas7ic: A full outbound queue can prevent even the message that tells the requester its response is finished from being queued. This retains that message and its reservations while the reactor continues other work, with cancellation and a deadline bounding the wait.

evan-forbes:
one thing we have to think about when merging to main is that if we cut a releases imbetween work here we'll have to update the blocksync / headersync / discovery versions a few times which is cumbersome

crates/zakura-network/src/zakura/block_sync/config.rs
```diff
@@ -316,6 +342,61 @@ pub struct ZakuraBlockSyncConfig {
...
+    /// Fully retained GetBlocks queue entries across all peers. A full queue
+    /// additionally permits one blocked decoded input per live session.
+    pub node_pending_requests: usize,
+    /// State queries and responses that may remain active across all peers.
+    pub node_active_requests: usize,
+    /// Response deadline for a state query. Timed-out reads keep their resource
+    /// charges until the underlying state work finishes.
+    #[serde(with = "humantime_serde")]
+    pub query_timeout: Duration,
+}
```
evan-forbes, line 338: one of the ideas in 747 was to reuse a lot of these logic instead of making one per message

not sure if we can do this for everything but many of these seem like they could be generalized

I'm biased, and not sure how feasible it is irl, but the idea of defining the message then adding and configuring the functions then and there seems nice

crates/zakura-network/src/zakura/block_sync/config.rs
```diff
@@ -316,6 +342,61 @@ pub struct ZakuraBlockSyncConfig {
...
+/// drops, or when its queued frame leaves the application-owned transport path.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, default)]
+pub struct GetBlocksRegulationConfig {
+    /// Byte-equivalent work retained for each admitted request regardless of response size.
+    pub request_overhead_bytes: u64,
+    /// Sustained byte-equivalent serving allowance for one authenticated peer.
+    pub peer_rate_bytes_per_second: u64,
+    /// Burst allowance retained across reconnects for one authenticated peer.
+    pub peer_rate_capacity_bytes: u64,
```
evan-forbes: I think we can get rid of these byte/s rate limits

one of our main goals is to have high throughput while not being DoS'd

if we are getting DoS'd, then limits per peer don't do much as its easy to spin up many peers (also at 1Gib/s this would not be helpful for most connections after one or two by default.

the rest is handled by Quic's congestion control fairness algo, where one peer is limited in how much bandwidth they can absorb relative to the other peers

the best thing here would be a prioritization mechanism.

not opposed to limits if user's ask for them, but then they should global, and not burst. rate limiting messages could also be simpler in that case. while we're low throughput this isn't an issue really, but as we get higher throughput, then users w/ metered bandwidth will likely want to avoid blocksync all together

crates/zakura-network/src/zakura/block_sync/peer_routine.rs
```diff
@@ -320,6 +325,22 @@ pub(super) struct PeerRoutine {
...
     routine_to_reactor: mpsc::Sender<RoutineToReactor>,
+    /// Completion of the one serving admission task currently waiting to hand
+    /// a request to the reactor.
+    pending_serving: Option<oneshot::Receiver<ServingAdmissionOutcome>>,
+    /// Decoded serving requests held behind `pending_serving`.
+    ///
+    /// Each entry owns its session and node pending-input slots. Once those
+    /// hard bounds fill, the routine holds one decoded request and stops reading
+    /// until capacity returns rather than dropping it or growing this queue.
+    queued_serving: VecDeque<PendingGetBlocksRequest>,
```
evan-forbes: is the reason to add this just for rate limiting messages?

if so, one tool that is useful is to use the quic buffer, or that was a our previous strategy. therefore, if we're strictly adding this for rate limiting then perhaps we can avoid another queue, keep the quic queue and window size etc full and instead only keep track of the rate, then decode lazily. this way when we back up the sender cannot send more bytes and throttles themselves

evan-forbes (commented):
nice!

crates/zakura-network/src/zakura/block_sync/config.rs
czarcas7ic: This is now reduced to the following config:

```
pub struct GetBlocksRegulationConfig {
    pub node_active_requests: usize,
    pub query_timeout: Duration,
}
```

Because we might want these to differ per message.

crates/zakura-network/src/zakura/block_sync/peer_routine.rs
czarcas7ic: The primary reason was that this allows us to keep receiving from the peer if we requested from them while block serving if we are too busy. But I don't think its worth the complexity so will go with your suggestion here!

czarcas7ic (commented):
NOTE: On the four questions we want to ask per message:

Safe
Authorized
Useful
Budgeted

For GetBlocks, I did not yet implement repeated range rejection from a single peer. So right now a single peer could keep asking for the same range. 

I think this is fine for now, since otherwise we need to start tracking what peers have asked us for (and vice versa), but wanted to flag.

crates/zakura-network/src/zakura/regulation/outstanding_bytes.rs
```diff
@@ -1,255 +0,0 @@
-//! Linear accounting for response bytes until their transport owner drops them.
```
czarcas7ic, line 1: As discussed, removing byte budgeting from our regulation package.

crates/zakura-network/src/zakura/handler.rs
```diff
@@ -148,8 +148,9 @@ pub const ZAKURA_DUPLICATE_EVICT_MIN_AGE: Duration = Duration::from_secs(300);
 /// resolution (milliseconds) so a genuine race keeps the transcript-tiebreak
 /// winner instead of flapping.
 pub const ZAKURA_SAME_IP_DUPLICATE_EVICT_MIN_AGE: Duration = Duration::from_secs(5);
-/// QUIC stream receive window used by Zakura endpoints.
-pub const DEFAULT_ZAKURA_STREAM_RECEIVE_WINDOW: u32 = 32 * 1024 * 1024;
+/// A paused stream may consume at most half the connection receive window,
+/// leaving credit for another service stream.
+pub const DEFAULT_ZAKURA_STREAM_RECEIVE_WINDOW: u32 = 16 * 1024 * 1024;
```
czarcas7ic, line 155: We reduced the per stream receive window to 16 MiB so a paused GetBlocks stream can’t use the entire 32 MiB connection window and stall other services on that peer connection.

crates/zakurad/src/commands/start/zakura/block_sync_driver.rs
```diff
@@ -489,90 +591,15 @@ pub(crate) async fn drive_block_sync_actions<ReadState, BlockVerifier>(
                     }
                 }
             }
-            BlockSyncAction::QueryBlocksByHeightRange { peer, start, count } => {
-                trace.trace_block_range_query_started(&peer, start, count);
-                let started = Instant::now();
-                match tokio::time::timeout(
-                    ZAKURA_BLOCK_SYNC_DRIVER_TIMEOUT,
```
czarcas7ic, line 496: Just to note, this used to be 30s timeout but we now specify the GetBlocks specific 8s timeout from config.rs line 110. Wanted to point this out explicitly.

czarcas7ic:
Saving these as follow ups

Screenshot 2026-09-07 at 5 58 02 PM

crates/zakura-network/src/zakura/block_sync/peer_routine.rs
```diff
@@ -2210,6 +2323,65 @@ fn outstanding_unreceived_through(
...
+                    "sync.block.serving.delayed",
+                    "bound" => blocked.kind().label()
+                )
+                .increment(1);
+                tracing::trace!(
+                    peer = ?peer,
+                    bound = blocked.kind().label(),
+                    "delaying GetBlocks at the work-admission bound"
+                );
+                acquired_slot = Some(blocked.wait().await);
```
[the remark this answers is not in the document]
czarcas7ic: I believe this is just following the design:

https://github.com/zakura-core/zakura/blob/4b5cab8936733c2944bf6d7e26e365030cc25c4a/docs/specs/peer-message-regulation.md?plain=1#L250-L253

crates/zakura-network/src/zakura/block_sync/config.rs
czarcas7ic: I left this unresolved because I could easily be convinced that maybe we just keep these global as well.

evan-forbes:
while talking with ai through it, it dug and noticed

<details>
```
**Recommendation: request changes for F001.** The update resolves the earlier queue and
admission-abstraction concerns. I reproduced a bidirectional serving stall through production
QUIC workers, peer routines, and reactors. The existing targeted suites pass.

## F001 — P1: break the dependency between paused reads and response writes

Primary location:
[`peer_routine.rs:494–504`](https://github.com/zakura-core/zakura/blob/fbf466b07dbc0e86c3ad1e8ae33c2f9e94f8a647/crates/zakura-network/src/zakura/block_sync/peer_routine.rs#L494-L504).

When both peers send a burst of GetBlocks before their response frames, each routine can stop
reading at its next request while its previous response owns the single producer. Each transport
reader fills its application channel with the remaining requests. Each response writer then fills
the remote stream receive window. Neither routine can consume the queued requests or reach the
response frames until its producer returns. Neither producer returns until its write progresses.

The ownership edge comes from
[`ResponsePermit::frame_guard`](https://github.com/zakura-core/zakura/blob/fbf466b07dbc0e86c3ad1e8ae33c2f9e94f8a647/crates/zakura-network/src/zakura/regulation/request.rs#L231-L239).
GetBlocks fixes the identity producer count at
[one](https://github.com/zakura-core/zakura/blob/fbf466b07dbc0e86c3ad1e8ae33c2f9e94f8a647/crates/zakura-network/src/zakura/block_sync/serving_regulation/policy.rs#L19-L20).
Independent writer tasks cannot resolve a dependency on the other routine resuming reads.

The local eight-second grace returns download work but leaves `pending_serving` in place.
The ten-second transport write timeout eventually cancels the connection. Thus the timeout turns
the stalled exchange into a disconnect; it does not restore useful progress on that connection.

### Reproduction

The harness sends a Status and unique, nonoverlapping, codec-valid one-block requests in each
direction. Each burst stays below the advertised 32,000-request limit. Both state adapters return
immediately. Production code performs admission, encoding, response delivery, and QUIC writes.

| Run | Requests per direction | Application queue per direction | Observed result |
| --- | ---: | ---: | --- |
| Production services, smaller negotiated queue | 128 | 32 | Total state queries stop at 18; both sessions remain stalled after the eight-second grace. |
| Production services, default two-ordered-stream queue split | 4,096 | 2,048 | Total state queries stop at 18; the transport write timeout then cancels the connection. |

The second run uses the default 16 MiB stream receive window, 32 MiB connection receive window,
32 MiB send window, 64 node producers, and 2,048-message/s transport bucket. Queue depth 2,048
matches the default 4,096-entry aggregate split across two ordered streams. The mock storage
returns 1,901,402-byte serializable blocks. Their repeated transactions are a size fixture, not
consensus-valid chain data. No disk, consensus validation, BBR-generated traffic, or full-node
sync runs in this reproduction. The result establishes a legal-request/response-size dependency
failure, not its frequency in an honest fleet.

Evidence:

- [Production reproduction](repro-bidirectional-production.log).
- [Default-queue reproduction and disconnect](repro-bidirectional-default-queues.log).
- [Reproduction source](repro-bidirectional-production.rs).
- [Patch and run instructions](reproduce-updated.md).

Before merge, define a progress rule for simultaneous request bursts that preserves bounded
query, result, and write ownership. Test it with both endpoints running production admission and
with request bursts large enough to fill the application channels. The rule must account for
requests and responses sharing one ordered stream. PR 747 should document any necessary sender
obligation or change to slot lifetime.

## Performance assessment

The update improves cross-peer storage concurrency by moving serving reads out of the serial
driver loop. The producer bound also limits retained query results and response frames. I found
no second demonstrated performance bug. Two measurements remain important:

1. **Receive-window throughput.** `handler.rs:153` halves every native stream's receive window
   from 32 MiB to 16 MiB. The dependency's `TransportConfig::stream_receive_window` documentation
   requires sizing this window against latency times desired throughput. The ideal flow-control
   ceiling at 200 ms RTT falls from about 1.34 Gbit/s to 0.67 Gbit/s. These are window/RTT estimates,
   not benchmark results. Size the stream and connection windows together against target paths
   and memory limits. Test sibling-stream progress at the same time.
2. **Concurrent serving memory and CPU.** At defaults, 64 response payload caps total about
   122 MiB. With 32 MiB responses, that envelope approaches 2 GiB. Decoded blocks, the extra block
   inspected at a byte boundary, serialization, and QUIC buffers add memory. Measure concurrent
   serving alongside block verification and commit work. The PR correctly avoids claiming an
   RSS bound or whole-node overload protection.
```
</details>

<details>
```md
# PR 892: performance, retained state, and correctness

Reviewed on 2026-09-08 at `fbf466b07dbc0e86c3ad1e8ae33c2f9e94f8a647`.
GitHub still reports this head. The review includes the surrounding download code.
The checkout is `/home/evan/src/valar/zakura.pr892-audit`.

## Findings and priorities

### 1. P1: the bidirectional progress failure remains

The [previous audit](audit-updated.md#f001--p1-break-the-dependency-between-paused-reads-and-response-writes)
reproduced this failure with production transport, routines, and reactors. Each peer
can pause reads while waiting for its preceding response to release a producer.
Both response writers can then exhaust the remote receive windows. The write timeout
disconnects the peers. This head has not changed since that reproduction.

The cleanup below does not fix this failure. A protocol or scheduling change must
break the dependency while preserving bounded response ownership. Increasing queue
sizes or replacing a ledger with a Bloom filter cannot establish progress.

### 2. P2: download registry publication repeatedly processes the entire window

Locations at the reviewed head:

- `peer_routine.rs:1213–1223`: append one request, then publish the entire window.
- `peer_routine.rs:2213–2239`: allocate a new BTreeMap and visit every retained height.
- `work_queue.rs:873–883`: acquire the shared work mutex for each owner lookup.
- `peer_registry.rs:664–676`: replace the old map under the shared registry mutex.
- `state.rs:599–600,656–660`: rescan outstanding reservations for byte-window decisions.

For an empty window filled to N one-block requests without completions, registry
publication alone visits N(N+1)/2 heights. At N=32,000, that means 512,016,000 visits.
It also performs tree insertions and per-height mutex acquisitions. Completions
rebuild the map again. Byte mode, which is the default, adds repeated reservation
scans. Actual window occupancy depends on BBR, byte limits, and peer progress.
This calculation describes the algorithm; it is not a measured full-sync result.

This problem predates PR 892. The PR adds full publication when admission grace
changes deadlines. The main cost comes from maintaining an exact derived copy,
not simply from choosing Vec for the authoritative requests.

Follow-up: publish generation- and owner-checked insert/remove/deadline updates.
Keep one authoritative reservation transition path and maintain its byte total.
The implementation must preserve the registry's protection against duplicate
issuance during rejection and reset races. Moving publication to the end of a
fill loop without examining those races is insufficient.

### 3. P2: request lookup and removal scale with the entire window

`state.rs:783–792` linearly searches the outstanding Vec by height or start height.
`peer_routine.rs:1696–1700` removes completed entries from that Vec. In-order
one-block completion often removes index zero, which shifts the remaining entries.
For N such removals, the Vec shifts N(N-1)/2 entries. Out-of-order arrivals also
pay for linear matching. This problem also predates PR 892.

Follow-up: use an exact ordered request index keyed by start height. A predecessor
lookup can match a height to a nonoverlapping range. Confirm the nonoverlap
invariant across retained, reset, and retried requests before relying on it.
Keep deadline processing and reservation settlement attached to the same records.
An additional index beside the existing Vec would add more synchronized state.

### 4. Serving serialization still runs on the shared reactor

`reactor.rs:1679–1705` loops over response blocks synchronously.
`service.rs:228–245` encodes each message before `try_send_guarded` checks queue
capacity. A full queue therefore discards an already encoded frame. Large
responses occupy the reactor while lifecycle and other completions wait.

The regulated method is new, but it retains the earlier encode-before-send
pattern. This is a concrete source of avoidable work; its throughput impact needs
measurement. Reserve transport capacity before encoding. Consider a bounded
encoding worker that carries the existing response ownership through handoff.
Do not release producer ownership when encoding finishes if writes remain pending.

The previous audit also records the 16 MiB stream-window throughput concern and
the concurrent response-memory envelope. Neither is a measured RSS or WAN result.

## Where a Bloom filter fits

| State | Required answer | Suitable representation |
| --- | --- | --- |
| Active serving request | Exact ID, range, elapsed time, and owned permit | One optional record under the current one-producer policy |
| Active download request | Expected hash, owner, deadline, and byte reservation | Exact request records with indexed range lookup |
| Received offsets within one request | Exact membership for at most 128 offsets | Existing u128 bitmap |
| Cross-peer outstanding claims | Exact height/hash/owner and retry eligibility | Exact records or an exact derived index with incremental updates |
| Retry suppression | Exact expiry and retry scope | Bounded expiring records; false suppression can delay the only supplier |
| Historical duplicate hints | Approximate membership, if a positive result is independently checked | Potential Bloom-filter use, but no replacement for active ownership |

A false positive cannot safely release a permit, declare a block received, or
suppress required work indefinitely. A Bloom filter also cannot return the owner,
deadline, expected hash, or reservation size. The current serving policy allows
completed ranges to be requested again, so a historical request filter must not
become a permanent rejection rule.

I did not find a history-only request collection in this serving path that merits
a Bloom-filter replacement. The large download Vec contains live obligations.

## Cleanup applied locally

The [patch](performance-state-cleanup.patch) changes three block-sync files:

- Replace the serving VecDeque with `Option<ServingBlockRequest>`. Admission permits
  one producer per authenticated peer. The ledger now expresses that bound directly
  and removes its heap allocation and linear removal. Completion still matches both
  request ID and start height before taking ownership. The record moves inline, so
  this change does not claim that an idle peer struct becomes smaller.
- Remove the cached received count. `len()` now counts the bits in the u128 bitmap.
  This removes redundant state while preserving exact duplicate and completion
  behavior. The compile-time 128-offset bound remains.
- Remove the unused serving-cap argument and VecDeque import.

These changes preserve query and frame permits. They do not collapse the separate
execution and transport lifetimes. The generic execution mutex protects the
start/cancel race; removing it because a cancellation token also exists would
require another synchronization mechanism.

## Validation and limits

The final checks use Rust 1.97:

- Block-sync suite: 315 passed; see [test log](tests-performance-state.log).
- Shared regulation suite: 13 passed; see [test log](tests-performance-regulation.log).
- Workspace formatting: passed; see [check log](fmt-performance-state.log).
- `git diff --check`: passed.

The block-sync suite exercises stale session completions, reconnect ownership,
queue saturation, terminal responses, reservation release, and the 128-bit boundary.
No new full-node, WAN, allocation, or CPU benchmark ran during this pass.
I did not establish a new Rust memory-safety failure. The outstanding correctness
blocker is the reproduced bidirectional progress failure.

The checkout also contains the previous audit harness in `handler.rs` and
`audit_bidirectional.rs`. The cleanup patch excludes that harness. No changes
were committed or published.
</details>

File: crates/zakura-network/src/zakura/block_sync/reactor.rs [hunk 17]
```diff
@@ -1449,69 +1561,129 @@ impl BlockSyncReactor {
...
+                .peers
+                .get(&peer)
+                .is_some_and(|peer_state| peer_state.session.session_id() == session_id)
+        {
             return;
         }
 
-        // `received_status` is now a registry fact (written reactor-side on
-        // `Status`); the serving slots stay on the reactor's thin peer handle.
+        let local_inflight_cap = self.startup.config.advertised_max_inflight_requests();
```
evan-forbes:
```suggestion
```

crates/zakura-network/src/zakura/block_sync/reactor.rs
```diff
@@ -1449,69 +1561,129 @@ impl BlockSyncReactor {
...
+            .get_mut(&peer)
+            .expect("the peer was checked before this non-awaiting admission path")
+            .try_start_serving_blocks(
+                local_inflight_cap,
+                request_id,
+                start_height,
+                count,
+                requested_count,
+                permit,
+            );
```
evan-forbes:
```suggestion
            .try_start_serving_blocks(request_id, start_height, count, requested_count, permit);
```

crates/zakura-network/src/zakura/block_sync/state.rs
```diff
@@ -809,8 +809,45 @@ pub(super) struct PeerBlockState {
     /// `status_reply_meter`; this half stays reactor-side because the reactor owns
     /// serving-tip advertisement.
     pub(super) refresh_meter: RateMeter,
-    pub(super) served_blocks_inflight: u32,
-    pub(super) served_block_requests: VecDeque<(block::Height, Instant)>,
+    served_block_requests: VecDeque<ServingBlockRequest>,
```
evan-forbes: Admission allows one response producer per authenticated peer, including across reconnects. An `Option` expresses that bound and removes the queue allocation. Apply this with the constructor, methods, and caller suggestions, plus the import edit in the review text.

```suggestion
    // Admission allows one response producer per peer.
    serving_request: Option<ServingBlockRequest>,
```

crates/zakura-network/src/zakura/block_sync/state.rs
```diff
@@ -819,41 +856,46 @@ impl PeerBlockState {
...
-        self.served_blocks_inflight = self.served_blocks_inflight.saturating_sub(1);
+    ) -> Option<ServingBlockRequest> {
         self.served_block_requests
             .iter()
-            .position(|(start, _)| *start == start_height)
+            .position(|request| request.id == request_id && request.start_height == start_height)
             .and_then(|index| self.served_block_requests.remove(index))
-            .map(|(_, started)| started.elapsed())
     }
 }
```
evan-forbes, line 825:
```suggestion
impl PeerBlockState {
    pub(super) fn new(session: BlockSyncPeerSession, config: &ZakuraBlockSyncConfig) -> Self {
        Self {
            direction: session.direction(),
            session,
            refresh_meter: RateMeter::new(config.status_refresh_interval),
            serving_request: None,
        }
    }

    /// Store the request, or return its permit if the peer already has one.
    pub(super) fn try_start_serving_blocks(
        &mut self,
        request_id: BlockRangeRequestId,
        start_height: block::Height,
        original_count: u32,
        requested_count: u32,
        permit: super::serving_regulation::GetBlocksServingPermit,
    ) -> Result<(), super::serving_regulation::GetBlocksServingPermit> {
        if self.serving_request.is_some() {
            return Err(permit);
        }
        self.serving_request = Some(ServingBlockRequest {
            id: request_id,
            start_height,
            original_count,
            requested_count,
            started: Instant::now(),
            permit,
        });
        Ok(())
    }

    /// Remove only the ledger entry matching both its ID and starting height.
    pub(super) fn finish_serving_blocks(
        &mut self,
        request_id: BlockRangeRequestId,
        start_height: block::Height,
    ) -> Option<ServingBlockRequest> {
        self.serving_request
            .take_if(|request| request.id == request_id && request.start_height == start_height)
    }
}
```

evan-forbes (commented):
Suggested cleanup for the serving ledger and received-block tracker. Apply the serving-ledger suggestions together. The combined changes pass 315 block-sync tests, 13 shared-regulation tests, and formatting checks.

GitHub cannot attach suggestions to the unchanged import and bitmap lines. These accompanying edits complete the cleanup:

Remove the unused `VecDeque` import after applying the serving-ledger suggestions.

```diff
diff --git a/crates/zakura-network/src/zakura/block_sync/mod.rs b/crates/zakura-network/src/zakura/block_sync/mod.rs
index 7d616ab1a..7dd4e204a 100644
--- a/crates/zakura-network/src/zakura/block_sync/mod.rs
+++ b/crates/zakura-network/src/zakura/block_sync/mod.rs
@@ -5,7 +5,7 @@
 //! commit pipeline, the registry, and the reactor.
 
 use std::{
-    collections::{BTreeMap, HashMap, HashSet, VecDeque},
+    collections::{BTreeMap, HashMap, HashSet},
     io::{self, Cursor, Read, Write},
     sync::{Arc, Mutex as StdMutex},
     time::{Duration, Instant},
```

In `crates/zakura-network/src/zakura/block_sync/state.rs`, derive the received count from the existing bitmap. This removes redundant state while preserving exact membership and completion for all 128 offsets.

```diff
@@ -1017,12 +1013,12 @@ const _: () = assert!(MAX_BS_BLOCKS_PER_REQUEST <= RECEIVED_TRACKER_OFFSET_CAPAC
 #[derive(Clone, Debug, Default)]
 pub(super) struct ReceivedBlockTracker {
     bits: u128,
-    count: usize,
 }
 
 impl ReceivedBlockTracker {
     pub(super) fn len(&self) -> usize {
-        self.count
+        // At most 128 set bits fit in usize on every supported target.
+        self.bits.count_ones() as usize
     }
 
     fn contains_offset(&self, offset: u32) -> bool {
@@ -1037,7 +1033,6 @@ impl ReceivedBlockTracker {
             return false;
         }
         self.bits |= bit;
-        self.count = self.count.saturating_add(1);
         true
     }
```

evan-forbes:
refactor proposal

<details>
```md

# PR #892: bounded serving with independent request intake

Status: proposed design; no implementation or benchmark claims.

Basis: reviewed PR head `fbf466b07dbc0e86c3ad1e8ae33c2f9e94f8a647`.

## Proposal

Remove the serving query-result timer. Give each response one owner from admission
through its final application write. Put inbound `GetBlocks` requests on a separate
QUIC stream from inbound block responses. Bound lifecycle state by session slots.

These changes let us remove the scheduling machinery that compensates for pausing
downloads during serving admission. They preserve the rule that unfinished work
keeps its resource reservation.

This proposal changes the block-sync wire layout. It requires a negotiated service
version. The implementation must not silently change the existing version.

The requested removal of the query-result timer supersedes the repository's general
guidance to time out external waits. Transport write timeouts and download progress
deadlines remain because they govern different resources and behavior.

## Required behavior

1. Reject malformed requests before expensive allocation or storage work.
2. Admit at most one response producer per authenticated peer across reconnects.
3. Admit at most A response producers per node; retain the current default A = 64.
4. Bound each response by the advertised count and body-byte limits.
5. Keep capacity reserved while storage work, returned data, or application writes remain.
6. Let a peer download while its incoming request waits for serving capacity.
7. Cancel delivery to an obsolete session without redirecting it to a replacement.
8. Bound retained request and lifecycle metadata independently of request history.
9. Preserve exact download ownership, late-response handling, and reservation settlement.

Bounded state does not promise service under unlimited identities or failed storage.
An attacker can occupy serving slots. The design must contain that load and expose
it through metrics.

## 1. Separate request intake from response intake

Use two fixed bidirectional streams for each block-sync session:

| Stream | Messages | When its reader may pause |
| --- | --- | --- |
| Block data and control | `Status`, `Block`, and terminal responses | Existing bounded download processing |
| Block requests | `GetBlocks` | While serving capacity or the preceding response is outstanding |

Both peers use the request stream's respective send direction. Both peers use the
data stream's respective send direction for responses. Stream roles belong to the
negotiated service version. The transport admits exactly one stream of each role.
Duplicate or unexpected roles cannot create additional readers, workers, or budgets.

The request reader retains one decoded request. It waits for the peer permit and
then the node permit. It dispatches one serving job and waits for that job's resource
ownership to end before admitting another response. A bounded request queue at the
requester may preserve existing request pipelining.

The download routine uses nonblocking request-queue admission. A full outgoing
request queue stops new issuance. It must not stop response reads or hold an
outstanding download reservation for a request that was never queued.

Each stream needs an independently polled writer. A shared writer that awaits a
blocked request write before polling response writes would recreate the dependency.
The same rule applies to application channels and connection-level scheduling.

### Initial Status ordering

Separate streams do not preserve ordering between `Status` and `GetBlocks`. Retain
at most one early request and wait for the session's validated initial Status before
admission. Cancel this wait when session establishment fails or the session closes.
Use the service's establishment deadline; if the implementation has no such deadline,
add it at session establishment rather than adding a timer to each serving request.
Do not penalize a valid peer merely because QUIC delivered its request stream first.

### QUIC connection credit

Stream separation alone does not reserve connection receive credit. Give the request
stream a small receive window, sized from the supported request pipeline and frame
size. Do not copy the 16 MiB data-stream window onto it by default.

Before choosing constants, account for every stream that can pause on the connection.
The sum of their possible retained receive credit must leave credit for a response
reader that can make progress. Account for header sync and other services too.
Verify whether the QUIC API supports the required per-stream limits. If it does not,
use explicit request credits or revise connection allocation before shipping this
layout. A larger connection window alone is not a proof.

### Scheduling removed

Delete serving admission's download-grace deadline, accumulated admission delay,
temporary download-deadline extensions, and grace-expiry work returns. Keep normal
download deadlines and normal reassignment. Keep reset-specific accounting wherever
view resets still need it.

The existing download timeout may expire while the remote waits for its serving
slot. That is an ordinary download timeout. Existing exact ownership rules handle
late results; the local server does not invent a compensating deadline.

## 2. One serving job owns the response

Move serving execution out of the shared block-sync reactor. The session's serving
task processes admitted requests sequentially. It captures the exact session sender
and cancellation handle when it admits the request.

The job owns:

- The peer and node permits.
- The accepted range and response byte allowance.
- The session generation and cancellation handle.
- The storage result while it is being encoded and sent.

The reactor publishes the current serving view through a bounded latest-value
interface. The job snapshots that view to clamp the requested range. State storage
remains responsible for returning the available bounded prefix. A changing tip must
not turn a snapshot into authorization to exceed the advertised response limits.

The serving job follows structured control flow:

```text
one decoded request
    -> peer permit
    -> node permit
    -> bounded storage read
    -> bounded encoding and ordered writes
    -> terminal write
    -> release permits
```

This sequence does not require a separately stored enum for every stage. Types and
local ownership should express the transitions.

Acquire the peer permit first. A session may hold that permit while waiting for the
node permit. It then has no storage work and consumes no node slot. This order
prevents a blocked old peer from hoarding node slots. Cancellation drops the waiter
and the peer permit. Retain FIFO semaphore behavior; do not add a polling retry loop.
The wait must be independent of response reads.

### Query cancellation without a timer

Remove `query_timeout`, the timeout result branch, and the timeout-generated serving
completion event. A storage error produces the existing terminal error response.
A slow query remains pending until it completes or session cancellation ends delivery.

Give the blocking storage closure a strong reference to the response resources.
The async serving job holds another reference. If session cancellation drops the
async receiver, the worker still owns capacity. The result channel must hold at most
one result. A failed delivery drops the result and its resource ownership.

Check cancellation inside the blocking closure before the first read and between
block reads. Cancellation can stop remaining work; it cannot interrupt an already
blocked database call. Keep the permit until the closure exits.

Use the existing dependency direction: network creates a job-specific closure that
captures its resource guard, and the state execution seam runs that closure or an
equivalent dependency-neutral job envelope. Do not make `zakura-state` depend on
`zakura-network`. Confirm that the chosen API keeps the guard inside the actual
blocking operation, including panic unwinding and failed result delivery.

The current execution mutex protects cloned leases that can start the same query.
Remove that mutex only when the new job is non-cloneable and the worker owns the
single dispatch. The cancellation check then bounds wasted work but does not need
to prove that no read can ever start concurrently with cancellation. A racing read
is acceptable because it still owns capacity and cannot deliver to a new session.

### Transport ownership

Reserve bounded output capacity before encoding a block. Run expensive encoding
outside the shared reactor. Prefer extending the bounded serving worker path over
creating an unbounded task per block.

Keep `FrameGuard` while the transport owns a queued or partially written frame.
The serving job shares its resource ownership with each frame. The final guard
releases capacity after the application write completes or the transport discards it.
QUIC acknowledgement is outside this lifetime.

Wait for queue capacity inside the serving task. Queue pressure should not truncate
a response into a partial reply merely because `try_send` failed. Send the bounded
available prefix and its terminal in order. Session cancellation and the existing
write timeout terminate a stopped transport.

This removes the reactor's `pending_serving_terminals` collection. A terminal wait
is simply the last write in its response's task. Await actual guard release before
admitting another producer for the same identity; queue insertion is not completion.

## 3. Bound lifecycle state by session slots

Replace the new unbounded peer-lifecycle event stream with an authoritative bounded
session table and a coalesced wakeup. Allocate table slots from the existing service
admission limit. Each slot stores its latest generation and current session handle.

The service updates a slot under the admission lock. It cancels an obsolete session
directly. It then signals that the table changed. The reactor reconciles its local
view against the current table. It detects removals by absence and replacements by
generation. Intermediate connect/disconnect events need no queue entry.

Use an epoch/watch notification or a correctly armed `Notify` loop so an update
during reconciliation cannot be lost. Never rely on dropping disconnect events from
a full channel. Keep generation validation on every asynchronous mutation.

A full reconciliation costs O(S) for S session slots. Coalescing bounds memory under
churn, but does not eliminate churn CPU cost. If measurement shows that cost matters,
use one dirty bit per fixed slot and reconcile changed slots. Do not introduce a
history of dirty generations.

The regulator's identity map remains separate because old workers can outlive
sessions. Keep weak references and prune expired entries. Live entries scale with
current sessions plus identities that still own response resources. Do not add an
identity history or a Bloom filter.

## 4. Reduce exact state and repeated work

Replace the serving deque with `Option<ServingBlockRequest>` as an intermediate
cleanup. Once the serving task owns the complete response, remove the reactor
serving record and its completion round trip. Keep internal request IDs only where
an asynchronous boundary still requires matching. Never match delayed work by
height alone. The serving job must never look up a replacement sender by peer ID.

Remove the received bitmap's cached count and compute it with `count_ones()`.
Keep the exact bitmap. It describes a bounded live request, not historical hints.

Treat download indexing as a separate implementation step within this proposal:

- Establish the nonoverlapping-range invariant across retained and retried work.
- Replace full-window linear matching with an exact ordered range index if that
  invariant holds. Otherwise select an index that represents actual overlap.
- Publish owner- and generation-checked claim changes incrementally.
- Maintain reserved bytes through the same exact settlement transitions.
- Preserve atomic duplicate-issuance protection during reset and rejection races.

These changes address preexisting costs. Removing serving admission grace already
removes one new caller of the expensive full-registry publication path. Do not
couple the stream migration to a speculative rewrite of all download scheduling.

## State and resource bounds

Let S be the admitted session bound, A the active-response bound, F the bounded
frames per response, and R the maximum encoded response payload.

| Resource | Proposed bound | Owner |
| --- | --- | --- |
| Waiting decoded requests and semaphore waiters | O(S), one request per session | Request reader |
| Lifecycle records | O(S), independent of connection history | Session table |
| Admitted response jobs and running serving reads | At most A | Node permits |
| Peer-budget identities | O(S + A), with expired entries pruned | Weak registry |
| Retained block results | O(A × R) encoded equivalent | Worker/job/result |
| Frame guard references | O(A × F), also constrained by queue capacity | Transport |
| Request bytes inside QUIC | Explicit request-window bound per connection | Transport |
| Download records | Existing configured request and byte bounds | Download scheduler |

Decoded memory can exceed encoded bytes. Encoding temporaries, allocator overhead,
transport buffers, cancelled task teardown, and other state clients need separate
accounting. Validate a concrete memory envelope before treating A × R as a memory
budget. Task teardown must release obsolete session resources promptly; add no
detached admission tasks that can accumulate during churn.

This design does not impose a request-rate limit. Cheap requests may run rapidly.
Do not add a refill timer without evidence that concurrency and normal connection
controls fail to protect useful work. Measure storage and reactor load first.

## Failure policy

| Failure | Behavior |
| --- | --- |
| No serving capacity | Pause only request intake |
| Slow storage | Keep the job and its permits; expose its age |
| Session cancelled before worker starts | Worker observes cancellation and exits |
| Session cancelled during a read | Stop delivery; retain permits until the read exits |
| Output queue full | Wait in the serving task while response reads continue |
| Peer stops reading | Existing transport write timeout closes the affected transport |
| Session replaced | Cancel old delivery; preserve old worker/frame ownership |
| Storage hangs indefinitely | Lose at most the configured serving capacity; report saturation |

A dedicated serving executor can isolate blocking thread occupancy if measurements
show contention with validation or RPC. It cannot guarantee storage progress or
isolate shared disk and database locks. Process isolation is a larger operational
choice and is outside the initial refactor.

## Alternatives and compatibility

Keeping the existing stream layout requires a different overload policy. We could
continue decoding into a strictly bounded request queue and disconnect a session
that exceeds its advertised pipeline. That avoids waiting behind queued requests,
but requires coordinated request-credit semantics and changes overload behavior.
Sending one busy reply per request adds another response queue unless the same
credit bound applies. Neither approach is a queue-size-only fix.

The preferred design uses stream separation because requests may then wait through
ordinary transport backpressure without obstructing responses. Negotiate the new
service version as a complete stream pair. Reject incomplete or duplicate pairs
within bounded establishment state.

If legacy support is required, keep it explicitly versioned. Do not claim that the
new progress guarantee covers the legacy path. A rollout must decide whether to
disable legacy block sync or maintain a separate bounded overload policy for it.

## Possible Implementation sequence

Two implementation questions need a small prototype before the architecture is
accepted: per-stream QUIC receive-credit control, and a state execution API that
carries resource ownership inside the actual blocking closure. Do not replace either
question with an assumption in the implementation. The requested timer removal and
bounded lifecycle work can proceed independently of those prototypes.

1. Record the invariants and preserve the existing bidirectional failure as a
   regression test. Add resource-owner counters to the test harness.
2. Remove the query-result timer while preserving current lease ownership. Apply
   the serving-record and bitmap simplifications. Update configuration docs.
3. Replace lifecycle event history with bounded session reconciliation. Test rapid
   generation replacement before proceeding.
4. Introduce the negotiated request stream and independent read/write tasks. Prove
   connection-credit behavior with production QUIC. Remove admission-grace state.
5. Move serving into a sequential task and move query ownership into the worker.
   Remove reactor query completion events, serving ledger, execution claim mutex,
   and terminal futures only as their last consumers disappear.
6. Reserve transport capacity before encoding. Validate response order, memory
   retention, and cancellation across the worker/transport boundary.
7. Implement measured download-index and incremental-publication improvements as
   a separately reviewable change with the same ownership tests.

Do not add a second permanent scheduling abstraction during migration. Keep each
compatibility adapter limited to a named service version or remove it when its last
caller moves.

## Acceptance tests and measurements

- Run the production bidirectional reproduction with both peers serving and
  downloading. Fill request windows and response queues. Both peers must finish
  without a timeout being needed to break a dependency cycle.
- Repeat with a third service sharing connection credit. Verify the proposed
  request-window bound and connection-credit reserve.
- Saturate A response slots. Flood further requests and verify one decoded waiter
  per session and no extra query dispatch.
- Advance time beyond eight seconds during a controllable read. Verify that no
  timeout completion occurs and that the original query retains its slot.
- Cancel before dispatch, during storage, after result delivery, during encoding,
  and during a partial write. Reconnect using the same identity at each boundary.
- Stall selected workers indefinitely. Verify that retained jobs plateau at A and
  that old identities cannot bypass the peer permit through reconnects.
- Block reactor reconciliation while repeatedly replacing sessions. Verify that
  lifecycle storage plateaus at the session-table bound and converges afterward.
- Reorder initial Status and request-stream arrival. Verify bounded waiting and
  correct session-establishment failure behavior.
- Fill the output queue. Verify that encoding starts only after capacity exists
  and that the response's terminal follows its blocks on the original session.
- Exercise response count/byte boundaries, unavailable ranges, overlapping retries,
  stale completions, reset races, and duplicate body arrivals.
- Measure allocations, retained bytes, reactor occupancy, request throughput, and
  useful download throughput at increasing S and A. Compare against the reviewed
  head under identical workloads. Report results before tuning pool sizes.

Completion requires the resource bounds and progress tests to pass together.
Formatting or unit tests alone cannot establish the transport progress property.

## Evidence and implementation anchors

- [Current serving design](../../../../zakura.pr892-audit/docs/design/getblocks-regulation.md)
- [Prior progress failure and audit](audit-updated.md)
- [Performance and state review](performance-state-review.md)
- [Production reproduction log](repro-bidirectional-production.log)
- [Default-queue reproduction log](repro-bidirectional-default-queues.log)

Primary implementation targets: `regulation/request.rs`,
`block_sync/serving_regulation.rs`, `block_sync/peer_routine.rs`,
`block_sync/reactor.rs`, `block_sync/service.rs`, `transport/io.rs`, `handler.rs`,
`zakurad/.../block_sync_driver.rs`, and `zakura-state/src/service.rs`.
```

</details>

File: crates/zakura-network/src/zakura/block_sync/peer_routine/trace.rs [hunk 21]
```diff
@@ -109,18 +109,6 @@ impl PeerRoutine {
         });
     }
 
-    pub(super) fn trace_queue_send_failed(&self, msg: &BlockSyncMessage, error: &OrderedSendError) {
```
czarcas7ic, line 112: Before we would reserve download resources, then try to queue the request, then if failed we would emit this and release the resources. Now, we reserve a queue slot first here

https://github.com/zakura-core/zakura/blob/3b8f8caf1dedd46dd972e23be5d44a0271ad8d9a/crates/zakura-network/src/zakura/block_sync/peer_routine.rs#L802

and if the queue is full, we stop before taking work or reserving download resources.

PR #895, opened 2026-09-04
fix(ops): consolidate continuous sync notifications

czarcas7ic:
## Motivation

The background sync tests repeatedly start from an empty database, catch up to the chain, and start over. Each successful cycle currently posts “Zakura sync complete” to the alerts channel, even though nobody needs to act. A failure can also be announced twice—once by the test controller and again by the scheduled audit—then repeated every six hours while nothing has changed.

## Solution

**Failures do not wait for the daily summary.** The controller attempts to send a Slack alert immediately when it detects a failure. If delivery fails, or the host is unreachable, the audit provides the fallback on its next check. That audit runs every 30 minutes, subject to GitHub Actions scheduling delays.

- Put routine sync successes and unchanged failures into one daily summary instead of separate completion posts and six-hour reminders. Include an unchanged failure only once at least 24 hours have passed since its last delivered alert or reminder; recent alerts wait for a later daily summary.
- Avoid the second failure announcement when the audit can confirm that the controller already delivered that exact failure to the same Slack destination. If it cannot confirm delivery, it still alerts. Cached delivery is tied to the Slack destination, so changing the destination cannot silence its first alert. New failures, changed failures, and recoveries are reported as soon as they are observed.
- Avoid repeating a canary result when only its notification job is retried. Running the test again is still a separate execution and can produce a new failure alert. A successful retry reports recovery once.

The controller must be updated before its completion messages move into the summary. Requester DMs are not part of this change.

## Testing

All 79 controller, audit, and canary tests pass on Linux. They cover duplicate announcements, separate runs, daily summaries, the 24-hour reminder boundary, destination changes, recoveries, missing evidence, and failed delivery retries. Workflow validation, Python compilation, Markdown lint, and changelog checks also pass.

Used Codex for implementation, tests, and this description.


docs/changelog/params.md
```diff
@@ -28,6 +28,7 @@ Keep entries **newest-first**. Each row records:
 
 | Parameter | Location | Old → New | PR | Why |
 | --- | --- | --- | --- | --- |
+| `--reminder-interval` / `REMINDER_INTERVAL` | `deploy/continuous-sync/deploy.py`, `.github/workflows/zakura-continuous-sync.yml` | `6 h` → `24 h` | [#895](https://github.com/zakura-core/zakura/pull/895) | Group unchanged failures and routine completions into a daily digest; new failures, changed failures, and recoveries remain immediate. |
```
p0mvn: I don't think we should pollute this file with infra param updates

The goal of this file was to track param changes for comple parts of the system

p0mvn (approved):

docs/changelog/params.md
czarcas7ic: https://github.com/zakura-core/zakura/pull/895/commits/0ffd263bf40242ca4f5a7231ecc1e339485372d5

Addressed so that future AI prompts don't automatically do this given this is the intent

PR #901, opened 2026-09-05
docs: add Dogwood block propagation design and specification

evan-forbes:
## Motivation

Define Dogwood block propagation before implementation. Proof of work makes the
next block's entry point unpredictable, while peer bandwidth and topology keep
changing.

## Solution

Add a design, protocol draft, and internal experiment report. The design
introduces three rough targets:

- **Throughput:** 50,000 TPS at 2 KiB per aggregated transaction requires 819.2
  Mbps of body traffic, or 1.024 Gbps with 25% parity before other overhead.
- **Latency:** replicate each block to 90% of nodes in about 400 ms, below 500
  ms. A geographic estimate explains the budget and relates propagation delay to
  the PoW block interval.
- **Robustness:** adapt subscriptions from any proposer to the bandwidth and
  paths available as the network changes.

Standing subscriptions push verified parts before receivers need to request
them. Receivers adjust supplier assignments, and parity covers missing parts.
Bounded requests to peers advertising `FullBlock` provide fallback.

The protocol draft defines finite grants, reconstruction checks, controller
rules, and candidate payload profile W1. W1 fixes frame encoding, commitments,
proofs, signatures, and separate seed cancellation. The proposer-key candidate
uses a zero-value coinbase output; a V5 txid proof cannot authenticate a key
carried only in the coinbase input script.

The targets are planning estimates. Internal simulations and synthetic workloads
do not establish production performance.

## Testing

The experiment report records finite-model checks, connected-relay simulations,
TCP feedback trials, signature fixtures, and reference-codec measurements with
their assumptions and failures. Experiment source and raw results remain in the
local Dogwood experiment worktree.

Checks cover the throughput arithmetic, section links, whitespace, and changelog
policy. No runtime code changed.

## Changelog

No fragment is required for these documentation-only changes.

### Follow-up work

Validate adaptation and failure coverage under realistic geography, concurrent
traffic, and changing peer capacity. Choose the block interval, body size, and
coding profile against the latency target. Complete transport negotiation, the
post-Tachyon chain adapter, and aggregate resource limits.


evan-forbes:
I'm not planning on adding all the the documents here, just wanted to show our work for now

the interesting thing is that I'm not sure we actually solved DOG's main issues, but I'm not sure DOG did now either after this.

Literally just subscribing to parts statically not inherently unlike gossip sub but different peers per part outperforms dynamic subscriptions!!!

PR #903, opened 2026-09-05
chore: bump MSRV to 1.97

ebfull:
## Motivation

Raise the minimum supported Rust version from 1.91 to 1.97 so the workspace can use APIs stabilized since 1.91, starting with `Atomic*::try_update` (stable since 1.95), which replaces `fetch_update` ahead of its 1.99 deprecation.

## Solution

- `rust-version = "1.97"` in the workspace manifest, `zakurad`, and the out-of-workspace `zakura-assets` crate (plus the `pack-assets-crate.py` self-test fixtures).
- MSRV CI job, `cargo xtask package ubuntu` default toolchain, and both Dockerfiles pinned to the 1.97 line (`rust:1.97.1-trixie` for the release image).
- AGENTS.md MSRV note.
- Replaced the four `fetch_update` calls with `try_update` and dropped the two `#[allow(deprecated)]` workarounds that were explicitly tied to the 1.91 MSRV.
- The MSRV job now checks `zakura-watchdog` too: its `sentry` tree builds on 1.97, so the exclusion added for the 1.91 floor is stale.
- Unblocked the `semver (zakura-consensus)` and `semver (zakura-rpc)` jobs, which started failing for every PR on 2026-09-05 for reasons unrelated to this change (details below):
  - Enabled rocksdb's `bindgen-runtime` feature on the workspace dependency. No-op for the normal build, where feature unification already turns `runtime` on; `Cargo.lock` is unchanged.
  - Extended the semver job's isolated Cargo home (the existing tinyvec patch) with a librocksdb-sys copy whose bindgen build-dependency enables `runtime`, so the published baselines build too. The script is renamed to `patch_registry_for_semver.sh`.

`docs/decisions/zakura/0001-iroh-dependency.md` still says 1.91; it is a historical decision record and is left unchanged.

### Why the semver jobs were failing

Not a semver finding: cargo-semver-checks has no lint on `rust-version`, and the other crates' semver jobs passed with the same MSRV bump. The `librocksdb-sys` build script panicked with "a `libclang` shared library is not loaded on this thread" while rustdoc was built for the current crate and then for the published baseline. cargo-semver-checks builds both in placeholder projects that resolve dependencies fresh, ignoring `Cargo.lock`. Since bindgen 0.73.0 was published (2026-09-05 16:27 UTC), that fresh resolution gives `libzcash_script` bindgen 0.73 with `runtime` while `librocksdb-sys` stays on bindgen 0.72 without it (our `rocksdb` dependency disables default features). The shared clang-sys ends up in runtime mode, but the 0.72 bindgen never loads libclang. Only crates depending on both rocksdb and zakura-script are affected, which is why zakura-state and zakura-script pass. The workspace feature fixes the current-crate build; the published baselines cannot change, so the Cargo-home patch covers them until the next release publishes crates that carry the feature.

## Testing

Locally on rustc 1.97.1 with `--locked`:

- `cargo check --workspace --bins` (the MSRV job command, including `zakura-watchdog`): passes.
- `cargo clippy -p zakura-network --all-targets -- -D warnings`: passes.
- `cargo fmt --all -- --check`: clean.
- `python3 scripts/pack-assets-crate.py --self-test`: 14 tests OK.
- `./scripts/changelog.py check` and markdownlint on AGENTS.md and the fragment: clean.
- The consensus semver placeholder (`cargo check -p librocksdb-sys` inside it) panics without the rocksdb feature and builds with it.
- A project depending on crates.io `zakura-consensus = "=7.0.1"` (what the baseline build resolves) panics with the CI message under a plain Cargo home and builds under the Cargo home produced by `patch_registry_for_semver.sh`.

CI: with only the rocksdb feature change, the current-crate build passed and the same panic moved to the baseline build; this push adds the Cargo-home patch for that.

## Changelog

`docs/changelog/unreleased/903.md` (Changed).


TalDerei (approved):
atomic API changes and semvar workaround seems fine to me.

czarcas7ic (approved):

PR #911, opened 2026-09-06
chore: add monetary-base defense in depth

czarcas7ic:
## Motivation

Add an aggregate monetary-base guard as defence in depth alongside the existing per-pool, subsidy, transaction-balance, and proof checks.

## Solution

Check that the combined balance of all six chain value pools stays within `MAX_MONEY`. This provides an additional accounting safeguard while preserving valid transfers and existing per-pool checks. Update the test fixture, bump `zakura-chain` for the new error variant, and patch-bump dependent libraries so their updated requirements can ship. The node release version remains managed by release preparation.

## Testing

Local validation passed 402 chain tests, 33 state/reorg tests, chain/state Clippy with warnings denied, formatting, and Markdown/changelog checks. Boundary tests cover the exact cap, one zatoshi over it in each pool, and transfers and reversals at the cap. These exercise the guard directly and do not demonstrate a way to exceed the cap through otherwise valid blocks. The dry-run publish graph resolves all nine affected libraries at their workspace versions; the existing node package is skipped pending its next release version.

## Changelog

Added `docs/changelog/unreleased/911.md`.

Used Codex for implementation and validation.

### DIFF
--- a/crates/zakura-chain/Cargo.toml
+++ b/crates/zakura-chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zakura-chain"
-version = "6.3.0-rc0"
+version = "7.0.0"
 authors.workspace = true
 description = "Core Zcash data structures for the Zakura node. Internal crate, published to support cargo install zakura"
 license.workspace = true
--- a/crates/zakura-chain/Cargo.toml
+++ b/crates/zakura-chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zakura-chain"
-version = "7.0.0"
+version = "7.0.0-rc0"
 authors.workspace = true
 description = "Core Zcash data structures for the Zakura node. Internal crate, published to support cargo install zakura"
 license.workspace = true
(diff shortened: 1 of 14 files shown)


crates/zakura-chain/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zakura-chain"
-version = "6.2.0"
+version = "7.0.0"
```
evan-forbes: what are doing to require a major instead of minor bump? are we just adding things?

evan-forbes (approved):
pre-approving as its a very minor nit in the worst case, but if we only need a minor bump then imo we should avoid

crates/zakura-chain/Cargo.toml
czarcas7ic: Agreed we can just make this a minor

ValarDragon (approved):
Fixing conflicts. The cargo bump is because the new error type is added, and these are configured to allow callers to exhaustively check across all enum options rn.

Generally think we need to change the way errors are handled (whitelist vs blacklist as we've discussed)

PR #918, opened 2026-09-07
fix(consensus): prevent serial UTXO lookup delays

evan-forbes:
## Motivation

Fix serial UTXO lookup delays during block verification. A high-input transaction previously waited for every external input's state lookup in sequence. The upstream 1,001-input reproducer with about 1 ms of injected latency per lookup took **2.28 seconds before overlap and 45 ms after it**. Our lookup-only benchmark measured **38.5 ms with the final implementation**, against the recorded 2.25-second serial baseline. These are diagnostic reproductions with injected latency.

## Solution

Overlap up to 64 external lookup responses per transaction, while waiting for state admission one request at a time. Block-local outputs resolve inline. Input order, v5 sighashes, mempool behavior, and state error mapping remain intact.

The first draft also queued up to 64 readiness waits per transaction. The block-limit probe exposed the cost: unrelated requests through the shared state buffer could wait about 400 ms. Serializing admission removes that extra queue while preserving overlapping responses. The new regression test catches a return to multiple admission waits per transaction.

The response bound remains per transaction. Concurrent lookups can start their existing timeout clocks earlier. Dropping verification cancels pending futures; already-started database work can finish.

## Testing

All 215 consensus tests passed. Tests cover ordering and sighashes, window boundaries and refill, mixed input sources, errors, timeouts, cancellation, one admission wait per transaction, and aggregate windows across transactions.

The state probe fills two serialized block layouts to the 2 MB limit with distinct external inputs. It compares serial lookup, the first draft, and the updated admission order against real RocksDB indexes. These are the largest observed delays for an unrelated `Tip` request through the shared state buffer:

| Block layout | Serial baseline | First draft | Updated |
| --- | ---: | ---: | ---: |
| 752 spending transactions, up to 64 inputs each; 48,115 inputs | 12.9 ms | 446.3 ms | 13.4 ms |
| 49 spending transactions, up to 1,001 inputs each; 48,696 inputs | 3.1 ms | 44.6 ms | 4.4 ms |

The blocks occupy 1,999,974 and 1,999,989 serialized bytes. The 64-input layout maximizes simultaneous lookups per byte for this transaction shape. The probe checks that unrelated state work completes during every block and gives each probe request a one-second deadline. With the update, at least 43 complete probe cycles finished during every measured block. Direct state reads stayed below 0.6 ms and blocking-queue delay stayed below 0.3 ms.

Each scheduling mode uses a fresh database, measures its first SST read, then repeats ten warm blocks. The probe runs with a 512-worker blocking limit and with 28 of 32 workers reserved, leaving four workers for state work. It measures the state lookup path with synthetic indexes; it does not run full consensus verification. The first SST read can still use the OS page cache. Rust 1.97.0, four Tokio workers, state optimization level 3, other dependencies in the development profile.

Reproduce with:

```sh
cargo +1.97.0 test -p zakura-consensus --locked --lib block_utxo
cargo +1.97.0 test -p zakura-consensus --locked --lib \
  --config 'profile.dev.package.zakura-consensus.opt-level=3' \
  block_utxo_lookup_timing -- --ignored --nocapture
cargo +1.97.0 test -p zakura-state --locked --lib \
  --config 'profile.dev.package.zakura-state.opt-level=3' \
  utxo_state_contention -- --ignored --nocapture
```

Clippy, formatting, and the changelog check passed.

## Changelog

Updated `docs/changelog/unreleased/918.md`; the concurrency-limit entry remains in `docs/changelog/params.md`.

Upstream fix and reproducer: https://github.com/ShieldedLabs/zero/pull/46

### Follow-up Work

Shared database-read budgeting and batching remain in #919. This PR fixes serial lookup delay and the admission-queue regression exposed by the block-limit probe.


crates/zakura-consensus/src/transaction.rs
```diff
@@ -818,20 +826,8 @@ where
...
-                            Ok(_) => TransactionError::TransparentInputNotFound,
-                            Err(boxed_error) => TransactionError::from(boxed_error),
-                        })?;
-
-                    if let zakura_state::Response::Utxo(utxo) = response {
-                        utxo
-                    } else {
-                        unreachable!("AwaitUtxo always responds with Utxo")
-                    }
+                    block_outpoints.push((input_idx, *outpoint));
```
p0mvn: Consider renaming to `block_outpoints_to_lookup` to make it clearer that these are looked up from state

crates/zakura-consensus/src/transaction.rs
```diff
@@ -61,6 +61,13 @@ mod tests;
 ///     chain in the correct order.)
 const UTXO_LOOKUP_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(6 * 60);
 
+/// Maximum pending state lookups per block transaction, not across the verifier.
+///
+/// Concurrent lookups start their timeout clocks together. During out-of-order sync,
+/// a later input can therefore time out sooner than with serial lookups. The existing
+/// sync restart recovers missing ancestors, as described on `UTXO_LOOKUP_TIMEOUT`.
+const MAX_CONCURRENT_UTXO_LOOKUPS: usize = 64;
```
p0mvn, line 69: Could we bound these lookups across the block/verifier rather than per transaction? 

Since transactions are verified concurrently, the current limit can produce transaction_count × 64 state requests and potentially saturate blocking database work. A shared budget, or evidence that an existing bound makes this safe, would address this risk.

p0mvn: I guess that's somewhat along the lines of what 619 attempts

crates/zakura-consensus/src/transaction.rs
```diff
@@ -841,6 +837,44 @@ where
...
+                        };
+                        Ok::<_, TransactionError>((input_idx, outpoint, utxo))
+                    }
+                });
+            // A single lookup cannot overlap another lookup, so skip the concurrency queue.
+            let lookups = if single_lookup {
+                lookups.then(std::convert::identity).left_stream()
+            } else {
+                lookups
+                    .buffer_unordered(MAX_CONCURRENT_UTXO_LOOKUPS)
```
p0mvn, line 877: The existing concurrency tests exercise one transaction at a time, but `SemanticBlockVerifier` verifies transactions concurrently, so four 65-input transactions can expose 256 pending `AwaitUtxo` requests with this per-transaction bound.

Could we add an aggregate test with several transactions containing distinct external outpoints, one shared mock state service, and withheld responses? After polling all `spent_utxos()` futures together, assert the aggregate pending count; then complete one request and verify bounded refill, and finally drop verification and check cancellation. Withholding responses keeps this deterministic without sleeps and would make the block-level resource behavior explicit.

p0mvn (commented):
Reviewing #619 before approving. Flushing comments so far

p0mvn:
If we want this to get in, I think it would be valuable to have a measurements showing worst-case multi-transaction blocks do not starve state work.

evan-forbes:
@p0mvn In response to https://github.com/zakura-core/zakura/pull/918#issuecomment-5575638201: the new measurements show no starvation in the tested workloads, working on the other benchmark now

edit: update

the latest fix in https://github.com/zakura-core/zakura/pull/918/commits/96db4a2e2f0687ba44c31b71df9412c1ab575f09 removes the slow down added here, good find

p0mvn (approved):
LGTM. The added tests have addressed my concerns.

PR #922, opened 2026-09-08
ci: add native Codex approval adapter for scoped changes

czarcas7ic:
## Motivation

Let eligible CI and deployment changes from authors with Write, Maintain, or Admin access satisfy the approval requirement through the existing native Codex review. Other PRs follow the repository's normal approval process.

## Solution

Add a GitHub Actions adapter that checks author permissions, verifies a clean native review of the current commit, and submits approval through a dedicated App. Automatic reviews and ordinary `@codex review` requests remain supported after fixes; review commands count only from accounts with Write access or higher. Eligible Rust and Cargo.toml changes require their own changelog fragment; release helpers and release-policy waivers stay excluded. The App withdraws approvals when evidence stops qualifying, including after retargeting away from main, and can restore its own withdrawals while respecting human dismissals. The first approval starts a Slack thread in #gh-alerts; withdrawals and reapprovals reply there. Delivery receipts prevent repeated notifications, and uncertain Slack delivery requires operator reconciliation without changing the GitHub approval.

Setup uses repository Actions secrets and identity variables, with no environment, reviewer team, or rule changes. The adapter starts in read-only mode. Manual-result correlation depends on visible comments; deleted-request ambiguity is an accepted limitation for trusted authors. Withdrawal after changes is asynchronous and preserves the existing merge policy.

## Testing

112 offline regression tests pass, covering native evidence, permissions, file eligibility, withdrawal recovery, and Slack approval cycles, retries, and failures. Actionlint, zizmor, Python lint, Markdown lint, and spelling checks pass. Earlier live read-only checks verified the author preflight and native evidence handling. Live App approval and Slack delivery still require activation validation. Used Codex for implementation, tests, and documentation.


ValarDragon (approved):

evan-forbes (commented):
One request before enabling writes, and three small notes.

**1. Keep the App key out of reach of branch workflows.** `6a8efc2ef` moved `CODEX_APPROVAL_APP_PRIVATE_KEY` from the `main`-restricted `codex-approval` environment into a repository secret. Any workflow on any branch that a write-access user pushes can read a repository secret. So one push lets someone mint an App token and approve an eligible PR without Codex. The secret exists today, so this applies before the PR merges. Please restore the environment, restricted to `main` with no reviewers. Every trigger this workflow uses runs on `main` (`pull_request_target`, `issue_comment`, `schedule`, and `workflow_dispatch` gated to `main`), so it adds no approval step. Then delete the repository secret. If deployment entries on PRs were the reason for dropping the environment, that trade is worth it here.

#### Notes

**2. Pending runs get replaced.** The workflow-wide concurrency group keeps only one pending run. Each new event, including any comment on any PR, cancels the run already pending. The withdrawal after a push can therefore wait for the hourly schedule. The ruleset doesn't dismiss stale approvals, so that approval still counts for up to an hour. Human approvals behave the same way today, and the trusted-author gate bounds the risk. Still, the README should say so, since the approval's premise is that Codex reviewed exactly that commit.

**3. The bypass-list check is a no-op.** GitHub omits `bypass_actors` from the ruleset response for callers without admin access. So `not full.get("bypass_actors")` always passes with `GITHUB_TOKEN`. Either drop the check or reword README lines 187–190, which say the adapter verifies the bypass list.

**4. Main advancing mid-run withdraws approvals.** When `check_trusted_revision` fails at the top of `reconcile`, the run withdraws every existing approval it visits. The next run restores them, so each merge during a reconciliation posts a withdraw/reapprove pair to Slack. It's deliberate and tested. Skipping the PR without writing anything would avoid the noise.

evan-forbes (approved):
LGTM, thanks for the quick turnaround!

One small thing: Codex's retarget finding is real, and it's wider than it says. A `pull_request_target` run on a stacked PR uses the PR's base as its ref. The `main`-only environment then fails `reconcile` with a red check on every stacked PR by a write-access author. Adding `github.ref == 'refs/heads/main'` to the `reconcile` condition would skip the job instead. Also, create `codex-approval` before setting `CODEX_APPROVAL_ENABLED`, otherwise GitHub auto-creates it with no branch restriction.

PR #923, opened 2026-09-08
fix(rpc): report block template transaction dependencies

ValarDragon:
## Motivation

getblocktemplate advertises transactions as mutable, but every transaction currently reports an empty depends list. Pools that remove or reorder transactions therefore cannot preserve parent-before-child ordering and can construct blocks that fail transparent spend validation.

## Solution

Convert selected mempool transactions after their final template order is known, map transaction hashes to one-based indexes, and populate each transaction template with its sorted direct dependency indexes from the mempool dependency graph.

## Testing

- cargo test -p zakura-rpc --lib transaction_templates_report_direct_dependency_indexes
- cargo test -p zakura-rpc --lib get_block_template -- --skip shielded_coinbase_paths
- cargo clippy -p zakura-rpc --lib -- -D warnings
- cargo fmt --all -- --check
- npm exec --yes markdownlint-cli -- docs/changelog/unreleased/923.md --config .github/.markdownlint.yaml
- git diff --check

The focused unit test covers independent transactions, multiple direct parents, and a transitive child.

## Changelog

- Added docs/changelog/unreleased/923.md.

- BIP 22 getblocktemplate transaction depends field

### DIFF
--- a/crates/zakura-rpc/src/methods.rs
+++ b/crates/zakura-rpc/src/methods.rs
@@ -2753,7 +2753,7 @@ where
 
                 precomputed_coinbase = wait_for_new_tip => {
                     let chain_info = fetch_chain_info(read_state.clone()).await?;
-                    let mempool_tx_deps = TransactionDependencies::default();
+                    let mempool_tx_deps = Default::default();
 
                     let server_long_poll_id = LongPollInput::new(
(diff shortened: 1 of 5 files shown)


ValarDragon:
Going to close for now, lets revisit once its time for GBT optimization

PR #930, opened 2026-09-08
refactor(mining): simplify the optimistic mined-block relay path

evan-forbes:
## Motivation

The optimistic mined-block relay path from #748 duplicated state bookkeeping and mixed RPC work IDs with consensus cache identity. Invalidation could disable early relay indefinitely. A rejected candidate could also block relay for its siblings.

## Solution

Use state write permits to guard unpublished writer transitions. Authorize relay per candidate after the existing proof-of-work and relay checks. Key the prepared-candidate cache by content and keep RPC work IDs outside consensus. Consolidate constructors, response handling, and template recovery.

Bound speculative preparation to one verification at a time. The worker retains its slot until verification and its dispatched checks finish. A missed deadline stops further speculation on that parent. Late validation failures still withdraw the template. Foreground recovery and submitted blocks retain their existing admission paths.

The final code addresses the eight findings in the [audit comment](https://github.com/zakura-core/zakura/pull/930#issuecomment-5589087027):

- F-270500: Poll completed verification before admission so an already-failed commit cannot emit an early-relay event.
- F-270501: Read withdrawal state and acknowledge its watch version together.
- F-270502: Recheck withdrawal state after scheduling preparation and recover a template if fallback became active.
- F-270503: Check the live chain tip before publishing a template.
- F-270504: Check the tip while holding the rejection-state lock so a stale request cannot overwrite the current parent's withdrawals.
- F-270505: Record a completed rejection when the chain has returned to that template's parent.
- F-270506: Notify withdrawal waiters when prepared-work eviction removes a fallback exemption.
- F-270507: Retain the waiter's starting revision so a parent change cannot erase an unread rejection.

The worker guard also releases the preparation slot on shutdown or panic. Preparation metrics now include semantic verification. State timing metrics again distinguish mined submissions from synced blocks.

Review decisions:

- A client proposal with content identical to a prepared server template can use the server cache entry and its relay authorization. Relay preflight still applies.
- `invalidateblock` returns a retryable error when all writer slots are occupied.
- Speculation has a concurrency bound. It shares the proof caches and CPU pools and cannot stop an in-progress cryptographic operation.

## Testing

Previous validation covered the state, consensus, RPC, sync, inbound, and internal-miner tests, plus clippy. CI on the previous pushed head passed unit tests, lint, and end-to-end checks.

Final validation after merging current `main`, with Rust 1.97.0:

- RPC library tests: 146 passed, one ignored. The withdrawal regression holds a waiter pending, records a rejection, changes the parent before polling the waiter, and verifies that the waiter still completes.
- Consensus library tests: 229 passed, one ignored.
- `cargo +1.97.0 clippy -p zakura-rpc -p zakura-consensus --lib --tests --locked -- -D warnings`
- Formatting, changelog validation, and `git diff --check` passed.
- The branch merges cleanly with current `main` and #931. All changes are committed and pushed.

CI semver checks currently fail while compiling published `stun-rs` dependencies because `precis-core` versions disagree. The failure occurs before API comparison.

Manual regtest checks for early relay after invalidation and contextual sibling rejection have not been run. Existing state regression tests cover the writer-permit behavior.

## Changelog

Updated `docs/changelog/unreleased/930.md` and the parameter ledger.

### Follow-up Work

Dedicated speculative CPU isolation, process-enforced cancellation, and foreground recovery admission limits remain separate work. #754 must adapt to content-keyed consensus preparation and retain work-ID-to-template lookup in the RPC layer.


evan-forbes:
still fixing things from ai audits

edit: sorry forgot to update, that the ai audits were fixed a while ago

czarcas7ic (approved):
I was talking through the v12 fixes you did, and I think there is a remaining edge case related to F-270506 and F-270507, but I don’t think it should hold up this PR.

The miner can still miss a “stop using this template” notification if a rejection happens between its two waits and a parent change clears the stored rejection before the next wait starts. There’s also a related case where eviction sends a withdrawal notification, but a parent change clears it before the miner reads it.

I had AI generate tests and reproduced both cases. The rejection case needs an invalid template to get through assembly first, followed by this specific timing. I saw that we only do the full verification of a candidate block we are mining on once we start work on it (I assume this is to save PoW time?), so its unclear to me how often that this verification actually comes back with an invalid block (I assume its low probability and would indicate a bug otherwise). The eviction case concerns previously validated work losing its retained exemption, so missing that cancellation doesn’t itself make the work invalid.

PR #933, opened 2026-09-08
fix(network): backport QUIC loss recovery datagram bound

czarcas7ic:
## Motivation

QUIC loss recovery can exceed the socket's datagram limit when a probe is smaller than the discovered packet size. On the macOS send path this can stall a connection even without GetBlocks regulation.

## Solution

Pin Iroh 0.92's existing three QUIC crate versions to a narrow backport that counts datagrams directly. The [source diff](https://github.com/zakura-core/iroh-quinn/compare/5c51e631ac64429ff5e94baa867bee09a7c80109...1dcc7a43488fecd199d343d47e93e9ed8319fcaa) contains one production condition and its regression. This workspace patch is not inherited by downstream library consumers; they must apply the same patch in their root workspace. A registry release carrying the fix is still needed before the next crates.io release; the existing release gate remains in place. This is a transport prerequisite for #892.

## Testing

The deterministic regression fails before the fix, and all 254 transport state-machine tests pass after it. A new bare-Iroh test transfers 64 MiB in each direction with production transport settings. Network Clippy, formatting, locked dependency checks, and the dependency source and audit checks pass. Used Codex for implementation and validation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2875,8 +2875,7 @@ dependencies = [
 [[package]]
 name = "iroh-quinn"
 version = "0.14.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0cde160ebee7aabede6ae887460cd303c8b809054224815addf1469d54a6fcf7"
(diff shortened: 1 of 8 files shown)


czarcas7ic:
From upstream, see:

https://github.com/quinn-rs/quinn/pull/2185/changes#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6

PR #934, opened 2026-09-08
fix(rpc): avoid fabricated difficulty values

ValarDragon:
## Motivation

When the state `ChainInfo` query fails, default-enabled difficulty callers
return the high 128 bits of the proof-of-work target as a floating-point
value. That raw target fragment is not normalized difficulty, so `getinfo` and
`getblockchaininfo` can report an enormous plausible-looking value.

Returning `1.0` for every state error would still fabricate minimum difficulty
when a chain tip exists. The successful calculation also discards the low 128
bits of both targets, which can produce `NaN` or infinity for valid configured
networks with small targets.

## Solution

- Confirm the state has no chain tip before using the zcashd-compatible `1.0`
  default. If a tip exists, preserve the original `ChainInfo` error.
- Let `getinfo` and `getblockchaininfo` propagate non-empty-state difficulty
  errors instead of assuming the fallback always succeeds.
- Use [`CompactDifficulty::relative_to_network`] for the normalized display
  calculation. It handles the full compact exponent range and is already used
  by the block RPCs.

This helper is not on the `getblocktemplate` call path. Template timestamp
admission remains a separate GBT-boundary concern.

## Testing

- `cargo test -p zakura-rpc --lib chain_tip_difficulty_`
- `cargo test -p zakura-rpc --lib rpc_getdifficulty`
- `cargo test -p zakura-rpc --lib`
- `cargo clippy -p zakura-rpc --all-targets -- -D warnings`
- `npm exec --yes markdownlint-cli -- docs/changelog/unreleased/934.md --config .github/.markdownlint.yaml`
- `git diff --check`

The regression tests cover the empty-state default, error propagation when a
tip exists, and finite normalized values for both a sub-`2^128` network limit
and a sub-`2^128` expected target.

## Changelog

- Added `docs/changelog/unreleased/934.md`.

- V12 finding F-270317: https://v12.sh/runs/7546/270317
- Follow-up V12 finding F-270551: https://v12.sh/runs/7560/270551
- zcashd difficulty calculation: https://github.com/zcash/zcash/blob/7b28054e8b46eb46a9589d0bdc8e29f9fa1dc82d/src/rpc/blockchain.cpp#L35-L73

### DIFF
--- a/crates/zakura-rpc/src/methods.rs
+++ b/crates/zakura-rpc/src/methods.rs
@@ -5457,7 +5457,8 @@ where
     // # TODO
     // - add a separate request like BestChainNextMedianTimePast, but skipping the
     //   consistency check, because any block's difficulty is ok for display
-    // - return 1.0 for a "not enough blocks in the state" error, like `zcashd`:
+    //
+    // Return the minimum difficulty when the state is unavailable, like `zcashd`:
     // <https://github.com/zcash/zcash/blob/7b28054e8b46eb46a9589d0bdc8e29f9fa1dc82d/src/rpc/blockchain.cpp#L40-L41>
     let response = state
         .ready()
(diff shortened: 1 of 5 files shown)


ValarDragon:
Addressed F-270551 and tightened the F-270317 fix in 793f2b020: the RPC now uses `CompactDifficulty::relative_to_network`, returns `1.0` only after confirming the state has no tip, and propagates the original `ChainInfo` error when a tip exists. Regression coverage includes both sub-`2^128` cases.

ValarDragon:
This is a problem, but im not happy with this PR and we have higher priority fixes

PR #935, opened 2026-09-08
feat(network): upgrade Iroh with retained Dalek dependencies

czarcas7ic:
## Motivation

Upgrade native P2P from Iroh 0.92 to the four published `zakura-iroh*` packages at `1.1.0-rc.1`, retaining BIP32, Dalek 2.2/4.1.3 and noq 1.2. Mainnet still defaults to legacy TCP.

## Solution

Native protocol 2 separates incompatible transport versions while mixed old/new dual-stack nodes retain TCP fallback. Checkpoint writes can compute tree updates normally when native VCT metadata is unavailable and the saved trees are current. Frozen trees still require authenticated native metadata. Per-IP admission now accounts for concurrent connections even when Iroh selects a path on the other connection. QUIC NAT traversal is explicitly disabled, preventing its new candidate-address exchange and peer-directed probes. Configured native binds use one address family, occupied ports fail startup, and router port forwarding must be configured manually. The removed public APIs require `zakura-network` 8.0.0 and `zakura-rpc` 10.0.1-rc1.

Known upstream defects retain their documented exact-version exemptions in the dependency decision. Rc1 changes only the transport opt-out and its tests in the fork. Other Rust source and cryptographic dependency versions are unchanged.

## Testing

The admission regression fails on the previous PR head because the accepted connection consumes no IP slot, and passes with the fix. The fork negotiation regression fails before zero is accepted and passes afterward. The tests cover both endpoint directions, an enabled control, direct data transfer, duplicate cleanup and rejection of a further identity at the IP cap. The updated node passes all 1,164 enabled network unit tests, the fork passes 115 unit tests and targeted Clippy, all four release archives rebuild, and registry checksums match. Formatting, cargo-vet and all-feature cargo-deny pass.

CI passes at `343f26c31`. On two isolated c-16 VMs, an rc1 native-only follower received and committed 5,172 Mainnet blocks from an rc1 provider, including 3,086 full-verification blocks. Block hashes and Sapling/Orchard trees matched an independent reference. All three tree pools matched between the nodes, and serving resumed after a provider restart. Public TCP peer rejection also occurred on main, so the final restart-serving check used a temporary main TCP source feeding the provider.

Six controlled native sync runs against main `9b83b94fd` covered the same 4,776 blocks with VCT disabled to compare equal work. Median active throughput was 20.96 blocks/s on main versus 20.82 on rc1 (-0.63%). Full verification was 12.69 versus 12.54 blocks/s (-1.11%). Median startup-to-target time was 239.1s versus 243.1s. One rc1 run took 456.1s after hitting the request-before-Status cooldown previously reproduced on the same main executable. It remains in the results, and this sample does not establish a change in its frequency.

A final VCT-enabled pair also passed, at 7.18 active blocks/s on main versus 20.97 on rc1. Main spent 446.6s waiting for metadata. Rc1 recomputed the first 1,000 checkpoint blocks, then used VCT for the remaining 1,086 without recorded metadata waits. This improvement comes from the sync fallback. Use the equal-work measurements above to assess transport throughput. These are snapshot catch-up tests, not a full genesis replay.

Used Codex and Claude for implementation, review, documentation and validation.


Cargo.toml
```diff
@@ -435,3 +435,10 @@ unwrap_in_result = "warn"
 
 # TODOs: fix this lint eventually
 result_large_err = "allow"
+
+# Temporary Iroh compatibility source; downstream workspaces must apply these patches too.
+[patch.crates-io]
+iroh = { git = "https://github.com/zakura-core/iroh", rev = "9e6cbf251d2edd6304c2d9799fc29bf39788d305" }
+iroh-base = { git = "https://github.com/zakura-core/iroh", rev = "9e6cbf251d2edd6304c2d9799fc29bf39788d305" }
+iroh-relay = { git = "https://github.com/zakura-core/iroh", rev = "9e6cbf251d2edd6304c2d9799fc29bf39788d305" }
+iroh-dns = { git = "https://github.com/zakura-core/iroh", rev = "9e6cbf251d2edd6304c2d9799fc29bf39788d305" }
```
czarcas7ic: Remove after we agree on the fork, create a crate for it, and tag

deny.toml
```diff
@@ -210,7 +212,8 @@ unknown-git = "deny"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
 
 # List of URLs for allowed Git repositories
-allow-git = []
+allow-git = ["https://github.com/zakura-core/iroh"]
+required-git-spec = "rev"
```
czarcas7ic: will remove

Cargo.toml
czarcas7ic: To change

.github/workflows/scripts/patch_registry_for_semver.sh
```diff
@@ -14,6 +14,23 @@ set -euo pipefail
 #   build script panics with "a `libclang` shared library is not loaded on this
 #   thread". The workspace enables `bindgen-runtime`, but published baselines
 #   cannot change, so give the crate's bindgen build-dependency the feature.
+# - precis-profiles 0.1.14 (published 2026-09-08) moved to precis-core 0.2,
```
czarcas7ic: The iroh stuff in this file is temporary, removing after crate is published

crates/zakura-network/src/peer/handshake/tests.rs
```diff
@@ -299,6 +299,71 @@ async fn start_test_zakura_endpoint_with_registry() -> (crate::zakura::ZakuraEnd
     (endpoint, peer_registry)
 }
 
+/// Process driver proving different native cohorts retain useful legacy TCP.
+#[allow(clippy::print_stdout)] // Machine-readable protocol for the process runner.
+#[tokio::test]
+#[ignore = "requires a separately started peer process"]
+async fn legacy_process_peer() -> Result<(), BoxError> {
```
czarcas7ic: Just highlighting here that Zakura v2 communicates through TCP with Zakura v1 peers to keep compatibility.

crates/zakura-rpc/src/methods/tests/vectors.rs
```diff
@@ -1540,13 +1540,10 @@ async fn rpc_getblock_preserves_historical_tree_error() {
         .respond(ReadResponse::TransactionIdsForBlock(Some(Arc::from(
             tx_hashes.into_boxed_slice(),
         ))));
-    read_state
```
czarcas7ic, line 1543: Changes in this file are fixing an existing flakey test

docker/zakura-regtest-e2e/node2.toml
```diff
@@ -28,6 +28,7 @@ initial_testnet_peers = []
 max_connections_per_ip = 10
 
 [network.zakura]
+listen_addr = "127.0.0.1:18334"
```
czarcas7ic, line 31: Iroh used to silently pick another port when one was taken, and it now fails startup, so we define it explicitly here so we dont have overlap.

czarcas7ic:
Failing tests will pass once crate is published

evan-forbes (approved):
pre-approving as we need to publish a crate before ci works

czarcas7ic:
Syncing nodes on this branch now

evan-forbes (approved):

PR #938, opened 2026-09-09
perf(network): prepare checkpoint prefixes before body starvation

czarcas7ic:
## Motivation

A selected-chain header request can keep accumulating pages while block validation runs out of admitted work. For example, with at most 400 admitted blocks remaining, two 250-header pages already provide enough headers for a checkpoint and its VCT successor, but the reactor previously waited for the target or a capacity boundary.

## Solution

Prepare that prefix through the existing validation and ownership path once more than 400 headers are staged. Apply this only to normal requests extending the exact selected tip in integrated mode; retain the existing behavior for competing branches, headers-only operation and auxiliary repair. Request and memory budgets are unchanged.

## Testing

Formatting, Markdown lint, changelog validation and release Clippy pass. The release network suite ran 1,119 tests: 1,116 passed (one reported a leaked handle), three existing listener tests failed binding `127.0.0.2`, and 45 were skipped. A standalone socket probe reproduces that macOS address error. Both added tests pass and cover the minimum-work boundary, backlog threshold, exact branch scope, excluded modes/purposes and preservation of owned capacity through normal preparation. Earlier isolated VCT-enabled genesis pairs for this optimization measured 144.0→161.0 BPS in dual mode and 144.2→167.9 BPS in Zakura-only mode; these single-pair observations do not establish repeatability. [Dual evidence](https://github.com/zakura-core/zakura/actions/runs/34355015102) · [Zakura-only evidence](https://github.com/zakura-core/zakura/actions/runs/34355150283).

## Changelog

Added `docs/changelog/unreleased/938.md` for the sync improvement.

### DIFF
--- a/crates/zakura-network/src/zakura/header_sync/reactor.rs
+++ b/crates/zakura-network/src/zakura/header_sync/reactor.rs
@@ -1495,6 +1495,10 @@ impl HeaderSyncReactor {
             self.report_misbehavior(peer, HeaderSyncMisbehavior::MalformedMessage);
             return;
         };
+        let checkpoint_prefix_ready = self
+            .committed_snapshot
+            .as_ref()
+            .is_some_and(|snapshot| Self::should_prepare_checkpoint_prefix(snapshot, active));
         let _ = active;
         debug_assert_eq!(
             self.peer_work_queue.owned_header_count(&peer),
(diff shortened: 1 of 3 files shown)


czarcas7ic:
@codex review

czarcas7ic:
v12 findings unrelated to additions of this fix

ValarDragon (approved):
idea makes sense + passes codex xhigh.

I have questions about why only "when admitted block work is running low", since to my knowledge were not CPU saturated though

PR #939, opened 2026-09-09
fix(consensus): preserve checkpoint progress on block rejection

evan-forbes:
## Motivation

A side-chain rejection can reset checkpoint progress to the state tip while a verified range is still committing. That reset reopens a range whose blocks the verifier already consumed, which can stall checkpoint verification.

A canceled caller can also discard recovery after a failed commit. The commit task survives cancellation, but the response future previously owned the reset request.

## Solution

Adapt Zebra #11385 to reset progress only after a non-duplicate `CommitCheckpointVerified` error. Preserve Zakura's duplicate-commit exclusion and reset-generation handling.

Keep reset handling inside the commit task so recovery survives caller cancellation.

## Testing

- `cargo +1.97.0 test -p zakura-consensus --lib checkpoint::tests --locked`: 16 passed.
- `cargo +1.97.0 clippy -p zakura-consensus --lib --tests --locked -- -D warnings`: passed.
- `cargo +1.97.0 fmt --all -- --check` and `git diff --check`: passed.
- The new regression injects a competing hash into the validated queue and holds state commits open. It checks that the rejection preserves checkpoint progress and that both ranges finish. With the original reset filter, it fails because progress rewinds from height 6 to height 3.
- Existing duplicate-request and reset-generation regressions pass.
- A new regression drops the response before a state commit fails. It confirms that the verifier resets from the final checkpoint to the durable state tip. The test fails without the cancellation fix because no reset arrives.
- Combined with #940: 89 sync tests, 5 block-sync driver tests, 5 UTXO lookup tests, 16 checkpoint tests, and 7 router tests passed. One UTXO timing benchmark remained ignored.

## Changelog

`docs/changelog/unreleased/939.md` records the checkpoint verification fix. `./scripts/changelog.py check` passes.

Backports https://github.com/ZcashFoundation/zebra/pull/11385.

Related sync timeout backport: https://github.com/zakura-core/zakura/pull/940. Both PRs target `main` independently.

### DIFF
--- a/crates/zakura-consensus/src/checkpoint.rs
+++ b/crates/zakura-consensus/src/checkpoint.rs
@@ -1261,6 +1261,8 @@ where
         // we don't reject the entire checkpoint.
         // Instead, we reset the verifier to the successfully committed state tip.
         let state_service = self.state_service.clone();
+        let recovery_state = self.state_service.clone();
+        let reset_sender = self.reset_sender.clone();
         let network = self.network.clone();
         let commit_checkpoint_verified = tokio::spawn(async move {
             let queued_result = req_block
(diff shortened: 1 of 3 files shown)


ValarDragon (approved):
I was confused by this until I realized the side-chain issue is a legacy only p2p problem.

ValarDragon:
And the dropped response bug affecting both paths checks out to me!

PR #942, opened 2026-09-09
feat(state): retain resources through bounded block range reads

czarcas7ic:
## Motivation

Cancelling an async block-range read must not release its resources while the database job or its undelivered result still retains blocks. This is the first refactor chunk extracted from #892, now targeting `main`. The merged Iroh upgrade from #935 is included through main.

## Solution

Add a bounded state read that carries caller-owned resources into the blocking job and its result. It returns a contiguous prefix within the byte limit and checks cancellation between lookups. Existing callers and protocol behavior remain unchanged.

## Testing

All owned-range storage tests pass on the combined stack at `6327f652e`, which passes 617 selected network, state, and node-driver tests without retries. Network, state, and node all-target Clippy with warnings denied, formatting, and changelog checks pass. Prepared with Codex assistance.

## Changelog

Added an internal API fragment for this PR.

### DIFF
--- a/crates/zakura-state/src/lib.rs
+++ b/crates/zakura-state/src/lib.rs
@@ -124,7 +124,7 @@ pub use service::read::{
 };
 pub use service::{
     finalized_state::{DiskWriteBatch, FallibleDiskValue, FromDisk, IntoDisk, WriteDisk, ZakuraDb},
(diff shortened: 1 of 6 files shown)


czarcas7ic:
@codex review

evan-forbes (approved):

PR #946, opened 2026-09-09
fix(rpc): prevent invalid submissions from exhausting block commit capacity

evan-forbes:
`submitblock` must reject invalid proof of work without exhausting the capacity used to verify and commit valid blocks. This PR separates bounded proof checking from bounded contextual verification and commit.

Previously, invalid submissions could occupy all 16 contextual commit slots. Checking proofs before reserving those slots avoids that failure, but proof checking also needs a limit: concurrent RPC calls could otherwise occupy async workers with expensive computation.

The RPC now runs proof checks on the blocking pool with a separate shared limit of 16 checks. It rejects concurrent checks for the same block hash. Only a successful check can reserve a contextual commit slot. If a caller cancels its RPC, the blocking task retains its slot until the computation finishes. Cancellation therefore cannot bypass the limit by leaving uncounted work running.

The tracked entries represent outstanding computation. Their lifetime follows that computation, including after an RPC ends.

The PR also enforces the active network's solution encoding rules on PoW-waived submissions and block proposals. Proposals continue to skip proof computation. The semantic verifier repeats the network policy check.

Validation:

- Five `rpc_submitblock` tests pass. Coverage includes invalid-proof floods, custom-network waivers, detached commit ownership, and cancellation while a proof check waits for the blocking pool.
- All 46 consensus block tests pass. Shape regressions cover Mainnet, Testnet, Regtest, and a configured disabled-PoW network.
- Rust 1.97 Clippy for both crates' libraries and tests passes with warnings denied. Formatting, diff, and changelog checks pass.

`docs/changelog/unreleased/946.md` records the behavior. The parameter ledger records the 16-check precheck bound.


ValarDragon:
`submitblock` as-is is supposed to be from a mining pool to the node correct?

If so, aren't we adding extra blocking equihash delay, when we really should just be submitting ASAP?

evan-forbes:
> submitblock as-is is supposed to be from a mining pool to the node correct?
If so, aren't we adding extra blocking equihash delay, when we really should just be submitting ASAP?

to also post here on slack, yeah generally this is the correct take. the issue with this method is that its stateful and requires an break to users to stop exposing it. miners ofc need it exposed, normal users definitely should not expose this.

perhaps the better approach is to keep by default serving this method, however have some sort of init that will auto turn this method off since its not secure to expose

PR #948, opened 2026-09-09
fix(state): handle repeated invalid verdicts and missing context

evan-forbes:
## Motivation

Three confirmed failure paths remain in the consensus audit:

- CA-004: a second invalid verdict for one block hash fails the header transition when its rule or evidence differs.
- CA-029: an interrupted VCT database can reopen on the legacy P2P stack with VCT enabled, then wait for roots that the stack cannot deliver.
- CA-021: mining-template construction panics when readable bodies do not provide the complete difficulty window.

## Solution

Preserve the first invalid-body tombstone when later invalid verdicts arrive. Keep rejecting attempts to replace invalidity with a valid or unknown state. Check header-root delivery configuration when reopening an interrupted VCT database for writes. Preserve read-only access when checkpoint sync, VCT fast sync, or native P2P is disabled. The startup diagnostic names the supported `zakura` and `dual` settings. Propagate incomplete difficulty context through the existing state/RPC error channel.

## Testing

Rust 1.97:

- 43 selected header-chain tests passed, including staged and persisted tombstone replay, eligibility invariants, and a repeated body-evidence transition with different evidence, rule, and supplier.
- 26 selected state tests passed, including interrupted VCT reopen on legacy versus v2, completed VCT reopen, difficulty windows from genesis through the full 28-header span on both networks, and existing anchor/UTXO checks.
- `cargo fmt --all -- --check` and `git diff --check` passed.
- `cargo clippy -p zakura-header-chain -p zakura-state --lib --tests --locked -- -D warnings` passed.
- `scripts/check-release-state.sh` passed without an override.
- `scripts/changelog.py check` passed.

V12 follow-up validation at `e5313ba20`: all 27 state pruning/startup tests passed, including the new eight-combination read-only interrupted-VCT regression and existing writable/completed reopen cases. State Clippy, formatting, diff, and fragment validation passed.

## Changelog

Added `docs/changelog/unreleased/948.md`.

### Follow-up Work

CA-001 remains a release-rollout deadline; this PR preserves the support policy deliberately set in #927. CA-002's shared writer-failure reporting already exists on main. Its coherence failures and CA-003's capacity recovery need a reproducer before replacing failure handling with retries. The exact-parent contextual checks do not establish CA-005's proposed transient path, so their classification stays unchanged.

CA-011 remains conditional on NU7. Section F's comparison harnesses remain separate work and were not run here. An oracle mismatch needs reachability triage before receiving FORK-HIGH severity.

This PR overlaps #924 in the mining-template helper. The two PRs address different failures and may need reconciliation before merge.

### DIFF
--- a/crates/zakura-header-chain/src/graph/mod.rs
+++ b/crates/zakura-header-chain/src/graph/mod.rs
@@ -639,10 +639,9 @@ impl MemHeaderStore {
     ///
     /// Returns `true` when the state changes and `false` when it is unchanged.
     /// A consensus-invalid body-validation state makes the node and its
-    /// descendants ineligible. Replaying the same consensus-invalid
-    /// body-validation state leaves the node unchanged. This method rejects a
-    /// different consensus-invalid body-validation state. It also rejects any
-    /// later non-invalid body-validation state.
+    /// descendants ineligible. Further consensus-invalid verdicts leave the
+    /// first verdict unchanged, even if their evidence or rule differs.
+    /// This method rejects any later non-invalid body-validation state.
     /// Consensus invalidity is permanent.
     ///
     /// This method does not enforce transitions between non-invalid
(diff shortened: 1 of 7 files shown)


ValarDragon (approved):
LGTM! Nice fixes

PR #950, opened 2026-09-09
fix(sync): reserve VCT repair capacity and supervise writer failures

evan-forbes:
## Motivation

A failed block writer could leave the daemon restarting sync against failed state. Auxiliary input could also exhaust VCT repair capacity, and peer churn or failed context reads could suppress the capacity deadline. A recovered repair could inherit an expired deadline from an earlier blockage.

## Solution

- Notify daemon supervision when the writer publishes its first failure. Preserve failure reporting to state-service clones and keep missing roots retryable.
- Reserve auxiliary slots for the finalized header and its first two selected successors on every integrated transition, including fork changes without auxiliary input. The planner and independent verifier enforce the reserve.
- Under reserve pressure, retention evicts only unprotected headers that hold input, with their descendants. Permanently ineligible holders go first, then the holder whose best descendant tip has the least work. Auxiliary pressure evicts no other header, so it no longer removes side branches without input or sweeps permanently ineligible subtrees.
- Bound the reserve by the larger of the aggregate limit and the pre-transition occupancy plus reserve. A database written before this PR can hold protected selected-path input inside the reserve. Transitions from that state can keep or reduce the deficit but cannot increase it: speculative input stays refused, commit-window input fills reserved slots, and finality drains the protected rows.
- Before publishing startup state, check the recovered auxiliary reserve and run settlement when it is deficient, even when finalized state already agrees and no deferred header is due. Preserve existing resource-stall alarms when the reserve is satisfied.
- Replace rejected or disputed input in a full per-header bucket. A selected repair may also replace unchecked input, including recovered rows whose outcome claims recovery discarded. Update the row and header index atomically. Preserve authenticated input on retained headers.
- Use reserve-aware repair preflight and range limits. Withdraw speculative sweep requests when only commit capacity remains. Preserve the committer's repair request.
- Let the repair task own its capacity deadline and keep its policy state private. Task transitions start the absolute capacity deadline on blockage. Preserve it through peer events and failed context reads. Clear it on positive capacity resolution before assignment. A new blockage gets a new deadline. Emit one fatal event after thirty continuous minutes.
- Request repair context immediately after a capacity refusal replays an already-observed state version. The capacity deadline replaced the idle maintenance wake that used to dispatch this recheck.

Startup settlement can reclaim unprotected rows from an older saturated database. It cannot discard protected evidence or reconstruct inconsistent authoritative data. When protected input keeps the reserve short, startup still succeeds and finality drains the deficit. Input replacement bounds rejection history; it grants no header validity or root authority.

## Testing

Rust 1.98.0, after merging main (#970):

- Header-chain, network, and state: 2,163 passed, 0 failed, 9 ignored.
- Generated task sequences check deadline preservation, reset, retirement, and exactly-once expiry against an independent model.
- Thirty-two generated persistent-storage cases mix saturation, rejection, replacement, fork changes, finality, and database reopen. An independent oracle checks reserve bounds, authenticated-input retention, row/index consistency, and refusal atomicity.
- Final strengthened deadline regression passed with a simultaneous snapshot update and exactly one fatal event.
- Clippy passed for header-chain, network, state, and `zakurad` libraries/tests with warnings denied.
- Conformance validated all 119 rules. Formatting, diff, and changelog checks passed.

The startup regression failed on an earlier PR head because all five auxiliary rows survived reopen. It now covers both startup entry points, reserve availability before publication, and persistence across a second RocksDB reopen. The existing resource-stall alarm reopen regression also passes.

These regressions fail on the code before their fixes:

- `startup_publishes_a_protected_reserve_deficit_without_discarding_evidence` and `protected_reserve_deficit_refuses_growth_but_drains_through_finality` fail against an absolute reserve bound. Startup returned `AuxiliaryLimitExceeded`, and the planner refused even commit-window repairs, so a deficit could never drain.
- `auxiliary_pressure_evicts_only_subtrees_that_hold_input` and `auxiliary_pressure_reclaims_input_on_a_shared_fork_header` fail against leaf-order eviction, which removed empty side branches before the input holder. `auxiliary_pressure_evicts_permanently_ineligible_input_first` pins the holder order.
- `superseded_resource_refusal_rechecks_context_without_a_timer` fails without the immediate context request.

Regressions cover fork changes at saturation, protected-branch retention, startup settlement of an older saturated selection, a protected reserve deficit at startup and through finality, input-targeted eviction, negative-input replacement, database reopen and selected replacement, reserve-aware F+3 preflight, speculative sweep withdrawal, continuous peer churn, failed context reads, a superseded capacity refusal, and recovery followed by immediate assignment and another blockage.

## Changelog

Updated `docs/changelog/unreleased/950.md`, the parameter ledger, LC-AUX-06/08, the conformance manifest, and transition recovery documentation.


czarcas7ic (approved):
I think the new cleanup works, but I believe there is a startup case that might miss this clean up logic.

For example, using a made up limit of five auxiliary records, suppose the database has one record for finalized block 100 and four on an obsolete side branch. The selected blocks 101 and 102 need auxiliary data before the node can continue. The new retention loop can remove that old branch and free the space:
https://github.com/zakura-core/zakura/blob/b03d2d3e44bc34a7a9d7406be3e31832f10530b8/crates/zakura-header-chain/src/transition/planner/retention.rs#L158-L176

But suppose we restart with the saved block state and header chain records already agreeing at block 100, and no deferred headers due for a recheck. Startup can skip the planner here:
https://github.com/zakura-core/zakura/blob/b03d2d3e44bc34a7a9d7406be3e31832f10530b8/crates/zakura-state/src/service/finalized_state/header_chain.rs#L762-L783

The store then stays full, and repair for block 101 is refused until another transition frees space. So this is an existing blockage that the startup recovery can leave unresolved.

The new test calls apply_transition directly, which proves the cleanup works when called, but doesn’t check whether startup reaches it:
https://github.com/zakura-core/zakura/blob/b03d2d3e44bc34a7a9d7406be3e31832f10530b8/crates/zakura-header-chain/src/transition/planner/tests/auxiliary.rs#L1404-L1415

Could we run the reserve cleanup before publishing startup state even when no deferred header is due, and cover this with a real database reopen test?

evan-forbes:
Addressed in 3e8f99082 and 745de8c84. Startup now checks the auxiliary reserve and runs cleanup before publication when needed, even when finalized state agrees and no deferred header is due.

The RocksDB reopen regression failed before the fix and now passes for both startup entry points and a second reopen. Existing resource-stall alarms remain preserved. All header-chain, network, and state library tests pass.

czarcas7ic (approved):
Confirmed previous comment was addressed

PR #957, opened 2026-09-10
fix: stop banning peers for non-final mempool transactions and harden misbehavior scoring

ebfull:
## Motivation

The peer misbehavior scoring around mempool transaction verification punishes some failures that are relative to the local chain view rather than evidence of peer misconduct, and the surrounding penalty plumbing has a few rough edges:

- `LockedUntilAfterBlockHeight` and `LockedUntilAfterBlockTime` were scored at 100, which is exactly the ban threshold (`MAX_PEER_MISBEHAVIOR_SCORE`). Lock-time finality is checked against the local next-block height and median-time-past, so an honest peer relaying a just-final transaction can be banned by a node whose tip or chain time lags slightly behind. The ban is IP-wide and has no expiry, so catching up does not undo it.
- Misbehavior scores were accumulated with ordinary unsigned addition in the report batcher and in the address book metadata merge, so peer-driven report streams could in principle wrap an accumulated score (or panic in debug builds), violating the monotonic-penalty invariant.
- Misbehavior reports are sent over a bounded channel with `try_send`, and drops were silent, so operators could not observe missed score updates.
- The `bans_by_ip` field comment claimed ban times are stored; the `BanList` type stores only an IP set and FIFO insertion order.
- The mempool activation log asserted "Zakura is close to the tip" as fact, although the sync status is a heuristic over recent sync lengths.

## Solution

- Classify both lock-time errors as non-punitive (score 0) in `mempool_misbehavior_score`. The transaction is still rejected and still enters the exact-tip rejection cache (cleared when the tip grows), so nothing becomes acceptable that wasn't before. This matches zcashd, which rejects non-final transactions as premature without a DoS penalty. The classifier is shared with block verification via `VerifyBlockError::misbehavior_score`, where the same tip-relative reasoning applies.
- Use `saturating_add` in both score-accumulation sites, matching the existing saturating sum in the sync-status length average.
- Count dropped misbehavior reports in a new `mempool.misbehavior.reports.dropped.total` metric.
- Correct the `bans_by_ip` comment to document the actual ban lifetime (process-local, no expiry, FIFO-evicted at capacity) and attribute the activation log's readiness claim to the heuristic.

## Testing

- `lock_time_errors_have_no_misbehavior_score` asserts the classifier scores both lock-time variants at 0, alongside the existing tests that keep policy rejections at 0 and context-independent verification failures at 100.
- `nonfinal_lock_time_rejections_have_no_misbehavior_score` asserts the mempool's misbehavior-report path returns no `(address, score)` pair for attributed lock-time failures, exercising the same helper that feeds the ban channel.
- `misbehavior_score_accumulation_saturates` applies misbehavior changes to never-attempted and attempted address book entries already near `u32::MAX` and asserts the score saturates instead of wrapping (the pre-fix code overflows in debug builds).
- Ran `cargo fmt --all -- --check` and `cargo clippy --locked -p zakura-network -p zakura-consensus -p zakura --all-targets -- -D warnings` (clean), the mempool test suite (96 passed), the consensus error/block score tests (15 passed), and the `zakura-network` lib suite (1156 passed). Three `zcashd_compat` listener tests failed locally on macOS because binding `127.0.0.2` needs an `lo0` alias; they fail identically on `main` in the same environment and are unrelated to this change (Linux CI can bind any `127.0.0.0/8` address).
- Acceptance tests match on the `"activating mempool"` log prefix, which is preserved.
- `./scripts/changelog.py check` validates the fragment.

## Changelog

`docs/changelog/unreleased/957.md` adds two `Fixed` entries.

### Follow-up Work

- Non-final lock times are one class of tip-relative rejection; upgrade-context-dependent failures (branch IDs outside the existing NU6.3 grace window, v4 signature-context mismatches) still score 100 and need a separate, more careful treatment.
- Ban lifetime remains unbounded; expiry/reconsideration is out of scope here.

### DIFF
--- a/crates/zakura-consensus/src/error.rs
+++ b/crates/zakura-consensus/src/error.rs
@@ -546,9 +546,15 @@ impl TransactionError {
             | OrchardProofSize
             | IronwoodProofSize
             | WrongConsensusBranchId
-            | MissingConsensusBranchId
-            | LockedUntilAfterBlockHeight(_)
-            | LockedUntilAfterBlockTime(_) => 100,
+            | MissingConsensusBranchId => 100,
+
+            // Lock-time finality is checked against the local next-block height
+            // and median-time-past, so a transaction that is final for the
+            // sender can be non-final on a node whose tip or chain time lags
+            // slightly behind. Like zcashd, treat non-final transactions as
+            // premature rather than malicious: reject them without a peer
+            // penalty.
+            LockedUntilAfterBlockHeight(_) | LockedUntilAfterBlockTime(_) => 0,
 
             // NU6.2 mempool transactions are invalid under NU6.3 rules, but
             // honest peers can relay them briefly while their chain tips converge.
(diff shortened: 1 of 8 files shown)


ebfull:
Agent opened a PR for this on its own... I hadn't reviewed the code yet.

PR #960, opened 2026-09-10
fix(mempool): preserve stateless bans and disconnect repeat offenders

evan-forbes:
## Motivation

The mempool currently bans peers for transaction failures that can depend on the node's local chain state. A node behind the network can reject an honest peer's transaction because it uses a different consensus branch or lock context. The resulting IP ban does not expire.

Stateless failures still justify bans. Context-dependent failures need a temporary penalty, with a disconnect if the peer repeats them after the penalty expires.

## Solution

- Keep bans for stateless transaction failures, including invalid Halo2/Groth16 proofs, malformed Groth16 proofs, invalid shielded proof sizes, and invalid transaction structure. These bans do not depend on tip freshness.
- Apply per-IP cooldowns to other previously scored failures only when the verification tip matches the current tip. Policy rejections, duplicate spends, consensus branch ID errors, and lock time errors start no cooldown.
- During a cooldown, ignore transaction advertisements and pushed transactions from that IP. Skip verification of transactions that the IP serves to queued downloads, including crawler downloads. Do not mark skipped transactions as rejected, so another peer can relay them.
- Start with a 10-minute cooldown. Each new eligible failure after the cooldown expires doubles the cooldown, up to 2 hours, and disconnects the peer without banning it. Failures from transactions already in verification during an active cooldown do not add strikes. Stateless failures still trigger bans.
- Route disconnects through a bounded local peer-set control channel. Disconnects work when all peers are busy and when the target is the last connected peer.
- Keep cooldown history across reconnects and mempool resets. Canonicalize IPv4-mapped IPv6 addresses. Forget history 2 hours after the last cooldown ends.
- Bound cooldown history at 20,000 IPs. Use an expiry index to evict expired entries. If every entry is active, pause transactions from untracked IPs until a slot expires.
- Preserve transaction source attribution and per-peer queue accounting across tip-reset retries.
- Reject download responses containing a different transaction than requested. This prevents the downloader from leaking the requested transaction's cancel handle and per-peer slot.
- Add `mempool.peer_cooldown.started.total` and `mempool.peer_cooldown.ignored.transactions.total` metrics.

The combined Sapling proof/signature error remains eligible for cooldowns because signature verification can depend on the consensus branch. Halo2 verification runs after the verifier checks the encoded branch ID for v5/v6 transactions.

Block verification keeps its existing ban policy.

## Testing

Tests cover stateless bans without a current tip, context-dependent errors avoiding bans, first-failure cooldowns, disconnects after repeated failures, and pending failures not adding strikes. Peer-set tests cover targeted disconnects without bans, busy peers, an unready network service, and closing the last peer through an idle buffer.

Existing mempool tests cover cooldown timing and capacity, source retention across tip resets, skipped crawler downloads, relay from another peer, and cleanup after a substituted transaction response. The consensus branch test checks v5 branch mismatch and grace-period boundaries.

Validation on Rust 1.98:

- `cargo test -p zakura --lib components::mempool --locked`: 118 passed.
- `cargo test -p zakura-network --lib peer_set::set::tests --locked`: 34 passed.
- `cargo test -p zakura-consensus --lib v5_consensus_branch_ids --locked`: 1 passed.
- `cargo clippy -p zakura -p zakura-network --all-targets --locked -- -D warnings`: passed.
- `cargo fmt --all -- --check`: passed.
- `./scripts/changelog.py check`: passed.
- `git diff --check`: passed.

## Changelog

`docs/changelog/unreleased/960.md`

- #27 addresses mempool activation and scoring while the node is behind the network.
- #957 proposed removing penalties for lock time failures. This PR excludes those failures from mempool cooldowns and preserves block scoring.

### Follow-up Work

- Attribute transaction failures from native Zakura peers. Their transactions still carry no advertiser address, so these failures cannot trigger a ban or cooldown.
- Consider grouping IPv6 cooldowns by /64 prefix.
- Bind Zakura transaction responses to requested IDs in the codec, as it already does for blocks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,7 +73,6 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
- "bytes",
  "crypto-common 0.1.7",
  "generic-array",
 ]
@@ -87,6 +86,20 @@ dependencies = [
  "cfg-if",
  "cipher",
  "cpufeatures 0.2.17",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.10.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
 ]
 
 [[package]]
@@ -210,19 +223,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
-name = "async-compat"
-version = "0.2.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1ba85bc55464dcbf728b56d97e119d673f4cf9062be330a9a26f3acf504a590"
-dependencies = [
- "futures-core",
- "futures-io",
- "once_cell",
- "pin-project-lite",
- "tokio",
-]
-
-[[package]]
 name = "async-trait"
 version = "0.1.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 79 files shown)


ValarDragon:
I don't think this is quite right. We should always be banning for anything that fails a stateless check imo.

This is true for stateful checks though

evan-forbes:
yeah that makes sense, not sure what I was fully thinking here for the complete removal of bans. I think that was a bit overzealous interpretation from the original prompt but I still should have gone through it more, apologies. 

refactored to keep more of the original intent, which is that we should not be perma-banning peers if there is a stateful error, since that can and does occur naturally.            

the thinking being that just as its important to disconnect from a malicious peers trying to waste resources, its similarly important to avoid disconnecting from honest peers from a false negative. that currently exposes attack surfaces, where an attacker could use that to partition the network.

PR #965, opened 2026-09-10
feat(trace): migrate all structured traces to CSV

evan-forbes:
## Motivation

Structured node traces repeated field names in every JSONL row. Migrating the storage format also exposed an evidence-contract problem: the regtest oracle could report PASS after event loss, rotation, missing tables, or incorrect cross-process matching.

## Solution

All twelve structured trace tables now use CSV with a shared schema, typed fields, UTC timestamps, and an extra column for undeclared fields. The trace crate advances to 2.0.0. The versioned envelope uses one monotonic clock per process. Readers preserve textual identities and reject malformed input within explicit resource limits.

Production tracing remains bounded and asynchronous. Validation captures use a fresh run ID, disable rotation, and record accepted rows, dropped rows, and writer failures. Before a planned restart or final validation, the harness seals each writer and waits for its drain acknowledgement. The oracle requires sealed counts, matching rows, and the expected node/table coverage. Missing or damaged evidence produces INCOMPLETE instead of PASS.

The oracle and flush guard share one ordered commit matcher that consumes each finish once within its process and source. The oracle and benchmark digest share a Python decoder. Rust and Python readers use the same contract fixtures. File access rejects symlinks and nonregular files, and snapshot locks protect reads from append and rotation.

Trace-runtime environment variables bypass the node's TOML loader. Pipeline-only snapshots have their own event name so full-snapshot checks cannot mistake omitted counters for zero. The benchmark deployment includes the shared decoder and schema.

The [design review](https://github.com/zakura-core/zakura/blob/evan/trace-csv/docs/design/trace-csv-evidence.md) maps all seventeen auditor findings to their root corrections and documents the remaining operational limits.

## Testing

- The full Docker `pr-gate` passes: restart, checkpoint handoff, catch-up, reorg, legacy compatibility, capture sealing, and the final trace oracle.
- 26 trace crate tests, 41 network trace tests, and 24 node configuration tests pass with Rust 1.97.
- The oracle self-test, 11 trace/archive Python tests, and 24 benchmark digest tests pass. Regression coverage includes cross-process matching, reused finishes, missing evidence, sealed counts, event loss, malformed input, symlinks/FIFOs, and resource limits.
- Trace and network Clippy checks pass with all targets; network checks enable `zakura-testkit`. The node binary builds successfully.
- Formatting, shell syntax, workflow actionlint, Markdown lint, changelog validation, and diff checks pass. Package contents include the schema and shared CSV fixtures.
- The extended restart matrix and full workspace test suite were not run. GitHub checks will run on the updated commit.

## Changelog

Updated `docs/changelog/unreleased/965.md`.


ValarDragon:
Reviewed all 17 findings against rebased head `541319f22`. The rebase itself
does not resolve these findings.

- **F-273501 — confirmed, directly in scope.** `wall_ts` is emitted but the
  oracle still orders and compares rows using emitter-local `ts`. Cross-table
  activity and checkpoint-handoff timing can therefore compare unrelated
  clocks, and appended process generations can overlap. The oracle should use
  parsed `wall_ts` for cross-emitter timing and retain process attribution where
  an invariant is process-local.
- **F-273502 — confirmed; overlaps F-273541.** A nonempty/header-only
  `commit_state.csv` passes the shell layout check and zero starts/zero finishes
  pass the balance checks. Required nodes need a validated commit header and
  the event coverage required by the selected test mode.
- **F-273519 — confirmed malformed-input handling gap.** `json.loads()` can
  raise `RecursionError`, which is not converted to `csv_decode_error`. This
  contradicts the PR's stated behavior for malformed CSV.
- **F-273520 — confirmed malformed-input handling gap.** `extra` can replace a
  key with a non-scalar value and make `row_key()` raise `TypeError`. Canonical
  fields need type validation before invariant evaluation; F-273544's collision
  rejection also removes this path.
- **F-273524 — valid export hardening, not a remote-node vulnerability.** The
  cited value comes from operator-controlled environment/hostname metadata and
  these traces are a machine-data format, not a spreadsheet interchange
  format. Spreadsheet-safe escaping or an explicit documentation warning would
  still be reasonable if opening raw traces in formula-evaluating software is a
  supported workflow.
- **F-273530 — confirmed.** Any recognized root entry suppresses nested
  discovery even when that root entry is empty or invalid. Discovery should
  select valid trace candidates, or fall back to nested candidates when the
  root produces no usable trace set.
- **F-273534 — confirmed.** The harness creates/reuses its trace directory and
  the oracle has no current-run boundary, so old rows can satisfy a later run.
  The harness/oracle should bind validation to the current run, using a recorded
  start boundary and `wall_ts`/`process_trace_id`.
- **F-273535 — technically valid local-filesystem hardening.** A FIFO or device
  at a recognized path can block the synchronous oracle. The trace directory is
  operator/CI-controlled rather than peer-controlled, so this is not a remote
  node attack path. Requiring regular files is a small, sensible hardening.
- **F-273536 — technically valid local-filesystem hardening.** The JSONL reader
  follows symlinks and includes a bounded source excerpt in diagnostics. This
  requires local control of the trace tree; it is not remotely reachable.
  Rejecting symlinks and enforcing root containment would make the oracle safer
  for arbitrary imported trace bundles.
- **F-273540 — confirmed resource-hardening issue under adversarial trace
  input.** Diagnostics rescan node tables for each failure, so many failures can
  produce quadratic work. Compute diagnostics once per node and cap retained or
  reported failures if arbitrary trace bundles are in scope.
- **F-273541 — confirmed; broader form of F-273502.** Per-node table and event
  coverage is not required, so unrelated trace rows can keep a node present
  while commit/frontier invariants pass vacuously. Required coverage should be
  explicit per test mode.
- **F-273542 — confirmed pre-existing JSONL ambiguity.** Default
  `json.loads()` accepts duplicate keys with last-value-wins semantics. If the
  oracle treats trace files as untrusted validation input, duplicate keys should
  become a decode/schema failure. This is not introduced by the CSV migration.
- **F-273544 — confirmed, directly in scope.** CSV headers are not checked
  against the table schema and `extra` can overwrite declared fields in both
  readers. Reject duplicate headers, unexpected/missing canonical columns, and
  collisions between `extra` and declared/envelope fields before materializing
  the row.
- **F-273545 — valid resource-hardening issue for arbitrary imported traces.**
  Both readers materialize all rows without a byte/record budget. Locally
  produced CI traces are trusted and bounded by test duration, so this is not a
  remote node issue, but imported/untrusted bundle support needs explicit
  limits or streaming validation.
- **F-273546 — confirmed, directly caused by the CSV migration.** The shell
  guards count physical lines and substrings rather than RFC 4180 records and
  commit identities. They can be confused by multiline fields or unmatched
  start/finish rows with equal totals. The guard should use the CSV parser and
  match the same identity used by the oracle.
- **F-273548 — technically valid local-filesystem hardening.** A prepopulated
  output symlink or special file can redirect or block the writer, but exploiting
  it requires local write access to the configured trace directory. Production
  event emission remains nonblocking; explicit guarded shutdown can wait on a
  blocked writer. No-follow regular-file opening and bounded shutdown would
  harden this local boundary.
- **F-273549 — confirmed pre-existing schema gap.** JSON syntax alone is not
  enough for rows consumed by invariants. Missing/non-scalar identity fields can
  be skipped or reach unsafe code. Validate the event discriminator and the
  scalar fields used by each invariant, reporting a trace-schema failure rather
  than silently dropping or crashing.

None of these findings exposes a peer-to-node or consensus attack path. I would
address the oracle-correctness findings—F-273501, F-273502/F-273541, F-273519,
F-273520, F-273530, F-273534, F-273544, and F-273546—before relying on this
oracle as a release gate. The remaining findings are either pre-existing JSONL
validation gaps or local/imported-trace hardening and can be prioritized under
an explicit trace-input trust model.

ValarDragon:
Follow-up commit [`b7634185e`](https://github.com/zakura-core/zakura/commit/b7634185e) addresses the parser/oracle correctness cluster from V12:

- **F-273502 / F-273541:** the e2e harness now names the nodes that require commit evidence (`node2` and `node4`). A header-only commit CSV, a required node with only unrelated traces, or a required commit table missing either `commit_start` or `commit_finish` fails instead of passing vacuously. Node1 is deliberately excluded because locally mined RPC blocks do not traverse the traced block-driver commit path.
- **F-273519:** excessive JSON nesting is converted into a trace decode failure instead of escaping as `RecursionError`.
- **F-273520:** correlation keys now accept only scalar string identity fields, and canonical field types are validated before invariant evaluation.
- **F-273542:** duplicate keys in JSONL rows and CSV `extra` objects are rejected.
- **F-273544:** `commit_state.csv` is checked against its exact producer schema; duplicate headers and collisions between `extra` and declared/envelope columns are rejected. The Rust test reader also requires the common envelope/trailing `extra` shape and rejects duplicate/colliding columns.
- **F-273549:** rows now require a nonempty string `event`, and known numeric, boolean, and text fields reject values of the wrong type. This closes the unsafe paths identified in V12; exhaustive event-specific required-field schemas remain outside this commit.

I intentionally left **F-273501** (cross-clock ordering), **F-273530** (nested discovery fallback), **F-273534** (current-run freshness), and **F-273546** (shell flush guard parsing) for separate changes because they alter oracle/harness behavior beyond input validation. The local-filesystem and resource-limit findings also remain unchanged.

Validation on this commit:

- Replayed the corrected oracle against the failed CI run's uploaded trace artifact; it passes with required evidence for node2/node4.
- `python3 docker/zakura-regtest-e2e/trace_oracle.py --self-test`
- `cargo test -p zakura-network --lib trace --locked` (37 passed)
- `cargo test -p zakura-jsonl-trace --locked` (18 passed)
- `bash -n docker/zakura-regtest-e2e/run.sh`

PR #972, opened 2026-09-12
fix(state): wake checkpoint handoff after durable completion

czarcas7ic:
## Motivation

Sync can stop at the final checkpoint when full-verification blocks are already queued. The checkpoint reaches disk, but an idle state buffer needs another request before it checks whether those blocks can start. A queued semantic child can also end checkpoint sync before its configured boundary, disabling further checkpoint writes.

## Solution

Notify state after the effective final checkpoint commits and require the configured boundary before switching writers. Retain the notification through caller cancellation or timeout, and report notification failures separately from durable commit success. Reject Regtest settings whose checkpoints do not cover mandatory verification during configuration parsing. The new public request and response variants require `zakura-state` 9.0.0 and updated consumer requirements.

## Testing

Focused checkpoint, router, state, network configuration, and chain tests pass. Coverage includes delayed handoff notifications, premature semantic children through mining and ordinary delivery, and valid and invalid Regtest checkpoint settings. Formatting, workspace Clippy, Markdown lint, and changelog validation pass. The mainnet sync trial has not been repeated for this revision.

Used Codex for implementation and validation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8147,7 +8147,7 @@ dependencies = [
 
 [[package]]
 name = "zakura-state"
-version = "8.0.0"
+version = "8.1.0"
 dependencies = [
  "bincode",
  "chrono",
--- a/crates/zakura-consensus/Cargo.toml
+++ b/crates/zakura-consensus/Cargo.toml
@@ -71,7 +71,7 @@ tower-fallback = { package = "zakura-tower-fallback", path = "../tower-fallback/
 tower-batch-control = { package = "zakura-tower-batch-control", path = "../tower-batch-control/", version = "1.3.0" }
 
 zakura-script = { path = "../zakura-script", version = "3.2.3" }
-zakura-state = { path = "../zakura-state", version = "8.0.0" }
+zakura-state = { path = "../zakura-state", version = "8.1.0" }
 zakura-node-services = { path = "../zakura-node-services", version = "3.2.4" }
 zakura-chain = { path = "../zakura-chain", version = "7.0.0" }
 zakura-header-chain = { path = "../zakura-header-chain", version = "2.1.0" }
@@ -96,7 +96,7 @@ toml = { workspace = true }
 
 tokio = { workspace = true, features = ["full", "tracing", "test-util"] }
 
-zakura-state = { path = "../zakura-state", version = "8.0.0", features = ["proptest-impl"] }
+zakura-state = { path = "../zakura-state", version = "8.1.0", features = ["proptest-impl"] }
 zakura-chain = { path = "../zakura-chain", version = "7.0.0", features = ["proptest-impl"] }
 zakura-test = { path = "../zakura-test/", version = "2.1.0" }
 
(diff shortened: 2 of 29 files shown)


p0mvn (approved):

evan-forbes (commented):
was digging with AI for alternatives and passing along

could be incorrect will take another look when I'm back at my computer 

```
I think the wake-up here is treating a symptom, and there may be a smaller fix underneath it.

Root cause as I read it: the writer transition is control-plane state, but it's only evaluated when something drives the data plane. `poll_ready` isn't called while the buffer queue is empty, so the transition can't happen on its own. This PR's answer is to manufacture a request. That works, but it has some costs:

- `Request::Tip` is no longer a read. It now performs an irreversible writer switch as a side effect for every caller holding the writable service, not just the checkpoint verifier.
- A durable commit can return `Err(VerifyCheckpointError::Tip(_))` when the wake-up times out. `failed_handoff_check_preserves_committed_checkpoint_progress` asserts exactly this — the block is on disk but the caller sees a failure, which everywhere else in this path means "not committed."
- The spawned task's `assert!(matches!(response, zs::Response::Tip(_)))` panics through `.expect("retained state tip task must preserve state invariants")` into the verifier. `S` is documented as pluggable, so that's a panic on a response we don't fully control.
- The notification fires at every `checkpoint_list.contains(height)`, not just at the boundary. On the mainnet list that's thousands of extra writable round trips, each taking a buffer slot ahead of real commits, and each one holding its permit past the 30s timeout because the task is detached. The 2,577-block run is too short to surface that.

Alternative: make the handoff a function of durable state rather than of who sends the next request. Two shapes, depending on what's reachable:

1. Call the handoff check at the end of the finalized write path, once the commit is durable. It already knows when the write landed — no second request, no detached task, no timeout, no new failure mode on the block's result.
2. Drive it off the finalized-tip watch channel, so writer selection recomputes whenever the durable tip moves. This fixes the whole "idle buffer is never re-polled" class rather than this one instance.

Separately, and independent of which route we take: F-275362 is still open. `try_handoff_to_non_finalized_write` switching writers irreversibly on a queued semantic child, with the finalized sender dropped and not restored when that child is rejected, seems worth resolving before this lands — this PR adds trigger paths into that helper rather than removing them. Either gate strictly on `finalized_tip_height >= max_checkpoint_height` and drop the queued-child fallback (the doc comment already calls it a fallback for configs with no finite checkpoint height), or hold the finalized sender until the non-finalized writer has actually accepted a block.

If the refactor turns out bigger than it looks from here, I'd rather land a reduced version of this PR — dedicated request variant instead of overloading `Tip`, non-fatal on wake-up failure, boundary-only — than the current shape.
```

czarcas7ic:
@evan-forbes Modified to take the smaller fallback you suggested

evan-forbes (approved):
Pushed 7eb102aa6 to fix the four Regtest fixtures from #1043 that the new mandatory-checkpoint check rejects.

PR #973, opened 2026-09-13
test: msg-reg: add shared regulation test instruments

czarcas7ic:
## Motivation

Message regulation tests need to measure real work separately from the node's own capacity counters. Otherwise, an incorrect counter could make a test pass by agreeing with itself.

## Solution

Add shared helpers for memory allocations, running background jobs, CPU use and lock delays. For example, a function that holds an 8 KiB temporary buffer and a 1 KiB result, then frees the temporary buffer, reports a 9 KiB peak and 1 KiB remaining at return. The helpers live in `zakura-test` so each message can reuse them with its own inputs and limits. The [testing guide](https://github.com/zakura-core/zakura/blob/adam/regulation-test-observations/docs/testing/regulation-observations.md) explains their use and measurement boundaries. Allocation measurements cover the calling thread. Process CPU and peak memory totals include other work in the same program. Prepared with Codex assistance.

## Testing

The helper tests cover:

- Allocation measurements distinguish temporary peaks from retained output and recover after a measured operation panics.
- Execution observations count actual job starts, keep completed work counted while its result is held, and release the count on completion or panic.
- Release guards unblock jobs paused at entry or before returning their results.
- Lock observations cover the guard's lifetime and leave no unmeasured interval between acquisition wait and hold time. Process CPU and peak memory totals do not decrease between samples.
- Load-round controls use the default only when the override is absent, clamp valid integers, and reject malformed integers or non-Unicode values.


czarcas7ic:
@codex review

czarcas7ic:
@codex review

czarcas7ic:
@codex review

evan-forbes:
From the review of the message-regulation stack (#973–#981). This PR carries the stack-wide overview plus the section on the test instruments.

# Property-testing review

Simplify PRs #973–#981 with direct resource limits and shared message rules, while closing the remaining DoS and testing gaps without unnecessarily limiting throughput.

We reviewed:

- Missing tests and tests that duplicate production behavior in a separate model.
- Simpler ways to bound allocations, queued bytes, execution, and retained requests.
- Spec compliance and compatibility with valid deployed encodings.
- DoS prevention while preserving useful throughput and other peers' progress.
- Reuse across services, shared-lock costs, and ownership through cancellation and reconnects.
- Dependency justification, CI coverage, and reproducible load qualification.

Suggested changes:

- Cap decoder allocations in bytes (#974) and remove network-policy plumbing.
- Share message declarations (#975), bound queued bytes (#975), and separate execution from output ownership (#976).
- Retain response identity with a local cap (#977), replacing the metadata planner (#979) and potentially the second session fence (#978).
- Apply the right traffic policy (#981) and limit dependency changes to demonstrated gaps (#981).
- Fix discovery reservations (#975), supplier attribution (#999), and remaining transport defects (#975).
- Remove duplicate test machinery and fix CI coverage (#973).
- Consider one stream per request (#977) to simplify response matching.


## PR #973 and test infrastructure: retain observations, remove duplicate models

The [allocation tracker](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-test/src/allocations.rs#L143) and [blocking-operation probe](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-test/src/execution.rs#L45) measure behavior that node counters cannot establish independently. The allocator-installation check prevents meaningless zero measurements. The blocking probe exposes work that survives waiter cancellation.

The execution probe's duplicate allocation aggregation can go. A counter suffices for lock-acquisition counts or proof that decoding never started. CPU, RSS, and lock timing belong in reproducible load experiments where they support a bound.

A small allocation helper can cover each production decoder family: peak and retained bytes against a payload-dependent bound plus fixed overhead. It complements the stricter collection checks. Header sync, discovery, and gossip need the same ingress coverage as block sync.

The [unit-test workflow filters](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/.github/workflows/tests-unit.yml#L9) exclude stack branches. Nextest profiles do not schedule jobs. Focused regressions belong on PRs; broader generated and load cases need scheduled runs. Existing Proptest shrinking and deterministic regressions can replace custom replay systems. One requirement-to-test map with runnable commands can replace the overlapping testing documents.

PR #974, opened 2026-09-13
fix(chain): msg-reg: bound nested decoding by available payload bytes

czarcas7ic:
## Motivation

A complete Block payload can declare 1,024 transactions without supplying any transaction bytes. The streaming decoder can reserve a collection before discovering that the data is missing.

## Solution

Configure a shared decoder once from the node's network. Nested reads inherit its rules and remaining byte count, so impossible collection counts fail before allocation without passing a network to every decoding call. Regtest uses 178-byte counted headers, while Mainnet and Testnet require 1,488 bytes. Their decoders use that larger minimum and reject Regtest solution lengths before allocating solution bytes. Native GetBlocks responses, buffered block replay, and legacy block and headers messages use this path. Generic streaming methods remain available. The [testing guide](https://github.com/zakura-core/zakura/blob/adam/bounded-payload-decoding/docs/testing/bounded-decoding.md) explains the contract. Prepared with Codex assistance.

## Testing

F03 allocation property tests cover:

- Fixed GetBlocks request and terminal fields decode without heap allocation.
- Missing transaction bytes cannot trigger reservation of the declared transaction collection.
- Retained decode allocations fit the object memory bound, including pointer overhead.
- Generated complete transactions preserve the streaming decoder's results.
- Nested collection counts that cannot fit in the supplied bytes fail before allocation.
- Truncated scripts, proofs and strings fail before allocating their buffers.
- Complete proof arrays stay within the declared allocation size at buffer growth boundaries.
- Header collections use the configured network's minimum size and reject missing bytes before allocation.


czarcas7ic:
@codex review

evan-forbes:

## PR #974: bound allocations in bytes, without carrying network policy through the decoder

[zcash_deserialize_external_count](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-chain/src/serialization/zcash_deserialize.rs#L119) [reserves up to 1,024 elements](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-chain/src/serialization/zcash_deserialize.rs#L160) before decoding them. For large element types, a tiny payload can trigger a large allocation. A byte cap on the initial reservation addresses that directly:

```rust
let initial_count = count.min(INITIAL_BYTES / size_of::<T>().max(1));
let mut items = Vec::with_capacity(initial_count);
```

Starting with `Vec::new()` is another option. Either approach belongs in the shared decoder so native `PushTransaction` and legacy transaction ingress receive the protection too. The slice-only path misses those callers. Subsequent growth still needs to stay within the validated count, and nested allocations need their own bounds. This changes allocation policy, not accepted protocol counts.

The spec separately requires checking counts against `remaining_bytes / minimum_item_size`. The [remaining-length reader](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-chain/src/serialization/reader.rs#L23) supplies that information. `Read::take(maximum)` supplies only a read ceiling; undocumented `Read::bytes().size_hint()` behavior is also a poor basis for this contract.

The reader can use conservative encoded minima that admit every supported encoding. That removes the need for [ZcashDecoder](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-chain/src/serialization/decoder.rs#L28), `min_serialized_size_for`, and network-specific decoder configuration through codecs, services, routines, and replay. Existing validation handles wrong-network data. Replay still needs the raw payload. Exact payload consumption, nested length checks, and explicit handling of zero minima and unsupported bounded decoding remain necessary.

Using minima only as allocation hints does not satisfy the spec's count check. The simpler design is a byte-capped reservation plus the remaining-length check, with no network policy in either.

The hand-written gossip decoders in [legacy_gossip.rs](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/legacy_gossip.rs#L819) have the same allocation problem. `read_hash_list`, `read_tx_id_list`, `read_block_locator`, and `read_header_list` reserve from peer-selected counts. A three-byte declaration can reserve space for 25,000 inventory entries. Their cursors already expose remaining bytes, so a small counted-list helper or capped lazy growth covers them.

The decoder tests also need an independent acceptance reference: comparing two new entry points can miss a shared regression. Existing encoding vectors or the base decoder provide one.

PR #975, opened 2026-09-13
fix(network): msg-reg: reject invalid frame declarations before payload reads

czarcas7ic:
## Motivation

A Block frame could declare 2,000,002 payload bytes and leave us waiting for them, even though a legal Block payload cannot exceed 2,000,001 bytes. Unsupported flags and conflicting frame and payload tags could also reach later processing than necessary. Outbound terminal messages accepted heights that inbound decoding rejected.

## Solution

Apply each service's declared flags and message limits in the shared frame reader before allocating or waiting for a payload. Check repeated message tags before Block decoding and validate terminal heights in both directions. The configured network decoder stays in place. Other messages use the same service declarations and reader. The [testing guide](https://github.com/zakura-core/zakura/blob/adam/native-frame-enforcement/docs/testing/frame-enforcement.md) explains the boundary. Prepared with Codex assistance.

## Testing

- F01 properties cover message caps, tighter negotiated caps, unsupported flags and custom flag masks while payloads are withheld. Legacy gossip and request/response controls cover every flag bit and valid zero-flag frames.
- F02 properties cover generated request and terminal fields, canonical round trips, truncation, tags, flags and the exact Block size limit.
- F04's frame property checks that conflicting message tags fail before Block allocation.
- T03 properties cover every request byte split, truncated or reset frames and finite read deadlines.


czarcas7ic:
@codex review

evan-forbes:
From the review of the message-regulation stack (#973–#981). Four sections land here: the shared message declarations (which also cover #981), the read-ahead byte bound, the discovery reservations, and the remaining transport gaps. They are filed on the lowest PR in the stack that they touch.

## PRs #975 and #981: one message declaration for type, size, and regulation

The [Block payload-cap declaration](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/block_sync/service.rs#L503) in #975 is the right fix: reject an oversized declaration from its header, before allocating or waiting for the payload. The same declaration can hold the allowlist and regulation policy:

```rust
enum Regulation {
    Cadence { burst: u32, period: Duration },
    Reserved,
    CapacityBound,
}

struct MessageRule {
    message_type: u16,
    max_payload: usize,
    regulation: Regulation,
}
```

An absent row rejects the type instead of falling back to the larger stream cap. The allowed flag mask remains a stream property. `Reserved` still requires the service to match and consume authorization; the table only selects the policy. Custom services need to retain their existing contract, and direct codec callers still need size checks outside transport.

Early flag checks and frame/payload tag agreement belong alongside these limits. [write_outbound_request_frame_inner](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/handler.rs#L4803) also needs the outbound flag check: it receives the mask but can write flags the receiver would reject.

The missing declarations are:

- **Header sync:** add caps to normal and passthrough services using the deployed `GetHeaders` encoding, which differs from the successor spec in places.
- **Discovery:** move the existing outer cap before allocation. V1 uses frame type 1 for every message, so per-message discrimination happens in the payload.
- **Gossip:** add per-stream allowlists and tight caps for fixed-size messages, retaining valid body/list maxima.

Discovery has a separate compatibility mismatch: the codec allows 16,384 bytes, while the spec allows a 20,738-byte `Peers` response containing 32 maximum-size records. Moving the current cap earlier fixes allocation timing. Supporting the larger response requires either a rollout to larger decoders or a spec change to the record/count limit.

Share header-boundary tests across production message declarations. Keep independent encoder samples and expected limits so a wrong declaration cannot validate itself. The existing header-only QUIC test covers rejection before payload reads.


## Read-ahead: bound retained bytes instead of counting frames

The [Block queue](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/block_sync/service.rs#L491) holds up to 128 frames and can retain about 256 MB of payload, plus a frame in the reader and another in the consumer. A byte permit acquired before [payload allocation](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/handler.rs#L4646) gives a much tighter bound for variable-size messages:

```text
validate header type, flags, and length
acquire payload-byte permit
allocate and read payload
queue payload together with permit
release permit when payload drops, or transfer it to another bounded owner
```

Dequeueing alone cannot release the permit because the consumer still holds the allocation. Decoding may temporarily retain both the wire payload and decoded objects. Local waits for capacity also belong outside peer read deadlines; otherwise our backpressure becomes their timeout.

The budget must fit the largest legal frame and leave room for independent streams. An 8 MiB read-ahead budget is a candidate, not a measured default. Its node cost depends on whether each stream, connection, or the node owns the allowance.

The existing real-QUIC paused-reader and sibling-progress tests can become shared transport helpers. The missing observation is total retained memory across reader, queue, and consumer, including cancellation and resumption at the byte limit.


## Discovery: use the existing session state for response reservations

[DiscoverySink](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/discovery/service.rs#L705) accepts unsolicited `Peers` and `Services`, sends refreshes without waiting for previous responses, and [silently drops replies on a full queue](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/discovery/service.rs#L817).

One reservation per exchange in the existing session-shared progress state is enough. `Peers` retains the requested record limit. Reservation publication precedes sending; rollback applies when sending definitely never started. A response consumes its reservation before importing records or marking progress. Duplicate, unsolicited, and over-limit responses then fail without touching the discovery book. `Services` still needs its author check.

An admitted response must remain queued or end with the applicable exchange/session closure, rather than disappear behind `Ok(())`. Local queue failure should not become an inappropriate peer penalty. Reservations also need the correct timeout/reopen lifetime so old replies cannot satisfy new requests. Signature verification stays outside the book lock.

Replace test-only discovery adapters with coverage of the production reservation and queue-full paths. The 16 KiB versus 20,738-byte compatibility question also applies here.


## Other gaps from the audits. (no action needed, just future reference) 

- Apply source-aware admission before one source fills every shared pending-handshake slot.
- Separate queued-request reservation from response eligibility using writer arbitration. Predicted responses can currently consume queued credit; legitimate fast responses still need to work.
- Separate receipt from validated progress so invalid deliveries cannot keep resetting liveness. Retain supplier retry/scoring evidence.
- Account for granted QUIC stream capacity even when the negotiated application ceiling is lower.
- Charge stream-open attempts before capacity rejection; rejected opens currently bypass the churn limit.
- Negotiate a compatible capability subset or reject configurations whose stream limit cannot support the layout.
- Include the serial local request stream in the stream count, or budget its extra slot explicitly.
- Classify FIN inside a frame header as truncation, rather than a clean boundary close.
- Preserve an observed fatal read cause when cancellation races with teardown.
- Set `drained` after the drain barrier succeeds, so a fatal interruption cannot produce graceful FIN.

PR #976, opened 2026-09-13
test: msg-reg: cover serving ownership through cancellation and reconnects

czarcas7ic:
## Motivation

Cancelling a serving request must not free its worker slot while a storage read, encoded result or unfinished write still owns it. For example, a peer that reconnects during a read must wait for that old operation to release capacity before starting another.

## Solution

Add independent ownership models and controlled tests of the existing serving path. GetBlocks and a test GetPeers policy share the same request and waiter checkers. GetBlocks adds real encoding, queued writes and reproducible reconnect histories. Tests are grouped by ownership, response ranges and local failures, with a [guide to their contracts](https://github.com/zakura-core/zakura/blob/adam/serving-ownership-properties/docs/testing/serving-ownership.md). Encoder observation is test-only. Prepared with Codex assistance.

## Testing

- Shared properties cover admission limits, waiting order, unobserved grants, cancellation, cloned leases, response byte limits and final release with both message policies.
- Ownership histories cover reconnects, queued and unfinished writes, concrete JSON replay and a broken-writer control.
- C01–C04 properties cover bounded operation starts, output pressure and capacity retained by running or completed operations after cancellation.
- C05–C06 serving properties cover allocation bounds and storage, encoding and output failure cleanup.
- C07 properties cover counts 1–128, response prefixes, byte boundaries, gaps and changing serving limits. R02's serving property keeps later requests waiting while an ending is owned.
- Generated request bursts and state range properties cover exact block hashes and endings, cancellation, contiguous reads, overflow and retained read results.


czarcas7ic:
@codex review

evan-forbes:

## PR #976: execution and output need different lifetimes

[ResponsePermit::frame_guard](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/regulation/request.rs#L203) clones the same `WorkResources` that holds the node execution permit:

```rust
FrameGuard::new(self.resources.clone())
``` That lets a slow reader potentially occupy a serving slot across a long response, even when storage and encoding have finished. The missing regression is straightforward: fill serving slots with barely-draining peers and ask another peer to complete a request.

Separate execution permits from result/output memory if that case confirms starvation:

```mermaid
flowchart LR
    A[Reserve result memory] --> B[Storage and encoding hold execution slot]
    B --> C[Work finishes: release execution slot]
    C --> D[Queued output retains memory permit]
    D --> E[Output drops: release memory permit]
```

Cancellation of an async waiter does not stop its blocking operation. The operation owns the execution permit until it finishes. Encoding needs that protection as well as storage. Memory reservation covers decoded objects, encoded buffers, and their overlap; a wire-size cap alone misses decoded heap usage.

Per-peer ownership must survive reconnects. A shared output pool also needs isolation, or slow readers simply move starvation from execution slots to memory. The same independent-peer test should exercise the output pool at capacity.

The production serving tests already cover valuable blocked-storage, encoding, cancellation, reconnect, and backpressure behavior. The separate `serving_regulation/properties/` model adds roughly 1,100 lines of adapters, JSON replay versions, and harness machinery. Short Proptest sequences against production transitions can retain its distinct interleavings with existing shrinking and failure storage. Small helpers for blocking and ownership are reusable; discovery adapters with no production integration add no discovery coverage.

PR #978, opened 2026-09-13
fix(network): msg-reg: fence request writers before replacing receivers

czarcas7ic:
## Motivation

Replacing a receiver could leave its queued writer able to start a request afterward. The peer would then have permission to respond without the original receiver to validate that response.

## Solution

Share one fence between publishing a request, starting its write and retiring its receiver. For example, replacing a receiver skips its old queued GetBlocks request. If writing started first, replacement closes the connection unless the response has ended. A new connection can proceed independently. Shared scope and writer permissions keep this rule reusable across messages. The [contract guide](https://github.com/zakura-core/zakura/blob/adam/response-session-fencing/docs/testing/response-session-fencing.md) explains ownership and the test adapters. Prepared with Codex assistance.

## Testing

- Generated lifecycle histories cover publication, first writes, endings, owner drop, cancellation and replacement while old writers survive.
- Scope and real GetBlocks queue properties cover both orders of retirement versus writing, and same-connection versus new-connection replacement.
- A discovery response adapter covers exact counts, bytes, excluded identities, duplicate responses and local interest changes using the real codec.
- A future subscription adapter covers bounded credit renewal, preserved consumption, updates crossing Close, and ending capacity independent of data credit or worker slots.


czarcas7ic:
@codex review

evan-forbes:

## PR #978: existing session ownership may make the second fence unnecessary

Three existing mechanisms may cover the second fence's job:

- [RequestWrite arbitration](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/block_sync/work_queue/request_write.rs#L195) checks the work owner under the work lock before changing `QUEUED` to `STARTED`. [Expiry](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/block_sync/work_queue/request_write.rs#L42) competes for the same state with `QUEUED` to `EXPIRED`.
- [PeerRoutine::drop](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/block_sync/peer_routine.rs#L2248) closes the connection before returning work if started responses remain:

```rust
if self.has_started_responses() {
    self.session
        .close_connection("unfinished_response_state_retired");
}
self.return_unreceived_requests("peer_routine_drop");
```

- [Transport session cleanup](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/handler/service_session.rs#L249) waits for every worker before reporting the session's exit:

```rust
while running.next().await.is_some() {
    cancel.cancel();
}
let _ = exits.send(exit);
```

The proposed removal targets [ResponseScope, ResponseAuthorization, and ResponseWritePermission](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/regulation/response_scope.rs#L30). Their additional scope lock serializes retirement with publication and writer startup. The current `RequestWrite::try_start` calls `response_write.try_start`, so the combined implementation does not prove that write arbitration alone suffices.

The deciding cases are duplicate-connection replacement and same-connection reopen, with queued and started writes, including legacy handoff. Old responses must not gain new-session authorization. The [unfinished-pair retirement test](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/handler/tests/paired_block_sync.rs#L948) exercises production connection closure. A fixture that creates two live sessions manually does not establish a reachable race. Moving cancellation earlier also does not make cancellation and the first write atomic.

If the existing mechanisms cover those transitions without the scope lock, remove the second fence. The [credit-renewal API](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/regulation/response.rs#L40) and test-only [discovery](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/regulation/response_properties/discovery.rs#L29) and [subscription](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/regulation/response_properties/subscription.rs#L11) adapters are additional removal candidates. A reachable race would justify a fix at the affected transition.

PR #979, opened 2026-09-13
fix(network): msg-reg: share response metadata budgets across services

czarcas7ic:
## Motivation

Response bookkeeping needs limits shared across messages and retained owners. A new service or replacement receiver must not obtain a fresh connection allowance while an old writer still owns memory.

## Solution

Share a 128 MiB node pool and a 16 MiB connection pool across services and replacements. Reserve setup and authorization memory before allocation, keep its charge until the last owner releases it, and wake waiting requests when capacity returns. Connections at their own limit ignore completions elsewhere. For example, if a replacement cannot fund its setup, the current receiver stays active. This layer covers shared bookkeeping. Request-specific buffers and retained collection growth remain a separate allocation-planning layer. The [contract guide](https://github.com/zakura-core/zakura/blob/adam/response-metadata-budgets/docs/testing/response-memory.md) explains the costs and boundary. Prepared with Codex assistance.

## Testing

- Generated allocation histories check both limits against independently tracked live owners across messages and connections.
- Allocation observations cover cold setup, exact authorization charges, allocation-free denial, transitions and unfinished cleanup.
- Lifetime and concurrency properties cover retained writers after endings or closure, racing reservations, counter boundaries and capacity wakeups.
- Wakeup properties cover 32 unrelated releases without scheduling the blocked receiver, recovery on its own release, switches between full limits, partial and early releases, and waiting without extra allocations.
- Service properties cover shared pools across stream activation, failed replacement setup preserving the current receiver, pending work under exhaustion and idle-loop progress.


czarcas7ic:
@codex review

czarcas7ic:
@codex review

evan-forbes:
From the review of the message-regulation stack (#973–#981). This section covers #979 and #980; it is filed on the lower of the two.

## PRs #979–#980: a retained-range bound can replace the metadata pool and planner

[ResponseMemoryPool::try_connection](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/regulation/response_memory.rs#L99) couples request retention to connection setup. It allows 16 MiB per connection from 128 MiB node-wide, so eight connections can consume the pool and deny unrelated new connections. A saturation regression would establish the runtime effect of that policy.

#979 initially covers only part of the metadata. #980 adds funded vectors, tree-layout estimates, growth plans, and permit transfers. Allocation under the shared registry lock increases contention. Allocator failure can also retry while logical counters still report room, creating a possible busy loop.

A local range cap gives a simpler bound:

```text
C = maximum retained owners, including replaced and draining owners
F = fixed bytes per owner
W(B) = worst retained bytes per range of B blocks
M = node metadata budget

R = floor((M - C * F) / (C * W(B)))
local_limit = min(R, protocol_limit, peer_advertised_limit)
```

This requires positive denominators, checked arithmetic, and enough budget for fixed storage. Clamping `R` upward to a minimum can violate the budget; incompatible configurations need a different budget or rejection.

`W(B)` includes heap allocations, spare capacity, expected-block records, queued writers, indexes, snapshots, and temporary growth copies. `size_of` covers only inline storage. Independently bounded storage can remain outside this budget. Earlier estimates of `740 + 48 * B` bytes per range, giving caps around 665 or 76, are unmeasured and insufficient as constants.

Ordinary collections with a logical maximum can replace [ResponseVec](https://github.com/zakura-core/zakura/blob/e729b89286e46f4ad059c05ddac6bef2e561fb5c/crates/zakura-network/src/zakura/regulation/response_vec.rs#L62), `CapacityPlan`, `ResponseIndexPlan`, `OutstandingCapacityPlan`, and `authorize_request_metadata`. Modest preallocation avoids growth; capped lazy growth avoids reserving every peer's maximum up front. `Vec::with_capacity` alone does not enforce a limit. Keeping an owner's allowance until its allocations disappear avoids per-entry refunds and shrink accounting.

The connection-count multiplier must include old owners that survive replacement. A narrow owner permit can cover that lifetime if the current connection permit ends too early. Moving request items instead of cloning them remains useful, as does moving allocations outside the global registry lock. Allocator failure needs a different path from logical capacity exhaustion.

The decision rests on peak allocation at full capacity and throughput at realistic RTTs. Both one-block and maximum-size ranges matter. Withheld endings, cancellation, repeated replacement, and another peer's connection/download expose bounds that steady-state tests miss. A cap of 16 or 64 has no demonstrated throughput justification.

PR #795, opened 2026-08-24
feat: add NuTachyon and V7 transactions

S1nus:
The bulk of Tachyon’s logic is `tachyon::coherence` (semantic verification), `TachyonMiningData` in the state service, and the `get_block_template` handler in RPC, `update_chain_tip_with_block_parallel` in crates/zakura-state/src/service/non_finalized_state/chain.rs; the rest of the PR is mostly plumbing.

## zakura-chain

- Adds a cfg-gated `zakura_chain::tachyon` module containing wrappers for Tachyon’s `Anchor` and `Tachygram` types.
- Updates V4 history tree computation to hash the Tachyon anchor into the tree.
- Adds the NuTachyon `NetworkUpgrade`, `ConsensusBranchId`, and `TX_V7_VERSION_GROUP_ID`.
- Adds the V7 transaction type, which carries Tachyon bundles.
- Updates value balance calculation to include Tachyon bundles.
- Adds `tachyon_anchor` and `tachyon_epoch_anchor` to `NoteCommitmentTrees`.

## zakura-consensus

- Modifies the mempool transaction verifier to enforce valid digests for Tachyon actions and accept only autonome transactions for now. In the future, we will perform peer-to-peer aggregation and allow other kinds of Tachyon transactions to propagate through the mempool.
- Adds parallelized validation of Tachyon signatures.
- Adds proof stamp verification.
- Adds error types for Tachyon’s invariants.
- Adds Tachyon coherence checks to semantic block verification.

## zakura-network

- Updates `CURRENT_NETWORK_PROTOCOL_VERSION_VALUE`.
- Adds NuTachyon and V7 transaction arms to exhaustive match statements in tests.

## zakura-state

- Adds `tachyon_anchor` and `tachyon_epoch_anchor` to `TreeState`.
- Updates `ChainInner` with Tachyon anchors, epoch anchors, tachygrams, and their height/epoch indexes.
- Executes Tachyon transactions while updating the chain tip in parallel.
- Adds `TachyonMiningData` to the read state service so miners can retrieve anchor data and revealed tachygrams for aggregation during block construction.
- Adds contextual transaction verification for nullifiers and anchors.
- Updates the database version and adds four finalized-state column families for Tachyon anchors, anchor indexes, epoch anchors, and tachygrams.
- Filters Tachyon-enabled blocks from VCT roots in `commit_finalized_direct_with_aux` (we may want to revisit how Tachyon affects VCT).
- Verifies Tachyon commitment roots.
- Updates history tree computation.
[...]

PR #983, opened 2026-09-14
feat(rpc): expose Tachyon pool state for explorers

S1nus:
## Motivation

Tachyscan needs the authoritative selected-chain Tachyon accumulator and retained Tachygram state. The existing verbose transaction RPC exposes Tachyon bundles, but reconstructing current pool state otherwise requires replaying every block since NuTachyon activation.

Depends on #795.

## Solution

Add a NuTachyon-gated `gettachyoninfo` RPC that reports the tip anchor, current epoch and next boundary, retained Tachygram count and payload size, and a bounded sample of the 42 most recently revealed Tachygrams.

The state query merges finalized and non-finalized best-chain data and applies the protocol's two-epoch scan window before producing the summary. A separate cfg-gated RPC extension keeps the method absent from non-NuTachyon builds, while the bounded sample keeps polling responses predictable.

## Testing

- `ROCKSDB_LIB_DIR="/opt/homebrew/lib" cargo +1.97.1 check -p zakura-state -p zakura-rpc --all-targets`
- `ROCKSDB_LIB_DIR="/opt/homebrew/lib" RUSTFLAGS='' cargo +1.97.1 check -p zakura-state -p zakura-rpc --all-targets`
- `ROCKSDB_LIB_DIR="/opt/homebrew/lib" cargo +1.97.1 clippy -p zakura-state -p zakura-rpc --all-targets -- -D warnings`
- `ROCKSDB_LIB_DIR="/opt/homebrew/lib" cargo +1.97.1 test -p zakura-rpc tachyon_info`
- RPC surface policy test with NuTachyon enabled and disabled
- `ROCKSDB_LIB_DIR="/opt/homebrew/lib" cargo +1.97.1 xtask check-rpc-artifacts`
- Markdown lint on `docs/changelog/unreleased/983.md`

The full listener surface test could not bind a local port in the sandbox (`Operation not permitted`); its non-network policy assertions pass in both cfg modes.

## Changelog

Added `docs/changelog/unreleased/983.md`.

- Tachyscan explorer mockup and implementation
- #795

### Companion Work

The local Tachyscan implementation now prefers this authoritative endpoint and retains its block index for historical footprint charting and compatibility with #795-only nodes.


S1nus:
Moved to S1nus/zakura#2, targeting the cnode-working branch.

PR #998, opened 2026-09-15
feat: prepare NU7 deployment parameters

evan-forbes:
## Motivation

The 2026-09-14 coinholder poll ratified NU7 and asked for it to ship as soon as possible. NU7 still has no activation heights and no production consensus branch ID. This PR makes the NU7 changes that depend on neither, so the deployment PR only has to add those values.

### Final NU7 peer protocol versions (ZIP 204)

ZIP 204 assigns protocol versions by rule from NU7 onward. The Testnet version is the least `170000 + 20·k` above the preceding upgrade's Mainnet version, and the Mainnet version is 10 higher. NU6.3 uses 170160 on both networks, so NU7 gets 170180 on Testnet and 170190 on Mainnet. Regtest uses the Testnet version.

Zakura had provisional values of 170170 (Testnet and Regtest) and 170180 (Mainnet). This PR sets `Version::min_specified_for_upgrade` to the ZIP 204 values and updates the NU7 comment in `constants.rs`. The deployment ZIP draft lists 170150 and 170160. Those values are stale: NU6.2 and NU6.3 already use them.

This PR leaves `CURRENT_NETWORK_PROTOCOL_VERSION` at 170160 and `INITIAL_MIN_NETWORK_PROTOCOL_VERSION` at NU6.2. Both change with the activation heights.

### History tree errors without a branch ID

Every ZIP 221 history tree node commits to its network upgrade's consensus branch ID. NU7 has a branch ID (`0xfffffffe`) only in test builds and builds with the `zakura-test` feature. A production build can reach NU7 on a configured Testnet or Regtest.

`NonEmptyHistoryTree::from_parts` and `NonEmptyHistoryTree::from_cache` return `HistoryTreeError::MissingBranchId` when the upgrade has no branch ID. The fallible primitive `Tree` constructors also reject a missing branch ID before they decode cache entries or build a leaf. The infallible `Version` conversion methods retain their branch-ID precondition, which their documentation now states.

VCT checkpoint verification propagates `MissingBranchId` without attaching a `VctCommitFailure`. The authentication sweep also excludes this local error from peer-delivery attribution. These paths preserve the local cause and avoid recording false delivery-rejection evidence.

Database startup checks stored history snapshots before format checks or state reads. If this build lacks the snapshot's branch ID, startup returns a `StateInitError::DatabaseFormatUpgrade` with the snapshot error as its source. The format quick-check also propagates snapshot decoding errors instead of calling an infallible accessor.

## What the deployment PR still needs

- The NU7 consensus branch ID in every build. This needs a `zcash_protocol` release that defines NU7 without `zcash_unstable = "nu7"`, using the deployment ZIP's `0x77190AD8`. Zakura then replaces the test-only `(Nu7, 0xfffffffe)` entry.
- NU7 activation heights for Mainnet and Testnet.
- `CURRENT_NETWORK_PROTOCOL_VERSION` set to the NU7 value, and `INITIAL_MIN_NETWORK_PROTOCOL_VERSION` moved from NU6.2 to the upgrade the heights make current.
- Removal of the `nu7` build gate from every NU7 rule.

## Testing

- Rust 1.97 library tests: chain 404 and network 1258 passed. The state run passed 595 tests; one new assertion needed correction to compare the typed source error. Both VCT missing-branch-ID regression tests passed on the final revision. Four state tests remain ignored.
- Constructor tests verify that fallible cache, block, and block-parts constructors reject an upgrade without a branch ID before decoding or leaf construction.
- The format-check regression verifies that snapshot decoding errors return through the fallible path.
- VCT regressions verify that local branch-ID failures preserve their cause and carry no delivery-failure attribution. Malformed supplied trees retain attribution.
- `cargo +1.97.0 clippy --locked -p zakura-chain -p zakura-state -p zakura-network --all-targets -- -D warnings` passed.
- Formatting and changelog checks passed.

The branch also includes the mock UTXO request/response ordering fix from #851/#859, which prevents a transaction-test race in CI.

## Changelog

Added `docs/changelog/unreleased/998.md`.


crates/zakura-chain/src/history_tree.rs
```diff
@@ -202,6 +209,11 @@ impl NonEmptyHistoryTree {
             | NetworkUpgrade::Blossom => {
                 panic!("HistoryTree does not exist for pre-Heartwood upgrades")
             }
+            // Every history tree node commits to its network upgrade's branch ID.
+            // Builds without an NU7 branch ID reach this arm at NU7 activation.
+            _ if network_upgrade.branch_id().is_none() => {
+                return Err(HistoryTreeError::MissingBranchId { network_upgrade });
```
p0mvn, line 215: Optional hardening (AI-generated): `MissingBranchId` is currently reachable only when a release build without the NU7 branch ID is paired with a custom Testnet/Regtest NU7 activation. Standard Mainnet/Testnet do not activate NU7 yet, so I don't consider this a current production blocker.

In the VCT paths, this local capability error is wrapped as `CurrentRoots`, which can attribute failure to an otherwise valid auxiliary delivery. Consider special-casing `MissingBranchId` before `vct_reject_supplied_root` and in `attribute_verification_failure`, propagating it without a `VctCommitFailure`; that would prevent custom NU7 testing from recording misleading delivery-rejection evidence.

p0mvn (approved):

czarcas7ic (approved):

PR #999, opened 2026-09-15
fix(consensus): bind body failures to header commitments

czarcas7ic:
## Motivation

A peer can modify a block body without changing its header hash: duplicate padding preserves the transaction Merkle root (CVE-2012-2459), and NU5 transaction IDs exclude authorization data, so swapped signatures or proofs keep the same IDs. The verifier classified those failures as permanent header invalidity, so one supplier could tombstone a valid header and its descendants. The branch-ID check also ran before the Merkle root check, and native sync let the same supplier reclaim the retry immediately.

## Solution

A body failure records permanent header invalidity only when the header fully binds the delivered body: the Merkle root matches without duplicates that padding could have introduced, and from NU5 the ZIP 244 authorization-data commitment matches against the exact parent's history (a new read-only `CheckBlockCommitment` state request, checked on the failure path so verification stays pipelined). Duplicates whose removal changes the root, or that the NU5 commitment proves the header commits to, remain consensus-invalid. Every other non-transient failure is a payload mismatch attributed to the supplier, which native sync now defers for the retry episode. A header whose bodies never bind falls back to the body-unavailable episode (LC-AVAIL-01/02), the same bound as a withheld body; the spec now states this, and partially bound failures are not tombstoned by design. Stored invalidity records from the old classification are not repaired (follow-up below). `zakura-state` moves to 9.0.0 because the new response variants extend exhaustive public enums.

## Testing

On Rust 1.98, `cargo fmt`, workspace Clippy with all targets, the consensus, state, header-chain, and network suites (the network suite except three listener tests that cannot bind 127.0.0.2 on this Mac), `cargo xtask header-conformance`, and `./scripts/changelog.py check`. Regressions cover padding versus intrinsic duplicates, NU5 auth-bound duplicates, altered NU5 authorization bytes with and without a matching commitment, missing parent history, partially bound expiry and sigops failures, the branch-ID ordering, misbehavior scores, and supplier deferral.

## Changelog

`docs/changelog/unreleased/999.md` (Security) and a `COMMITMENT_CHECK_TIMEOUT` row in `docs/changelog/params.md`.

### Follow-up Work

- Clear stored tombstones from the old classification: https://github.com/zakura-core/zakura/issues/1012
- Optional reconstruction and validation of a padded body's unique candidate: https://github.com/zakura-core/zakura/issues/1013
- Score repeated payload mismatches from one supplier: https://github.com/zakura-core/zakura/issues/1014

Prepared with Claude assistance. Codex implemented the original fix and tests; Claude planned and implemented the cut-down to the fully-bound rule.


czarcas7ic:
@codex review

czarcas7ic:
@codex review

evan-forbes:
From the review of the message-regulation stack (#973–#981), which also looked at this PR. Filing it here even though the PR is closed, so the attribution question is not lost.

## Consensus: preserve supplier fault when header validity is unresolved

The [#999 checkpoint error path](https://github.com/zakura-core/zakura/blob/1342f5af55fa380c30ae54bd7193bdec9a433d30/crates/zakura-consensus/src/checkpoint.rs#L1242) can replace duplicate-body attribution with a retryable missing-context result. This happens when a duplicate-padded NU5 body arrives before its parent's history commits.

Supplier fault and header validity need separate outcomes. Missing parent history cannot make the header invalid, but it should not erase the delivery's supplier fault. Deferring the commitment decision until exact-parent history arrives is another option.

The regression input is an odd-transaction-count NU5 block with its final transaction duplicated and its header hash unchanged. With the parent absent from committed state, the header should avoid a false tombstone while the delivery retains supplier attribution. Legacy scoring is a separate gap: its duplicate penalty was already absent on the base.

Existing false tombstones still need a recovery decision. Optional candidate reconstruction is separate from the attribution fix.

PR #1006, opened 2026-09-15
fix(sync): perf: start block requests after peer latency estimates expire

czarcas7ic:
## Motivation

After an idle interval, requester workers can defer the next missing block to each other because their latency estimates expire at different times. Ready work then waits for the periodic heartbeat despite available request capacity.

## Solution

Compare every peer using one published registry snapshot. When a published latency or availability change can alter the preferred provider, wake waiting workers to reconsider. Workers register before checking for work so an update between the comparison and the wait is retained.

## Testing

The independent branch on main `f8a62d4be` passed 1,262 network library tests on Rust 1.97.1, workspace all-target Clippy with warnings denied, formatting, and changelog validation. Three macOS loopback-binding tests were excluded after reproducing their failures on unchanged main. Three existing tests were ignored.

Earlier measurements of the initial ranking change showed about 27% higher committed BPS over heights 710k–749k. The wake-up change and this independent branch have not been rebenchmarked.


evan-forbes:
do we need this if we're already maximizing throughput and responsiveness via the congestion control? combined with that, it feels a bit over engineered. like a hat on a hat

czarcas7ic:
@evan-forbes looked into this. While congestion control determines how much a peer may request, this PR fixes cases where peers defer to one another or remain asleep despite available capacity. In this case, increasing capacity doesn't resolve that scheduling failure. 

From looking at PR 1023, it seems to extend the existing mechanism rather than introducing a replacement that makes #1006 unnecessary. Happy to be told otherwise, its's just that I have just been doing massive amount of A/B which point to this being important.

evan-forbes (approved):
Approving. I pushed `88f25e3cf` for Codex's second P1: `remove` and `upsert_status` now wake routines that deferred the floor. Without it, a routine that deferred to a peer kept sleeping after that peer disconnected or stopped serving the floor, until the 10 s heartbeat. `upsert_status` notifies only when `received_status` or the servable range changes, so a status flood cannot spin routines. `floor_ranking_notifies_when_a_carrier_changes_range_or_leaves` covers both paths.

On my earlier "hat on a hat" question: #1023 builds on this mechanism rather than replacing it, so this PR should land first.

**Not blocking:**

- V12 F-278738 (Medium) is real but small. A routine now ranks itself by its own published RTprop, which can lag its local filter by up to one heartbeat. A peer that measured fast and then went quiet can still take the floor inside that window. The floor then waits for the normal 8 s request timeout, because the peer has no fresh delivery measurement. Before this PR, the same window stalled the floor until the heartbeat, so the worst case barely moves. A cheap follow-up would publish the routine's own slots right before the ranking check, so its self score is fresh and still comes from the shared table.

PR #1016, opened 2026-09-16
fix(sync): admit completed header refill batches promptly

czarcas7ic:
## Motivation

Body commits keep freeing header capacity while a refill is still downloading.
The refill can keep extending its target instead of admitting headers already
received, letting the body queue drain.

## Solution

Prepare a completed 2,000-header normal extension batch immediately, while
preserving the 4,000-header capacity and the existing early checkpoint-prefix
escape. Forks, headers-only mode and auxiliary repair retain their existing
handling.

## Testing

The independent branch on main `f8a62d4be` passed 1,261 network library tests on Rust 1.97.1, workspace all-target Clippy with warnings denied, formatting, and changelog validation. Three macOS loopback-binding tests were excluded after reproducing their failures on unchanged main. Three existing tests were ignored.

Earlier paired Zakura-only runs measured 221.0 to 230.2 committed BPS over heights 705,000–790,000. These measurements predate the rebase onto main and have not been repeated for this independent branch.


evan-forbes (commented):
Thanks for tracking this down! The diagnosis is right! a running refill keeps extending into headroom that body commits free, and it holds its headers back until the body lag falls to 400. The fix is correct, and the tests and clippy pass locally.

I'd like to see if you're down to instead of merge only #1034 and #1032, since they fix the same case and a bit more. Here are the three versions of the admit condition in `should_prepare_checkpoint_prefix`, with the shared guards left out:

```rust
// main
entries > 400 && body_lag <= 400
// #1016
entries > 400 && (entries >= 2_000 || body_lag <= 400)
// #1034
entries >= 401
```

#1034 drops the body-lag condition, so a refill admits as soon as it holds one checkpoint range plus its successor. Any refill that #1016 would admit, #1034 admits earlier. Once #1034 lands, it replaces these lines. It also carries your moving-headroom regression as `completed_refill_does_not_chase_credits_returned_during_download`.

#1034 also changes when a refill starts, which this PR doesn't touch. In the 1.4M–1.5M trace, header lead averaged 2,856 of the 4,000-header window. About 1,144 body slots had no admitted header to download, because a refill starts only once the lead falls to 2,000. #1034 starts a refill once 401 slots are free, which is a lead of 3,599. The window and every other bound stay the same. It also wakes a refill when verified bodies free capacity, instead of waiting for a status refresh or a reanchor.

#1032 fixes a related gap. A refill that starts with one credit keeps requesting one header per continuation. The Sep 15 trace had 113 single-header requests.

Proposal:

1. Move #1032 and #1034 onto `main` as one PR. They don't depend on the drafts beneath them. Their commits replay onto current `main` with only a changelog conflict.
2. Run the experiment in #1034's design doc before merging. Admitting at 401 can make header transitions about five times more frequent, so the doc gates acceptance on CPU and RSS. If that gate fails, we pick a threshold between 401 and 2,000 inside that PR.
3. Close this PR once the combined PR is up.

Does that sound right to you? If you'd rather have this fix on `main` in the meantime, I'm open to that. The combined PR would then replace it.

czarcas7ic:
Closing in favor of https://github.com/zakura-core/zakura/pull/1136

PR #1017, opened 2026-09-16
perf(state): skip unused retained checkpoint validation context

czarcas7ic:
## Motivation

Ordinary checkpoint commits reconstruct durable header validation context even
when every promoted header is already retained and the planner does not
consume that context.

## Solution

Use the coherent retained graph under the existing writer locks to skip that
unused work. Missing headers and other transition causes keep the original
context path. Identity, continuity, eligibility, authority and atomic
publication checks are preserved.

## Testing

The independent branch on main `f8a62d4be` passed 99 state/header-chain tests on Rust 1.97.1, workspace all-target Clippy with warnings denied, formatting, and changelog validation.

Earlier paired Zakura-only runs over heights 705,000–735,000 measured 218.6 to 296.1 committed BPS. The control included the queue-refill patch. Those measurements have not been repeated for this independent branch.


evan-forbes (approved):
LGTM. 

Two small notes, neither blocking:

1. The perf numbers in the description come from the former stack, with the queue-refill patch in the control, and weren't repeated. The saved work is concrete, though: one parent read plus 27 predecessor reads and header hashes per checkpoint block. A one-line before/after BPS on this branch would still make the changelog claim easy to check later.
2. The skip assumes checkpoint commits carry no retention references. A future caller that passes one authenticated only by the parent lease would fail with `TransitionFailure::Authority`. It fails closed, so a comment on `retained_checkpoint_headers` saying so is enough.

czarcas7ic:
Thanks, added the fail-closed comment on `retained_checkpoint_headers` in b350c0f8b, and kept the changelog to the qualitative "reduced work" claim rather than rerunning BPS.

PR #1018, opened 2026-09-16
perf(headers): bound checkpoint pin checks to affected heights

czarcas7ic:
## Motivation

Each header transition scans every configured checkpoint pin, including
heights absent from both projections and every changed node.

## Solution

Bound the sorted pin slice by the affected heights, then run the existing
checks in their original order. Changed nodes outside the selected and
verified projections remain included.

## Testing

The independent branch on main `f8a62d4be` passed 291 header-chain library tests on Rust 1.97.1, workspace all-target Clippy with warnings denied, formatting, and changelog validation. Two existing header-chain tests were ignored.

Earlier paired Zakura-only runs over heights 705,000–735,000 measured 218.3 to 241.4 committed BPS with the queue-refill patch in both arms. Those measurements have not been repeated for this independent branch.


evan-forbes (approved):
LGTM. The bounded loop is exactly equivalent to the old one.

I pushed `96a6cebc3`, which adds a `debug_assert!` that trust pins stay strictly sorted by height, because the bounded loop would otherwise skip pins silently.

PR #1019, opened 2026-09-16
perf(headers): derive retained height from checkpoint finality

czarcas7ic:
## Motivation

Checkpoint finality scans the retained header graph to recover its minimum
height after pruning, although the new finalized anchor already determines
that minimum.

## Solution

Use the finalized anchor height when checkpoint finality actually appends a
finalized prefix. Keep the existing calculation for all other transition
shapes.

## Testing

The independent branch on main `f8a62d4be` passed 290 header-chain library tests on Rust 1.97.1, workspace all-target Clippy with warnings denied, formatting, and changelog validation. Two existing header-chain tests were ignored.

Earlier paired Zakura-only runs over heights 705,000–735,000 measured 214.1 to 236.3 committed BPS with the queue-refill patch in both arms. Those measurements have not been repeated for this independent branch.


crates/zakura-header-chain/src/transition/planner/write_set.rs
```diff
@@ -157,22 +157,27 @@ pub(super) fn derive_plan(
...
-            .into_iter()
-            .map(|node| node.height)
-            .min()
-            .unwrap_or(graph.view_finalized_frontier().height)
-    };
+    metadata.oldest_retained_height =
+        if effect.is_checkpoint_finality() && finality_append.is_some() {
+            // Finality removes ancestors and sibling subtrees. The new anchor is
+            // retained, and every remaining header descends from it.
+            graph.view_finalized_frontier().height
```
evan-forbes: Optional: recovery recomputes the floor as the graph minimum, so a debug assertion would catch a future finality path that leaves a lower header.

```suggestion
            let anchor_height = graph.view_finalized_frontier().height;
            debug_assert_eq!(
                graph
                    .view_header_nodes()
                    .into_iter()
                    .map(|node| node.height)
                    .min(),
                Some(anchor_height),
                "checkpoint finality must leave the anchor as the lowest retained header",
            );
            anchor_height
```

evan-forbes (approved):
LGTM

PR #1020, opened 2026-09-16
fix(sync): discard block size hints from another fork

czarcas7ic:
## Motivation

During fork recovery, block metadata pairs the selected branch's hash with the old branch's size. In the captured testnet incident, a valid 7,592-byte replacement was rejected against a 3,340-byte limit derived from the old 1,670-byte block, preventing sync progress.

## Solution

Discard the size hint when the stored block and requested block have different hashes, using the existing unknown-size allowance. This fixes the download rejection. Recovery of already-stuck verification requests remains under investigation.

## Testing

The state regression fails before the fix and passes afterward. A replay using the captured blocks confirms rejection with the old size and submission with an unknown size. Targeted tests, state-library Clippy, and formatting pass. An existing reanchor test failed once concurrently and passed serially; CI still needs to confirm the broader result.

Implemented and tested with Codex.


czarcas7ic:
@codex review

ValarDragon:
Sanity checking with AI that size hints can be none, but this seems like an important fix!

ValarDragon:
Screenshot 2026-09-21 at 2 37 39 PM

ValarDragon (approved):

PR #1027, opened 2026-09-16
feat(consensus): enforce ZIP 218 shielded action limits at NU7

evan-forbes:
## Motivation

The Zcash coinholder poll that closed on 2026-09-14 adopted [ZIP 218](https://zips.z.cash/zip-0218) for NU7. ZIP 218 has two parts: a 25-second block target spacing, and per-block shielded action limits. This PR is the action limits, split from #851 so each part can be reviewed and merged on its own. #851 keeps the target spacing.

NU7 has no activation heights yet, so these rules stay dormant on Mainnet and the default Testnet.

## Solution

From the NU7 activation height (`NetworkUpgrade::is_nu7_active`, ZIP 218's `IsNU7Activated`), the node:

- rejects blocks with more than 330 Orchard and Ironwood actions or 300 Sapling spends plus outputs, and blocks containing any Sprout JoinSplit;
- rejects blocks whose global shielded cost exceeds 330, where each JoinSplit counts twice;
- rejects mempool transactions whose own counts exceed a per-block limit; and
- reserves the coinbase transaction's shielded counts in `getblocktemplate`, then selects only transactions that fit the remaining limits.

The limits don't depend on the 25-second spacing. Neither PR depends on the other, and they can merge in either order.

## Beyond the ZIPs

- **Ironwood actions count against the Orchard limit and the global budget.** ZIP 218 names only Orchard actions. NU6.3 (ZIP 258) moves new Orchard-protocol value to the Ironwood pool, so uncounted Ironwood actions would bypass the limits.
- **The Sprout JoinSplit limit is zero, not the 25 ZIP 218 names.** ZIP 2003 disallows version 4 transactions at NU7, and only version 2, 3, and 4 transactions can contain JoinSplits, so no block at or after NU7 can contain one.
- **The mempool rejects a transaction that exceeds a per-block limit on its own.** ZIP 218 does not specify this rejection. No block can include such a transaction, so the verifier gives it misbehavior score 100.

## Review history

This code was reviewed in #851. The V12 audit finding [F-280623](https://v12.sh/runs/7913/280623) is fixed here: `is_nu7_active` now compares the current network upgrade with NU7, instead of also requiring a literal NU7 entry in the activation list.

## Testing

- Per-pool and global limits, over and exactly at each limit, in block validation, mempool admission, and block-template selection, including coinbase shielded counts.
- Ironwood actions alone and mixed with Orchard actions.
- Historical blocks satisfy the limits.
- `is_nu7_active` is true exactly from the NU7 activation height.

Local runs passed on this branch alone: `cargo fmt --all -- --check`; `cargo clippy --locked -p zakura-chain -p zakura-consensus -p zakura-rpc -p zakura-state -p zakura --all-targets -- -D warnings`, with and without `--features nu7`; and `cargo test --locked --lib` for `zakura-chain`, `zakura-consensus` and `zakura-rpc`.

## Changelog

Adds `docs/changelog/unreleased/1027.md`.

`zakura-consensus` moves to 9.0.0 because the limit checks add variants to the exhaustive `TransactionError` enum.

- [ZIP 258](https://zips.z.cash/zip-0258)
- #851 (ZIP 218 target spacing)


crates/zakura-consensus/src/block/check.rs
```diff
@@ -404,6 +405,92 @@ pub fn time_is_valid_at(
...
+            ShieldedActionCounts::default(),
+            ShieldedActionCounts::saturating_add,
+        );
+
+    if totals.orchard_and_ironwood_actions > ORCHARD_BLOCK_ACTION_LIMIT {
+        return Err(TransactionError::OrchardActionsExceedBlockLimit {
+            actions: totals.orchard_and_ironwood_actions,
+            limit: ORCHARD_BLOCK_ACTION_LIMIT,
+        });
+    }
```
ValarDragon, line 467: Ok this is a consensus divergence from my ZIP update if global ever changes (https://github.com/zcash/zips/pull/1361). IMO we merge, then update after.

crates/zakura-consensus/src/error.rs
```diff
@@ -272,6 +272,24 @@ pub enum TransactionError {
     #[error("wrong tx format: tx version is ≥ 5, but `nConsensusBranchId` is missing")]
     MissingConsensusBranchId,
 
+    #[error(
+        "Orchard and Ironwood action count {actions} exceeds the per-block Orchard limit of {limit}"
+    )]
+    OrchardActionsExceedBlockLimit { actions: u32, limit: u32 },
```
ValarDragon, line 278: Same comment as for Ironwood being a separate limit

ValarDragon (approved):

ValarDragon:
Lets resolve my two comments in follow-up's and the mempool @ upgrade boundary

PR #1028, opened 2026-09-16
fix(state): reserve database format 29 for NuTachyon

S1nus:
## Motivation

NuTachyon development assigns database format major version 29 while it remains outside `main`. Without a reservation on `main`, a future production database migration could independently assign version 29 to an incompatible format.

## Solution

Reserve database format major version 29 for NuTachyon in the production version constants. A compile-time assertion prevents the production database format from selecting that reserved number, so existing CI compilation enforces the reservation.

## Testing

- `ROCKSDB_LIB_DIR="/opt/homebrew/lib" cargo +1.97 check -p zakura-state --all-targets`
- `ROCKSDB_LIB_DIR="/opt/homebrew/lib" cargo +1.97 fmt --all -- --check`
- `npx --yes markdownlint-cli docs/changelog/unreleased/1028.md --config .github/.markdownlint.yaml`
- `git diff --check`

## Changelog

Added an internal-only changelog fragment because this reservation has no operator-visible effect.

- NuTachyon implementation: #795


evan-forbes (approved):

S1nus:
closed as premature

let's just assume no user will make the unlikely error of using a NuTachyon DB on a non-NuTachyon build, and vice versa.

PR #1030, opened 2026-09-16
perf(headers): reuse the unchanged selected chain path

czarcas7ic:
## Motivation

Header settlement reconstructs the selected chain path even when a body update
leaves its tip and finalized boundary unchanged.

## Solution

Reuse the existing path when both endpoints are unchanged. Canonical parent
links are immutable and retention protects that path. Changed endpoints keep
the existing reconstruction.

## Testing

The independent branch on main `f8a62d4be` passed 290 header-chain library tests on Rust 1.97.1, workspace all-target Clippy with warnings denied, formatting, and changelog validation. Two existing header-chain tests were ignored. The broader workspace run had three macOS loopback-binding failures, also reproduced on unchanged main, and one peer-reliability scenario failure. That scenario passed on both main and this branch after rebuilding with the installed Command Line Tools. The full workspace suite was not rerun after that rebuild.

Earlier matched runs over heights 3,000,000–3,025,000 measured 444.0 to 465.7 committed BPS (+4.9%). That is a range result measured before this rebase, not an isolated full-sync improvement.


evan-forbes (approved):
seems fine

PR #1032, opened 2026-09-16
fix(sync): restore header continuation capacity as credits return

evan-forbes:
## Motivation

A header request that starts with one available credit keeps requesting one header on every continuation. The reactor stores the first issued count as the continuation maximum, even after shared credits return. A retained sync trace records hundreds of one-header requests to a supplier advertising 1,000 headers per response. Apply exhausts the published prefix during a 19.2-second starvation episode.

## Solution

Store the negotiated page maximum before applying temporary credit limits. Apply current shared staging credits, durable headroom, and the body window to each issued request. Keep negotiated peer and byte limits intact. No parameters or prefix-publication thresholds change.

This PR stacks on #1031. It addresses a separate header producer defect exposed by that PR's retained node test.

## Testing

The regression starts with one shared credit, releases another request's credits, and checks that the next wire request grows to the byte-limited page size. It also checks that a peer advertising a one-header limit stays capped at one. The existing wire test keeps continuations within the remaining durable headroom. All 191 header-sync tests pass. Network Clippy passes with warnings denied.

The retained fresh-database run keeps sequencer submissions at 401 throughout 417.54 measured seconds. The driver averages 400.93 active applies out of 401, with zero submission starvation. All 30 metadata queries publish work. The archive is downloaded and checksum-verified.

The longest one-header run is five requests versus 275 in the preceding late-chain case. The fresh run does not independently reproduce the first-request-one-credit trigger; the deterministic regression covers it. Different heights and a different state volume prevent a direct throughput comparison. Both state databases and all traces remain preserved.

## Changelog

`docs/changelog/unreleased/1032.md`.


evan-forbes:
Superseded by #1136, which carries #1032 and #1034 onto `main` as one PR.

PR #1040, opened 2026-09-17
feat(consensus): seed and specify the NSM value balance

evan-forbes:
## Motivation

The 2026-09-15 ZIP Editor call settled two questions that #859 left open, and
reopened one it had already answered.

1. **The NSM value balance is seeded.** It holds the block subsidy and fees that
   pre-NU6 coinbase transactions never claimed. #859 follows
   `draft-valargroup-deploy-nu7.md` (zips#1363), whose only commit predates the
   call by under two hours, and starts the balance at zero.
2. **The balance has a name.** zips#1354 calls it `NSMValueBalance`, not the
   issuance deficit.
3. **The payout fraction is wrong.** `BLOCK_SUBSIDY_FRACTION_NUMERATOR` is a flat
   4126, calibrated for 75-second blocks. ZIP 218 makes NU7 blocks 25 seconds
   apart, so the flat numerator pays the balance out three times too fast: a
   half-life of about 1.33 years instead of four. A spacing fix landed earlier in
   the stack and was lost in the explicit-pool refactor.

## Solution

Six commits, each landing green on its own.

1. **Rename** `issuance_deficit` to `nsm_value_balance` across 32 files. No
   behaviour change.
2. **Add `INITIAL_NSM_VALUE_BALANCE`** as a network parameter: 36,858,445,520
   zatoshi on Mainnet, 55,768,414,957 on Testnet, zero everywhere else, with a
   builder knob and a config field for configured networks.
3. **Seed the balance** on the last block below NU7, and offset the format 29
   backfill to match.
4. **Derive the fraction** as ZIP 234 defines it, `floor(LN2_SCALED /
   HalvingInterval(height))`, with the halving interval derived from the target
   spacing. One expression covers every era: 840,000 / 1,680,000 / 5,040,000
   blocks on Mainnet, giving 8252 / 4126 / 1375.
5. **Reject a negative balance from NU7**, not from the later reissuance start.
6. **Rename the start height** to `nsm_reissuance_height`, zips#1363's name.

### Where the measured constants come from

At the last pre-NU6 height on each chain, the halving schedule's cumulative
subsidy minus the chain value pools:

| Network | Height | Scheduled | Issued | Seed |
| --- | --- | --- | --- | --- |
| Mainnet | 2,726,399 | 15,750,000 ZEC | 15,749,631.41554480 ZEC | 36,858,445,520 zat |
| Testnet | 2,975,999 | — | — | 55,768,414,957 zat |

Both terms are already in every node's own state: the left is a closed form over
the halving schedule, and the right is the chain value pool total, a running
total maintained at commit time. No archive state is needed, which two nodes
confirmed — a 278 GB archive node and a snapshot-bootstrapped node holding about
11,000 blocks returned the same number. The ZIP fixes the value as a constant so
nodes agree without consulting anything.

zips#1354 leaves the constant as a TODO and estimates 350.8 ZEC. The measured
value is about 17.8 ZEC higher.

### Fee recycling scope

Fee recycling is implemented in follow-up #1056, stacked on this PR. The NU7 deployment draft proposes
contributing `floor(6 * TransactionFees(h) / 10)` from aggregate block fees to NSM
without adding transaction fields. This PR retains the full fee claim in coinbase
validation and block templates. The deferral does not exclude fee recycling from
the intended NU7 scope.

For example, with 1,000 zatoshi in fees, this implementation requires the coinbase
to claim all 1,000. The draft would leave 400 for the miner and contribute 600 to
NSM. Excluding the old transaction-field mechanism does not establish that
`removed(h)` is zero.

The current rule makes `removed(h)` zero for semantically valid blocks from NU7
onward. The block's change across the six monetary pools is therefore the halving
subsidy plus the bonus. Subtracting that change from the halving subsidy yields
`-AdditionalBlockSubsidy(h)`. Once coinbases withhold the fee contribution, the
same calculation yields `-AdditionalBlockSubsidy(h) + removed(h)`. Crediting the
contribution separately would count the same fees twice. That follow-up updates
coinbase validation, block templates, activation tests, and the accounting comments
together.

### No chain changes behaviour

The seed is zero on every network but Mainnet and the default Testnet, and
neither has an NU7 activation height yet.

## Testing

The fee recycling scope clarification changes comments and documentation only.
Formatting and Markdown lint pass. Existing behavior tests were not rerun for
this clarification.

- `cargo test --lib` with no features, after the stack stopped using the
  `nu7` feature: zakura-chain 422, zakura-consensus 239, zakura-header-chain
  284, zakura-rpc 144 and zakura-state 640 pass
- `cargo test -p zakura-network` — 1257 pass
- `cargo clippy --all-targets -- -D warnings` on the five crates above
- `markdownlint` on both changed Markdown files

New tests:

- the seed lands on the last block below NU7 and on no earlier block
- rolling back below the seed height undoes it, and replay restores it
- the format 29 backfill seeds the same block, as a unit test on
  `eligible_balance` and against a real database
- the halving interval and fraction numerator per spacing era, which is the test
  that would have caught the flat 4126
- one halving interval leaves about half the balance
- a positive balance always pays at least one zatoshi, so it drains in finite time
- a block over-claiming at NU7 is rejected below the reissuance start height

The fixtures now read the fraction from the network instead of repeating 4126.

## Changelog

Added after this draft opens.

- zcash/zips#1354 `draft-judah-nsm-halving-preserving-issuance.md`
- zcash/zips#1363 `draft-valargroup-deploy-nu7.md`
- ZIP 218, ZIP 234, ZIP 236
- 2026-09-15 ZIP Editor call, reported on zcash/zips#1354

### Follow-up Work

- **Known divergence:** the reissuance start height is still derived from a
  crossing rule that no ZIP defines. Both drafts take it from the deployment ZIP,
  and both leave it TBD. A `TODO` in `nsm_reissuance_height` records this.
- Post the measured constants on zcash/zips#1354 to close its `INITIAL_NSM_VALUE_BALANCE`
  TODO, after a zcashd cross-check.
- Expose the balance over `getblockchaininfo`. Additive, so it belongs in its own PR.

Used Codex to clarify fee recycling scope in code comments, the design document,
and this description.


crates/zakura-chain/src/block.rs
```diff
@@ -346,17 +349,35 @@ impl Block {
...
+    ///
+    /// zips#1354 seeds the balance immediately before NU7 and then draws it down by the
+    /// bonus each block claims:
+    ///
+    /// > NSMValueBalance(NU7ActivationHeight - 1) = INITIAL\_NSM\_VALUE\_BALANCE
+    /// >
+    /// > NSMValueBalance(height) = NSMValueBalance(height - 1)
+    /// >   - AdditionalBlockSubsidy(height) + removed(height)
+    ///
+    /// NU7 deploys neither ZIP 233 nor ZIP 235, so `removed(height)` is zero throughout.
```
czarcas7ic: AI flagged, I know this is still draft but leaving it here

---

The current NU7 deployment draft proposes routing 60% of aggregate transaction fees into the NSM balance without adding transaction fields:
https://github.com/zcash/zips/blob/32f447759aba83acfb20aab0757b68147643de22/zips/draft-valargroup-deploy-nu7.md#L120-L149

For example, 1,000 zatoshi in fees would leave 400 for the miner and contribute 600 to NSM, so excluding the old transaction-field mechanism doesn't by itself establish that removed(height) is zero. Could we clarify whether fee recycling is deferred to a separate change or excluded from the intended NU7 scope, and reflect that decision here and in the PR description?

evan-forbes: fixed in the recylcing in #1056 correct? so can resolve?

czarcas7ic: Correct yes!

crates/zakura-chain/src/block.rs
```diff
@@ -365,10 +395,21 @@ impl Block {
...
+        let Some(nu7) = NetworkUpgrade::Nu7.activation_height(network) else {
+            return Ok(Amount::zero());
+        };
+
+        // The seed lands on the last block below NU7, so the balance already holds it when
+        // the first NU7 block claims its bonus. NU7 at genesis has no such block, and a
+        // chain with no history before NU7 has nothing to seed.
+        if nu7.0.checked_sub(1) == Some(height.0) {
+            return network
+                .initial_nsm_value_balance()
```
evan-forbes, line 407: Consensus seed insertion: add the historical seed at `NU7 - 1`, so the first NU7 block reads it from its parent. The constants are 36,858,445,520 zatoshis on Mainnet and 55,768,414,957 on default Testnet. Configured networks default to zero and can override it. NU7 at genesis has no preceding seed block. Review this together with `nsm_value_balance_pool.rs`, which must reconstruct the same balance during migration and restart.

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -469,19 +480,50 @@ pub fn halving(height: Height, network: &Network) -> u32 {
...
+/// would pay out three times as fast under ZIP 218's 25-second blocks.
+///
+/// On Mainnet this is 4126 after Blossom and 1375 from NU7.
+///
+/// [ZIP 234]: https://zips.z.cash/zip-0234
+pub fn block_subsidy_fraction_numerator(height: Height, network: &Network) -> u128 {
+    let interval = u128::try_from(halving_interval(height, network).max(1))
+        .expect("a positive halving interval fits in u128");
+
+    LN2_SCALED / interval
```
evan-forbes, line 524: Consensus payout arithmetic: first floor `6,931,680,000 / HalvingInterval(height)` to obtain the numerator, then round the bonus upward in `reissuance_amount`: `ceil(parent_balance * numerator / 10^10)`. The numerator changes from 4126 at 75-second spacing to 1375 at 25-second spacing. These are two separate rounding steps. This is the follow-up correction to #1053’s fixed fraction and preserves the intended approximate four-year half-life.

crates/zakura-state/src/service/check.rs
```diff
@@ -45,48 +45,53 @@ mod tests;
...
+pub(crate) fn nsm_value_balance_is_non_negative(
     network: &Network,
     height: block::Height,
     value_pools: &ValueBalance<NonNegative>,
     block_value_pool_change: &ValueBalance<NegativeAllowed>,
 ) -> Result<(), ValidateContextError> {
-    if !is_zip234_active(network, height) {
+    let nu7_active = NetworkUpgrade::Nu7
+        .activation_height(network)
+        .is_some_and(|nu7| height >= nu7);
```
evan-forbes, line 73: Consensus rejection boundary: reject `parent_balance + block_balance_change < 0` from NU7, even when reissuance starts later. #1053 applies this check only from reissuance; this PR closes that intermediate difference. Both finalized and non-finalized commit paths must apply this rule before changing pools. The signed storage type does not authorize a negative post-NU7 balance.

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -520,25 +566,25 @@ pub const ZIP234_START_HALVING: u32 = 3;
...
 /// on a network without NU7. Configured testnets and Regtest can replace the crossing
-/// height with [`ParametersBuilder::with_zip234_start_height`]. Mainnet and the default
+/// height with [`ParametersBuilder::with_nsm_reissuance_height`]. Mainnet and the default
 /// Testnet always use the crossing rule.
 ///
 /// [ZIP 234]: https://zips.z.cash/zip-0234
-/// [`ParametersBuilder::with_zip234_start_height`]: super::testnet::ParametersBuilder::with_zip234_start_height
-pub fn zip234_start_height(network: &Network) -> Option<Height> {
+/// [`ParametersBuilder::with_nsm_reissuance_height`]: super::testnet::ParametersBuilder::with_nsm_reissuance_height
+pub fn nsm_reissuance_height(network: &Network) -> Option<Height> {
```
evan-forbes, line 574: Remaining specification difference: this function still derives the reissuance start from the crossing rule and clamps it to NU7. The halving-preserving draft takes that height from the deployment ZIP. Renaming the parameter does not resolve the policy difference. Configured networks can override it; neither this PR nor fee-recycling #1056 assigns the deployment height.

evan-forbes (commented):
Review navigation: [13c96f705](https://github.com/zakura-core/zakura/commit/13c96f705c6001e6625c24b41ecb9c5c5c63d1bc) contains mechanical NSM names; [e36e863a0](https://github.com/zakura-core/zakura/commit/e36e863a0ce4e0d693f35d4ffc0c2a2217680e06) contains the consensus changes, matching migration, caller adaptations, and tests; [d9351324a](https://github.com/zakura-core/zakura/commit/d9351324a0b895494f0c52025e182eaa98c5ef70) contains documentation. The final tree is identical to the previous PR head.

This PR closes three intermediate differences identified in [#1053](https://github.com/zakura-core/zakura/pull/1053): the zero historical seed, fixed per-block payout fraction, and negative-balance rejection starting only at reissuance. Compared with [Shielded Labs’ halving-preserving draft](https://github.com/ShieldedLabs/zips/blob/60720df9e971e19d8b6f67f1869426d78c96d250/zips/draft-judah-nsm-halving-preserving-issuance.md), the derived reissuance crossing height remains unresolved. This PR supplies measured seed constants where that draft has a provisional estimate. Fee recycling is implemented separately in [#1056](https://github.com/zakura-core/zakura/pull/1056).

Approved parent #1054 and approved dependent #1056 retain their histories. #1056 still points through the previous #1040 ancestry; its eventual restack needs reviewer coordination.

Validation: Rust 1.97 passed four reissuance tests, two fraction/half-life tests, and 25 NSM state/migration tests. The rename-only commit passed `cargo check --locked -p zakura-consensus -p zakura-rpc --tests`. Formatting, changelog, and diff checks passed.

PR #1041, opened 2026-09-17
fix(consensus): give Ironwood its own ZIP 218 action limit

evan-forbes:
## Motivation

A follow-up from the review of #1027, Testnet blocking.

#1027 counted Ironwood actions against the Orchard per-pool limit, because ZIP
218 named only Orchard actions and NU6.3 (ZIP 258) moves new Orchard-protocol
value to Ironwood. The ZIP 218 update in
[zcash/zips#1361](https://github.com/zcash/zips/pull/1361) instead gives Ironwood
its own per-pool limit, equal to Orchard's, with both pools drawing on the one
global shielded budget.

This changes no block's validity while `GlobalShieldedBudget` equals
`OrchardProtocolBlockActionLimit`: a block with more than 330 Orchard plus
Ironwood actions already costs more than the global budget, so the merged counter
and the split counters accept exactly the same blocks. The divergence appears
only if the global budget is ever raised above the per-pool limit, which is what
the review comment on #1027 called out. What does change today is which error a
mixed overflow reports, now `ShieldedCostExceedsBlockBudget` rather than
`OrchardActionsExceedBlockLimit`.

## Solution

`ShieldedActionCounts` splits `orchard_and_ironwood_actions` into
`orchard_actions` and `ironwood_actions`, and `ORCHARD_BLOCK_ACTION_LIMIT` becomes
`ORCHARD_PROTOCOL_BLOCK_ACTION_LIMIT`, checked once per pool. The new
`IronwoodActionsExceedBlockLimit` reports an Ironwood overflow. The shielded cost
still sums both pools, so the global budget of 330 keeps a block from spending
the per-pool capacity twice. Block template selection tracks the two pools
separately.

### Removed during review

Earlier revisions of this PR also did two other things. Both are gone.

- **An activation grace period for mempool relay** (V12
  [F-280823](https://v12.sh/runs/7955/280823)). It forgave a transaction at the
  first activation-dependent check it failed, without verifying the transaction
  under the previous upgrade. So it forgave transactions that were also invalid
  for other reasons, and it missed other activation-dependent failures. #960's
  per-IP cooldowns cover the honest-peer case: a ZIP 218 rejection there starts
  a temporary cooldown rather than a ban.
- **A ZIP 218 check in checkpoint verification** (V12
  [F-280824](https://v12.sh/runs/7955/280824)). Checkpoint verification trusts a
  block by its hash and already skips most transaction rules, so the check
  protected nothing on a real network.

## Testing

- Ironwood above its own per-pool limit, Orchard and Ironwood together above the
  shared global budget, and each pool exactly at its limit, in block validation,
  mempool admission, and block-template selection, including coinbase counts.

Local runs passed on this branch: `cargo fmt --all -- --check`; `cargo clippy
--locked -p zakura-chain -p zakura-consensus -p zakura-rpc -p zakura
--all-targets -- -D warnings`; `cargo test --locked --lib` for `zakura-chain`,
`zakura-consensus` and `zakura-rpc`; the same `zakura-consensus` tests with
`--features nu7`; and the `zakurad` mempool tests.

## Changelog

Adds `docs/changelog/unreleased/1041.md`.

`zakura-chain` moves to 9.0.0 for the `ShieldedActionCounts` field split and the
constant rename. `zakura-consensus` moves to 10.0.0 for the new
`TransactionError::IronwoodActionsExceedBlockLimit` variant, because
`TransactionError` is not `#[non_exhaustive]`.

- [zcash/zips#1361](https://github.com/zcash/zips/pull/1361) (Ironwood per-pool limit)
- #1027 (ZIP 218 shielded action limits)
- #960 (mempool cooldowns instead of bans)


p0mvn (approved):

ValarDragon (approved):
LGTM, we have to do another thing to get grace period E2E working, but its fine for this one.

ValarDragon (changes requested):
actually I think we shouldn't merge as-is. The consensus critical part is perfect, the other part about grace periods I think should not go in until more fixed. Namely:

>   This is the fundamental grace-classification problem.

  The verifier encounters certain checks early—branch ID and ZIP 218 limits—and immediately returns the zero-scored grace error. It does not finish verification or verify the
  transaction under the previous upgrade.

  That causes errors in both directions:

  - Too broad: a transaction can violate ZIP 218 and independently be invalid, such as being a coinbase transaction submitted to the mempool. The early ZIP 218 failure hides the later
    independent failure and gives it score 0.

ValarDragon:
> [F-282392](https://v12.sh/runs/8012/282392)

AFAIU this is an edge case where I checkpoint sync under the wrong binary, but commit NU7 blocks, then restart with NU7 binary?

I think thats fine. We just make the first NU7 checkpoint more than 1k blcoks ahead

EDIT: Its also fine because its literally in the checkpoint. The relevant rules are already checked by everyone else.

evan-forbes:
makes sense, leaving as is then. 94269cba6 also removes the ZIP 218 check from checkpoint verification.

ValarDragon (approved):

PR #1044, opened 2026-09-18
perf(state): read difficulty context from headers

p0mvn:
## Motivation

Difficulty checks load full block bodies even though they only consume header timestamps and targets.

## Solution

Use the existing ancestor iterator for headers in contextual validation, prepared-mined relay checks, and mining-template history. Obtain the parent height from the iterator. Keep the 28-header context and all consensus parameters unchanged.

## Testing

- `cargo +1.97.0 test --locked -p zakura-state --lib`: 596 passed, 4 ignored.
- `cargo +1.97.0 clippy --locked -p zakura-state --all-targets -- -D warnings`: passed.
- GitHub checks are running on the latest head; this PR remains draft.

## Changelog

Added `docs/changelog/unreleased/1044.md`.

### Related extractions

#1044 (header reads), #1045 (context retention/migration), #1046 (subsidy), and #1047 (node timing) each target `main` independently. The coordinated NU7 timing switch remains in #851.


ValarDragon (approved):
Cool, good perf update. TY @evan-forbes too, glad we split this out, makes it very clear

PR #1045, opened 2026-09-18
feat(state): retain and backfill wider difficulty context

p0mvn:
## Motivation

A future wider averaging window needs more retained history, including on databases created with the current 27-predecessor context.

## Solution

Retain 113 ancestor headers (112 below the separately retained anchor), authenticate and backfill older contexts before the startup audit, and update context-bound fixtures. The active difficulty window remains 17 and spacing remains unchanged. Migration coverage includes an idempotent second resize/startup.

Extracted from #851 as an independent PR against `main`. This branch does not activate the 25-second / 102-block NU7 rules.

## Testing

- `cargo +1.97.0 test --locked -p zakura-header-chain -p zakura-state --lib`: 284 header-chain tests and 597 state tests passed; 6 ignored in total.
- `cargo +1.97.0 clippy --locked -p zakura-header-chain -p zakura-state --all-targets -- -D warnings`: passed.
- Formatting, `git diff --check`, changelog validation, and fragment Markdown lint passed.
- GitHub checks are running on the latest head; this PR remains draft.

## Changelog

Added `docs/changelog/unreleased/1045.md`.

### Related extractions

#1044 (header reads), #1045 (context retention/migration), #1046 (subsidy), and #1047 (node timing) each target `main` independently. The coordinated NU7 timing switch remains in #851.

PR #1046, opened 2026-09-18
refactor(chain): derive subsidy and halvings from the spacing schedule

p0mvn:
## Motivation

The existing subsidy formulas hardcode two spacing eras, making future changes harder to review. The halving-height inverse also adds the slow-start shift twice when a halving occurs before Blossom on a custom network.

## Solution

Sum elapsed block seconds over the existing spacing schedule, derive subsidy from its spacing ratio, and find halving heights by monotonic inversion. Preserve the existing 150/75-second schedule, including configured NU7. Add comparisons against the previous halving formula and a regression for the custom-network inverse (110 instead of 120). Reject non-positive custom halving intervals and intervals whose conversion to target seconds or post-Blossom blocks would overflow, before storing them in the network parameters.

## Testing

- `cargo +1.97.0 test --locked -p zakura-chain --lib`: 408 passed, including invalid and boundary custom halving intervals.
- `cargo +1.97.0 clippy --locked -p zakura-chain --all-targets -- -D warnings`: passed.
- GitHub checks are running on the latest head. The new interval regression was also run against the old builder and failed as expected.

## Changelog

Added `docs/changelog/unreleased/1046.md`.

### Related extractions

#1044 (header reads), #1045 (context retention/migration), #1046 (subsidy), and #1047 (node timing) each target `main` independently. The coordinated NU7 timing switch remains in #851.


p0mvn:
This is only related to the custom testnets, does not touch mainnet code

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -419,28 +417,46 @@ pub fn halving_divisor(height: Height, network: &Network) -> Option<u64> {
...
+            .peek()
+            .map(|(next_start, _)| *next_start)
+            .unwrap_or(height);
+        total_block_seconds += (era_end - era_start) * era_spacing.num_seconds();
+    }
 
-        (scaled_pre_blossom_height + post_blossom_height) / network.post_blossom_halving_interval()
-    };
+    let pre_blossom_denominator =
+        network.pre_blossom_halving_interval() * pre_blossom_spacing_seconds;
```
czarcas7ic, line 451: AI found comment

---

The Rust builder accepts with_halving_interval(122_978_293_824_730_345), whose existing multiplication by 2 fits in i64.
https://github.com/zakura-core/zakura/blob/5069d6472635c5f1e99110ffd2d9b19d190f9451/crates/zakura-chain/src/parameters/network/testnet.rs#L918-L930

With slow start disabled and Blossom/Canopy at height 1, converting that interval to seconds panics with overflow checks or wraps to 134 in release arithmetic. Height 2 then gets halving index 1 instead of 0. Can we reject intervals that aren't positive and representable in seconds when constructing the network, and add a regression test?

czarcas7ic (approved):
I mentioned in person but feels like if this logic is for custom testnets only we should make it more clear to prevent misuse, but non blocking.

Saying this because even reviewing the small 181 line PR it wasn't immediately clear that I was reviewing logic that was not being used for mainnet

crates/zakura-chain/src/parameters/network/subsidy.rs
p0mvn: AI comment: Fixed in b24c4b8a0. ParametersBuilder::with_halving_interval now rejects non-positive intervals and uses checked multiplication to validate both the conversion to pre-Blossom target seconds and the post-Blossom block interval before storing either field. Invalid inputs return InvalidHalvingInterval.

The regression covers your 122_978_293_824_730_345 value, zero, negative values, i64 extremes, and the first interval above the representable boundary. It also accepts the largest valid interval and checks that its halving index remains zero at heights 2 and Height::MAX with slow start disabled and Blossom/Canopy at height 1.

All 408 zakura-chain library tests and all-target Clippy passed. The regression fails against the old builder. Formatting, changelog validation, and Markdown lint also passed.

PR #1047, opened 2026-09-18
refactor(node): derive timing policies from target spacing

p0mvn:
## Motivation

Several node policies hardcode post-Blossom timing, which would diverge from consensus after a future spacing change. The Testnet template heuristic also reads the parent height instead of the candidate height.

## Solution

Derive mining-template allowance, the RPC averaging-window lookup, sync lookahead, and progress timing from the existing schedule. Add an averaging-window accessor that still returns 17. Preserve current NU7 spacing at 75 seconds; cover the existing Blossom boundary and unchanged NU7 behavior. The candidate-height lookup intentionally fixes activation-boundary behavior.

## Testing

- `cargo +1.97.0 test --locked -p zakura-state -p zakura-rpc -p zakura --lib`: 598 state, 149 RPC, and 432 node tests passed; 5 ignored in total.
- `cargo +1.97.0 clippy --locked -p zakura-chain -p zakura-state -p zakura-rpc -p zakura --all-targets -- -D warnings`: passed.
- `cargo +1.97.0 xtask check-rpc-artifacts`: passed.
- Full-suite results above were obtained before rebasing. After rebasing onto main, all 3 difficulty tests, 149 RPC tests (1 ignored), and 112 sync tests passed. Formatting, changelog validation, and RPC artifact checks also passed after resolving the conflicts.

## Changelog

Added `docs/changelog/unreleased/1047.md`.

### Related extractions

#1044 (header reads), #1045 (context retention/migration), #1046 (subsidy), and #1047 (node timing) each target `main` independently. The coordinated NU7 timing switch remains in #851. End-of-support timing is already on `main` through #1043, and header-only reads are inherited from #1044.


crates/zakura-chain/src/parameters/network_upgrade.rs
```diff
@@ -522,6 +522,13 @@ impl NetworkUpgrade {
         }
     }
 
+    /// Returns the difficulty averaging window at the selected network height.
+    ///
+    /// All currently configured upgrades use the same 17-block window.
+    pub fn averaging_window_for_height(_network: &Network, _height: block::Height) -> usize {
+        POW_AVERAGING_WINDOW
+    }
```
czarcas7ic, line 530: nit: do we ever expect this to not be the case? Seeing it used in code would make me think it varies per network when it doesnt.

czarcas7ic (approved):

crates/zakura-chain/src/parameters/network_upgrade.rs
p0mvn: AI comment: Yes—#851 plans to change the averaging window from 17 to 102 blocks at NU7 activation. This PR only prepares the caller and deliberately keeps the window at 17 everywhere, so you are right that it does not vary today.

The network parameter is needed to select that network’s activation schedule, and the height determines which upgrade applies. The window size itself is an upgrade parameter. We split this preparation from the consensus switch so each can be reviewed separately.

PR #1049, opened 2026-09-18
fix(state): preserve deferred funding during slow-start replay and rollback

evan-forbes:
## Motivation

only relevant on custom testnets

Block commits always apply the ZIP 1015 deferred (lockbox) funding stream to the chain value pools, including during slow start. Rollback and the legacy BlockInfo replay in the disk-format upgrade instead skip it for every height up to slow_start_interval(). 

On Mainnet and the public Testnet, this gap has no effect: funding streams (ZIP 207, ZIP 214) start at Canopy, well after slow start ends. On configured networks, a funding stream can start inside slow start. There, a rolled-back or replayed database computes a deferred pool that differs from the pool the original commits wrote. The ZIP 234 stack makes this mismatch consensus-relevant. #1052 derives and persists the issuance deficit from the chain value pools. #1053 then makes each block's subsidy depend on its parent's deficit. 

Any pool drift after rollback or replay would therefore change the subsidies the node accepts. This PR computes the deferred change, net of the ZIP 271 lockbox disbursement, at every height in both paths. It also makes replay fail loudly on a subsidy error instead of substituting zero. Consensus behavior on Mainnet and the public Testnet stays unchanged.

Chunk 1 of the #859 split (#1049 → #1050 → #1051 → #1052 → #1053 → #1054). Targets the ZIP 218 branch (#851).

## Solution

Calculate deferred funding at every height in both paths. Propagate subsidy errors during legacy replay instead of substituting zero. This is chunk 1 of the split of #859 and depends on the split-branch CI prerequisite (#1055), which targets the ZIP 218 branch (#851).

## Testing

The helper regression and direct database replay regression pass. Restoring the original slow-start omission makes the replay regression fail.

## Changelog

Added `docs/changelog/unreleased/1049.md`.


p0mvn (approved):

evan-forbes:
Responses to V12 run [8014](https://v12.sh/runs/8014):

[F-282384](https://v12.sh/runs/8014/282384): Fixed in 95fb67db0 on #1052, where the migration now refuses those databases and requires a resync.

p0mvn (approved):

PR #1050, opened 2026-09-18
feat(state): await committed parent state across forks

evan-forbes:
## Motivation

Before ZIP 234, the block subsidy was a pure function of height and network, so the block verifier never needed chain state to check a coinbase. ZIP 234 reissuance changes that: once reissuance starts, the subsidy for a block depends on the issuance deficit recorded in its parent's value pools. 

The verifier therefore needs the committed state of the block's actual parent, not the best chain tip. That parent can sit on a side fork, and because verification is pipelined, it may still be in the queue to commit when its child arrives. The existing hash lookup only searches the best chain and returns immediately, so it misses both cases. 

This PR lets the verifier look up a hash across every non-finalized fork and, through AwaitBlockInfo, wait for the parent to commit. Height lookups stay best-chain-only, because a height has no single answer across forks. Later chunks in the #859 stack use this to compute parent-aware subsidies; this PR adds only the state plumbing and does not change consensus behavior on its own.

## Solution

Add AwaitBlockInfo and commit notifications. Search every non-finalized fork for hashes, but resolve heights only on the best chain. Bump the state crate major version for the new public request and response variants.

Chunk 2 of #859. Depends on #1049.

## Testing

Tests cover competing forks and best-chain height lookup. The async regression covers multiple readers, unrelated commits, pending parents, and parents committed before registration. Every test wait has a timeout. All four block-info tests pass on this chunk.

## Changelog

Added `docs/changelog/unreleased/1050.md`.


crates/zakura-state/src/service.rs
```diff
@@ -2005,6 +2010,34 @@ impl Service<Request> for StateService {
...
+                        };
+
+                        if block_info.is_some() {
+                            return Ok(Response::BlockInfo(block_info));
+                        }
+
+                        block_commit_receiver
+                            .changed()
+                            .await
+                            .map_err(BoxError::from)?;
```
p0mvn: **AI review comment (Codex):** `AwaitBlockInfo` can remain pending after its requested parent becomes available through `reconsiderblock`. Sequence: commit P, invalidate P, start `AwaitBlockInfo(P)` and let it settle into this wait, then reconsider P. The successful `NonFinalizedWriteMessage::Reconsider` path publishes the restored state through `update_channels_after_operator_change`, but never signals `block_commit_sender` and skips the ordinary commit notification path. A fresh lookup finds P while the existing waiter stays asleep until an unrelated block commits, potentially indefinitely. This is a liveness defect in the new API; this PR does not yet connect it to subsidy verification. Identified by code inspection, not a new executable reproduction.

Notify `block_commit_sender` after successful reconsideration publishes the restored state. Add a regression that ensures a waiter is pending while P is invalidated, reconsiders P, and requires the waiter to complete without any further block commit. The existing checkpoint-wait and reconsideration tests pass, but do not cover this combination.

p0mvn (approved):

crates/zakura-state/src/service.rs
evan-forbes: Fixed in 3b1791733. A successful reconsideration now wakes waiting `AwaitBlockInfo` requests, and `await_block_info_wakes_after_reconsider` covers your sequence without a further commit. The same commit also makes waiters fail when the state rejects the block, and time out after two minutes.

evan-forbes:
Responses to V12 run [8015](https://v12.sh/runs/8015):

[F-282386](https://v12.sh/runs/8015/282386): Fixed in 3b1791733, which wakes `AwaitBlockInfo` waiters after reconsideration.
[F-282388](https://v12.sh/runs/8015/282388): Fixed in 3b1791733, where the legacy writer republishes the state after finalization.
[F-282390](https://v12.sh/runs/8015/282390): Fixed in 3b1791733, where `AwaitBlockInfo` fails on a rejected block and times out after 2 minutes.

p0mvn (approved):

ValarDragon:
I think we should revert this, if I understand it correctly. We don't want to add any additional latency due to NSM here

PR #1051, opened 2026-09-18
feat(chain): calculate cumulative scheduled issuance

evan-forbes:
## Motivation

ZIP 234 measures reissuance against the halving schedule. zips#1354 defines the NSM balance through 
```
ExpectedIssuedSupply(height): the total block subsidy the schedule issues for blocks 0..=height. 
```
#1052 needs that total twice. Once to set each block's balance change, and once to backfill existing databases in the format 29 migration. Summing every historical subsidy per block would cost O(height). 

This PR adds expected_issued_supply, which sums the schedule piecewise instead. The subsidy is linear through slow start and constant between halving and ZIP 218 target-spacing boundaries, so the sum is exact in a bounded number of steps. 

The PR also extracts halving_block_subsidy, the subsidy without any reissuance bonus, so #1053 can add the bonus on top. scheduled_issuance_zatoshis returns a u128 before the Amount limit applies. Configured schedules can exceed MAX_MONEY, and clamping early would lose value the migration later subtracts. block_subsidy keeps its existing API and behavior.

Chunk 3 of the #859 split. Depends on #1050.

## Solution

Extract the halving subsidy and add cumulative schedule arithmetic. Preserve the existing block_subsidy API and behavior. Keep wide arithmetic available before applying Amount limits.

Chunk 3 of #859. Depends on #1050.

## Testing

Compare cumulative issuance to direct per-block sums across slow start, halvings, and NU7 spacing changes. Check adjacent-height differences at schedule boundaries and the maximum height on Mainnet, Testnet, and Regtest. All 410 chain tests pass on this chunk. The maximum-height case exposed a loop through permanently zero subsidies; cumulative issuance now stops at that boundary.

## Changelog

Added `docs/changelog/unreleased/1051.md`.


crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -460,10 +460,135 @@ pub fn halving(height: Height, network: &Network) -> u32 {
...
+
+        // `rate * h` for h in 1..=min(height, shift - 1).
+        let first_phase_end = height.min(slow_start_shift - 1);
+        total += triangle(first_phase_end) * rate;
+
+        // `rate * (h + 1)` for h in shift..=min(height, interval - 1), which is
+        // `rate * k` for k in shift + 1..=that end + 1.
+        if height >= slow_start_shift {
+            let second_phase_end = height.min(slow_start_interval - 1);
+            total += (triangle(second_phase_end + 1) - triangle(slow_start_shift)) * rate;
```
p0mvn: **AI review comment — P2: Stop slow-start issuance at the halving-divisor cutoff.** `scheduled_issuance_zatoshis()` counts the entire slow-start triangle, but `halving_block_subsidy()` returns zero once the halving index reaches 64, including during slow start. I reproduced this against the compiled PR using the accepted configuration `testnet::Parameters::build().with_halving_interval(1).unwrap().clear_funding_streams().to_network().unwrap()` (default slow-start interval: 20,000). At height 10,064, the actual block subsidy is zero, while cumulative issuance increases by 629,062,500 zatoshi. At height 19,999, cumulative issuance is 12,500,000,000,000 zatoshi versus a direct per-block sum of 3,164,817,500,000. This overstates issuance on supported custom schedules and breaks the accounting helper's stated equivalence to summed subsidies; the default Mainnet/Testnet/Regtest schedules did not exhibit this failure.

Truncate the slow-start sum before the first height where `halving_divisor()` returns `None`, preserving the existing per-block subsidy behavior. Add a regression with the cutoff inside slow start that checks both direct-sum equality and adjacent-height differences. The existing 38 network parameter tests pass but do not cover this configuration.

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -460,10 +460,135 @@ pub fn halving(height: Height, network: &Network) -> u32 {
...
+        (boundary, None) | (None, boundary) => boundary,
+    }
+}
+
 /// `BlockSubsidy(height)` as described in [protocol specification §7.8][7.8]
 ///
 /// [7.8]: https://zips.z.cash/protocol/protocol.pdf#subsidies
 pub fn block_subsidy(height: Height, net: &Network) -> Result<Amount<NonNegative>, SubsidyError> {
+    halving_block_subsidy(height, net)
+}
```
czarcas7ic, line 612: Confirmed we use this wrapper later with different logic in 1053

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -460,10 +460,135 @@ pub fn halving(height: Height, network: &Network) -> u32 {
...
+    let slow_start_shift = u128::from(net.slow_start_shift().0);
+    let slow_start_interval = u128::from(net.slow_start_interval().0);
+    let height = u128::from(height.0);
+    let mut total: u128 = 0;
+
+    // The slow start issues `rate * h` below the shift and `rate * (h + 1)` from the shift
+    // up to the interval, so each phase is a triangular number rather than a rectangle.
+    if slow_start_interval > 0 && slow_start_shift > 0 {
+        let rate = u128::from(MAX_BLOCK_SUBSIDY) / slow_start_interval;
+        let triangle = |n: u128| n * (n + 1) / 2;
```
p0mvn: What does triangle mean?

p0mvn: I don't think we should call it a triangle. Perhaps `sum_from_one_through`?

p0mvn (approved):
LGTM. Pending comments

czarcas7ic (approved):
Confirmed the v12 finding is the same as what Roman's AI posted

evan-forbes:
Responses to V12 run [8016](https://v12.sh/runs/8016):

[F-282446](https://v12.sh/runs/8016/282446): Fixed in 4a5f96fbb, where slow-start issuance stops where the halving divisor overflows.

crates/zakura-chain/src/parameters/network/subsidy.rs
evan-forbes: Fixed in 4a5f96fbb. Slow-start issuance now stops before the halving divisor overflows. The regression uses your configuration and checks cumulative issuance against a direct sum around the cutoff and through the end of slow start.

crates/zakura-chain/src/parameters/network/subsidy.rs
evan-forbes: It computes `1 + … + n = n * (n + 1) / 2`. Renamed it to `sum_from_one_through` and updated the comment in 4a5f96fbb.

crates/zakura-chain/src/parameters/network/subsidy.rs
evan-forbes: Correct. #1053 adds the reissuance bonus in `block_subsidy`; `halving_block_subsidy` continues to return only the scheduled subsidy.

p0mvn (approved):

PR #1052, opened 2026-09-18
feat(state): persist signed issuance deficits and migrate existing databases

evan-forbes:
## Motivation

NSM reissuance needs a persisted issuance-deficit counter that follows commits, forks, finalization, rollback, and restart. Existing databases must reconstruct the same counter as fresh sync.

## Solution

Track signed scheduled issuance minus monetary pool changes from NU7. Store the counter alongside chain value pools, excluding it from monetary totals and excluding the pre-NU7 baseline. Keep runtime accounting, expanded records, format 29, and retryable backfill together. The migration validates Deferred changes from the NU7 baseline and skips backfill when NU7 is unscheduled or has not activated.

This PR neither pays bonuses nor rejects negative deficits; #1053 adds those rules. Historical seeding remains a separate policy decision. Chunk 4 of #859, building on #1051 and the replay/rollback corrections in #1049.

General cache discovery, multi-major reuse, startup locks, retained-source checkpoints, and ephemeral version-path fixes are extracted into the follow-up #1075, which targets this branch. This PR keeps the existing general startup infrastructure.

### V12 findings split between PRs

From the [V12 audit comment](https://github.com/zakura-core/zakura/pull/1052#issuecomment-5742334680), these general database findings and their fixes moved to **#1075**:

| Finding | Follow-up scope |
| --- | --- |
| [F-282735](https://v12.sh/runs/8055/282735): newest cache masks reusable state | Validate eligibility before selection, skip malformed candidates, and report unusable existing caches. The proposed ranking by chain height was not adopted: newest eligible major remains the policy. |
| [F-282737](https://v12.sh/runs/8055/282737): ephemeral probes leak directories | Skip ephemeral discovery and use the open database's actual path for version IO. |
| [F-282739](https://v12.sh/runs/8055/282739): cleanup race deletes sibling cache | Retain the source and remove shared-parent recursive cleanup from reuse. |
| [F-282740](https://v12.sh/runs/8055/282740): reuse renames before locking | Acquire the source's native RocksDB lock and publish a checkpoint under a startup lock. |

The NSM-specific findings remain addressed here: [F-282727](https://v12.sh/runs/8055/282727) (healthy Deferred histories rejected), [F-282728](https://v12.sh/runs/8055/282728) (unnecessary pre-NU7 scans), and [F-282731](https://v12.sh/runs/8055/282731) (slow-start endpoint rejection). The exact endpoint regression test stays in #1052.

The earlier general multi-major cache-reuse/version-preservation changes also moved to #1075. This split moves fixes; it does not assert that every original audit claim was accepted unchanged.

## Testing

Tests cover signed accounting, monetary-total exclusion, activation and genesis boundaries, forks, finalization, rollback, replay, reopen, legacy records, cancellation/retry, malformed or missing rows, invalid totals, Deferred boundaries, and tip agreement.

Post-split verification on this branch:

- `cargo +1.97.1 test --locked --features nu7 -p zakura-consensus -p zakura-state --lib --no-fail-fast`: 252 consensus tests passed (1 ignored), 645 state tests passed (4 ignored).
- `cargo +1.97.1 clippy -p zakura-state --all-targets --locked -- -D warnings`: passed.
- Formatting, Markdown lint, changelog validation, and whitespace checks passed.
- Independent testing caught that the deliberately corrupt history-snapshot test needs cleanup before format-29 shutdown validation. That existing cleanup stays in #1052; the full corrected suite passes.
- CI on `1585dc56b`: 36 successful checks, 16 skipped, no failures or pending checks.

Checkpoint fixtures isolate accounting and do not establish production NU7 transaction verification.

## Changelog

`docs/changelog/unreleased/1052.md` contains only issuance-accounting and format-29 migration entries. General cache-reuse entries belong to #1075.

### Follow-up Work

- #1075: general database cache reuse and startup hardening.
- #1053: consensus rules for reissuance.


crates/zakura-state/src/service/finalized_state/disk_format/upgrade.rs
```diff
@@ -144,7 +145,8 @@ fn format_upgrades(
         )),
         Box::new(drop_header_root_auth_frontier::Upgrade),
         Box::new(unauthenticated_commitment_roots::Upgrade),
-    ] as [Box<dyn DiskFormatUpgrade>; 10])
+        Box::new(issuance_deficit_pool::Upgrade),
```
czarcas7ic, line 148: AI comment, checked against Fable + manual review

---

Pre-existing, but this bump is the first time the v28 to v29 reuse path runs. `try_reusing_previous_db_after_major_upgrade` rewrites the moved directory's recorded major to the previous one:
https://github.com/zakura-core/zakura/blob/fe61e7805b76509fff5ab99a3cf4739245e7b55f/crates/zakura-state/src/service/finalized_state/disk_db.rs#L1529-L1547

A v28 directory can still record `27.3.0` after an interrupted v27 to v28 upgrade. Reuse turns that into `28.3.0`, which skips `28.0.0`, `28.1.4` and `28.1.5`. The v29 backfill then runs and `29.0.0` is written before the final validity check fails on the missing genesis Ironwood tree, so a cache that a v28 restart would have resumed now needs a resync.

Same issue as V12's F-282403. Fine as a follow-up issue rather than a change here; the fix is small (only rewrite `major` for a legacy minor.patch version file, or read the version from the old path before the rename).

crates/zakura-state/src/constants.rs
```diff
@@ -95,7 +95,7 @@ pub const MAX_PRUNE_HEIGHTS_PER_COMMIT: u32 = 100;
 ///
 /// Instead of using this constant directly, use [`constants::state_database_format_version_in_code()`]
 /// or [`config::database_format_version_on_disk()`] to get the full semantic format version.
-const DATABASE_FORMAT_VERSION: u64 = 28;
+const DATABASE_FORMAT_VERSION: u64 = 29;
```
czarcas7ic: AI comment

---

A node upgrading directly from v27 will start a fresh v29 database because cache reuse only probes the immediately preceding major version, now v28.
https://github.com/zakura-core/zakura/blob/fe61e7805b76509fff5ab99a3cf4739245e7b55f/crates/zakura-state/src/service/finalized_state/disk_db.rs#L1468-L1473

The registered migrations still cover v27 through v29. Can we locate and migrate the newest supported existing cache instead of requiring an intermediate binary to avoid another sync?

crates/zakura-state/src/service/finalized_state/disk_format/upgrade/issuance_deficit_pool.rs
```diff
@@ -0,0 +1,567 @@
...
+        let Some(tip_height) = db.finalized_tip_height() else {
+            return Ok(Ok(()));
+        };
+
+        let network = db.network();
+        let tip_pools = read_tip_pools(db)?;
+
+        check_cancelled(cancel_receiver)?;
+        let tip_info = read_block_info(db, tip_height)?;
+        if *tip_info.value_pools() != tip_pools {
```
czarcas7ic, line 69: AI comment, checked against Fable + manual review

---

With debug_validity_check_interval enabled, a block can commit between reading tip_height and reading tip_pools. That compares the new tip's pools with the previous tip's BlockInfo and reports a healthy database as inconsistent. The periodic-check path turns that result into a panic.
https://github.com/zakura-core/zakura/blob/fe61e7805b76509fff5ab99a3cf4739245e7b55f/crates/zakura-state/src/service/finalized_state/disk_format/upgrade.rs#L618-L626

Could we read the height, pools, and block metadata from one consistent database snapshot?

evan-forbes (commented):
Review navigation: the consensus-critical accounting and migration are isolated in [0d75ab866](https://github.com/zakura-core/zakura/commit/0d75ab866675afa719dac1623230b244593a3d66). The surrounding commits are [288f1162c](https://github.com/zakura-core/zakura/commit/288f1162c74a80742ad02493028ea62ce3f3501e) (database reuse across major versions, with its tests) and [b6cdfc34f](https://github.com/zakura-core/zakura/commit/b6cdfc34f0eb76b1afb72d01eb9ecf076240aa71) (documentation). The final tree is identical to the previous PR head.

The zero historical balance is an intermediate stack boundary: [#1040](https://github.com/zakura-core/zakura/pull/1040) supplies the historical seed, spacing-dependent payout fraction, and rejection from NU7. [#1053](https://github.com/zakura-core/zakura/pull/1053) adds parent-dependent reissuance validation; [#1054](https://github.com/zakura-core/zakura/pull/1054) supplies mining/RPC integration; [#1056](https://github.com/zakura-core/zakura/pull/1056) supplies fee recycling. The derived reissuance crossing height remains an unresolved deployment-specification difference after #1040.

#1053 is restacked onto this history. Approved #1054 retains its existing history; restacking it requires separate coordination with its reviewers.

Validation: Rust 1.97 passed 19 value-balance tests and 20 issuance accounting/migration tests. The infrastructure commit passed all six database-reuse tests independently. Formatting, changelog, and diff checks passed.

crates/zakura-chain/src/block.rs
```diff
@@ -333,11 +340,51 @@ impl Block {
...
+        } else {
+            halving_block_subsidy(height, network)
+                .map_err(ValueBalanceError::ScheduledIssuance)?
+                .constrain::<NegativeAllowed>()
+                .map_err(ValueBalanceError::IssuanceDeficit)?
+        };
+
+        let issued = change.total().map_err(ValueBalanceError::IssuanceDeficit)?;
+
+        (scheduled - issued).map_err(ValueBalanceError::IssuanceDeficit)
```
evan-forbes, line 387: Consensus arithmetic starts here: from NU7, `delta_D = scheduled_halving_subsidy - delta_monetary_pools`, then state accumulation adds that signed delta to the parent balance. Genesis contributes zero scheduled issuance. Before NU7 the delta is zero. Historical funds are deliberately deferred to [#1040](https://github.com/zakura-core/zakura/pull/1040); zero here does not describe the final stack policy.

crates/zakura-chain/src/value_balance.rs
```diff
@@ -157,13 +177,17 @@ where
...
     /// Returns the sum of all value pool balances.
     ///
     /// Returns an error if the final sum does not satisfy the amount constraint `C`.
     /// Signed balances are summed before applying that constraint.
+    ///
+    /// [`ValueBalance::issuance_deficit`] is excluded: it holds value that is in no pool,
+    /// so adding it would overstate the monetary base this sum is used to bound.
     pub fn total(self) -> Result<Amount<C>, amount::Error> {
         let total: i128 = [
             self.transparent,
```
evan-forbes, line 193: Review the accounting boundary here. The signed deficit is a bookkeeping counter, so `total()` excludes it from the six monetary pools. Including it would overstate issued value and also feed the deficit back into its own update equation. Serialization, addition, subtraction, and constraint conversion must preserve the signed counter without treating it as spendable value.

crates/zakura-state/src/service/finalized_state/tests/issuance_accounting.rs
```diff
@@ -0,0 +1,584 @@
...
+        commit(&mut fresh, &fork).unwrap();
+        proptest::prop_assert_eq!(state.db.finalized_value_pool(), fresh.db.finalized_value_pool());
+    }
+}
+
+proptest::proptest! {
+    #![proptest_config(proptest::test_runner::Config::with_cases(64))]
+
+    #[test]
+    fn non_finalized_forks_keep_independent_deficits(shortfall in 1i64..1_000_000) {
```
evan-forbes, line 368: This property locates the fork-specific accounting requirement: each non-finalized fork accumulates its own balance from its parent. Review it with `deficit_rollback_replay_and_fork_equivalence` above. Neither a best-tip lookup nor a global counter may substitute for the selected parent state.

crates/zakura-state/src/service/finalized_state/disk_format/upgrade/issuance_deficit_pool.rs
```diff
@@ -0,0 +1,661 @@
...
+    if !NetworkUpgrade::Nu7
+        .activation_height(network)
+        .is_some_and(|start| height >= start)
+    {
+        return Ok(Amount::zero());
+    }
+    let eligible = deficit_at(network, height, pools)
+        .and_then(|raw| {
+            let eligible = i64::try_from(raw - baseline)
+                .map_err(|_| zakura_chain::parameters::subsidy::SubsidyError::Overflow)?;
```
evan-forbes, line 256: Migration must reproduce runtime accounting: subtract the pre-NU7 baseline from `scheduled_issuance(height) - monetary_pools(height)`. Perform the subtraction before applying the signed amount bound; do not clamp the operands or a negative result. The migration rewrites per-block records and the separate tip balance, then validates their agreement. [#1040](https://github.com/zakura-core/zakura/pull/1040) changes both this baseline and the runtime rule together to insert the historical seed.

crates/zakura-state/src/service/finalized_state/disk_format/upgrade/issuance_deficit_pool.rs
```diff
@@ -0,0 +1,661 @@
+//! Backfill the unseeded reissuance balance from NU7 onward.
+//!
+//! The balance equals the signed schedule deficit minus its value immediately before
+//! NU7. Earlier records carry zero. Historical funds remain excluded pending guidance.
```
ValarDragon, line 4: IMO we shouldn't do this, and just do it as a DB write at the relevant height onwards. But I'm ok fixing that in a follow-up, pre-release

crates/zakura-state/src/service/finalized_state/disk_format/upgrade.rs
czarcas7ic: https://github.com/zakura-core/zakura/commit/90b4b4192eecbeea76273dbbb5a52af933546f80

crates/zakura-state/src/service/finalized_state/disk_format/upgrade/issuance_deficit_pool.rs
czarcas7ic: https://github.com/zakura-core/zakura/commit/086a15be07711d699d0946df14f3f5e9f6a99202

crates/zakura-chain/src/value_balance.rs
```diff
@@ -157,13 +177,17 @@ where
             orchard: zero,
             deferred: zero,
             ironwood: zero,
+            issuance_deficit: Amount::zero(),
```
p0mvn, line 180: nit: Why not reuse the `zero` var? Wondering if there is a perf benefit we are leaving

ValarDragon:
@v12sec pls review

p0mvn:
Flagging that there is a consensus follow-up here:

INITIAL_NSM_VALUE_BALANCE is set to 0, but it is 36,858,445,520 zat on Mainnet. I see the issue is fixed in #1040

p0mvn:
Implemented the remaining fixes in [97666526c](https://github.com/zakura-core/zakura/commit/97666526cd2a743c5e32bb6d6ed13d697a0ec550), then completed a read-only review of that commit against `311d822`.

**Review verdict: approve for this change. No blocking regression found.** This is a review of the cache-reuse and ephemeral-path changes, not a new approval of the PR's broader consensus policy.

### Audit disposition

| Finding | Result |
| --- | --- |
| F-282727 — healthy Deferred databases rejected | Already fixed by `618759f` / `311d822`. The blanket refusal is gone; healthy histories and inactive-NU7 startup remain covered. |
| F-282728 — unnecessary slow-start scan | Already fixed. The large-tip/no-history regression passes; no additional scan or optimization was introduced. |
| F-282731 — slow-start endpoint | The faulty guard was already removed. Added the exact boundary fixture: the first Deferred payment occurs at the slow-start endpoint, with NU7 active, future, or unscheduled. |
| F-282735 — candidate selection | Fixed malformed/ineligible-candidate fallback and explicit failure when no existing eligible cache is usable. **Newest-major preference is intentionally retained**, as agreed; caches are not ranked by height. |
| F-282737 — ephemeral probe directories | Fixed. Ephemeral startup skips persistent discovery; live format-version I/O uses the open database's actual path. An isolated subprocess verifies one directory per database and cleanup on drop. |
| F-282739 — sibling deletion race | Removed the reuse path's old-parent cleanup. The source and sibling network caches remain in place. |
| F-282740 — rename before locking | Replaced source renaming with an exclusively opened source and staged RocksDB checkpoint. A per-network startup lock coordinates writable opens and publication. The checkpoint contains the full source version before it becomes the current database. |

### Regression review

Inspected discovery, version inference, source locking, publication, startup-error propagation, migration retry, read-only behavior, and ephemeral shutdown. Failure tests cover a live source in another process, malformed markers, unsupported upgrade chains, competing startup, checkpoint/version/publication failures, and an externally created destination. They preserve the source or return an explicit error without silently creating fresh state.

The checkpoint test verifies all column families survive, SST hard links are used on the local filesystem, and subsequent source/destination writes are independent. The existing full startup-failure regression now also verifies that repairing and upgrading the destination leaves the retained source's version and legacy records unchanged.

### Validation

On macOS, using Rust 1.97.1:

- `cargo test --locked --features nu7 -p zakura-consensus -p zakura-state --lib --no-fail-fast`: **661 state tests + 252 consensus tests passed**; five pre-existing ignored tests remained ignored.
- `cargo check --locked -p zakura --features nu7`: passed.
- `cargo clippy -p zakura-state --all-targets --locked -- -D warnings`: passed.
- Rust formatting, Markdown lint, changelog validation, and diff whitespace checks: passed.

[Linux CI](https://github.com/zakura-core/zakura/actions/runs/35454866633) also passed all three unit-test shards, NU7 tests, config compatibility, and network-dependent acceptance tests. The other completed PR checks are green; the runtime Docker-image job is still running at the time of this comment.

The initial verification attempt hit local disk exhaustion. Disposable incremental-build artifacts were cleared and the CI-equivalent suite was rerun successfully. No node databases were removed.

Source-cache retention and its disk-space cost are documented. Failure injection and restart tests passed; abrupt power loss was not simulated. No code changes were needed during the final review.

p0mvn:
@v12sec review

p0mvn:
The general migration changes have been extracted into the stacked follow-up **#1075**, targeting #1052. The PR description now records this mapping.

From the [V12 audit comment](https://github.com/zakura-core/zakura/pull/1052#issuecomment-5742334680), these general database findings and their fixes moved to **#1075**:

| Finding | Follow-up scope |
| --- | --- |
| [F-282735](https://v12.sh/runs/8055/282735): newest cache masks reusable state | Validate eligibility before selection, skip malformed candidates, and report unusable existing caches. The proposed ranking by chain height was not adopted: newest eligible major remains the policy. |
| [F-282737](https://v12.sh/runs/8055/282737): ephemeral probes leak directories | Skip ephemeral discovery and use the open database's actual path for version IO. |
| [F-282739](https://v12.sh/runs/8055/282739): cleanup race deletes sibling cache | Retain the source and remove shared-parent recursive cleanup from reuse. |
| [F-282740](https://v12.sh/runs/8055/282740): reuse renames before locking | Acquire the source's native RocksDB lock and publish a checkpoint under a startup lock. |

The NSM-specific findings remain addressed here: [F-282727](https://v12.sh/runs/8055/282727) (healthy Deferred histories rejected), [F-282728](https://v12.sh/runs/8055/282728) (unnecessary pre-NU7 scans), and [F-282731](https://v12.sh/runs/8055/282731) (slow-start endpoint rejection). The exact endpoint regression test stays in #1052.

The earlier general multi-major cache-reuse/version-preservation changes also moved to #1075. This split moves fixes; it does not assert that every original audit claim was accepted unchanged.

ValarDragon (approved):
Merging, a bunch of this DB issues are esoteric / not that applicable, we can do more followups as separate PR's. Like fixing restarts during the sub-ms migration

PR #1053, opened 2026-09-18
feat(consensus): enforce parent-dependent NU7 reissuance rules

evan-forbes:
> Replacement draft stack: #1070 → #1071 → #1072 → #1073 → #1074, based on #1052.
> These PRs split seeded accounting, reward arithmetic, start-height policy,
> verification/recovery, and mining/RPC integration. The final source matches
> this PR at `b73630b2d` combined with #1052's newer `618759f4c` fix.
> All five replacement drafts have green CI. This PR remains open as the original
> reference; use the replacement stack for focused review.

splitting up, I'll defer to others if we want to break this up, however all of the changes here are directly related to reissuance.

- Reissuance arithmetic, start-height calculation, and configuration overrides.
- Parent-state lookup in semantic and checkpoint verification.
- Negative-deficit rejection in both commit paths.
- Rollback, replay, and migration changes. These ensure stored balances remain consistent with the new rules.
- Activation, rounding, fork, and recovery tests.
- The two RPC call-site adaptations. These accommodate the changed function signature; the actual RPC/mining integration already lives in #1054 (https://github.com/zakura-core/zakura/pull/1054).

## Motivation

NU7 reissuance changes the permitted coinbase reward and rejects blocks that overdraw the eligible issuance deficit. Each calculation must use the actual parent's state, including side forks.

This PR turns on ZIP 234 reissuance as a consensus rule. From the reissuance start height, each block may claim its halving subsidy plus
 
```
ceil(NSMValueBalance(parent) × 4126 / 10¹⁰)
```

The balance comes from the actual parent's pools (#1052), which the semantic and checkpoint verifiers await through #1050's AwaitBlockInfo, so side forks and pipelined commits get the correct value. The start height follows the 2026-09-14 coinholder poll (Q2: February 2031). No ZIP defines that date. Zakura derives it by applying ZIP 234's own start rule after the third halving instead of the second. On the 75-second schedule this gives Mainnet height 5,342,746 (about 2031-02-14). 

The height then moves to the matching point on the ZIP 218 25-second schedule. Both commit paths and the migration reject a block that leaves the balance negative from the start height. Configured testnets and Regtest can override the start height. Every rule activates by height alone, in every build; this PR adds no build feature. NU7 has no production branch ID, so the semantic test mocks transaction verification. These tests don't cover ZIP 233 transactions, and NU7 ships no ZIP 233 or ZIP 235 value removal.

Chunk 5 of the #859 split. Depends on #1052. #1054 wires the same rules into mining and the RPCs; don't activate reissuance without it. #1040 changes three rules here after the 2026-09-15 ZIP Editor call. It derives the fraction as floor(LN2_SCALED / HalvingInterval), because the flat 4126 is calibrated for 75-second blocks and pays out three times too fast at ZIP 218 spacing. It rejects a negative balance from NU7 instead of from the reissuance start. It renames the start height to zips#1363's nsm_reissuance_height.

## Solution

Add the rounded reissuance bonus, crossing-height policy, and configured start overrides. Semantic and checkpoint verification await the actual parent. Both commit paths reject negative deficits from the reissuance start. Migration applies the same rejection boundary. Rollback reconstructs deferred funding from each parent's pools.

Chunk 5 of #859. Depends on #1052. The mining/RPC integration follows in chunk 6; this PR only adapts its callers to compile. Do not activate reissuance without that integration.

## Testing

Add a 100-case contextual activation/sign/arithmetic-limit matrix and a subsidy matrix spanning activation and rounding boundaries. Semantic verification checks exact claims and one-zatoshi overclaims across multiple deficits. Retain active-bonus histories for fork independence, finalization, rollback, replay, and migration. The default-build suites pass: 419 chain tests, 238 consensus tests, 284 header-chain tests, 143 RPC tests, and 633 state tests. Extended validation uses 10,000 arithmetic cases and 256 histories per rollback/replay property. Mutations to rounding, parent lookup, and negative-deficit rejection all trigger failures.

The semantic test mocks transaction verification because NU7 lacks a production branch ID. These tests do not establish ZIP 233 transaction support.

## Changelog

Added `docs/changelog/unreleased/1053.md`.

### Follow-up Work

Confirm historical seeding, the later rejection height, and the fraction across spacing changes. Production activation requires the branch ID, activation heights, independent accounting review, historical balance verification, representative migration measurements, and a private-network test with real transactions.


evan-forbes:
Responses to V12 run [8021](https://v12.sh/runs/8021):

[F-282405](https://v12.sh/runs/8021/282405): Fixed in #1054, which replaces these RPC adapters.
[F-282406](https://v12.sh/runs/8021/282406): Fixed in 525578137, where the migration refuses a database whose tip is at or above the ZIP 234 start.
[F-282407](https://v12.sh/runs/8021/282407): Fixed in 3b1791733 on #1050, where `AwaitBlockInfo` fails on a rejected parent and times out after 2 minutes.
[F-282408](https://v12.sh/runs/8021/282408): Fixed in 3b1791733 on #1050, where `AwaitBlockInfo` fails on a rejected parent and times out after 2 minutes.

czarcas7ic:
Did not get time to complete this one as its getting late. I am going to leave the Fable review that just finished up that I was going to start sifting through. Feel free to ignore or use if helpful

<details>
  <summary>Fable review</summary>

Checkpoint reset gap, on `checkpoint.rs`:

```
The new `deferred_pool_balance_change(..)?` here runs after `queued_result.result?` has already advanced verifier progress, but its `SubsidyError`/`AmountError` is not mapped to `CommitCheckpointVerified`, so the reset guard below never fires for it and the error scores the advertiser 100 for a local-state condition. The `AwaitBlockInfo` error a few lines up is mapped. The trigger is a negative deficit at `start - 1`, which `issuance_deficit_is_non_negative` does not cover, so only configured networks with over-claiming checkpointed history reach it, but the guard invariant is worth keeping. Same `map_err` as the wait.

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-consensus/src/checkpoint.rs#L1315-L1319

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-consensus/src/checkpoint.rs#L1362
```

Negative deficit, on `block.rs`:

```
`.constrain().ok()` turns a negative parent deficit into `None`, so `block_subsidy` reports `MissingIssuanceDeficit` for a value that is present, and the new `SubsidyError::NegativeIssuanceDeficit` is never constructed. The comment above assumes `issuance_deficit_is_non_negative` covered the parent, but that check exempts heights below the start, so the parent of the first active block is unchecked. Same pattern in `checkpoint.rs`, `rollback.rs` and the v27 replay, and PR 1054 adds three more. Suggest `block_subsidy` take `Option<Amount<NegativeAllowed>>`, constrain inside, and return the existing variant; the `verify(-1, ..)` test can then assert `NegativeIssuanceDeficit`.

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-consensus/src/block.rs#L471-L479

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-chain/src/parameters/network/subsidy.rs#L391-L394

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-consensus/src/block/tests.rs#L1771-L1779
```

Config test, on `config.rs`:

```
The sibling fields have serialization round-trip tests (`max_block_time_start_height_serialization_roundtrip`), but nothing sets `zip234_start_height` in a config and reads it back, so dropping it from this `From` impl or either deserializer arm would pass CI and a configured start would silently revert to the crossing rule. A copy of the sibling test covers it. While here, the regtest arm builds `Height` without the `try_into` bounds check the testnet arm uses.

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-network/src/config.rs#L1103-L1105

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-network/src/config/tests/vectors.rs#L791
```

Docs, on `issuance_deficit_pool.rs`:

```
This refusal ends startup with a panic on every restart until the directory is deleted, but `reissuance-accounting.md` still says restarting recomputes every balance and not to move the database back, and the changelog mentions neither the refusal nor the new `zip234_start_height` key. In this case no record was rewritten (the rename runs before the refusal), so moving back is actually the cheap recovery. Also `{:?}` prints `Some(Height(n))` in the operator message, and the new `[`BlockInfo`]` doc link in `request.rs` does not resolve under `-D warnings`.

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-state/src/service/finalized_state/disk_format/upgrade/issuance_deficit_pool.rs#L157-L163

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-state/src/request.rs#L1444
```

Follow-ups, as one PR-level comment:

```
Non-blocking, for follow-up issues before any network activates reissuance:

1. The parent wait sits before the transaction fan-out, so at active heights the syncer's 60 concurrent verifications serialise to one, and the deepest waiter can hit the 2 minute `AwaitBlockInfo` timeout on slow hosts. Nothing in the fan-out needs the subsidy, so the wait can be joined with the drain and `subsidy_is_valid` moved next to `miner_fees_are_valid`.

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-consensus/src/block.rs#L446-L452

2. Checkpoint commit tasks now wait per block, so the write queue gets one block per round trip, and the watch in `AwaitBlockInfo` wakes every parked waiter on every commit (N^2/2 reads, N up to 3000 with the legacy syncer, about 40 ms per commit inside the timeout). The state already has the parent pools in `prepare_chain_value_pools_batch`, so the deferred change could be computed there with no wait.

https://github.com/zakura-core/zakura/blob/6cb03dd17f42e9981415f30cd048b7e2e7c3beaf/crates/zakura-consensus/src/checkpoint.rs#L1298-L1323

3. Tests: the rollback and replay tests pass with the parent lookup pointed at the block's own pools (regtest has no Deferred stream); `issuance_deficit.rs` duplicates 484 lines and six test names from `issuance_accounting.rs`, so the extended CI filter runs the property twice; the ceil oracle literal is hand-written nine times and PR 1040 changes the fraction.

4. The v29 refusal is keyed on the running binary's `nu7` feature, so a non-nu7 sync past the start followed by an nu7 binary passes `validate`; stamping the feature into the version build metadata like `indexer` would catch it.
```
</details>

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -461,6 +469,319 @@ pub fn halving(height: Height, network: &Network) -> u32 {
...
+}
+
+/// Applies the [ZIP 234] reissuance fraction to `amount`, rounding up.
+///
+/// [ZIP 234]: https://zips.z.cash/zip-0234
+fn reissuance_amount(amount: Amount<NonNegative>) -> Result<Amount<NonNegative>, SubsidyError> {
+    let subsidy = amount_to_u128(amount)
+        .checked_mul(BLOCK_SUBSIDY_FRACTION_NUMERATOR)
+        .ok_or(SubsidyError::Overflow)?
+        .div_ceil(BLOCK_SUBSIDY_FRACTION_DENOMINATOR);
```
evan-forbes, line 815: Consensus arithmetic navigation (1/4): the reissuance amount.

This computes `ceil(D(parent) * 4126 / 10^10)` with integer arithmetic. Review the fraction, upward rounding, and checked conversions here. [block_subsidy](https://github.com/zakura-core/zakura/blob/7c1557d7a0f611b7e0c1fab14bd388b7f4bd19cb/crates/zakura-chain/src/parameters/network/subsidy.rs#L932-L954) adds this bonus to the halving subsidy once reissuance activates; it requires the caller to supply the parent deficit.

The fraction is fixed per block in this PR, including after the change to 25-second spacing. The PR description identifies #1040 as the follow-up that changes the fraction to depend on the halving interval. Review the constants in this revision with that distinction in mind.

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -461,6 +469,319 @@ pub fn halving(height: Height, network: &Network) -> u32 {
...
+                return None;
+            }
+
+            // `ceil(fraction * reserve) < subsidy` holds when `reserve <= max_reserve`.
+            let max_reserve = (subsidy - 1) * BLOCK_SUBSIDY_FRACTION_DENOMINATOR
+                / BLOCK_SUBSIDY_FRACTION_NUMERATOR;
+            let reserve =
+                max_money.saturating_sub(supply + u128::from(candidate - height) * subsidy);
+            let crossing =
+                u128::from(candidate) + reserve.saturating_sub(max_reserve).div_ceil(subsidy);
```
evan-forbes, line 648: Consensus arithmetic navigation (2/4): the activation crossing.

This finds the first eligible height where `ceil(fraction * (MAX_MONEY - ScheduledSupply(h - 1))) < halving_subsidy(h)`. The `subsidy - 1` converts the strict inequality into an integer reserve threshold. The ceiling division then finds the first height that satisfies it within a constant-subsidy run.

Review the parent-height supply, strict inequality, and run boundaries together. [ZIP234_START_HALVING](https://github.com/zakura-core/zakura/blob/7c1557d7a0f611b7e0c1fab14bd388b7f4bd19cb/crates/zakura-chain/src/parameters/network/subsidy.rs#L485-L491) selects the third halving. [Pre218Schedule](https://github.com/zakura-core/zakura/blob/7c1557d7a0f611b7e0c1fab14bd388b7f4bd19cb/crates/zakura-chain/src/parameters/network/subsidy.rs#L629-L733) evaluates the crossing on the pre-ZIP-218 schedule and maps it to the real schedule; for a crossing after NU7, the mapping is `N + ratio * (H - N)`. [zip234_start_height](https://github.com/zakura-core/zakura/blob/7c1557d7a0f611b7e0c1fab14bd388b7f4bd19cb/crates/zakura-chain/src/parameters/network/subsidy.rs#L528-L545) applies configured overrides and clamps the result to NU7 activation. These choices determine the consensus activation boundary.

crates/zakura-consensus/src/block.rs
```diff
@@ -494,8 +494,53 @@ where
...
+                        // AwaitBlockInfo only returns after the parent block commits.
+                        return Err(VerifyBlockError::ValidateProposal(
+                            format!("proposal parent {parent_hash} has not committed").into(),
+                        ));
+                    };
+
+                    // `issuance_deficit_is_non_negative` exempts heights below the start,
+                    // so the parent of the first active block can have a negative deficit.
+                    Some(zakura_chain::parameters::subsidy::parent_issuance_deficit(
+                        parent_info.value_pools().issuance_deficit_amount(),
```
evan-forbes: Consensus arithmetic navigation (3/4): the source of D(parent).

The bonus uses the issuance deficit from the block's actual parent hash. Review this lookup together with the formula in `reissuance_amount`; substituting the best-chain tip would change the permitted subsidy on side forks. `parent_issuance_deficit` rejects a negative parent deficit instead of clamping it to zero.

The checkpoint verifier also obtains the actual parent's deficit before calculating Deferred funding. Rollback and the format-27 replay adapt their Deferred calculations to the parent-dependent subsidy. The underlying signed-deficit accounting and historical-baseline policy come from #1052.

crates/zakura-state/src/service/check.rs
```diff
@@ -43,6 +45,54 @@ mod tests;
...
+
+    let deficit_after = (deficit_before + deficit_change).map_err(|_| {
+        ValidateContextError::NegativeIssuanceDeficit {
+            height,
+            deficit_before,
+            deficit_change,
+        }
+    })?;
+
+    if deficit_after.zatoshis() < 0 {
```
evan-forbes: Consensus arithmetic navigation (4/4): the post-block deficit constraint.

This checks `D(after) = D(before) + delta_D(block)` and rejects a negative result. Both finalized and non-finalized commits call this check. Review it alongside the subsidy calculation: the permitted reward uses the parent deficit, while this constraint checks the resulting deficit after applying the block.

This revision activates the constraint at the reissuance start height and permits signed deficits before that height. The PR description identifies #1040 as the follow-up that moves rejection to NU7. The format-29 migration also adds the activation-dependent constraint and refuses histories already past the reissuance start.

evan-forbes (commented):
Review navigation: the four inline comments mark the bonus formula, activation arithmetic, parent-deficit input, and post-block deficit constraint. These comments identify review entry points; they do not assert that the arithmetic or policy has been independently validated.

evan-forbes:
The branch now has five commits on the current #1052 base. Review the consensus rules, their tests, and recovery handling together in 2c8cb5aa9. The surrounding commits contain dependency version requirements (530da4c6f), the non-waiting state lookup API (be82e4034), extended CI (ce08a3406), and documentation (2e460cb99).

This rewrite preserves the previous head tree exactly. Formatting, changelog validation, and whitespace checks pass. The earlier arithmetic navigation comments still describe the same code.

evan-forbes:
Consensus review note: this PR implements the core formula from Shielded Labs’ halving-preserving proposal, but reviewers should distinguish this PR in isolation from the complete stack.

Compared with the [current halving-preserving draft](https://github.com/ShieldedLabs/zips/blob/60720df9e971e19d8b6f67f1869426d78c96d250/zips/draft-judah-nsm-halving-preserving-issuance.md), #1053 and its base have four material differences:

| Rule | Behavior through #1053 | Follow-up in the stack |
| --- | --- | --- |
| Payout fraction | Uses a fixed numerator of 4126 over 10^10, including at 25-second spacing. The draft derives the fraction from the halving interval, giving 1375 over 10^10 at 25 seconds. | #1040 derives the fraction from the halving interval. The fixed fraction here is an intermediate state of the stack. |
| Historical balance | Starts at zero and excludes historical unclaimed subsidy and fees. The draft seeds the balance with pre-NU6 unclaimed subsidy and fees. | Historical seeding is deliberately handled in #1040. Starting at zero here does not mean the complete stack excludes those funds. #1040 adds the seed parameters and updates runtime accounting, migration, and tests. |
| Negative-balance rejection | Rejects negative balances from the reissuance start height. The draft requires rejection from NU7. | #1040 moves the rejection boundary to NU7. |
| Reissuance start | Derives a crossing after the third halving and maps it through the ZIP 218 spacing change. The draft takes an assigned height from the deployment ZIP. | #1040 renames the parameter to nsm_reissuance_height but retains the derived crossing. This remains an unresolved divergence after that follow-up. |

**The first three differences are addressed by #1040**, stacked after the mining/RPC integration in #1054. Reviewers should assess those rules across the stack. The derived reissuance start height still needs resolution; it is not specified by the current halving-preserving draft.

cc @ValarDragon on the reissiance start

evan-forbes:
Restacked this PR onto the isolated #1052 history. The final tree is unchanged. The current review order is:

- [a00d83ef5](https://github.com/zakura-core/zakura/commit/a00d83ef5): internal dependency requirements.
- [2e65046f6](https://github.com/zakura-core/zakura/commit/2e65046f6): non-waiting committed block lookup.
- [1c497a350](https://github.com/zakura-core/zakura/commit/1c497a350): consensus-critical reissuance rules, migration/recovery, caller adaptations, and tests.
- [9b2e254e7](https://github.com/zakura-core/zakura/commit/9b2e254e7): extended CI properties.
- [5feb144c3](https://github.com/zakura-core/zakura/commit/5feb144c3): documentation.

Approved #1054 retains its original head, as requested. Its GitHub comparison retains the old ancestry until a separately coordinated restack. The previous arithmetic and specification review notes still apply. #1040 now has its own isolated consensus commit and review map: https://github.com/zakura-core/zakura/pull/1040#pullrequestreview-5255002725

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -461,6 +469,319 @@ pub fn halving(height: Height, network: &Network) -> u32 {
         .expect("halving index is non-negative and fits in u32")
 }
 
+/// The numerator of [ZIP 234]'s `BLOCK_SUBSIDY_FRACTION`.
+///
+/// [ZIP 234]: https://zips.z.cash/zip-0234
+pub const BLOCK_SUBSIDY_FRACTION_NUMERATOR: u128 = 4_126;
```
p0mvn: `BLOCK_SUBSIDY_FRACTION_NUMERATOR = 4_126` is calibrated for 1,680,000 blocks at 75-second spacing: `(1 - f)^1,680,000 ≈ 0.5`. ZIP 218 produces 5,040,000 blocks in the same four-year wall-clock period, so applying this numerator unchanged gives `(1 - f)^5,040,000 ≈ 0.125`: the NSM balance has a ~1.33-year half-life and about 87.5% is reissued in four years, rather than ZIP 234's intended ~50%.

Please ensure this consensus rule cannot activate with the flat numerator. Derive it from the spacing-adjusted halving interval (post-NU7 approximately `1375` with the existing `10^10` denominator) and retain a regression asserting that one current halving interval leaves approximately half the balance, as #1040 proposes.

p0mvn:
Superseded by the focused replacement stack based on #1052: #1070 → #1071 → #1072 → #1073 → #1074. The replacement stack preserves this PR’s implementation and has green CI. Keeping the source branch because #1056 still targets it. The replacement stack separately needs the newer #1052 commit 311d822a incorporated and revalidated before merging.

PR #1054, opened 2026-09-18
feat(rpc): use parent issuance deficits in mining templates and subsidy RPCs

evan-forbes:
## Motivation

Mining templates and subsidy RPCs must use the same parent deficit as consensus verification. A future template cannot precompute its coinbase before the new parent's pools exist.

After #1053, the valid coinbase for block h depends on the NSM balance of its parent (ZIP 234). The RPCs must compute the same value, or miners build blocks that consensus rejects. getblocktemplate now snapshots the parent's pools with the tip. It passes the balance through transaction selection and coinbase construction, so the template's subsidy, fees and coinbase outputs agree with the verifier. 

Where reissuance applies, a template for a future parent can't precompute its coinbase, because that parent's pools don't exist yet. The template therefore builds the coinbase after the parent commits. getblocksubsidy reads the best-chain parent's BlockInfo. It answers only up to tip + 1, because later subsidies depend on blocks that don't exist yet. This replaces #1053's compile-only None adapters with real parent pools.

Chunk 6 of the #859 split and the last PR in the stack. Depends on #1053. It replaces #859 together with #1049–#1053.

## Solution

Include parent pools in the template state snapshot. Pass the deficit through transaction selection and coinbase construction. Disable precomputed coinbases where reissuance applies. Resolve getblocksubsidy through best-chain parent metadata and explain its tip-plus-one height limit.

Chunk 6 of #859. Depends on #1053 and completes the replacement stack.

## Testing

Mining RPC tests cover zero, one-zatoshi, larger, negative, and missing parent deficits at and after activation. They compare RPC subsidy and serialized template coinbase values with an independent integer bonus calculation. All 144 RPC tests pass with nu7.

## Changelog

Added `docs/changelog/unreleased/1054.md`.


czarcas7ic (approved):
ACK. Highlighting that once reissuance activates, we stop precomputing future coinbases. This increases template creation latency, particularly for shielded coinbases, because proof generation must wait until the new parent’s deficit is known.

evan-forbes:
Follow-up from the #1052 audit (V12 [F-282398](https://v12.sh/runs/8019/282398)): the RPC doesn't report the issuance deficit.

#1052 stores the deficit as a seventh leg of the chain `ValueBalance`. #1040 renames it to `nsm_value_balance`. This PR reads it to compute subsidies in `getblocktemplate` and `getblocksubsidy`, but no RPC returns the balance itself:

- `getblockchaininfo` passes the tip balance to `GetBlockchainInfoBalance::value_pools`, which returns only the six monetary pools (transparent, Sprout, Sapling, Orchard, lockbox, Ironwood).
- Verbose `getblock` computes a delta across all seven legs, then drops the deficit leg in the same six-entry projection.

Leaving it out of `valuePools` is deliberate. The deficit is a signed counter, not a pool of funds, so listing it there would break clients that sum `valuePools` to get the chain supply. zcashd has no equivalent field.

Operators and explorers will still want to see the NSM balance once ZIP 234 reissuance is live. A separate field would expose it without touching `valuePools`, for example `nsmValueBalance` on `getblockchaininfo` and on verbose `getblock`. That can go in this PR or in a follow-up after #1040 settles the name.

PR #1056, opened 2026-09-18
feat(consensus): recycle NU7 transaction fees into NSM

czarcas7ic:
## Motivation

Builds on #1074 to complete NU7 fee recycling. The earlier implementation required miners to claim all fees. From NU7's first block, a block with 1,000 zatoshi in fees pays 400 to the miner and contributes 600 to NSM.

## Solution

Coinbase validation and mining templates share one calculation that contributes 60% of aggregate block fees to NSM, rounded down once per block. It activates by height alone in every build, including the interval before reissuance starts. The existing balance calculation captures the contribution exactly once. The stack uses the fixed NU7 payout fraction from #1081. The obsolete Cargo feature checks are removed, and the comments, design document, and changelog describe the same activation rule.

## Testing

The updated stack passes all 439 chain, 260 consensus, 152 RPC, and 665 state library tests with NU7 enabled, with six ignored. This includes the full migration batch/retry test, fee activation, aggregate rounding, exact coinbase claims, and mining template checks. Clippy passes for all targets in the affected libraries and node crate. Formatting, Markdown lint, and changelog validation also pass.

Follow-up CI fix: the mining verifier mock now shares the test’s 10-second deadline instead of timing out after 300 ms during template construction. The regression holds construction for 600 ms; it fails with the old deadline and passes with the fix. All 165 RPC library tests pass locally (two ignored), along with Clippy, formatting, and changelog validation. CI on `fb156f0d8` is green: all 37 executed checks passed, including all 1,238 tests in the previously failing shard.

After merging the latest main at `a37384d15`, all 167 RPC library tests, five fee-allocation and consensus tests, and three mining-template state tests passed. RPC Clippy for all targets, formatting, Markdown lint, and changelog validation also passed. The merge preserves the shared template-construction limit and cancellation tests from #1074 together with the verifier timeout regression.

## Changelog

Updated `docs/changelog/unreleased/1056.md`.

Used Codex for implementation, tests, documentation, and this description.


czarcas7ic:
@codex review

czarcas7ic:
In ref to the v12 finding, NU7 is intended to activate with fee recycling included from its first block, so this is a non issue

evan-forbes (commented):
lgtm

evan-forbes (approved):

evan-forbes:
@czarcas7ic, the NU7 stack no longer has a `nu7` Cargo feature. #1053 no longer adds it, and #1054 and #1040 have merged that change and removed their own uses. Every NU7 rule on the stack now activates by height alone, in every build. #1060 removes the last use on `main`, the ZIP 2003 V4 gate.

Could you do the same on this branch after merging `feat/nsm-value-balance-seed`? The feature no longer exists in `zakura-chain`, so the `cfg!(feature = "nu7")` checks here evaluate to false in every build, and the fee recycling rule never runs. The `unexpected_cfgs` lint also flags each check, so clippy fails with `-D warnings`.

What changes:

- `subsidy/fees.rs`: drop `!cfg!(feature = "nu7") ||` from the NU7 check, so the rule applies from NU7 in every build.
- Tests in `subsidy/fees/tests.rs`, `consensus/src/block/tests/nsm_fees.rs` and `rpc/.../get_block_template/tests/nsm_fees.rs`: remove `#[cfg(feature = "nu7")]`, and keep the with-feature side of each `cfg!(feature = "nu7") && ...` expectation.
- Docs: the `block.rs` and `fees.rs` doc comments, `docs/changelog/unreleased/1056.md`, and `docs/design/reissuance-accounting.md` say "with the `nu7` feature enabled". Those sentences can drop that clause.

Once #1060's heights land, every build recycles 60% of each block's fees into the NSM from NU7, as D8 decided. Miners will notice this change at NU7.

p0mvn (approved):

PR #1062, opened 2026-09-19
fix(state): harden header-chain context migration and startup

p0mvn:
## Motivation

The wider retained difficulty context introduced by #1045 needs an explicit header-chain disk-format boundary. Without it, an older release can open the database and fail with a row-limit error instead of reporting that the database was written by a newer binary. Startup also resized the context before checking whether the full state had rolled back below persisted header finality, hiding the more useful finalized-divergence error.

## Solution

Bump the header-chain disk format from version 4 to version 5 before widening the retained validation context, and migrate released version-4 metadata before backfilling context rows. Older binaries now reject the format marker before reading the wider context.

Check finalized-state divergence before resizing the validation context, and log the number of rows changed when resizing succeeds.

This is extracted from #851 and does not change target spacing, the active difficulty window, or other NU7 consensus parameters.

## Testing

- `cargo +1.97.0 test --locked -p zakura-state --lib` (603 passed, 4 ignored)
- `cargo +1.97.0 clippy --locked -p zakura-header-chain -p zakura-state --all-targets -- -D warnings`
- Markdown lint for `docs/changelog/unreleased/1062.md`

## Changelog

Added `docs/changelog/unreleased/1062.md`.

- #851
- #1045


p0mvn:
I'm reviewing @evan-forbes's work in #851 and splitting up the non-consensus-critical chunks. 

Since this is not consensus-critical, I'm going to merge with only one approval

PR #1064, opened 2026-09-19
test(zakurad): add the v1.4.0 compatibility config

p0mvn:
## Motivation

Configuration compatibility tests keep released configuration files as fixtures, but the v1.4.0 fixture was added inside the larger ZIP 218 change in #851.

## Solution

Add the released v1.4.0 configuration as a standalone compatibility fixture. This is extracted from #851 and does not change runtime behavior.

## Testing

- No Rust code changes; CI will exercise the configuration compatibility suite.

## Changelog

No fragment is required because this PR adds only a test fixture.

- #851


p0mvn:
Split up from #851 to narrow down the focus. I am the reviewer of Evan's work in #851 and am merging it with one approval

PR #1066, opened 2026-09-19
feat(consensus): reduce block target spacing to 25 seconds at NU7

p0mvn:
## Motivation

ZIP 218 reduces the target block spacing from 75 seconds to 25 seconds at NU7. The consensus schedule, difficulty window, subsidy calculations, and consumers of those parameters must change at the same activation height.

This PR supersedes #851 after its preparatory and non-consensus changes were split into focused PRs. It is stacked on #1063, which centralizes candidate-height difficulty-context reads.

## Solution

Activate the 25-second NU7 spacing and 102-block difficulty averaging window, retain the required difficulty context, and apply the new spacing era to subsidy and halving calculations. Update Regtest halving derivation, RPC and block-template consumers, operational timing expectations, and boundary/formula tests.

## Testing

- `cargo +1.97.0 test --locked -p zakura-chain -p zakura-header-chain -p zakura-state --lib` (416 chain, 286 header-chain, and 605 state tests passed; 6 ignored)
- `cargo +1.97.0 test --locked -p zakura --lib lookahead_limit_scales_with_target_spacing`
- `cargo +1.97.0 test --locked -p zakura --lib progress::tests`
- `cargo +1.97.0 test --locked -p zakura-rpc --lib rpc_getnetworksolps_uses_averaging_window_at_height`
- `cargo +1.97.0 clippy --locked -p zakura-chain -p zakura-header-chain -p zakura-state -p zakura-rpc -p zakura --all-targets -- -D warnings`
- `cargo +1.97.0 fmt --all -- --check`
- `git diff --check`

## Changelog

Added `docs/changelog/unreleased/1066.md` with the user-visible ZIP 218 activation, including provenance back to #851.

- Supersedes #851
- Depends on #1063
- ZIP 218: https://zips.z.cash/zip-0218


crates/zakura-state/src/service/check/tests/vectors.rs
```diff
@@ -191,15 +191,79 @@ fn difficulty_context_follows_the_candidate_height() {
...
+    let candidate_time = DateTime::from_timestamp(15_000, 0).expect("test timestamp is in-range");
+    let context = daa_context(&network, parent_height, candidate_time);
+    let expected = AdjustedDifficulty::new_from_header_time(
+        candidate_time,
+        parent_height,
+        &network,
+        context.clone(),
+    )
+    .expect("the test supplies the complete post-NU7 difficulty context")
+    .expected_difficulty_threshold();
```
czarcas7ic: AI comment, stepped through and agreed

---

This test can't fail on the difficulty threshold. `Network::new_regtest` sets `disable_pow(true)`, and with PoW disabled the validator only checks that `difficulty_threshold` is a valid encoding and never calls `expected_difficulty_threshold`:
https://github.com/zakura-core/zakura/blob/2a2b1819955a8424d9f2d965e8ee38694447decf/crates/zakura-chain/src/parameters/network/testnet.rs#L1223
https://github.com/zakura-core/zakura/blob/2a2b1819955a8424d9f2d965e8ee38694447decf/crates/zakura-header-chain/src/validation/contextual/validate.rs#L77-L92

I confirmed this locally: leaving block 347499's real Mainnet `nBits` in the header instead of `expected` still passes. Even with PoW on, the test would be circular, since `expected` comes from the same calculation it checks, and `daa_context` uses one threshold with even spacing, which gives the same answer for a 17-block or 102-block window.

The only test that pins a fixed DAA answer stops at NU6.3:
https://github.com/zakura-core/zakura/blob/2a2b1819955a8424d9f2d965e8ee38694447decf/crates/zakura-header-chain/src/validation/contextual/tests/validation.rs#L1007-L1040

Can we add post-NU7 versions on a configured Testnet built with `Parameters::build()` (PoW on), with expected answers worked out separately and mixed thresholds and timestamps? Ideally: just before, at, and after NU7; a check that the 17-block answer is rejected; heights 102/103 with NU7 active from height 1 (the `PoWLimit` cutoff moves from 17/18); and Testnet blocks 150 versus 151 seconds after their parent (6 × 25 s).

p0mvn: Addressed in aaf582130. The state integration test now uses a PoW-enabled configured Testnet and fixed, independently calculated nBits with mixed targets and timestamps; it accepts the 102-block result and rejects the old 17-block result. I also added fixed vectors just before/at/after NU7, the height 102/103 PoWLimit cutoff, and the strict 150/151-second Testnet minimum-difficulty boundary. `cargo +1.97.0 test --locked -p zakura-header-chain -p zakura-state --lib` passes (289 header-chain and 605 state tests; 6 ignored).

czarcas7ic: Verified aaf5821

czarcas7ic (approved):

PR #1067, opened 2026-09-19
test(consensus): add independent NU7 difficulty vectors

p0mvn:
## Motivation

PR #1066 changes the active difficulty averaging window at NU7, but its integration test calculated the expected threshold with the same implementation under test. A shared window, median, damping, or compact-target error could therefore pass unnoticed.

## Solution

Add fixed, independently calculated `nBits` vectors at NU7 - 1, NU7, and NU7 + 1 using nonuniform targets and timestamps. Each vector also rejects an adjacent incorrect compact target.

This PR changes tests only.

## Testing

- `cargo +1.97.0 test --locked -p zakura-header-chain nu7_difficulty_vectors_match_expected_nbits --lib`
- `cargo +1.97.0 clippy --locked -p zakura-header-chain --all-targets -- -D warnings`
- `cargo +1.97.0 fmt --all -- --check`
- `./scripts/changelog.py check`
- `markdownlint docs/changelog/unreleased/1067.md --config .github/.markdownlint.yaml`
- `git diff --check`

## Changelog

Added `docs/changelog/unreleased/1067.md` with an explicit no-changelog marker because this PR only adds tests.

### Specifications & References

- Stacked on #1066
- ZIP 218: https://zips.z.cash/zip-0218

PR #1068, opened 2026-09-19
fix(chain): apply NU7 spacing at activation boundary

p0mvn:
## Motivation

The near-tip and network-height estimators counted the NU7 activation block with the pre-NU7 75-second spacing. ZIP 218 defines the activation-height block as the first block using the 25-second spacing, so estimates crossing the boundary were 50 seconds too high.

## Solution

- Switch target spacing before counting the activation block in segmented near-tip calculations and network-height estimation.
- Update the property oracle to use height-dependent spacing transitions.
- Add exact activation-boundary regressions for near-tip classification and network-height estimation.

## Testing

- `cargo +1.97.0 test --locked -p zakura-chain chain_tip::tests --lib`
- `cargo +1.97.0 clippy --locked -p zakura-chain --all-targets -- -D warnings`
- `cargo +1.97.0 fmt --all -- --check`
- `markdownlint docs/changelog/unreleased/1068.md --config .github/.markdownlint.yaml`
- `git diff --check`

## Changelog

Added `docs/changelog/unreleased/1068.md` with an explicit no-changelog marker because this PR corrects unreleased behavior from #1066.

- Stacked on #1067
- ZIP 218: https://zips.z.cash/zip-0218


crates/zakura-chain/src/chain_tip/tests/prop.rs
```diff
@@ -35,18 +29,7 @@ proptest! {
...
-                network_height.min(blossom_activation_height),
-                NU_BEFORE_BLOSSOM,
-            )
-            // Estimate time difference for heights after Blossom activation.
-            + estimate_time_difference(
-                current_height.max(blossom_activation_height),
-                network_height.max(blossom_activation_height),
-                NetworkUpgrade::Blossom,
-            );
+            estimate_time_difference(&network, current_height, network_height);
```
czarcas7ic, line 32: AI comment

---

This test can add enough time to reach the next block, then incorrectly expect the height to stay unchanged. Just before Blossom, it can add 112.5 seconds even though the next block’s spacing is 75 seconds. Can we limit the extra time using the next block’s spacing and add a regression test for this case?

docs/changelog/unreleased/1068.md
```diff
@@ -0,0 +1,4 @@
+<!-- changelog: none -->
+
+This PR corrects unreleased behavior from #1066; that PR's fragment already
```
czarcas7ic: AI comment, is technically correct since Blossom changed target spacing, seems worth just adding to changelog.

---

This changes `getblockchaininfo` height estimates when syncing across Blossom. The old estimator behavior shipped in v1.4.0.
https://github.com/zakura-core/zakura/blob/1e36d1bb6a8a9778a1bd316704b9c8cb75182de6/crates/zakura-chain/src/chain_tip/network_chain_tip_height_estimator.rs#L70-L72

Could we replace the exclusion with a `## Fixed` entry describing the corrected height estimates and near-tip window across spacing changes?

crates/zakura-chain/src/chain_tip/network_chain_tip_height_estimator.rs
```diff
@@ -68,7 +68,11 @@ impl NetworkChainTipHeightEstimator {
     /// estimation.
     pub fn estimate_height_at(mut self, target_time: DateTime<Utc>) -> block::Height {
         while let Some((change_height, next_target_spacing)) = self.next_target_spacings.next() {
-            self.estimate_up_to(change_height);
+            self.estimate_up_to(
+                change_height
+                    .previous()
+                    .expect("target spacing changes after genesis"),
+            );
```
czarcas7ic: AI comment

---

With this change `estimate_up_to` no longer does what its doc says ("advance ... to the next change in target spacing"), and its `time_to_activation` local now names the wrong thing, since the only caller stops one block short. Upstream Zebra still calls `estimate_up_to(change_height)`, so the next merge, or a reader trusting the doc, could revert this without a test noticing on Mainnet or Testnet.

https://github.com/zakura-core/zakura/blob/9ff314014fb9004d6ceeec79a94dc96db5020998/crates/zakura-chain/src/chain_tip/network_chain_tip_height_estimator.rs#L87-L102

Could we fold the offset into the helper (advance to the last block before `change_height`), rename the local, and state once here that the activation block already uses the new spacing per ZIP 218 (`IsNU7Activated(height) := height >= NU7ActivationHeight`)? Nit: the `expect` message states the invariant but not why it holds; `new()` consumed the genesis entry and configured upgrades cannot activate at height 0.

crates/zakura-chain/src/chain_tip/network_chain_tip_height_estimator.rs
```diff
@@ -68,7 +68,11 @@ impl NetworkChainTipHeightEstimator {
...
 
             if self.current_block_time >= target_time {
```
czarcas7ic: AI comment

---

Two pre-existing issues in this function, not regressions from this PR, flagging in case we want them here or in a follow-up:

1. When `target_time` is at or before the tip block time, this `break` fires on the first iteration before any later spacing is adopted, so every backwards estimate uses the 150 second genesis spacing. Regtest with NU7 at 1000, tip 1001, estimate at tip time minus 125 seconds returns 1000 where the 25 second model gives 998. Callers only use negative distances as "at tip" and `getblockchaininfo` clamps to the tip, so the visible impact is the progress log.

2. `num_seconds()` truncates toward zero, so the blanket `-= 1` below is off by one at exact block times and never fires for a sub-second negative remainder. With NU7 at 1000 and tip 998, this PR returns 999 at tip time plus 74.5 seconds (model 998), while the base returned 998 at plus 75.0 seconds (model 999). The one second window just moved with the new stopping point. Flooring the signed duration (for example `num_milliseconds().div_euclid(1000)`) and dropping the `-= 1` fixes both cases.

https://github.com/zakura-core/zakura/blob/9ff314014fb9004d6ceeec79a94dc96db5020998/crates/zakura-chain/src/chain_tip/network_chain_tip_height_estimator.rs#L114-L130

crates/zakura-chain/src/chain_tip.rs
```diff
@@ -70,11 +70,13 @@ fn is_at_or_near_tip_with_spacing_changes(
     let baseline_spacing = i64::from(POST_BLOSSOM_POW_TARGET_SPACING);
     let time_limit = AT_OR_NEAR_TIP_THRESHOLD * baseline_spacing;
     let mut elapsed = 0;
-    let mut segment_start = local_tip;
+    let mut segment_start = local_tip
+        .next()
+        .expect("an estimated tip above the local tip means the local tip is below Height::MAX");
```
czarcas7ic: AI comment

---

Moving `segment_start` to `local_tip + 1` uses up the one block of slack in the raw `u32` subtraction below, and this helper's doc does not say the schedule must be strictly above `local_tip` and non-decreasing:

https://github.com/zakura-core/zakura/blob/9ff314014fb9004d6ceeec79a94dc96db5020998/crates/zakura-chain/src/chain_tip.rs#L83

`is_at_or_near_tip_with_spacing_changes(Height(1000), Height(1010), 75, [(Height(1000), 25)])` now panics in debug builds and wraps in release (no `overflow-checks` in our profiles), returning "far from tip". Production is safe because `is_at_or_near_tip` filters `height > local_tip`, but this is the injectable test seam. Could we use the checked `Height - Height` form the proptest oracle already uses, or document the precondition with a `debug_assert`? Longer term, a shared `target_seconds_between_heights` in `zakura-chain` (`zakura-rpc` already has a private one with the same convention) would replace this loop and the proptest oracle.

crates/zakura-chain/src/chain_tip/tests/vectors.rs
```diff
@@ -91,3 +98,27 @@ fn near_tip_threshold_handles_spacing_change_boundaries() {
...
+        DateTime::from_timestamp(2_000_000_000, 0).expect("test timestamp is in range");
+    let estimate_after = |seconds| {
+        NetworkChainTipHeightEstimator::new(current_time, Height(ACTIVATION - 1), &network)
+            .estimate_height_at(current_time + Duration::seconds(seconds))
+    };
+
+    assert_eq!(estimate_after(24), Height(ACTIVATION - 1));
+    assert_eq!(estimate_after(25), Height(ACTIVATION));
+    assert_eq!(estimate_after(49), Height(ACTIVATION));
+    assert_eq!(estimate_after(50), Height(ACTIVATION + 1));
```
czarcas7ic: AI comment

---

Starting at `ACTIVATION - 1` makes both `estimate_up_to` calls no-ops (remaining blocks -999 and 0), so this vector never exercises the advance whose argument the PR changed; an implementation that also counted the blocks below activation at 25 seconds would pass all four asserts. Could we add a start at `ACTIVATION - 10`, where 9 × 75 + 24 seconds gives `ACTIVATION - 1` and 9 × 75 + 25 gives `ACTIVATION`? Nit: this rebuilds the regtest network from the test above with a third spelling of the activation height; a shared helper and constant would do.

crates/zakura-chain/src/chain_tip/tests/vectors.rs
```diff
@@ -52,16 +57,18 @@ fn near_tip_threshold_handles_spacing_change_boundaries() {
         )
     };
 
-    // Sixteen old-spacing blocks are exactly the baseline 20-minute window.
+    // The activation block already uses the new spacing.
     assert!(crosses_activation(ACTIVATION.0 - 16, ACTIVATION.0));
-    assert!(!crosses_activation(ACTIVATION.0 - 16, ACTIVATION.0 + 1));
+    assert!(crosses_activation(ACTIVATION.0 - 16, ACTIVATION.0 + 2));
+    assert!(!crosses_activation(ACTIVATION.0 - 16, ACTIVATION.0 + 3));
```
czarcas7ic: AI comment

---

These boundary cases go through the injectable seam with a literal schedule, and `near_tip_threshold_follows_nu7_spacing` mocks the distance inside one era, so nothing drives `is_at_or_near_tip` (the `height > local_tip` filter plus the new `local_tip.next()`) across an activation height. Relaxing that filter to `>=` or dropping the `.next()` keeps every test green. Could we add, through `is_near` in the NU7 test: `is_near(NU7 - 16, 18)` true, `is_near(NU7 - 16, 19)` false, `is_near(NU7 - 1, 2)` true?

crates/zakura-chain/src/chain_tip/tests/prop.rs
```diff
@@ -63,21 +46,34 @@ proptest! {
...
+        NetworkUpgrade::target_spacing_for_height(network, start_height).num_seconds();
+    let mut elapsed_seconds = 0;
 
-    if height_difference > 0 {
-        Duration::seconds(height_difference * spacing_seconds)
-    } else {
-        Duration::zero()
+    for (change_height, next_spacing) in NetworkUpgrade::target_spacings(network) {
+        if change_height < segment_start {
+            segment_spacing = next_spacing.num_seconds();
```
czarcas7ic: AI comment

---

Two things about the new oracle:

1. The initial `segment_spacing` is a dead store: `target_spacings` always yields the genesis entry at `Height(0)` first and `segment_start >= 1`, so the `change_height < segment_start` branch overwrites it before it is read. It also reads the spacing at `start_height` rather than at the first counted block, which is the convention this PR is removing. Dropping it and filtering the iterator to `height > start_height`, as `is_at_or_near_tip` does, leaves one assignment per iteration.

2. The oracle now reads the same `target_spacings` table the estimator consumes, where the old one derived Blossom independently from `target_spacing()` and `activation_height()`. A wrong entry in that array (say `Nu7` paired with the 75 second constant) would pass. And since `any::<Network>()` only yields Mainnet and the default Testnet, neither with an NU7 height, the property never sees a 25 second segment and crosses Blossom in roughly 14% of 256-case runs. Could we add a configured-NU7 regtest arm to the strategy and derive the oracle's spacings from `target_spacing_for_height`?

czarcas7ic (approved):

PR #1072, opened 2026-09-19
feat(chain): define the NSM reissuance start height

p0mvn:
## Motivation

NSM reissuance needs an explicit start height selected by the deployment ZIP. Mainnet and default Testnet must leave it unset until that decision is made.

## Solution

Add per-network `NSM_REISSUANCE_HEIGHT` constants set to `None`, with TODOs to fill them in once the NU7 deployment ZIP selects the heights. Custom Testnets and Regtest can configure their own heights. Reissuance requires both NU7 and an explicit height, and never starts before NU7.

Used Codex to implement and validate the change.

## Testing

All 433 chain library tests and 31 network configuration tests pass. Coverage includes unset heights with NU7 enabled, activation boundaries, and configuration round trips for custom Testnets and Regtest. Chain and network Clippy pass for all targets, along with formatting, Markdown lint, and changelog checks.

## Changelog

Added `docs/changelog/unreleased/1072.md`.


crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -524,11 +524,271 @@ pub fn block_subsidy_fraction_numerator(height: Height, network: &Network) -> u1
...
+    Some(Height(Pre218Schedule::new(network).real_height(crossing.0)))
+}
+
+/// Returns the lowest height after halving `halving` at which
+/// `ceil(BLOCK_SUBSIDY_FRACTION * (MAX_MONEY - ScheduledSupply(height - 1)))` is less than
+/// the block's halving subsidy, on `network`'s 75-second schedule.
+///
+/// Returns `None` if no such height exists. See [`nsm_reissuance_height`].
+pub(crate) fn zip234_crossing_height(network: &Network, halving: u32) -> Option<Height> {
+    let schedule = Pre218Schedule::new(network);
```
czarcas7ic: Would it be better to calculate the crossing directly on the 25s schedule, since that’s what NU7 will use? Seems like we could avoid the intermediate 75s calculation and height conversion, unless there’s a reason to preserve that exact starting point. I could be missing something here though on why we chose to use 75s schedule.

czarcas7ic (approved):

evan-forbes (approved):

czarcas7ic:
@codex review

crates/zakura-chain/src/parameters/network/subsidy.rs
czarcas7ic: Actually, do we need to calculate the crossing at all? Since we're keeping halvings and the coinholder poll chose February 2031, could we just use an agreed start height and leave it unset for now? Then we wouldn't need either the 75s calculation or the conversion.

czarcas7ic: Are we intentionally keeping the old 75s crossing as the reference point for the start date, even though we're preserving halvings? If so, that explains why we'd preserve this calculation instead of recomputing it at 25s. It would help to make that policy choice explicit here.

czarcas7ic: I think there were some discussions on this last night I was just working on something else

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -503,11 +503,271 @@ pub const BLOCK_SUBSIDY_FRACTION_NUMERATOR: u128 = 1_375;
...
+            _ => real,
+        })
+    }
+
+    /// Returns the halving subsidy at `height`.
+    fn subsidy(&self, height: u32) -> Option<u128> {
+        match self.spacing_change {
+            Some(change)
+                if height >= change.height
+                    && Height(height) >= self.network.slow_start_interval() =>
```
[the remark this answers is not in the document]
czarcas7ic: No longer needed

crates/zakura-chain/src/parameters/network/subsidy.rs
czarcas7ic: Addressed after in person discussion https://github.com/zakura-core/zakura/pull/1072/commits/ccd2f50e3dfb075de48008704571c2fb80eec3d6

czarcas7ic:
@codex review

p0mvn (commented):
LGTM. Discussed addressing the equation in a follow-up

PR #1079, opened 2026-09-19
fix(state): safely upgrade released format-28 caches

czarcas7ic:
## Motivation

Format-29 startup can move a released format-28 database before acquiring its RocksDB lock. Preserve the release upgrade path without adding discovery or recovery of older development formats.

## Solution

Reuse only the immediately previous major format. Hold the source's native RocksDB lock while publishing a checkpoint with its original format marker, then run the existing migrations. Report source and publication failures instead of silently starting fresh. The source stays in place, so later writes and compaction can use additional disk space. Used Codex to narrow the implementation, tests, and documentation.

## Testing

All 669 state and 253 consensus library tests pass with NU7 enabled, with five ignored. Coverage includes every released 28.x format marker, a source locked by another process, publication failure and retry, migration failure and reopen, and refusal to discover older development formats. All-target Clippy for both crates, formatting, Markdown lint, and changelog checks pass.

## Changelog

Updated `docs/changelog/unreleased/1079.md` for released format-28 upgrade safety.


czarcas7ic:
@v12-auditor review

czarcas7ic:
Closed this one since its addressing an existing "issue" that is not unique to v29 db

PR #1082, opened 2026-09-20
feat(chain): derive NSM activation from the NU7 crossover

p0mvn:
## Motivation

NSM reissuance should be determined by the configured NU7 activation height. A diagnostic calculator alone leaves reissuance disconnected from that schedule. ZIP 218 changes target spacing to 25 seconds, so the crossover must use the actual spacing-aware halving clock and scheduled issuance.

## Solution

Wire the spacing-aware crossover into `nsm_reissuance_height` and the existing block-subsidy activation path. Remove the Mainnet/Testnet reissuance constants and the node configuration override. When NU7 is configured, production code always uses the calculated crossing. Without NU7 or a qualifying crossing, reissuance remains inactive.

Cache the derived result per immutable network. Artificial heights remain available only behind the existing test feature for short-chain consensus and database fixtures; node configuration cannot set them.

For NU7 activation height `N`, let `H3` and `H4` be the third and fourth halving heights under the actual spacing schedule, `S(h)` cumulative scheduled issuance through height `h`, and `B(h)` scheduled block subsidy. Use `F = 1375`, `D = 10,000,000,000`, and `M = MAX_MONEY`:

```text
A(h) = ceil(F * max(0, M - S(h - 1)) / D)
H = min { h | h >= max(H3 + 1, N) and h < H4 and A(h) < B(h) }
```

Search through `Height::MAX` when the fourth halving is not representable. Return no crossing when the third halving is not representable or the bounded search has no solution. Actual NSM balances and amounts removed from circulation do not affect the activation height.

Inside the constant-subsidy run:

```text
first = max(H3 + 1, N)
B = B(first)
R0 = max(0, M - S(first - 1))
Rmax = floor((B - 1) * D / F)
k = ceil(max(0, R0 - Rmax) / B)
H = first + k
```

The parent supply avoids a circular dependency. With illustrative Mainnet NU7 height `3,543,000` at 2026-11-05 12:00 UTC, the result is `8,940,474`, estimated at 2031-02-14 06:27:30 UTC. These inputs are illustrative, not assigned activation constants.

This third-era deployment rule is a proposed adaptation of ZIP 234, not an existing normative requirement of the halving-preserving draft. Agreement in the deployment ZIP remains outstanding.

## Suggested ZIP Normative Specification

> Let `ScheduledSupply(height)` be the cumulative block subsidy under the scheduled halving rules, including ZIP 218 target-spacing changes, but excluding NSM reissuance and amounts removed from circulation.
>
> Let `ThirdHalvingHeight` be the lowest height at which `Halving(height) = 3`.
>
> Define:
>
> `ReferenceNSMSubsidy(height) := ceiling(BLOCK_SUBSIDY_FRACTION(height) * (MAX_MONEY - ScheduledSupply(height - 1)))`.
>
> `DEPLOYMENT_BLOCK_HEIGHT` MUST be the lowest height greater than `ThirdHalvingHeight`, and no lower than the NU7 activation height, for which:
>
> `ReferenceNSMSubsidy(height) < ScheduledBlockSubsidy(height)`.
>
> This calculation MUST apply ZIP 218's target-spacing schedule using the assigned NU7 activation height. It MUST assume that no funds have been removed from circulation. Actual NSM balance changes MUST NOT alter `DEPLOYMENT_BLOCK_HEIGHT`. The resulting Mainnet and Testnet heights MUST be published as constants in the NU7 deployment ZIP.

This is a proposed spacing-aware adaptation of ZIP 234's crossover selection rule for the halving-preserving design. It is not currently normative text in the draft.

## Testing

- `cargo test -p zakura-chain --locked`: 445 unit tests and 9 doc tests passed.
- Network configuration round-trip and removed-override rejection test passed.
- Production chain/network library compile checks passed.
- Chain/network Clippy for all targets with warnings denied passed.
- Formatting and diff checks passed.
- Coverage includes the actual subsidy activation boundary, parent-balance requirement, multiple network schedules, NU7 inside the third era and at/after the fourth halving, the final permitted run height, missing NU7, no crossing, and an unrepresentable fourth halving.
- Property tests check the closed form against a linear oracle and randomized NU7 schedules against the earliest-crossing inequality.
- Additional consensus/state test-build validation did not complete: the native RocksDB build failed while local disk space was nearly exhausted. This is not reported as passing.

## Changelog

Updated `docs/changelog/unreleased/1082.md` to describe automatic activation and removal of the configuration override.

- [ZIP 234](https://zips.z.cash/zip-0234)
- [Halving-preserving NSM draft](https://github.com/zcash/zips/pull/1354)
- [NU7 deployment draft](https://github.com/zcash/zips/pull/1363)

### Follow-up Work

Agree on the bounded third-era deployment-height rule in the ZIP drafts, including explicit no-solution behavior, before deploying these consensus rules.


p0mvn:
@codex review

crates/zakura-chain/src/parameters/network/subsidy.rs
```diff
@@ -503,23 +503,131 @@ pub const BLOCK_SUBSIDY_FRACTION_NUMERATOR: u128 = 1_375;
...
+    };
+    let Some(third_halving) = height_for_halving(NSM_REISSUANCE_START_HALVING, network) else {
+        return Ok(None);
+    };
+    let run_end = match height_for_halving(NSM_REISSUANCE_START_HALVING + 1, network) {
+        Some(fourth_halving) => fourth_halving
+            .previous()
+            .map_err(|_| SubsidyError::UnsupportedHeight)?,
+        None => Height::MAX,
+    };
```
evan-forbes, line 538: worth noting that if this doesn't occur before before the 4th halvening then it I don't think it can activate afaict

https://github.com/zakura-core/zakura/blob/45865c11a03f94b6e8a28823d3ae67564b49c729/crates/zakura-chain/src/parameters/network/subsidy.rs#L592-L593

evan-forbes (approved):
might have just been a sync convo thing/ me being out of loop but just noting that the current zebra implementation differs here, so just making sure we communicate with them 

and presumably update the spec as I didn't see it there either

- The halving-preserving draft (https://github.com/zcash/zips/blob/60720df9e971e19d8b6f67f1869426d78c96d250/zips/draft-judah-nsm-halving-preserving-issuance.md#L181-L185) says the deployment ZIP assigns the height.

- The deployment draft (https://github.com/zcash/zips/blob/32f447759aba83acfb20aab0757b68147643de22/zips/draft-valargroup-deploy-nu7.md#L260-L262) lists Mainnet and Testnet heights as TBD, corresponding to February 2031.

https://github.com/ZcashFoundation/zebra/blob/8353ce4cec562b5b537ff3365030dc0c2d18c212/zebra-chain/src/parameters/network/subsidy.rs#L493-L507

otherwise math lgtm and astra

crates/zakura-chain/src/parameters/network/subsidy.rs
evan-forbes: this can't occur afaict tbc, just noting as it stuck out

p0mvn: This cannot occur is my understanding as well

czarcas7ic (approved):

Issue #1080, opened 2026-09-20
bug: getblocktemplate withholds fresh work during an equal-height non-finalized reorg, serving a stale parent until the next block


learnprogram2:
### What happened?

I expected to see this happen:

> After the chain tip advances, `get_block_template` returns work built on the new best tip.

Instead, this happened:

> During an **equal-height non-finalized reorg**, `get_block_template` stops returning new work and keeps answering with the **previous height's parent**, until the next block breaks the tie — even though the state's best tip has already advanced.

### What were you doing when the issue happened?

Found by source review of the `get_block_template` path (v1.4.0). No repro steps provided.

### Zakura Version

v1.4.0 (same path on current master)

### Additional information

The `get_block_template` long-poll loop gates emission on two conditions that both keep failing during an equal-height reorg.

**1. Tip-agreement guard.** The template parent comes from `fetch_chain_info` (`methods.rs:3031`, def `get_block_template.rs:970`), which reads `ReadRequest::ChainInfo`. Before returning, the loop checks the chain-tip watch against it: `latest_chain_tip.best_tip_hash().is_some_and(|tip| tip != chain_info.tip_hash)` → `continue` (`methods.rs:3032-3033`; the other path at `methods.rs:2872-2875`). On an equal-work fork, `best_chain` is decided by the raw tip-hash tie-break (`ChainScore::cmp`, `frontier.rs:93-96`; `Chain::cmp`, `chain.rs:2659`). Each same-height candidate committed through `update_latest_chain_channels` (`write.rs:1371`; `non_finalized_state_sender.send` at `1392`, `set_best_non_finalized_tip` at `1394`) can flip the best tip between the two observation points, so the guard keeps hitting `continue`.

**2. Self-validation withhold.** `prepare_template_in_background` (`methods.rs:1286`) submits the built template to `ReadRequest::CheckBlockProposalValidity`, whose handler requires the proposal's `previous_block_hash == best_tip` (`service.rs:3404`; check at ~`3421`: "proposal is not based on the current best chain tip"). If the best tip flipped between build and validate, this fails and `finish_mining_template` (`methods.rs:1206`) withholds / returns an error, so the miner keeps the previous job (`TemplateRejections`, `get_block_template.rs:71`; `reject` at `89`, `needs_fallback` at `109`).

[...]

PR #1083, opened 2026-09-20
fix(rpc): retry superseded mining recovery

czarcas7ic:
## Motivation

After a rejected mining template triggers fallback recovery, a tip or rejection change can leave `getblocktemplate` waiting on obsolete validation or returning its stale error. That can trigger the internal miner's error backoff even when fresh work is available. This follows #1088 and focuses on fallback recovery. Refs #1080.

## Solution

Abandon validation and state-tip waits when recovery is superseded, recheck freshness before propagating results or errors, and record successful recovery under the same lock as its context check. Validation has 29 seconds, reserving the final second of the shared 30-second budget for the committed-tip read. That read can detect a new parent even when validation times out and tip notifications lag. Retries use the existing four-rebuild limit. Abandoning a wait does not preempt dispatched proof work.

## Testing

The RPC library suite passed with 185 tests and two ignored. All 13 focused recovery regressions passed, covering stale results, delayed tip notifications, pending validation and state reads, current-context errors, and the shared deadline. The two new validation-timeout cases failed before reserving time for the read and passed afterward. RPC Clippy passed for all targets with warnings denied. Workspace formatting, diff checks, and changelog validation passed. A live testnet storm was not rerun.

## Changelog

Updated the Unreleased entry and mining-template withdrawal specification for fallback recovery.

AI assistance: Codex implemented the fixes, tests, and PR description.


crates/zakura-rpc/src/methods.rs
```diff
@@ -2869,17 +2911,9 @@ where
                 cur_time,
                 ..
             } = fetch_chain_info(read_state.clone()).await?;
-            if latest_chain_tip
-                .best_tip_hash()
-                .is_some_and(|tip| tip != tip_hash)
-            {
+            if !self.select_mining_template_parent(tip_hash) {
                 continue;
             }
```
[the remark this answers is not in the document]
czarcas7ic: Let's defer this for now. The retry already existed, and the real-node burst tests recovered without another block. A stuck tip publisher is still untested, so we'd want a repro showing excessive load or delayed recovery before adding a wait.

czarcas7ic:
Note that this is the best guess from the 1080 report. I was not able to reproduce the claim of an equal height non finalized reorg causing us to return the previous heights parent, but the issue this PR addresses would cause an error and delay which I think could result in the symptom that was reported in 1080.

czarcas7ic:
@codex review

crates/zakura-rpc/src/methods.rs
```diff
@@ -1252,34 +1275,54 @@ where
...
+            }
+            if !matches!(validation, Ok(Ok(_))) {
+                // State commits precede tip notifications. A failed proposal can already
+                // have a stale parent even while both watches still name the old one.
+                let response = tokio::time::timeout_at(
+                    deadline,
+                    call_service(self.read_state.clone(), zakura_state::ReadRequest::Tip),
+                )
+                .await
+                .map_misc_error()??;
```
p0mvn: The freshness check above precedes another await. If fallback validation fails on A, the context check passes, the Tip read captures A, and B becomes current before that response is consumed, this path still propagates A's validation error. A timeout/service error from the tip read also exits through `?` without checking whether the context changed. The internal miner then clears its work and enters its 20-second backoff. This is a narrow race, but it leaves the superseded-request retry incomplete.

Keep the tip-read result and recheck parent/revision freshness after the await, before propagating either the read error or the validation error; return `Ok(None)` when superseded. Add a case that holds the Tip response across a change to B and then returns A, plus an error/timeout variant. Failures in an unchanged context should still propagate.

crates/zakura-rpc/src/methods.rs
```diff
@@ -1252,34 +1275,54 @@ where
...
-                return Err(ErrorObject::owned(
-                    0,
-                    "template changed during recovery; retry",
-                    None::<()>,
-                ));
-            }
-            drop(current);
-            self.gbt.template_rejections.send_if_modified(|state| {
-                state.mark_prepared(chain_info.tip_hash, template.work_id());
+            .await;
```
p0mvn: Foreground fallback recovery waits for verification to finish or reach its 30-second deadline before checking whether the parent changed. If verification on A is queued or slow and B arrives, this request continues waiting on obsolete work even though it could rebuild on B. This is bounded, not a permanent stall, and becomes significant when recovery verification is slow. Background preparation already abandons its wait when the parent becomes stale.

Select between foreground validation, relevant parent/revision changes, and the existing deadline. Abandon the obsolete wait and return `Ok(None)` when superseded, while retaining validation requirements for current work. Add a regression case that publishes B while leaving A's verifier response pending and requires a fresh-state retry before the deadline; the current stale-fallback cases release the verifier themselves.

crates/zakura-rpc/src/methods/tests/mining_recovery.rs
```diff
@@ -0,0 +1,604 @@
...
+            .expect_request(mempool::Request::FullTransactions)
+            .await
+            .respond(make_mempool(parent_b));
+        let replacement = tokio::time::timeout(Duration::from_secs(2), pending)
+            .await
+            .unwrap()
+            .unwrap()
+            .unwrap()
+            .try_into_template()
+            .unwrap();
```
p0mvn, line 508: This verifies replacement work reaches the RPC caller, but the internal solver can still remain idle after recovery. In the existing [`run_mining_solver` no-template branch](https://github.com/zakura-core/zakura/blob/d898b9c7ef8ff29ca371f3947c74e717d7e081b0/crates/zakurad/src/components/miner.rs#L528-L555), observing `None` starts an unconditional 20-second sleep. If withdrawal publishes `None` and recovery publishes `Some(template)` shortly afterward, the channel update cannot wake that sleep. Once a solver observes the empty interval, replacement work can sit unused for almost 20 seconds. This is an existing internal-miner issue outside the PR diff.

As a focused follow-up, wait on `template_receiver.changed()` in that branch, retaining a timer only if needed for shutdown checks. Add a paused-time solver-wait test for `None -> Some(template)` that requires prompt resumption; successful RPC recovery or channel publication alone does not exercise this delay.

p0mvn (commented):
Three remaining mining-work availability findings from source review. The first is a remaining race in this change; the other two are existing delays worth addressing as follow-ups. These are conditional failure paths, not measured production frequencies or executed reproductions.

czarcas7ic:
Closed from split out PRs

PR #1084, opened 2026-09-20
fix(mining): prefer the first received block on equal work

czarcas7ic:
## Motivation

Equal-work forks currently select the greatest raw tip hash. Mining should prefer the valid block received first. If A arrives before B but B finishes verification first, A should become the mining tip once it passes validation. A chain with more work still wins.

## Solution

Record receipt order before asynchronous verification and preserve it through duplicates, forks, reconsideration, and temporary failures with a bounded retry cache for blocks whose proof of work was checked. Keep header download ordering independent while publishing the full-state choice. This PR is stacked on #1102, which owns mirror snapshot recovery and its tests. Merge that first, then update this branch from main and retarget it. This diff adds source receipt metadata to that protocol. The independent semver CI fix is in #1101. Receipt order is runtime metadata, so restored forks use the hash fallback after restart. Upgrade trusted secondaries alongside their primary to preserve the new preference.

## Testing

The split retains mining, retry, header/state coherence, and receipt transport regressions with this implementation. All 1,420 library tests across consensus, state, header-chain, and RPC pass locally on Rust 1.97.1, with nine ignored. All-target Clippy passes for all four crates with warnings denied. Formatting, Markdown lint, changelog validation, generated RPC artifact checks, and all 118 header conformance rules pass. Used Codex for implementation and tests.

## Changelog

Added `docs/changelog/unreleased/1084.md`.


czarcas7ic:
On finding 4, I don't think the dropped-header case loops forever. The client saves `None` after the header disappears, so its next request can't match the server's session and gets a full snapshot. It can cause a reset and replay, but it recovers.

czarcas7ic:
@codex review

crates/zakura-state/src/request.rs
```diff
@@ -395,6 +395,12 @@ pub struct SemanticallyVerifiedBlock {
     /// finalized committer. `None` means the committer falls back to computing
     /// it from the block's transactions.
     pub auth_data_root: Option<AuthDataRoot>,
+    /// Local order in which the full-block verifier received this submission.
+    ///
+    /// Earlier receipts win equal-work ties after validation. This is not block
+    /// data and is not persisted. Restored blocks have no order and precede new
+    /// receipts, with hash ordering between unstamped blocks.
+    pub receipt_order: Option<u64>,
```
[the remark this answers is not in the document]
czarcas7ic: 9.0.0 hasn’t been published yet. The latest crates.io release is 8.0.0, so the major bump already on this branch covers this API change.

czarcas7ic:
@codex review

czarcas7ic:
@v12 review

czarcas7ic:
Fixed the three issues introduced here, each in its own commit:

- F-283601: fixed in [7f1877c](https://github.com/zakura-core/zakura/commit/7f1877c84ab0465c7d98cc9be849128590c3bfce). Temporary failures keep receipt priority in a bounded retry cache.
- F-283602: fixed in [b669c06](https://github.com/zakura-core/zakura/commit/b669c06e88f56d14835e360aec4ceb3e4770863b). Empty legacy failover clears the old fork and keeps the finalized tip current.
- F-283603: fixed in [b40a5ac](https://github.com/zakura-core/zakura/commit/b40a5ac877c46b5c4591315242e4d8039bff400c). Mirrors publish complete snapshots and discard interrupted batches.
- F-283605: deferred. The ten-versus-eleven fork retention mismatch already exists on main and needs a separate fix to align header and full-state retention.

czarcas7ic:
@v12-auditor review

czarcas7ic:
@codex review

czarcas7ic:
@v12-auditor review

crates/zakura-state/src/service/queued_blocks.rs
```diff
@@ -262,13 +262,17 @@ impl QueuedBlocks {
...
-            .get(&hash)
+            .get_mut(&hash)
             .expect("replacement was inserted under the same hash");
         assert_eq!(old.0.height, replacement.0.height);
         assert_eq!(
             old.0.block.header.previous_block_hash,
             replacement.0.block.header.previous_block_hash
         );
+        // The queue can retain a body longer than the verifier's retry receipt cache.
+        if old.0.block == replacement.0.block {
```
p0mvn: **AI-generated review comment:** `QueuedBlocks::replace` still lets a later unequal body replace the earlier body solely because both share a header hash. Under ZIP-244, authorizing data (for example, non-height coinbase script data) can change without changing the transaction ID or Merkle root, so both variants can reach the orphan queue before parent context is available. The altered later body can evict the honest one and then fail the parent-dependent block-commitment check, losing the miner's valid body.

Key queued variants by `(header hash, complete-body digest)` and retain a bounded set until contextual commitment identifies the valid body, or otherwise ensure an unvalidated unequal variant cannot evict the existing body. Add a regression using semantically admissible same-header auth-data variants rather than a duplicate-transaction body.

czarcas7ic: Agreed, a later invalid body can replace the valid one. Given this already happens on main, I am tracking the queue fix separately in issue #1129.

czarcas7ic: Addressing in this PR https://github.com/zakura-core/zakura/pull/1130

evan-forbes (commented):
Makes sense, this change has grown on me. I still don't like the inevitable increase in one block reorgs, but as mentioned in the body / by ValarDragon its worth the extra selfish mining protection. 

I have two requests.

**1. Make the header engine and full state agree on equal work.** Right now `header_best` breaks ties by raw hash and `verified_best` by receipt order. A node can mine on A while it downloads and serves B's branch. Supporting that split takes the `verified_tip` override, `InvalidVerifiedPreference`, the header-branch serving path in `block_range.rs`, and the two-comparator spec text. It also defeats the policy when both headers arrive before either body: native sync downloads only the higher-hash body, so A's body may never reach the verifier (#1131).

Headers are different because a node can hold a header without its body. So the rule should be that the first body received wins, and having only a header never wins a tie. On equal work, the header engine would:

- prefer a tip whose body full state has verified over a header-only tip, and
- between two verified tips, take full state's receipt-order choice.

The raw hash would then break ties only between header-only tips. Selection stays a deterministic function of the graph plus body-verification evidence, and the engine already consumes that evidence. `header_best` and `verified_best` then agree on every tie where a body exists. Native sync downloads and serves the mining branch, so the serving change should no longer be needed.

**2. Drop the retry receipt cache** (`RetryReceipts`, `allow_retry`, `retains_error`, and the body digest). Without it, a block that fails before it reaches the state queue gets a fresh receipt when someone redelivers it.

- **Most of the audit churn comes from the cache.** Of the 27 V12 findings across runs 8107–8207, 14 concern retry-receipt retention: F-283675, F-283679, F-283740, F-283741, F-283742, F-283743, F-283789, F-283794, F-283795, F-284871, F-284872, F-284875, F-284876, and F-284877. Each fix added another classification rule or bound: the error allowlist, the Merkle-before-retry reorder, the per-header variant cap, the TTL, and the pending-duplicate exception. Run 8207 still lists F-284875 (a resubmission renews the TTL forever) and F-284877 (the global lock hashes 2 MB bodies) against the current design.
- **The cache decides #1127 early.** It exists mainly so that a block which fails because its parent is missing keeps its priority. That is the orphan-priority question the PR defers to #1127. The design doc also notes that it lets a miner reserve priority for a tip while withholding an ancestor. zcashd assigns priority only once the block and its ancestors are available, so a fresh receipt on redelivery is closer to zcashd, not further from it.
- **Without the cache, the common paths still work.** Overlapping deliveries still share one receipt through the active registry. A block waiting for its parent in the state queue still keeps its receipt on the prepared block. Transient local failures (`StateService`, `Io`, `QueueFull`) are rare, and losing a tie there costs little.

Removing the cache deletes most of `receipt.rs`, the retry tests in `block/tests.rs`, and the three `params.md` rows. It also closes every open retry finding at once. The queue behavior from p0mvn's thread belongs in #1130, as you said.

czarcas7ic:
Will drop the retry cache here; for (1), serving would still need to change because mismatched bodies are penalized (https://github.com/zakura-core/zakura/blob/1b8b57cf42d0757357fbbf7c657c9e63ca738c93/crates/zakura-network/src/zakura/block_sync/peer_routine.rs#L1592-L1602), so can we make it a follow-up that lands before 9.0.0 along with #1131?

czarcas7ic:
Retry cache removed in 86cc1371b0bb4feae2a83f4b165e238d0bcd1925, and header selection agreement is tracked in #1137.

evan-forbes (approved):
LGTM, thanks for dropping the retry cache, and you're right about serving on (1).

evan-forbes:
trying this out atm on the 80 node cluster

Issue #1086, opened 2026-09-20
perf(rpc): preserve coinbase precomputation when ZIP 234 reissuance starts


evan-forbes:
## Motivation

Preserve the `getblocktemplate` long-poll fast path when ZIP 234 reissuance starts. This is deferred performance work, not a blocker for NU7 activation.

The current NU7 stack schedules the default mainnet reissuance start around February 2031. Before that start, the stack retains coinbase precomputation. Configured test networks can activate reissuance earlier. Recheck the schedule and accounting rules before implementing this fix.

On `split/859-reissuance-rpc` at `54094c64a`, `crates/zakura-rpc/src/methods.rs:2987` disables coinbase precomputation when `is_zip234_active(precomputed_height)` is true. After a tip change, the RPC then builds the provisional empty-block coinbase inline. Shielded rewards require proof generation, so miners must wait for that build before receiving work on the new tip. We have not benchmarked the added latency on the current stack.

## Proposed fix

Predict the next tip's issuance deficit while waiting for a tip change. Validate that prediction against committed state before reusing the precomputed coinbase.

For a child that claims its full available reward, including fees, with no additional burn, the current accounting gives:

```text
child_deficit = parent_deficit - reissuance_bonus(parent_deficit)
```

Before reissuance starts, the bonus is zero. Before NU7, the deficit accounting records no change. The predictor therefore carries the parent deficit through both regimes.

1. Add `full_claim_child_issuance_deficit(network, child_height, parent_deficit)` beside `parent_issuance_deficit` in `crates/zakura-chain/src/parameters/network/subsidy.rs`. Use the existing bonus calculation and checked amount arithmetic.
2. At tip height `h`, predict the deficit after the child at `h+1`. Pass that prediction to the zero-fee coinbase build for height `h+2`.
3. Carry the prediction alongside the precomputed coinbase through the long-poll wait.
4. After fetching chain information on wakeup, reuse the coinbase only if the candidate height and actual parent deficit match the prediction. Extend the existing wrong-height discard in `methods.rs`.
5. Fall back to the existing inline build when prediction fails or either check mismatches. Preserve error propagation. Do not require a deficit before the subsidy calculation needs one.

Underclaims, unclaimed fees, burns, and unexpected tip changes can invalidate the prediction. The committed-state check makes prediction accuracy a performance concern rather than a validity assumption.

[...]

PR #1087, opened 2026-09-20
fix(rpc): preserve the precomputed template fast path

evan-forbes:
## Motivation

PR #1074 routes template construction through one shared proof slot. A tip-change response with a completed precomputed coinbase also waits for that slot, even though it needs no proof. An unrelated build can therefore delay the existing fast path before reissuance activation.

## Solution

Build empty templates with precomputed coinbases directly. Keep proof construction on the existing blocking pool with its shared limit. This adds 14 lines to the template helper and changes no reward calculation.

The post-reissuance prediction optimization remains deferred in #1086. This follow-up fixes a separate queue delay introduced by #1074, which merged during the audit.

## Testing

- Added a regression that occupies proof capacity and verifies that a precomputed template returns immediately with the same coinbase. The test failed before the fix.
- Updated the long-poll regression to require immediate completion for matching precomputed work while retaining blocking-pool checks for activation boundaries, height mismatches, and reorgs.
- RPC library suite: 167 passed, two existing tests ignored, using Rust 1.97.
- Formatting, diff whitespace, and changelog checks passed.

## Changelog

Added a fix fragment for this PR.

AI assistance: Codex audited the mining RPC changes and implemented this fix and its tests.


evan-forbes (commented):
likely needs more thought but something I noticed while talking w/ astra from #1074 

opening from review to see v 12 findings

p0mvn (approved):

p0mvn:
This trades a small chance of mining an empty block for less wasted hash power and faster convergence on the newest chain tip.

This makes sense to me, but not merging without more reviews

czarcas7ic (approved):

PR #1090, opened 2026-09-21
fix(rpc): rebuild superseded fallback templates instead of erroring

p0mvn:
## Motivation

#1088 stopped the fast path from surfacing a transient supersede to miners, and
deliberately left the fallback-recovery branch alone. That branch still has two
exits that reach the client, and under tip churn they dominate:

- the recovery re-check returns `template changed during recovery; retry`
- a failed proposal validation propagates the verifier's own error verbatim, even
  when the only reason it failed is that the tip moved

Lever E in #1089 forces template rejections so the branch is reachable at all —
the node's own templates are valid by construction, so nothing else gets there.
With rejections re-armed per parent, two block producers and 32 clients for 30s,
on #1088:

| Outcome | run 1 | run 2 |
| --- | --- | --- |
| Templates served | 2 936 | 3 567 |
| Raw verifier error reaching the client | 3 065 | 6 305 |
| `template changed during recovery; retry` | 209 | 184 |

Over half the responses were errors, and most were not a retry signal but
`"block could not be full-verified due to: ... proposal is not based on the current
best chain tip"` — the proposal verifier's internals, handed to a miner who can do
nothing with them.

## Solution

Ported from #1083, adapted to #1088's `Result<Option<_>>` shape:

- The context re-check and the successful-recovery record happen under one write
  lock, so a rejection cannot land between them and mark work prepared against a
  parent that is no longer selected. A superseded context returns `Ok(None)` and
  rebuilds instead of erroring.
- A failed validation is confirmed against committed state before being surfaced.
  State commits precede tip notifications, so a proposal can fail against a parent
  that both watches still name; if committed state has moved on, the caller
  rebuilds.
- Validation and that tip check share one 30 second deadline, so a slow validation
  cannot hand the tip read an already-expired budget.

## Testing

Measured before opening, same Lever E parameters, two runs each:

| | #1088 | this PR |
| --- | --- | --- |
| Raw verifier errors to client | 3 065 / 6 305 | **0 / 0** |
| `template changed during recovery` | 209 / 184 | **0 / 0** |
| Rebuild-budget exhaustion | 0 / 0 | 841 / 993 |
| Total withholds | 3 274 / 6 489 | 841 / 993 |
| Templates served | 2 936 / 3 567 | 3 428 / 2 994 |

Both error classes go to zero. What remains is the bounded-rebuild exhaustion
message from #1088, which is the intended transient signal rather than verifier
internals.

Note the residual is larger here than in #1088's own measurements, because this
scenario re-arms a rejection on every parent — each rebuild lands back in fallback
mode. That is an artifact of the injection, not of normal operation.

Local checks:

- [x] `cargo fmt -p zakura-rpc --check`
- [x] `cargo test -p zakura-rpc mining` — 22 passed, 1 ignored.
      `recovery_construction_yields_and_rechecks_parent` asserted the old error on a
      parent change during recovery; renamed to
      `recovery_construction_yields_and_rebuilds_on_parent_change` and updated, the
      same way #1088 updated `tip_change_during_construction_does_not_publish_stale_work`.
- [x] `cargo clippy -p zakura-rpc --all-targets`
- [x] New regression test `failed_recovery_validation_rebuilds_when_committed_state_moved_on`
      drives the case only the committed-tip read can distinguish: both watches still
      name the old parent while committed state has moved on. Confirmed to fail
      against the parent commit and pass with the fix.

## Changelog

`docs/changelog/unreleased/1090.md`. `./scripts/changelog.py check` passes.

- Issue [#1080](https://github.com/zakura-core/zakura/issues/1080)
- Based on [#1088](https://github.com/zakura-core/zakura/pull/1088), which this stacks on
- Ported from [#1083](https://github.com/zakura-core/zakura/pull/1083), by @czarcas7ic
- Harness and measurements: [#1089](https://github.com/zakura-core/zakura/pull/1089)

### Follow-up Work

- `saturated` looks unreachable: the first rejection sets `needs_fallback`, which
  stops `prepare_template_in_background` — the only producer of rejections. Asking
  for 500 rejections with 64 concurrent clients landed 6. Worth confirming and
  then either removing the exit or documenting why it stays.
- The V12 findings on #1083 are not addressed here.


p0mvn:
## Why the remaining withholds are fine

This PR takes the two error classes that reached miners from the fallback branch to
zero. What is left — 841 and 993 across two runs — is **rebuild-budget exhaustion**,
the transient signal #1088 introduced on purpose. Established three ways rather than
inferred from the error string:

1. **One emission site.** `template parent changed; retry` is returned from a single
   place, after the `'rebuild` loop falls through. It is unreachable unless every
   attempt in the budget was superseded in turn.
2. **The counter agrees.** `mining.template.rebuilt` recorded 8 652 rebuilds across
   3 999 requests, 2.16 each, with 21.4% of requests reaching the cap.
3. **Raising the bound removes them.** Rebuilt with `MAX_TEMPLATE_REBUILDS = 24`:
   3 517 requests, **zero** withholds, 2.75 rebuilds each.

So the race is handled; the budget simply runs out under the test's conditions.

### Why the count is higher than #1088's own numbers

Two effects compound, and both are properties of the scenario rather than the fix:

- The harness re-arms a forced rejection on **every** parent, so each rebuild lands
  straight back in fallback mode. Nothing does that in normal operation — the node
  builds its own templates and they are valid by construction, which is why the
  branch needed fault injection to reach at all.
- The fallback path is slow by design: it builds a *second* template and waits on a
  verifier round trip. That is ample time for the tip to move again, so attempts are
  consumed faster than on the fast path.

### Why this does not argue for raising the constant

The scenario runs a block roughly every 76 ms against a target spacing of 25 s
post-NU7 (`POST_NU7_POW_TARGET_SPACING`, ZIP 218) — a few hundred times faster than
the real network — with rejections injected on every parent. A withhold needs a tip
change to land inside a build window, and the budget needs that to happen five times
consecutively. At real spacing, with fallback mode rarely entered in the first place,
that does not happen.

Raising the bound also is not free: template construction sits behind a single-slot
semaphore (`MAX_TEMPLATE_BUILDS = 1`), so each extra rebuild is another pass through
a globally serialized queue. The current bound is the right trade.

### What this leaves open

`saturated` still looks unreachable: the first rejection sets `needs_fallback`, which
stops `prepare_template_in_background` — the only producer of rejections. Asking for
500 with 64 concurrent clients landed 6. Worth confirming separately, then either
removing that exit or documenting why it stays.

Measurements are reproducible from the harness in #1089 (Lever E).

crates/zakura-rpc/src/methods.rs
```diff
@@ -1329,30 +1332,52 @@ where
...
+            if !matches!(validation, Ok(Ok(_))) {
+                // State commits precede tip notifications, so a proposal can fail against
+                // a parent that both watches still name. Confirm against committed state
+                // before surfacing a verifier error the caller can do nothing about.
+                let response = tokio::time::timeout_at(
+                    deadline,
+                    call_service(self.read_state.clone(), zakura_state::ReadRequest::Tip),
+                )
+                .await
+                .map_misc_error()??;
```
czarcas7ic: AI comment

---

If fallback validation fails on parent A, the tip read can capture A and finish after the tip watch or selected parent has moved to B. We still propagate the old validation error, and read errors or timeouts also exit without checking whether the context changed.

Could we retain the tip-read result and recheck parent, revision, and tip freshness after the await, before propagating either error? A regression should hold the read across a parent change and cover both an old successful response and a read failure.

p0mvn:
@codex review

czarcas7ic (approved):

